PR #13, opened 2024-06-06
github: Enable syntax highlighting for Daml

stas-sbi:
This is done by making GitHub believe it's Haskell


cocreature (approved):
thanks!

waynecollier-da (approved):

PR #14, opened 2024-06-06
Generalized diagram for Splice + Canton

waynecollier-da:
Prior diagrams used the term "Global Synchronizer" which is a brand name controlled by the Global Synchronizer Foundation, and is just one example of how Splice may be implemented to provide governance and tokenomics for a decentralized Canton synchronizer.

This diagram also replaces the term "SV App", which was used to the describe the Validator node operated by a Super Validator. The updated diagram calls this node the SV Validator node, for better consistency. It also replaces the terms "SV 1" and "SV 2" with "Super Validator 1" and "Super Validator 2"


waynecollier-da:
Once this PR has been committed, use the attached diagram to replace the referenced diagram in the Splice readme file.

isegall-da (approved):
Thanks

PR #17, opened 2024-06-28
Changes from release 0.1.13

ray-roestenburg-da:


waynecollier-da:
@isegall-da I'm good with these changes in terms of their intent. Can you confirm that they are correct, and that they are consistent with the state of 0.1.13 as it will be deployed this week by the GSF node?

isegall-da:
Confirmed, thanks

isegall-da (approved):
Thanks

PR #296, opened 2025-03-26
feat: add K8S+Helm recommended/standard labels

SLoeuillet:
Add standard/recommended labels as per :
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
https://helm.sh/docs/chart_best_practices/labels/

I did keep the original labels like "app" but best practice is to "namespace" them like : canton.network/app

@isegall-da  && @moritzkiefer-da, it supersedes : https://github.com/digital-asset/decentralized-canton-sync/pull/365


SLoeuillet:
Example output :

```
+ metadata:
+   labels:
+     app.kubernetes.io/component: wallet-web-ui
+     app.kubernetes.io/managed-by: Helm
+     app.kubernetes.io/instance: canton-val-devnet-wallet-web-ui
+     app.kubernetes.io/name: splice-validator
+     app.kubernetes.io/version: 0.3.17
+     helm.sh/chart: splice-validator-0.3.17
+     app.kubernetes.io/part-of: canton-val-devnet
+     app: wallet-web-ui
+     migration: "1"
```

cluster/helm/splice-participant/templates/participant.yaml
```diff
@@ -5,6 +5,13 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
+    app.kubernetes.io/component: participant
```
martinflorian-da, line 8: I suggest we extract this whole block into a reusable definition in [cluster/helm/splice-util-lib/templates/_helpers.tpl](https://github.com/hyperledger-labs/splice/blob/main/cluster/helm/splice-util-lib/templates/_helpers.tpl). This avoids code duplication (this is a fairly large block that we'll want in many places).

I'd also argue that this is more consistent to our goal state for managing these Helm templates. See for example `include "splice-util-lib.service-account"` which we added recently.

SLoeuillet:
I must admit that I didn't touch other components as I'm on Validator deployment only for now
But yes, best way is to use a template 

I open PRs once I find some things which would make multiple deployments on the same K8S cluster fail

Exemple : in previous PR, hardcoded PVC volume name, but it could be hardcoded DB name or DB username in case you use a unique PG server

PR #318, opened 2025-04-09
canton, docker image: Add libjemalloc2

giner:
Allow operators to use an efficient memory allocator libjemalloc2 without necessity to rebuild images

### DIFF
--- a/cluster/images/canton/Dockerfile
+++ b/cluster/images/canton/Dockerfile
@@ -14,7 +14,7 @@ ARG TARGETPLATFORM
 
 # Install screen for running the console in a headless server
 RUN xx-apt-get update \
-   && DEBIAN_FRONTEND=noninteractive xx-apt-get install -y screen tini \
+   && DEBIAN_FRONTEND=noninteractive xx-apt-get install -y screen tini libjemalloc2 \
    && xx-apt-get clean \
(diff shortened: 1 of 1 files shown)


moritzkiefer-da:
@giner how are you using this in practice? i.e. where do you enable it for what?

new-programmist:
I used libjemalloc2 in other projects but can't do this easily in with splice unless it's available in docker images. The main use is decreasing memory fragmentation but it is also useful for memory profiling. In order to use it environment variable `LD_PRELOAD` has to be set to `/usr/lib/x86_64-linux-gnu/libjemalloc.so.2`. Splice helm charts allow to add arbitrary environment variables.

giner:
I used libjemalloc2 in other projects but can't do this easily in with splice unless it's available in docker images. The main use is decreasing memory fragmentation but it is also useful for memory profiling. In order to use it environment variable `LD_PRELOAD` has to be set to `/usr/lib/x86_64-linux-gnu/libjemalloc.so.2`. Splice helm charts allow to add arbitrary environment variables.

moritzkiefer-da:
the LD_PRELOAD part was the one I was interested in, ok seems reasonable. Merging

moritzkiefer-da:
merged internally, will be included in the next release, thank you!

PR #322, opened 2025-04-10
fix: put a real message when user tries to access the WalletUI in insecure context

SLoeuillet:
Signed-off-by: Stephane Loeuillet <stephane.loeuillet@kaiko.com>


SLoeuillet (commented):
Actually, when one tries to take a look at the wallet UI, it won't work if not using localhost or https as its code is using Web Crypto API from browser that is only available from a secure context.

Tell it explicitly to the user instead of dying with a cryptic error message

moritzkiefer-da:
Thanks for the contribution! We've merged this internally and it will be included in the next release.

PR #332, opened 2025-04-16
cluster, helm: Add new Helm-chart "splice-info"

giner:
This helm charts runs a simple web-server to provide static deployment and dynamic runtime information as JSON files.

- Static deployment configuration is published as JSON under /
- Runtime part supports fetching DSO information from Scan and publishes it under /runtime/dso.json


isegall-da:
Thanks @giner . I haven't yet looked at it closely, but my main concern is that there's no tests for it. 
Unfortunately that's still a bit hard to do in Splice directly, as there's no CI here (coming soon...), so I see two options: either we park this for a while, until we have enough testing infra in Splice for you to add tests, or we have someone from DA add tests in the internal repo. WDYT?

At a high level, we need to: deploy this in Pulumi (the Pulumi code is not yet in Splice though 😢 coming soon...), and add some logic to one of the preflight tests in `apps\app`

giner:
Whatever gets it merged (or rejected, also works for me :)) faster. If DA folks could add tests let's do that.

isegall-da:
Definitely aiming at merge, not reject. Let me see if someone can carve out a bit of time for this then.

cluster/helm/splice-info/templates/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+              else
+                exit_code=$?;
+                echo "ERROR: Failed to fetch DSO from $SCAN_URL";
+              fi;
+
+              end_time=$(date +%s);
+              sleep "$((period - (end_time - start_time)))";
+
+              if [ "$exit_code" -ne 0 ]; then
+                exit $exit_code;
```
giner, line 72: This `exit` makes the updater container restart if update fails (which makes the problem easily visible) while the main container keeps serving (for some time at least) the latest successfully fetched data

giner:
What sort of test would be useful here?

nicu-da:
@giner There's a few things missing to complete this:


1. We could use a helm test for the generated config map at least, to make sure all the values are treated correctly. You can check https://github.com/hyperledger-labs/splice/tree/main/cluster/helm/splice-participant/tests for some helm tests examples
2. We need to also add this to the deployment to include it as part of the DA deployments but also to be able to test the  deployment of the helm chart. This should be fairly similar to the docs deployment. You can see that one here https://github.com/hyperledger-labs/splice/blob/main/cluster/pulumi/canton-network/src/docs.ts. In the same pulumi project we need to add a new spliceInfo deployment. Note that the docs is installed globally while the new chart should be installed as part of an individual SV here https://github.com/hyperledger-labs/splice/blob/b2d78f62adbb565e5ac430349f665153d267ff45/cluster/pulumi/canton-network/src/sv.ts#L123
3. We need to test that the new chart API is accessible after the deployment. It should be part of https://github.com/hyperledger-labs/splice/blob/a91c38d296d8742561a4f4c4c25756bebae8608b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/runbook/DsoPreflightIntegrationTest.scala and we just need to add a new test scenario similar to the docs one that validates the new url. For the docs one we load the UI, in the new test probably we want to do a simple http call and check the response. You can check the way we build the http client and do a http request in the cometbft client https://github.com/hyperledger-labs/splice/blob/b2d78f62adbb565e5ac430349f665153d267ff45/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/cometbft/CometBftHttpRpcClient.scala#L190 

Just let me know if you need any help around this.

giner:
@nicu-da , thank you. I'll check p1. As for the rest, wouldn't it make sense to do that separately? This PR is only to add the helm chart and make ot available, not making it part of the deployment.

nicu-da:
> @nicu-da , thank you. I'll check p1. As for the rest, wouldn't it make sense to do that separately? This PR is only to add the helm chart and make ot available, not making it part of the deployment.

@giner We would need the rest of the points as well to test that the helm chart is deployable as part of CI.

nicu-da:
> @nicu-da, For p1 (helm tests) I'm having hard time coming up with something that would justify the need of maintaining these tests. What sort of tests do you think would be valuable?

My thoughts were only around the generated config map in `cluster/helm/splice-info/templates/configmap-content-static.yaml` as we have a lot of templating there and it's easy to mess out the output yaml

giner:
> My thoughts were only around the generated config map in cluster/helm/splice-info/templates/configmap-content-static.yaml as we have a lot of templating there and it's easy to mess out the output yaml

Noted, will do.

As for p2 and p3. I have more or less idea of the changes I need to make however I'm not certain on how to iteratively test them. The deployment doesn't seem to be suitable for local deployment (it looks like it's not cloud agnostic, i.e. I won't be able to run it on my local k8s, and it may also require some external dependencies that I don't have) so I can't run it locally. Pushing through CI is likely to be slow because the deploy process takes time and requires an approval for each change. Any suggestions?

cluster/helm/splice-info/tests/configmap-content-static_test.yaml
giner, line 1: @nicu-da something like this?

nicu-da: Yup, LGTM!

giner:
@nicu-da  I've setup my local Scala environment, created a simple http endpoint call test using ScalaTest and I think I'm pretty comfortable with this part. The biggest issue for me now is that I can't easily test the actual changes against splice repo.

cluster/expected/canton-network/expected.json
```diff
@@ -464,6 +464,90 @@
...
+            "staging": {
+              "chainIdSuffix": "4",
+              "migrationId": 4,
+              "version": "0.3.21"
+            }
+          }
+        },
+        "imageRepo": "us-central1-docker.pkg.dev/da-cn-shared/ghcr/digital-asset/decentralized-canton-sync-dev/docker",
+        "istioVirtualService": {
+          "gateway": "cluster-ingress/cn-http-gateway",
```
giner, line 531: @nicu-da @isegall-da How to make the endpoint fully public? Would replacing `cluster-ingress/cn-http-gateway` with `cluster-ingress/cn-public-http-gateway` be enough?

isegall-da: Not sure that's well maintained, it's a bit of a legacy.
I would claim that if you don't need to expose publicly for tests to pass (and I believe you don't), then you don't need to worry about that. That's when it becomes DA's problem for our deployment as opposed to part of testing.

giner: So actual nodes don't use this configuration, only tests. I see, thanks. Nvm then.

isegall-da: > actual nodes don't use this configuration

Well, to be precise, ours do, but that's not for you to solve for us ;)

cluster/pulumi/infra/src/config.ts
```diff
@@ -82,17 +76,8 @@ function extractIpRanges(x: IpRangesDict): string[] {
 }
 
 export function loadIPRanges(): pulumi.Output<string[]> {
-  if (spliceConfig.pulumiProjectConfig.isExternalCluster && !PRIVATE_CONFIGS_PATH) {
-    throw new Error('isExternalCluster is true but PRIVATE_CONFIGS_PATH is not set');
-  }
-
-  const externalIpRanges = spliceConfig.pulumiProjectConfig.isExternalCluster
```
nicu-da, line 89: I think I'm missing where this check is still performed

cluster/pulumi/common/src/utils.ts
```diff
@@ -183,22 +183,63 @@ export function fixedTokens(): boolean {
...
+  if (!net) {
+    throw new Error(`Unknown cluster network: ${clusterNetwork}`);
+  }
+
+  return net;
+}
+
+export const clusterDirectory = getClusterDirectory();
+
+function getPathToPrivateConfigFile(fileName: string): string {
```
nicu-da:
```suggestion
function getPathToPrivateConfigFile(fileName: string): string | undefined {
```

cluster/pulumi/common/src/utils.ts
```diff
@@ -183,22 +183,63 @@ export function fixedTokens(): boolean {
...
+  if (PRIVATE_CONFIGS_PATH) {
+    const svPublicConfigsClusterDirectory = `${PRIVATE_CONFIGS_PATH}/configs/${clusterDirectory}`;
+    return `${svPublicConfigsClusterDirectory}/${fileName}`;
+  } else {
+    if (spliceConfig.pulumiProjectConfig.isExternalCluster) {
+      throw new Error('isExternalCluster is true but PRIVATE_CONFIGS_PATH is not set');
+    }
+
+    return '';
```
nicu-da:
```suggestion
    return undefined
```

giner:
> Thanks for the great work. A few details that I think also cause the cluster test to fail with:
> 
> ```
> 
>     error: Running program '/home/********/project/splice/cluster/pulumi/canton-network/src/index.ts' failed with an unhandled exception:
>     Error: ENOENT: no such file or directory, open ''
>         at Object.readFileSync (node:fs:441:20)
>         at fileDigestMd5 (/home/********/project/splice/cluster/pulumi/canton-network/src/info.ts:32:21)
>         at installInfo (/home/********/project/splice/cluster/pulumi/canton-network/src/info.ts:48:18)
>         at /home/********/project/splice/cluster/pulumi/canton-network/src/sv.ts:247:14
>         at Generator.next (<anonymous>)
>         at fulfilled (/home/********/project/splice/cluster/pulumi/canton-network/src/sv.ts:28:58)
>         at processTicksAndRejections (node:internal/process/task_queues:105:5) {
>       errno: -2,
>       code: 'ENOENT',
>       syscall: 'open',
>       path: ''
>     }
> ```

How can I run this check locally? I ran `npm run check; echo $?` and it didn't fail

cluster/pulumi/infra/src/config.ts
giner: It is used in `getPathTo...` functions

cluster/pulumi/common/src/utils.ts
giner: thanks, updated

cluster/pulumi/common/src/utils.ts
giner: thanks, updated

cluster/pulumi/infra/src/config.ts
nicu-da: But that's just for the env variables if the env is set, you can have the env vars check but still we should not load the external ips as it's not an external cluster

giner: I see what you mean. It won't load if not provided https://github.com/hyperledger-labs/splice/blob/e1bb6c8690aa24243a540fd4eb88ad8eb85a5bbe/cluster/pulumi/infra/src/config.ts#L80

nicu-da: If `isExternalCluster` is false, it should not load the file, I still fail to see where we don't provide the file if `isExternalCluster` is false

nicu-da: No, I'm talking about the value of `externalIpRanges`. 
Previosly if `spliceConfig.pulumiProjectConfig.isExternalCluster` was false it was always empty, now we load it ignoring `spliceConfig.pulumiProjectConfig.isExternalCluster` as `externalIpRangesFile` checks only 
```
if (PRIVATE_CONFIGS_PATH)
```

but that can still be set even if `isExternalCluster` is false

giner: Are you talking about the error from fileDigestMd5 you shared above? If so I fixed it here https://github.com/hyperledger-labs/splice/pull/332/commits/e1bb6c8690aa24243a540fd4eb88ad8eb85a5bbe

stas-sbi: > No, I'm talking about the value of `externalIpRanges`. Previosly if `spliceConfig.pulumiProjectConfig.isExternalCluster` was false it was always empty, now we load it ignoring `spliceConfig.pulumiProjectConfig.isExternalCluster` as `externalIpRangesFile` checks only
> 
> ```
> if (PRIVATE_CONFIGS_PATH)
> ```
> 
> but that can still be set even if `isExternalCluster` is false

Got it. I'll fix that. Thanks.

stas-sbi: I didn't realize that there was an intention to ignore file even if provided for non-external cluster. I thought the only intention is to make sure the file is provided if the cluster is external.

giner: > No, I'm talking about the value of `externalIpRanges`. Previosly if `spliceConfig.pulumiProjectConfig.isExternalCluster` was false it was always empty, now we load it ignoring `spliceConfig.pulumiProjectConfig.isExternalCluster` as `externalIpRangesFile` checks only
> 
> ```
> if (PRIVATE_CONFIGS_PATH)
> ```
> 
> but that can still be set even if `isExternalCluster` is false

Got it. I'll fix that. Thanks.

giner: I didn't realize that there was an intention to ignore file for non-external cluster even if provided. I thought the only intention was to make sure the file is provided if the cluster is external.

giner: fixed

giner: please wait, I've realized that I need to do the same for the checksum

nicu-da: Ah right, we can return the checksum as part of reading the external ips, wrapping the ip list in an ojbect that also has the checksum so we don't have to do the check twice

giner: or I just return file path as undefined for non-external cluster

giner: updated

giner: ~ah, no, sorry, still missing something~

giner: no, it's all good

nicu-da:
@giner The last deployment failed with:

```
  error: Running program '/home/********/project/splice/cluster/pulumi/canton-network/src/index.ts' failed with an unhandled exception:
    Error: isExternalCluster is false but PUBLIC_CONFIGS_PATH is not set
        at getPathToPublicConfigFile (/home/********/project/splice/cluster/pulumi/common/src/utils.ts:222:11)
        at approvedSvIdentitiesFile (/home/********/project/splice/cluster/pulumi/common/src/utils.ts:241:10)
        at approvedSvIdentities (/home/********/project/splice/cluster/pulumi/common/src/utils.ts:245:16)
        at /home/********/project/splice/cluster/pulumi/canton-network/src/dso.ts:162:41
        at Dso.<anonymous> (/home/********/project/splice/cluster/pulumi/canton-network/src/dso.ts:166:7)
        at Generator.next (<anonymous>)
        at /home/********/project/splice/cluster/pulumi/canton-network/src/dso.ts:31:71
        at new Promise (<anonymous>)
        at __awaiter (/home/********/project/splice/cluster/pulumi/canton-network/src/dso.ts:27:12)
        at Dso.installDso (/home/********/project/splice/cluster/pulumi/canton-network/src/dso.ts:94:16)
```

You can test this locally by changing [this](https://github.com/hyperledger-labs/splice/blob/main/cluster/deployment/mock/config.yaml#L70)  to false and running update-expected again.

Let me know when I can do another cluster test.

Also it would be helpful to avoid force pushing as I don't know what changes were introduced since the last deployment and can't tell if there's fixes that should already cover that.

cluster/pulumi/canton-network/src/info.ts
```diff
@@ -0,0 +1,94 @@
...
+          : null,
+      },
+    },
+    istioVirtualService: {
+      host: host,
+      gateway: gateway,
+    },
+  };
+
+  installSpliceHelmChart(
```
nicu-da, line 87: @giner You're missing the `appsAffinityAndTolerations` for the helm chart

```
lastTimestamp: "2025-07-16T12:54:34Z"
message: "pod didn't trigger scale-up: 1 node(s) had untolerated taint {cn_infra: true}, 1 node(s) had untolerated taint {components.gke.io/gke-managed-components: true}, 1 node(s) had untolerated taint {cn_apps: true}"
```

giner: ~Will check. Thanks. I'm wondering if there was a way I could've known it's needed.~ Turns out it's a common practice and helm suggests these by the demo template.

PR #386, opened 2025-05-08
Issue-385: Update validator deployment ports with metrics-api

RRaule:


moritzkiefer-da (commented):
Thank you! It looks like it's also missing for SV, scan and possibly some other charts. Could you please also add it to those.

RRaule:
Will do!

stas-sbi:
[cluster/helm/splice-global-domain/templates/sequencer.yaml](cluster/helm/splice-global-domain/templates/sequencer.yaml) as well?

PR #387, opened 2025-05-09
compose, validator: Fix restoration from dump

giner:
Fix location of volume path with identities dump (identities.json)

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeValidatorFrontendIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeValidatorFrontendIntegrationTest.scala
@@ -191,7 +191,7 @@ class DockerComposeValidatorFrontendIntegrationTest
       "recovering from identities dump",
       Seq(
         "-i",
-        backupsDir.toAbsolutePath.toString,
+        identities.toAbsolutePath.toString,
         "-P",
         "da-composeValidator-13",
       ),
--- a/cluster/compose/validator/compose-restore-from-id.yaml
+++ b/cluster/compose/validator/compose-restore-from-id.yaml
@@ -12,4 +12,4 @@ services:
           ADDITIONAL_CONFIG_MIGRATE_VALIDATOR_PARTY=
             canton.validator-apps.validator_backend.migrate-validator-party.scan.admin-api.url = "${SCAN_ADDRESS}"
     volumes:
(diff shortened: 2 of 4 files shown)


isegall-da:
Thanks @giner 
I looked into this a bit. This will currently fail CI, but that's because the test and the docs are inconsistent: 

[In the test](https://github.com/hyperledger-labs/splice/blob/969355682b47274f425338511ab8bf8476b1a4d3/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeValidatorFrontendIntegrationTest.scala#L194) we actually pass the directory where the file resides, thus mounting the directory as the volume passes the test.

But indeed the docs say that the `-i` should be the file itself, not the directory. 

I'm not sure whether we should change the docs or the implementation (and test) :thinking:

giner:
I'd suggest that passing file is more obvious

moritzkiefer-da:
> I'm not sure whether we should change the docs or the implementation (and test) 🤔

changing the docs is backwards compatible, changing the behavior of `-i` is not. So I'd slightly lean towards changing the docs.

giner:
It's good to keep backwards compatibility. At the same time I think this not that sort of change which requires users to do a lot. The change in behavior could be mentioned in the release notes and the ux will become more user-friendly since that point.

moritzkiefer-da:
> It's good to keep backwards compatibility. At the same time I think this not that sort of change which requires users to do a lot. The change in behavior could be mentioned in the release notes and the ux will become more user-friendly since that point.

Fair, if we were starting from scratch the file definitely seems like a nicer UX. No strong opinion.

giner:
@isegall-da friendly ping

giner:
> @giner since CI is now all open, want to take a shot at completing this yourself?

updated tests and release notes

isegall-da:
Hi @giner , sorry, I missed this on Friday. 
Can you please merge the latest main, so that CI runs by default? (See https://github.com/hyperledger-labs/splice/pull/1127)

isegall-da:
Actually, since it's on: pull_request_target, just pushing an empty commit probably suffices, to wake up main on the latest ci logic

giner:
rebased

giner:
I've removed `[ci]` from the last commit or it's still needed?

giner:
@isegall-da there seem to be an unrelated to my change test failure

isegall-da:
Yes, there's a known flaky test there that we're looking into. I hit "rerun from failed" now.

giner:
@isegall-da I don't see any tests failing this time but final result has returned failure

PR #438, opened 2025-06-04
Most doc and helm values are Auth0 specific, make it more generic + …

SLoeuillet:
make it easier to have multiple wallet users by simple string replacement

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/pack/examples/sv-helm/participant-values.yaml
+++ b/apps/app/src/pack/examples/sv-helm/participant-values.yaml
@@ -5,8 +5,8 @@ persistence:
   databaseName: participant_MIGRATION_ID
   schema: participant
 auth:
-  # replace OIDC_AUTHORITY_URL with your provider's OIDC URL
-  jwksUrl: "https://OIDC_AUTHORITY_URL/.well-known/jwks.json"
+  # replace OIDC_AUTHORITY_URL & OIDC_AUTHORITY_JWKS_SUB_URL with your provider's OIDC base URL and JWKS sub-URL
+  jwksUrl: "https://OIDC_AUTHORITY_URL/OIDC_AUTHORITY_JWKS_SUB_URL
   # replace OIDC_AUTHORITY_LEDGER_API_AUDIENCE with the audience of your choice.
   targetAudience: "OIDC_AUTHORITY_LEDGER_API_AUDIENCE"
 enableHealthProbes: true
--- a/apps/app/src/pack/examples/sv-helm/validator-values.yaml
+++ b/apps/app/src/pack/examples/sv-helm/validator-values.yaml
@@ -13,18 +13,17 @@ scanAddress: "TRUSTED_SCAN_URL"
 # useSequencerConnectionsFromScan: false
 # TRUSTED_SINGLE_SEQUENCER_END
 
-# Replace OPERATOR_WALLET_USER_ID with the user id in your IAM that you want to use to log into
-# the wallet as the SV party. Note that this should be the full user id, e.g., ``auth0|43b68e1e4978b000cefba352``
-# not only the suffix ``43b68e1e4978b000cefba352``:
-# You can specify multiple user ids if you want multiple users to be able to log into your wallet.
-validatorWalletUsers:
-  - "OPERATOR_WALLET_USER_ID"
+# Replace OPERATOR_WALLET_USER_IDS with the quoted user id in your IAM that you want to use to log into
+# the wallet as the SV party. Note that this should be the full user id, e.g., ``"auth0|43b68e1e4978b000cefba352"``
+# not only the suffix ``"43b68e1e4978b000cefba352"``:
+# You can specify multiple user ids if you want multiple users to be able to log into your wallet : ``"userid1", "userid2"``
+validatorWalletUsers: [OPERATOR_WALLET_USER_IDS]
 auth:
   # replace OIDC_AUTHORITY_VALIDATOR_AUDIENCE with the audience of your choice
   audience: "OIDC_AUTHORITY_VALIDATOR_AUDIENCE"
 
-  # replace OIDC_AUTHORITY_URL with your provider's OIDC URL
-  jwksUrl: "https://OIDC_AUTHORITY_URL/.well-known/jwks.json"
+  # replace OIDC_AUTHORITY_URL & OIDC_AUTHORITY_JWKS_SUB_URL with your provider's OIDC base URL and JWKS sub-URL
+  jwksUrl: "https://OIDC_AUTHORITY_URL/OIDC_AUTHORITY_JWKS_SUB_URL
 
 # ENABLEWALLET_START
 # This will disable the wallet HTTP server and wallet automations when set to false
--- a/docs/src/validator_operator/validator_helm.rst
+++ b/docs/src/validator_operator/validator_helm.rst
@@ -121,14 +121,15 @@ In the following, we will refer to this URL as ``OIDC_AUTHORITY_URL``.
 Both your Validator node and any users that wish to authenticate to a web UI connected to your Validator node must be able to reach the ``OIDC_AUTHORITY_URL``.
 We require your OIDC provider to provide a `discovery document <https://openid.net/specs/openid-connect-discovery-1_0.html>`_ at ``OIDC_AUTHORITY_URL/.well-known/openid-configuration``.
 We furthermore require that your OIDC provider exposes a `JWK Set <https://datatracker.ietf.org/doc/html/rfc7517>`_ document.
-In this documentation, we assume that this document is available at ``OIDC_AUTHORITY_URL/.well-known/jwks.json``.
+In this documentation, we assume that this document is available at ``OIDC_AUTHORITY_URL/OIDC_AUTHORITY_JWKS_SUB_URL``.
+For Auth0, ``OIDC_AUTHORITY_JWKS_SUB_URL=/.well-known/jwks.json``.
 
 For machine-to-machine (Validator node component to Validator node component) authentication,
 your OIDC provider must support the `OAuth 2.0 Client Credentials Grant <https://tools.ietf.org/html/rfc6749#section-4.4>`_ flow.
 This means that you must be able to configure (`CLIENT_ID`, `CLIENT_SECRET`) pairs for all Validator node components that need to authenticate themselves to other components.
 Currently, this is the validator app backend - which needs to authenticate to the Validator node's Canton participant.
 The `sub` field of JWTs issued through this flow must match the user ID configured as ``ledger-api-user`` in :ref:`helm-validator-auth-secrets-config`.
-In this documentation, we assume that the `sub` field of these JWTs is formed as ``CLIENT_ID@clients``.
+In this documentation, we assume that the `sub` field of these JWTs is formed as ``CLIENT_ID@clients``. (``@clients`` suffix is Auth0 specific)
 If this is not true for your OIDC provider, pay extra attention when configuring ``ledger-api-user`` values below.
 
 For user-facing authentication - allowing users to access the various web UIs hosted on your Validator node,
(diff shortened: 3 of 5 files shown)


SLoeuillet:
Actually, if we have several wallet users, we have to replace

<pre>  - &quot;OPERATOR_WALLET_USER_ID&quot;=
validatorWalletUsers:=validatorWalletUsers: [&apos;&quot;stephane.loeuillet@kaiko.com&quot;&apos;, &apos;&quot;charles.desmonty@kaiko.com&quot;&apos;, &apos;&quot;ugo.plouviez@kaiko.com&quot;&apos;]
</pre>

After that patch, for one or several users
OPERATOR_WALLET_USER_IDS="stephane.loeuillet@kaiko.com"
OPERATOR_WALLET_USER_IDS="stephane.loeuillet@kaiko.com", "charles.desmonty@kaiko.com", "ugo.plouviez@kaiko.com"

docs/src/validator_operator/validator_helm.rst
```diff
@@ -260,6 +262,7 @@ OIDC_AUTHORITY_URL                 ``https://AUTH0_TENANT_NAME.us.auth0.com``
 OIDC_AUTHORITY_LEDGER_API_AUDIENCE The optional audience of your choice for Ledger API. e.g. ``https://ledger_api.example.com``
 VALIDATOR_CLIENT_ID                The client id of the Auth0 app for the validator app backend
 VALIDATOR_CLIENT_SECRET            The client secret of the Auth0 app for the validator app backend
+VALIDATOR_LEDGER_API_USER          Same as VALIDATOR_CLIENT_ID (with ``@clients`` suffix for Auth0 only)
```
stas-sbi: This doesn't seem correct to me. `VALIDATOR_CLIENT_ID` is ID of the client while `VALIDATOR_LEDGER_API_USER` needs a user ID

SLoeuillet: Well, it works that way with Okta. The "@clients" suffix is 100% Auth0 specific

apps/app/src/pack/examples/sv-helm/participant-values.yaml
```diff
@@ -5,8 +5,8 @@ persistence:
   databaseName: participant_MIGRATION_ID
   schema: participant
 auth:
-  # replace OIDC_AUTHORITY_URL with your provider's OIDC URL
-  jwksUrl: "https://OIDC_AUTHORITY_URL/.well-known/jwks.json"
+  # replace OIDC_AUTHORITY_URL & OIDC_AUTHORITY_JWKS with your provider's OIDC base URL and JWKS sub-URL
+  jwksUrl: "https://OIDC_AUTHORITY_URL/OIDC_AUTHORITY_JWKS
```
stas-sbi: Should `OIDC_AUTHORITY_JWKS` be `OIDC_AUTHORITY_JWKS_PATH` or just:
```
jwksUrl: https://OIDC_AUTHORITY_JWKS_URI  # Example: "https://OIDC_AUTHORITY_URL/.well-known/jwks.json"
```
?

~Also I'm curious what OIDC providers have this path different from `/.well-known/jwks.json`?~ This is not a standard path in any way

SLoeuillet: OIDC_AUTHORITY_JWKS is just sub-URL after OIDC_AUTHORITY_URL

Okta is one.

But to be frank, the code should get the JWKS endpoint from well-known/openid-configuration endpoint 

```
wget https://kaiko-prod.okta.com/oauth2/ausmq0zn6eZMgabPv417/.well-known/openid-configuration -O oidc_okta > /dev/null 2>&1 | cat oidc_okta | python -m json.tool | egrep jwks_uri
"jwks_uri": "https://kaiko-prod.okta.com/oauth2/ausmq0zn6eZMgabPv417/v1/keys",
```

SLoeuillet: "jwks_uri": "https://kaiko-prod.okta.com/oauth2/ausmq0zn6eZMgabPv417/v1/keys",

SLoeuillet: Actually, templating existing values, I have to first replace
```
https://OIDC_AUTHORITY_URL/.well-known/jwks.json=https://kaiko-prod.okta.com/oauth2/ausmq0zn6eZMgabPv417/v1/keys
```

then

```
OIDC_AUTHORITY_URL=https://kaiko-prod.okta.com/oauth2/ausmq0zn6eZMgabPv417/
```

apps/app/src/pack/examples/sv-helm/validator-values.yaml
```diff
@@ -13,18 +13,17 @@ scanAddress: "TRUSTED_SCAN_URL"
...
 # TRUSTED_SINGLE_SEQUENCER_END
 
-# Replace OPERATOR_WALLET_USER_ID with the user id in your IAM that you want to use to log into
+# Replace OPERATOR_WALLET_USER_IDS with the user id in your IAM that you want to use to log into
 # the wallet as the SV party. Note that this should be the full user id, e.g., ``auth0|43b68e1e4978b000cefba352``
 # not only the suffix ``43b68e1e4978b000cefba352``:
 # You can specify multiple user ids if you want multiple users to be able to log into your wallet.
-validatorWalletUsers:
-  - "OPERATOR_WALLET_USER_ID"
+validatorWalletUsers: [OPERATOR_WALLET_USER_IDS]
```
stas-sbi, line 20: I'm afraid this could be confusing, it's not very clear what form the replacement for `OPERATOR_WALLET_USER_IDS` should have for multiple users and that each user id should be quoted to prevent YAML from trying to parse it as a non-string.

SLoeuillet: I can add an example for multi user

apps/app/src/pack/examples/sv-helm/participant-values.yaml
giner: > Okta is one.

Right, I forgot we modifying this for Keycloak as well.

giner: > But to be frank, the code should get the JWKS endpoint from well-known oidcconfig endpoint

agreed

docs/src/validator_operator/validator_helm.rst
giner: > Well, it works that way with Okta

That probably means that Okta has a user id associated the client matching the client id. This is different in case of Keycloak and could be different for other providers

> The "@Clients" suffix is 100% Auth0 specific

Yes, it is. CLIENT_ID is a client ID while CLIENT_ID@clients is a user ID in case of Auth0

apps/app/src/pack/examples/sv-helm/validator-values.yaml
SLoeuillet: Noted the user_id should be quoted, gave example of having multiple user_ids

docs/src/validator_operator/validator_helm.rst
SLoeuillet: That's why a new variable is necessary here. Current doc example is too Auth0 specific. I had to discover those differences while testing with Okta. KeyCloack & EntraID & other OIDC users had their own pain

giner: I agree with the variable, my concern is the comment

SLoeuillet: Well, the issue is that don't know how to phrase that better. And I don't know changes required by KeyCloack/EntraId/others. Could be good to have input on what people using them had to change

giner: My input above is from our experience (we use keycloak). The case where client id and user id match is implementation specific, the comment saying this would be misleading

SLoeuillet: @giner what would you put instead ? 
that it is a CLIENT_ID (and with Auth0, they end by @clients) ?
That's it is an OIDC sub ?

giner: How about this?
```suggestion
VALIDATOR_LEDGER_API_USER          ID of the user associated with the client VALIDATOR_CLIENT_ID (value of `sub` in the access token)
```

moritzkiefer-da: I agree with Stas concerns and the suggestion makes sense to me.

apps/app/src/pack/examples/sv-helm/validator-values.yaml
```diff
@@ -13,18 +13,17 @@ scanAddress: "TRUSTED_SCAN_URL"
...
+# You can specify multiple user ids if you want multiple users to be able to log into your wallet : ``"userid1", "userid2"``
+validatorWalletUsers: [OPERATOR_WALLET_USER_IDS]
 auth:
   # replace OIDC_AUTHORITY_VALIDATOR_AUDIENCE with the audience of your choice
   audience: "OIDC_AUTHORITY_VALIDATOR_AUDIENCE"
 
-  # replace OIDC_AUTHORITY_URL with your provider's OIDC URL
-  jwksUrl: "https://OIDC_AUTHORITY_URL/.well-known/jwks.json"
+  # replace OIDC_AUTHORITY_URL & OIDC_AUTHORITY_JWKS_SUB_URL with your provider's OIDC base URL and JWKS sub-URL
+  jwksUrl: "https://OIDC_AUTHORITY_URL/OIDC_AUTHORITY_JWKS_SUB_URL
```
moritzkiefer-da, line 26: why not just add a single parameter for the JWKS url instead of trying to combine it from two? I'm not sure this adds much.

apps/app/src/pack/examples/sv-helm/validator-values.yaml
moritzkiefer-da: I'm not quite sure I'm sold on this. The previous example made it very clear what you have to fill in and for the common case of one user makes sense. Now we're replacing that by asking users to under how to do a somewhat confusing string replacement.

Why can't you just overwrite this from a separate values file instead of trying to template this?

apps/app/src/pack/examples/sv-helm/validator-values.yaml
SLoeuillet: Well, I can change OIDC_AUTHORITY_JWKS_SUB_URL to OIDC_AUTHORITY_JWKS_URL and it 
would contain the OIDC_AUTHORITY_URL.
It is just that OIDC_AUTHORITY_JWKS_SUB_URL won't change between ENVs (OIDC provider specific) but OIDC_AUTHORITY_URL would (as no one should re-use the same auth server between ENVs)

apps/app/src/pack/examples/sv-helm/validator-values.yaml
SLoeuillet: Because if I do that, I have no garanty of which content would be taken into account, my separate file or the provided default values. I would have to edit the provided value file to comment that block. What I try to avoid is any manual modification of the provided fles. I want to be able to automate as much as possible and cover single and multi-user

moritzkiefer-da: > Because if I do that, I have no garanty of which content would be taken into account, my separate file or the provided default values

huh? later values overwrite earlier ones. So just specify your values last.

SLoeuillet: Works if you use helm manualy, specifying the values file names in a specific order. 
My target is a CI that deploys everything, taking all relevant *values*.yaml files, with no specific order

moritzkiefer-da: what CI are you using? It seems odd to me that you have no control over order.

nicu-da:
@SLoeuillet There are still a number of comments open, do you want to rebase and fix the existing review to push this through?

Issue #340, opened 2025-04-21
Display Transfer Offers that have not been accepted


waynecollier-da:
In the Transfers page of the Splice Wallet UI, list all transfer offers that have not yet been accepted, or that have been accepted but not completed.

PR #441, opened 2025-06-05
feat: show list pending offer in Transfers page

AngelHackAthur:
## Purpose

Display transfer offers that have not been accepted.

## Key Changes

- Added component to render unaccepted transfer offers
- Updated API integration to fetch unaccepted offers

## Reference

Closes [#340]
(https://github.com/hyperledger-labs/splice/issues/340)


moritzkiefer-da:
@AngelHackAthur your commit history looks a bit messed up, you have 96 commits in this PR where only the last one actually seems new. Can you please fix it

apps/wallet/frontend/src/components/PendingOffers.tsx
```diff
@@ -0,0 +1,192 @@
...
+  sender: string;
+  expiresAt: string;
+  isTokenStandard: boolean;
+};
+export const PendingOffers: React.FC = () => {
+  const [offers, setOffers] = useState<WalletTransferOffer[]>([]);
+  const amuletPriceQuery = useAmuletPrice();
+  const primaryPartyId = usePrimaryParty();
+
+  const toWalletTransferOffer = useCallback(
```
fayi-da: The logic in this function exists in the `TransferOffers` component. Rather than duplicating it here, we should extract into a util and reuse in both components. I believe the only difference is the predicate for the filter. Can we parameterise that?

apps/wallet/frontend/src/components/PendingOffers.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  const transferOfferContractsQuery = useTransferOffers(amuletPriceQuery.data);
+  const { data: transferOfferContracts } = transferOfferContractsQuery;
+  const tokenStandardTransfersQuery = useTokenStandardTransfers();
+  const { data: tokenStandardTransferContracts } = tokenStandardTransfersQuery;
+  const amuletPrice = amuletPriceQuery.data;
+
+  useMemo(() => {
+    if (transferOfferContracts && tokenStandardTransferContracts && amuletPrice) {
+      const allTransfers: PartialWalletTransferOffer[] = transferOfferContracts
```
fayi-da: Similar to my comment above, the logic for all transfers here is exactly the same as that in `TransferOffers`. We should extract this into a util and reuse in both components instead.

apps/wallet/frontend/src/components/PendingOffers.tsx
```diff
@@ -0,0 +1,192 @@
...
+    tokenStandardTransfersQuery.isLoading;
+  const isError =
+    amuletPriceQuery.isError ||
+    transferOfferContractsQuery.isError ||
+    tokenStandardTransfersQuery.isError;
+
+  return (
+    <Stack spacing={4} direction="column" justifyContent="center" id="transfer-offers">
+      <Typography mt={6} variant="h4">
+        Action Needed{' '}
```
fayi-da: There's no action needed for these items so the header here should say "Pending Offers"

apps/wallet/frontend/src/components/PendingOffers.tsx
```diff
@@ -0,0 +1,192 @@
...
+      )}
+    </Stack>
+  );
+};
+
+interface TransferOfferProps {
+  transferOffer: WalletTransferOffer;
+}
+
+export const TransferOfferDisplay: React.FC<TransferOfferProps> = props => {
```
fayi-da: Can we rename this to `PendingOfferDisplay` or something better that explicitly describe what this component is? Same for all the ids and classnames within the component so it's easier to distinguish in tests.

apps/wallet/frontend/src/components/PendingOffers.tsx
```diff
@@ -0,0 +1,192 @@
...
+          direction: 'row',
+          justifyContent: 'space-between',
+          alignItems: 'center',
+        }}
+      >
+        <ArrowCircleLeftOutlined fontSize="large" />
+        <Stack direction="row" alignItems="center">
+          <Stack direction="column">
+            <BftAnsEntry
+              partyId={offer.senderId}
```
fayi-da: Showing the sender here will not be useful. If Alice is logged in, it'll always say Alice. The recipient will be a more useful Party to display here.

apps/wallet/frontend/src/components/PendingOffers.tsx
```diff
@@ -0,0 +1,192 @@
...
+                currency={offer.convertedCurrency.currency}
+              />{' '}
+              @ {offer.convertedCurrency.amuletPriceToShow.toString()}{' '}
+              {config.spliceInstanceNames.amuletNameAcronym}/USD
+            </>
+          </Typography>
+        </Stack>
+
+        <Stack direction="row" alignItems="center" spacing={2}>
+          <Button variant="outlined" size="small" className="transfer-offer-accept">
```
fayi-da: There's no need to have the button here. All items on this list are pending and there's no action to be taken.

fayi-da (changes requested):
Thanks for the PR. I've left some comments on the code. 

For frontend changes, it's very helpful to include a screenshot in the PR to assist with the review process. Please add one after addressing comments.

Also, let’s add some tests for this please. There are some examples in the `__tests__` directory of the wallet using vitest.

Let me know if you have any questions or need any clarification!

AngelHackAthur:
Hi @fayi-da  Thank you for reviewing!
I've updated the code based on the feedback from the preview process. Please let me know if there's anything else I should adjust.

AngelHackAthur:
> > Hi @waynecollier-da The "pending acceptance" feature has been implemented as follows: When alice_wallet_user transfers to bob_wallet_user, a pending item is shown in Alice's wallet (see the screenshot below). image Once bob_wallet_user either accepts or rejects, the corresponding pending entry disappears from Alice’s view. I'd appreciate your feedback on this — does this approach look good to you, or would you suggest any improvements or additions?
> 
> Please show the recipient, not the sender, in each pending transfer line. The user knows they are the sender; to make it precise you can put that in the header: Pending Offers: Sender alice. But the helpful thing is to know at a glance who hasn't accepted the transfer yet.

Thank you for the clarification!
I've updated the code so that each pending transfer now shows the recipient instead of the sender.
Attaching a screenshot for your reference. Please let me know if anything needs further adjustment.
image

apps/wallet/frontend/src/components/TransferOffers.tsx
```diff
@@ -186,23 +209,31 @@ export const TransferOfferDisplay: React.FC<TransferOfferProps> = props => {
...
+                size="small"
+                onClick={() => reject(offer.contractId)}
+                className="transfer-offer-reject"
+              >
+                Reject
+              </Button>
+            </>
+          ) : (
+            <Typography variant="pill" className="pending-offer">
+              Pending acceptance
```
fayi-da: We don't need this as it's constant across all items in this list. The title of the section already indicates that each item in the list is pending acknowledgement(acceptance/rejection).

apps/wallet/frontend/src/components/TransferOffers.tsx
```diff
@@ -110,11 +121,12 @@ export const TransferOffers: React.FC = () => {
     amuletPriceQuery.isError ||
     transferOfferContractsQuery.isError ||
     tokenStandardTransfersQuery.isError;
+  const heading = mode === 'received' ? 'Pending Offers ' : 'Action Needed ';
```
fayi-da: We should flip this, see my comment below about the modes.

apps/wallet/frontend/src/components/TransferOffers.tsx
```diff
@@ -46,7 +52,9 @@ export const TransferOffers: React.FC = () => {
       amuletPrice: BigNumber
     ): Promise<WalletTransferOffer[]> => {
       return items
-        .filter(item => item.sender !== primaryPartyId)
+        .filter(item =>
+          mode === 'received' ? item.sender === primaryPartyId : item.sender !== primaryPartyId
```
fayi-da: We should flip this, see my comment below about the modes.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletTransfersFrontendIntegrationTest.scala
```diff
@@ -122,6 +122,10 @@ abstract class BaseWalletTransfersFrontendIntegrationTest
                 offerCard.childElement(className("transfer-offer-sender"))
               ) should matchText(expectedAns(aliceUserParty, aliceAnsName))
 
+              seleniumText(
+                offerCard.childElement(className("transfer-offer-received"))
+              ) should matchText(expectedAns(bobUserParty, bobAnsName))
+
```
fayi-da, line 128: Does this pass? You shouldn't see both the sender and receiver at the same time according to the rendering logic below

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletTransfersFrontendIntegrationTest.scala
```diff
@@ -122,6 +122,10 @@ abstract class BaseWalletTransfersFrontendIntegrationTest
                 offerCard.childElement(className("transfer-offer-sender"))
               ) should matchText(expectedAns(aliceUserParty, aliceAnsName))
 
+              seleniumText(
+                offerCard.childElement(className("transfer-offer-received"))
```
fayi-da, line 126: `transfer-offer-receiver`

fayi-da (changes requested):
We need a separate test that simulates :
- Alice creates 2 transfer offers to Bob
- Bob logs in and goes to Transfer page
- Bob sees both transfer offers
- Each item in the list renders expected details

If there's a way to add these checks to the existing test, that's fine too.

fayi-da:
> I'll let you decide whether to go ahead with this format for now, or push for a cleaner transaction list format rather than separate UI boxes for each transaction.

I'm ok with the separate boxes for now. It's low friction and consistent with how we show the other transfer offers.

fayi-da:
Thanks @AngelHackAthur , you're still missing this though:

> We need a separate test that simulates :
> 
> Alice creates 2 transfer offers to Bob
> Alice logs in and goes to Transfer page
> Alice sees both pending transfer offers
> Each item in the list renders expected details
> If there's a way to add these checks to the existing test, that's fine too.

PR #1014, opened 2025-06-06
Support running tests with an older initial package version set

### LINKED ISSUE #452: Improve our testing to catch missing Daml version guards
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#19814). Original author: @moritzkiefer-da ]

Daml upgrading requires that clients do not:

1. Exercise choices before the corresponding versions are vetted.
2. Set new optional fields to Some before the corresponding versions are vetted.

This is accomplished through the feature support endpoint to first check the version client side.

However, it is very easy to forget to add such a check when you write code and then code might blow up until the new versions are vetted. Sometimes ciupgrade/ciperiodic help to catch them but it is unrealistic to test everything there. We also have select integration tests but you're obviously not gonna write one if you forget about the check being needed.

The best option I can come up with is that we have a dedicated job (run e.g. daily but we can debate frequency, should definitely be possible to run it on a PR on demand ofc) that runs all integration tests but sets the initial package version to the one from the latest release instead of the current one.

We generally test all functionality in integration tests so this gives us pretty decent coverage.

To make this work in practice, I would imagine:

1. The CI job sets an environment variable (or several) to switch to the old versions.
2. We pick that up in the environment setup and use it to change SV1's initial package config to the latest DARs in the LATEST_RELEASE version.
3. We add [scalatest tags](https://www.scalatest.org/user_guide/tagging_your_tests) to fully exclude some tests, e.g., the BoostrapPackageConfigTest makes no sense to run here.
4. We add scalatest tags like Amulet_0_1_8, DsoGovernance_0_1_12. Tests that should only run with versions >= those are tagged like this. The CI job then excludes tests with tags that have versions higher than what is in LATEST_RELEASE.

moritzkiefer-da:
to catch missing Daml version guards

fixes #452

[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


.github/actions/tests/scala_test/action.yml
```diff
@@ -133,6 +136,35 @@ runs:
       with:
         cache_version: 4
 
+    - name: Set Daml package versions
```
moritzkiefer-da, line 140: This is not as precise as we probably should be: It relies on the version shipped to mainnet, not the versions vetted on mainnet. The latter is a bit of a mess since we also need a splitwell version which is not in the packageConfig, definitely solvable but annoying. So for now I'd suggest we keep it as is and open a separate issue for making the version we test against it more precise. This still gives us a two week compatibility window at a minimum which is much better than what we had so far.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/AmuletConfigUtil.scala
```diff
@@ -37,27 +36,34 @@ trait AmuletConfigUtil extends TestCommon {
...
   )(implicit
       env: SpliceTests.SpliceTestConsoleEnvironment
   ): splice.amuletconfig.AmuletConfig[splice.amuletconfig.USD] = {
-    val activeSynchronizerId =
-      AmuletConfigSchedule(amuletRules)
-        .getConfigAsOf(env.environment.clock.now)
-        .decentralizedSynchronizer
-        .activeSynchronizer
-    val domainFeesConfig = defaultSynchronizerFeesConfig
-    defaultAmuletConfig(
```
moritzkiefer-da, line 50: this was just buggy, this used the default config instead of the existing one.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/ConfigScheduleUtil.scala
```diff
@@ -40,27 +40,31 @@ trait ConfigScheduleUtil extends TestCommon {
...
-      domainFeesConfig.extraTrafficPrice.value,
-      domainFeesConfig.minTopupAmount.value,
-      domainFeesConfig.baseRateBurstAmount.value,
-      domainFeesConfig.baseRateBurstWindow,
-      domainFeesConfig.readVsWriteScalingFactor.value,
-      holdingFee = holdingFee,
-      nextSynchronizerId = nextSynchronizerId,
+    val existingAmuletConfig = AmuletConfigSchedule(amuletRules)
+      .getConfigAsOf(env.environment.clock.now)
+    val existingTransferConfig = existingAmuletConfig.transferConfig
```
moritzkiefer-da, line 48: this was just buggy, this used the default config instead of the existing one.

canton/community/app-base/src/main/scala/com/digitalasset/canton/admin/api/client/commands/ParticipantAdminCommands.scala
```diff
@@ -193,7 +193,7 @@ object ParticipantAdminCommands {
                 if (dar.description.isEmpty)
                   PathUtils.getFilenameWithoutExtension(Path.of(filename))
                 else dar.description
-              _ = logger.info(s"Sending upload doar for ${descriptionOrFilename}")(
+              _ = logger.info(s"Sending upload dar for ${descriptionOrFilename}")(
```
moritzkiefer-da, line 196: doar

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SplitwellTestUtil.scala
```diff
@@ -176,7 +176,14 @@ trait SplitwellTestUtil extends TestCommon with WalletTestUtil with TimeTestUtil
   ): AcceptedAppPayment.ContractId = {
     senderSplitwell.initiateTransfer(key, receiverAmounts)
     val request = eventually()(getSingleAppPaymentRequest(senderWallet))
-    senderWallet.acceptAppPaymentRequest(request.contractId)
+    val (cid, _) = actAndCheck(
+      "Sender accepts payment request",
+      senderWallet.acceptAppPaymentRequest(request.contractId),
+    )(
+      "Splitwell completes payment",
```
moritzkiefer-da, line 183: not sure this actually mattered in the end but it makes debugging easier at least to synchronize on this

moritzkiefer-da:
I had a green build on this and only changed the from triggering on the PR to triggering periodically in the last commit https://github.com/hyperledger-labs/splice/pull/1014/commits/0a13a13dfa9aeff57adae7c5768915d408accdef

moritzkiefer-da:
@nicu-da @isegall-da could I get a review on this please

.github/actions/tests/scala_test/action.yml
```diff
@@ -83,6 +83,9 @@ inputs:
   failure_notifications_slack_channel:
     description: "Slack channel for failure notifications (for failure notifications)"
     required: true
+  daml_base_version:
```
nicu-da, line 86: Nit: can we add a description as to what is this used? Makes easier compared to having to read the action

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/plugins/TokenStandardCliSanityCheckPlugin.scala
```diff
@@ -42,77 +46,89 @@ class TokenStandardCliSanityCheckPlugin(
...
-            val validator = environment.validators.local
-              .find(_.config.adminApi.port.unwrap == wallet.config.adminApi.url.effectivePort)
-              .getOrElse(
-                throw new RuntimeException(s"Failed to find validator for wallet of $ledgerApiUser")
-              )
-            // JSON API port = gRPCport + 1000
-            val participantHttpApiPort =
-              validator.participantClient.config.ledgerApi.port.unwrap + 1000
+        val sv1 = environment.svs.local.find(_.name == "sv1").value
+        val amuletVersion = inside(sv1.config.onboarding) {
```
nicu-da, line 59: Nit: any reason why we don't read the current vetted version from the topology state?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BootstrapPackageConfigIntegrationTest.scala
```diff
@@ -42,6 +42,7 @@ import java.time.Instant
 import java.time.temporal.ChronoUnit
 import scala.jdk.CollectionConverters.*
 
+@org.lfdecentralizedtrust.splice.util.scalatesttags.NoDamlCompatibilityCheck
```
nicu-da, line 45: I LOVE IT! 

Can we adda readme on how to add new tags and when to use them?  (can be done later in a different PR)

scripts/initial-package-config.py
nicu-da, line 1: Can we add a comment as to what it does?

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/util/ScalaTestTags.scala
```diff
@@ -0,0 +1,10 @@
+package org.lfdecentralizedtrust.splice.util
+
+import org.scalatest.Tag
+
+object Tags {
```
nicu-da, line 5: Can we add a comment for each tag as to why they exist and when they're used?

nicu-da (approved):
Really nice! Thank you!

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BootstrapPackageConfigIntegrationTest.scala
moritzkiefer-da: did you see the readme I added?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/plugins/TokenStandardCliSanityCheckPlugin.scala
moritzkiefer-da: I find this a bit easier to debug. I had quite a few cases where a prior test messed up the vetting state by vetting too much but the next one ran with an older initial package config change or something like that. This isn't gonna fix it but at least it makes it easier to figure out what is going wrong™

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BootstrapPackageConfigIntegrationTest.scala
nicu-da: In DEVELOPMENT.md or something else that I'm missing?
I'm seeing there instructions to tag a test if it doesn't work with this setup but I don't see anything about how are tags used, when to add a new type of tag and how to make the test run aware of it.

.github/actions/tests/scala_test/action.yml
moritzkiefer-da: added

scripts/initial-package-config.py
moritzkiefer-da: added

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/util/ScalaTestTags.scala
moritzkiefer-da: added

meiersi-da (commented):
Very nice work! Thank you @moritzkiefer-da .

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BootstrapPackageConfigIntegrationTest.scala
moritzkiefer-da: extended

.github/actions/tests/scala_test/action.yml
moritzkiefer-da: https://github.com/hyperledger-labs/splice/issues/1101

PR #1017, opened 2025-06-06
[ci] Update pulumi k8s operator to v2

nicu-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


cluster/deployment/scratchnetd/.envrc.vars
```diff
@@ -35,6 +38,6 @@ export APPROVE_SV_RUNBOOK=true
 
 # TODO(#14679): Remove
```
moritzkiefer-da, line 39: why do we have this stuff at all?

cluster/deployment/scratchnetd/config.yaml
```diff
@@ -13,3 +13,24 @@ validator1:
     maxDuration: "5m"
     retention: "1h" # Keep 1h
   deduplicationDuration: 30m
+synchronizerMigration:
```
moritzkiefer-da: presumably we don't want to commit this

cluster/pulumi/common/src/operator/stack.ts
```diff
@@ -152,6 +206,86 @@ export function createStackCR(
...
           // retry if the stack is locked by another operation
           retryOnUpdateConflict: true,
+          // https://github.com/pulumi/pulumi-kubernetes-operator/blob/v2.1.0/docs/stacks.md#stackspecworkspacetemplatespec
+          workspaceTemplate: {
+            metadata: {
+              name: `${name.replaceAll('.', '-')}`,
+              namespace: namespaceName,
+            },
+            spec: {
+              image: `pulumi/pulumi:${semver.gt(pulumiVersion, minimumPulumiVersionRequired) ? pulumiVersion : minimumPulumiVersionRequired}-nonroot`,
```
moritzkiefer-da, line 216: does this still go through our cache?

moritzkiefer-da (commented):
Thanks! Do you have a link to a successful HDM test?

I am slightly nervous about just replacing this. It seems ideally we would get a few weeks of experience running this in internal clusters instead of just forcibly rolling it out to dev/test/mainnet with the next upgrade.

How hard is it to keep this is an option that we only enable on internal clusters for now?

nicu-da:
> Thanks! Do you have a link to a successful HDM test?
> 

I have [one](https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/13012/workflows/84169122-2ed6-464f-9953-5d5a41e24027/jobs/56307) where the steps that were touched by this did succeed but the full workflow failed because it was during the switch to splice. 
I will trigger a new one though to test it out with the new repo. 

> I am slightly nervous about just replacing this. It seems ideally we would get a few weeks of experience running this in internal clusters instead of just forcibly rolling it out to dev/test/mainnet with the next upgrade.
> 
> How hard is it to keep this is an option that we only enable on internal clusters for now?

It's actually fairly simple as we can keep the operator deployment on an old reference  if needed. Though I wouldn't postpone it more than required tbh. Once it's merged to main updating CILR should provide the feedback we need in a few days of testing.

moritzkiefer-da:
> It's actually fairly simple as we can keep the operator deployment on an old reference if needed. Though I wouldn't postpone it more than required tbh. Once it's merged to main updating CILR should provide the feedback we need in a few days of testing.

fair, I guess we could just revert if needed reasonably easily as well.

cluster/deployment/scratchnetd/.envrc.vars
nicu-da: It does not make sense indeed, let me clean it up

cluster/pulumi/common/src/operator/stack.ts
nicu-da: My understanding is that yes as we basically cache everything as it always goes through that proxy. @isegall-da is that correct?

.github/workflows/cluster_tests.yml
```diff
@@ -39,5 +39,5 @@ jobs:
         run: |
           set -euo pipefail
           result=$(curl -sSfL -H "Authorization: Bearer ${{ steps.auth.outputs.id_token }}" \
-            "${{ secrets.CLUSTER_TEST_INVOKER_URL }}?workflow=${{ inputs.workflow }}&splice-git-ref=${{ inputs.sha }}")
+            "${{ secrets.CLUSTER_TEST_INVOKER_URL }}?workflow=${{ inputs.workflow }}&splice-git-ref=${{ inputs.sha }}&public-splice=true")
```
isegall-da: This seems outdated. We removed this, didn't we?

nicu-da: Yes, messed up the merge. Will re-remove it

cluster/pulumi/common/src/operator/stack.ts
isegall-da: I don't think so, I think only the DinD is configured to use the cache, so that images we pull in "docker pull" in tests are pulled from there.
But I'm not sure we care. How often do we expect this to be re-pulled per cluster? (the rate limit is per IP AFAIK)

isegall-da: We needed the cache for jobs in CI that were pulling multiple images per each CI run, hence hit rate limits on docker.io

cluster/pulumi/operator/src/operator.ts
```diff
@@ -37,13 +24,13 @@ export const operator = new k8s.helm.v3.Release(
...
-    version: '0.7.3',
+    version: '2.1.0',
     namespace: namespace.ns.metadata.name,
     values: {
       resources: {
         limits: {
-          cpu: 5,
-          memory: config.optionalEnv('OPERATOR_MEMORY_LIMIT') || '20G',
+          cpu: 2,
+          memory: config.optionalEnv('OPERATOR_MEMORY_LIMIT') || '4G',
```
isegall-da, line 33: This is now per stack, which is why we need 80% less, right?

isegall-da (approved):
LGTM, thank you!

cluster/pulumi/operator/src/operator.ts
nicu-da: Yes, will update it as well after we deploy to prod and see actual usage but the main container should be fairly light

cluster/pulumi/common/src/operator/stack.ts
moritzkiefer-da: >  But I'm not sure we care. How often do we expect this to be re-pulled per cluster? (the rate limit is per IP AFAIK)

I guess ciperiodic might be the most with one pod per stack? Hopefully still infrequent enough

isegall-da: Rate limit seems to be 100 per IP address per 6 hours. I think we should be ok. Famous last words...

isegall-da: And I think that k8s has some internal cache too, doesn't it? (didn't help with docker pull in DinD, but will for the image for a pod)

PR #1026, opened 2025-06-06
Retry npm commands on TS sanity plugin

OriolMunoz-da:
No changes since the PR in the private repo

(copied description:)

What I found:
- Martin's logs exit with status code 139 and absolutely no logs from the command itself (only the npm run line)
- Itai's log straight up show a native stack trace

So this seems to be some issue with node, or some native library that we use, or who knows. I'm hoping that a retry would fix the issue, but of course I can't know that because it's not even reproducible 🤷
Some online docs pointed at maybe node_modules being corrupt somehow, but slapping an npm rebuild or whatever doesn't seem great either... why can't npm just work...


moritzkiefer-da (approved):
i still hate it

PR #1069, opened 2025-06-10
Faster getLastImportUpdateId

rautenrieth-da:
We have a partial index
```
create index updt_hist_tran_hi_mi_ui_import_updates
    on update_history_transactions (history_id, migration_id, update_id)
    where record_time = -62135596800000000
```

so it should be fast to fetch all import updates, but that still leads to sequential scans over all creates (because the planner thinks we have thousands of create events per update on average).

New query plan is
```
Aggregate (cost=19246.47..19246.48 rows=1 width=32) (actual time=72.576..72.578 rows=1 loops=1)
--
-> Index Scan using updt_hist_crea_hi_mi_rt on update_history_creates c (cost=0.57..19220.99 rows=10191 width=142) (actual time=35.798..63.413 rows=14464 loops=1)
Index Cond: ((history_id = 1) AND (migration_id = 2) AND (record_time = '-62135596800000000'::bigint))
Planning Time: 12.030 ms
Execution Time: 72.734 ms
```


OriolMunoz-da (approved):
thanks! that denormalization paying dividends :zany_face:

Issue #1061, opened 2025-06-10
Scan's `/v0/holdings/summary` will return a wrong summary if there are too many Amulet contracts - and log a WARN for the operator


OriolMunoz-da:
It should fail the request instead.

`/v0/holdings/summary` loads up to 1000 Amulets (locked or not) into memory to compute the summary.
If there are more than those in the database, a WARN is logged, but the computation carries on. This means that in the following situations the user will get a wrong summary:
- A party owns more than 1000 Amulets (unlikely to happen, given Amulet merging)
- Many parties are requested

Additionally, due to the same behavior, for a request where no parties are requested, all amulets (up to 1001) are loaded into memory and then nothing is returned, but a WARN is logged:
``` 
POST https://scan.sv-2.global.canton.network.digitalasset.com/api/scan/v0/holdings/summary
Accept: application/json
Content-Type: application/json

{
    "migration_id": 2,
  "record_time": "2025-06-10T15:00:00Z",
  "owner_party_ids": []
}
```
returns
```
{
  "record_time": "2025-06-10T15:00:00Z",
  "migration_id": 2,
  "computed_as_of_round": 48615,
  "summaries": []
}
```
which is correct, but also logs a WARN, and it shouldn't.

Thus, the request should fail (4xx):
- When there are no parties provided, as the summary returned will be empty anyway
- When the number of amulets over which the summary is being computed is greater than the HardLimit, as then it won't be properly computed

PR #1072, opened 2025-06-11
Fix getHoldingsSummary returning bad counts when too many contracts

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/1061


apps/scan/src/main/openapi/scan.yaml
OriolMunoz-da, line 1: technically a breaking change but:
- no changes to scala code that i've seen
- it wasn't doing anything useful anyway

isegall-da: The only change is that we reject empty party ID lists, right? Seems fine. But let's mention it in the release notes.

isegall-da (approved):
Nice, thank you

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/AcsSnapshotStoreTest.scala
```diff
@@ -543,6 +551,39 @@ class AcsSnapshotStoreTest
...
+          )
+          _ <- store.insertNewSnapshot(None, DefaultMigrationId, timestamp1)
+          _result <- store.getHoldingsState(
+            DefaultMigrationId,
+            timestamp1,
+            None,
+            HardLimit.tryCreate(1),
+            NonEmptyVector.of(owner),
+          )
+        } yield fail("should not get here, call should've failed")).failed.futureValue shouldBe a[
```
rautenrieth-da, line 582: Why do you need the `fail()` part, instead of something like `yield succeed`? Would `.failed.futureValue` not fail the test with a useful description if it didn't throw an exception?

rautenrieth-da (approved):
Thanks!

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/AcsSnapshotStoreTest.scala
OriolMunoz-da: I guess it doesn't matter much what I put in the yield branch - seems more explicit than just a `succeed`

Issue #451, opened 2025-06-05
allow creating an allocation with user-specified parameters


canton-network-da:
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#19815). Original author: @OriolMunoz-da ]

PR #1082, opened 2025-06-11
[wip] Allocations create

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/451

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
@@ -1285,11 +1285,14 @@ object SpliceLedgerConnection {
     * @param parties       : list of parties whose method calls should be considered distinct,
     *                      e.g., "Seq(directoryProvider)"
     * @param discriminator : additional discriminator for method calls,
-    *                      e.g., "digitalasset.splice" in case of deduplicating directory entry requests relating to directory name "digitalasset.splice". Beware of naive concatenation
+    *                      e.g., Seq("digitalasset.splice") in case of deduplicating directory entry requests relating to directory name "digitalasset.splice". Beware of naive concatenation
     *                      strings for discriminators. Always ensure that the encoding is injective.
     */
-  case class CommandId(methodName: String, parties: Seq[PartyId], discriminator: String = "")
-      extends PrettyPrinting {
+  case class CommandId(
+      methodName: String,
+      parties: Seq[PartyId],
+      discriminator: Seq[String] = Seq.empty,
+  ) extends PrettyPrinting {
     require(!methodName.contains('_'))
 
     override def pretty: Pretty[this.type] =
@@ -1304,13 +1307,17 @@ object SpliceLedgerConnection {
           parties.length.toString
         ) // prepend length to avoid suffixes interfering with party mapping, e.g., otherwise we have
         // CommandId("myMethod", Seq(alice), "bob").commandIdForSubmission == CommandId("myMethod", Seq(alice,bob), "").commandIdForSubmission
-        .appended(discriminator)
+        .appendedAll(discriminator.flatMap(str => Seq(str.length.toString, str)))
         .mkString("/")
       // Digest is not thread safe, create a new one each time.
       val hashFun = MessageDigest.getInstance("SHA-256")
       val hash = hashFun.digest(str.getBytes("UTF-8")).map("%02x".format(_)).mkString
       s"${methodName}_$hash"
     }
+  }
+  object CommandId {
+    def apply(methodName: String, parties: Seq[PartyId], discriminator: String): CommandId =
+      CommandId(methodName, parties, Seq(discriminator))
   }
 
   def decodeExerciseResult[T](
--- a/apps/wallet/frontend/src/components/CreateAllocation.tsx
+++ b/apps/wallet/frontend/src/components/CreateAllocation.tsx
@@ -1,0 +1,301 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useState } from 'react';
+import { useWalletClient } from '../contexts/WalletServiceContext';
+import { useMutation } from '@tanstack/react-query';
+import {
+  AllocateAmuletRequest,
+  AllocateAmuletRequestSettlement,
+  AllocateAmuletRequestTransferLeg,
+} from 'wallet-openapi';
+import { Alert, Button, Card, CardContent, Stack, TextField, Typography } from '@mui/material';
+import { DisableConditionally } from '@lfdecentralizedtrust/splice-common-frontend';
+import BftAnsField from './BftAnsField';
+import AmountInput from './AmountInput';
+import { Add, Remove } from '@mui/icons-material';
+import { DateTimePicker } from '@mui/x-date-pickers/DateTimePicker';
+import dayjs from 'dayjs';
+
+const CreateAllocation: React.FC = () => {
+  const { createAllocation } = useWalletClient();
+  const [error, setError] = useState<object | null>(null);
+  const [allocation, setAllocation] = useState<PartialAllocateAmuletRequest>(emptyForm());
+  const validated = validatedForm(allocation);
+  const createAllocationMutation = useMutation({
+    mutationFn: async () => {
+      return validated && (await createAllocation(validated));
+    },
+    onSuccess: () => setError(null),
+    onError: error => {
+      console.error('Failed to submit allocation', error);
+      setError(error);
+    },
+  });
+
+  return (
+    <Stack mt={4} spacing={4} direction="column" justifyContent="center">
+      <Typography mt={6} variant="h4">
+        Create Allocation
+      </Typography>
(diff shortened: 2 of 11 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AllocationsFrontendIntegrationTest.scala
```diff
@@ -0,0 +1,165 @@
...
+      .truncatedTo(ChronoUnit.MINUTES)
+      .toInstant(ZoneOffset.UTC)
+    def wantedAllocation(requestedAt: Instant) = new AllocationSpecification(
+      new SettlementInfo(
+        validatorPartyId.toProtoPrimitive,
+        new SettlementReference("some_reference", Optional.empty),
+        requestedAt,
+        settleAndAllocateBefore,
+        settleAndAllocateBefore,
+//        new Metadata(java.util.Map.of("k1", "v1", "k2", "v2")),
```
moritzkiefer-da: leftover?, same below

OriolMunoz-da: PR description: I am currently testing, but since you're off tomorrow I opened PR without those while I work on it

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/WalletFrontendTestUtil.scala
```diff
@@ -241,9 +241,9 @@ trait WalletFrontendTestUtil extends WalletTestUtil { self: FrontendTestCommon =
       receiver.toProtoPrimitive,
     )
 
-    click on "create-offer-amulet-amount"
-    numberField("create-offer-amulet-amount").value = ""
-    numberField("create-offer-amulet-amount").underlying.sendKeys(transferAmount.toString())
+    click on "amulet-amount"
```
moritzkiefer-da: how badly are you gonna break ciupgrade/ciperiodic preflights by renaming this?

OriolMunoz-da: ...ugh. You're right, I'll make that ID configurable or something...

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AllocationsFrontendIntegrationTest.scala
OriolMunoz-da: nvm that was quick, done

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
```diff
@@ -1304,14 +1307,18 @@ object SpliceLedgerConnection {
           parties.length.toString
         ) // prepend length to avoid suffixes interfering with party mapping, e.g., otherwise we have
         // CommandId("myMethod", Seq(alice), "bob").commandIdForSubmission == CommandId("myMethod", Seq(alice,bob), "").commandIdForSubmission
-        .appended(discriminator)
+        .appendedAll(discriminator.flatMap(str => Seq(str.length.toString, str)))
```
moritzkiefer-da: I probably need more coffee but can you convince me that this never collides even without a separator between length and str if str can contain numbers?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AllocationsFrontendIntegrationTest.scala
```diff
@@ -0,0 +1,176 @@
...
+  private def createAllocation(sender: PartyId)(implicit
+      ev: SpliceTestConsoleEnvironment,
+      webDriver: WebDriverType,
+  ) = {
+    val validatorPartyId = aliceValidatorBackend.getValidatorPartyId()
+    val receiver = validatorPartyId
+    val settleAndAllocateBefore = LocalDateTime
+      .now()
+      .plusSeconds(3600)
+      .truncatedTo(ChronoUnit.MINUTES)
```
moritzkiefer-da, line 52: dumb question, why do you need to truncate?

moritzkiefer-da: ah because the UI only supports minutes? maybe add a comment

apps/wallet/frontend/src/components/CreateAllocation.tsx
```diff
@@ -0,0 +1,349 @@
...
+              </Button>
+            </DisableConditionally>
+          </Stack>
+        </CardContent>
+      </Card>
+    </Stack>
+  );
+};
+
+interface PartialAllocateAmuletRequest {
```
moritzkiefer-da, line 230: Can we not reuse the Daml type `AllocationSpecification` here?

apps/wallet/src/main/openapi/wallet-internal.yaml
```diff
@@ -1281,3 +1332,102 @@ components:
           type: boolean
         transfer_preapproval_description:
           type: boolean
+
+    AllocateAmuletRequest:
```
moritzkiefer-da, line 1336: this looks a bit dodgy to me. I would have expected we pass a Daml value here, why do we need to pass a whole bunch of individual fields?

OriolMunoz-da: There's two things:
- we'd have to define it as `type: object` and parse the json (which means the FE would be `object` and thus have no typecheck)
- note the absence of `sender` and `instrumentId`, as those are hardcoded in the BE to be the wallet's user party and Amulet

apps/wallet/frontend/src/components/CreateAllocation.tsx
OriolMunoz-da: no for the same reason as the other comment:

> note the absence of sender and instrumentId, as those are hardcoded in the BE to be the wallet's user party and Amulet

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
OriolMunoz-da: ugh, even worse, as you said on Slack:

> Note that you need to be careful that you don't change the hash for cases where only a single discriminator is passed. Backwards compatibility matters here since we need to dedup across upgrades

so this would break backwards compatibility.

We need to keep: discriminator of one item => just the item (previous behavior, untouched)

>  that this never collides even without a separator

but there is a separator? I'm not sure what you're asking for here

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AllocationsFrontendIntegrationTest.scala
OriolMunoz-da: (stumbled upon while writing IT test) it's not only that: Instant has more precision than daml seems to support:
- request: `2025-06-13T15:26:24.534755221Z`
- daml:::::: `2025-06-13T15:26:24.534755Z`

I assume this might be the reason why we use int64 in some cases ([example](https://github.com/hyperledger-labs/splice/blob/b2d78f62adbb565e5ac430349f665153d267ff45/apps/wallet/src/main/openapi/wallet-internal.yaml#L1206)) as opposed to date-time?
...though [we're not entirely consistent](https://github.com/hyperledger-labs/splice/blob/b2d78f62adbb565e5ac430349f665153d267ff45/apps/wallet/src/main/openapi/wallet-internal.yaml#L1055)

apps/wallet/frontend/src/components/CreateAllocation.tsx
moritzkiefer-da: Hm fair enough, I think I would have gone for the Daml field and just checked that the extra fields like the user match what is passed. Still seems like less duplication than what we have here. But don't feel very strongly about it.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
moritzkiefer-da: > We need to keep: discriminator of one item => just the item (previous behavior, untouched)

right, it looks like this change is also still missing from your change?

moritzkiefer-da: > but there is a separator? I'm not sure what you're asking for here

duh, clearly wasn't paying attention when reviewing. all good

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AllocationsFrontendIntegrationTest.scala
moritzkiefer-da: expiresAt for creating token standard transfers has an int64 so I'd also use that one here. Being inconsistent in the token standard endpoints seems needlessly confusing.

PR #1089, opened 2025-06-11
Upgrade to Canton 3.4

moritzkiefer-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


moritzkiefer-da:
cometbft is disabled as it's still broken, I opened https://github.com/hyperledger-labs/splice/issues/1145 to track fixing this

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/LocalSynchronizerNode.scala
```diff
@@ -394,18 +399,25 @@ final class LocalSynchronizerNode(
...
-              onboardingState
-            )
-          case NodeStatus.Success(_) =>
+            for {
+              _ <- sequencerAdminConnection.initializeFromOnboardingState(
+                onboardingState
+              )
+              status <- sequencerAdminConnection.getStatus
+            } yield status.successOption.fold(
+              throw Status.INTERNAL
```
nicu-da, line 414: When could this happen? Wouldn't this make sense to switch to `ensureThat`?

nicu-da (approved):
Thank you for pushing this mountain forward! :partying_face:

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/LocalSynchronizerNode.scala
moritzkiefer-da: No the endpoint is synchronous afaik so an ensuresThat just seems confusing.

nicu-da: I am still confused as to when would that call return ok but the sequencer would not be initialized

moritzkiefer-da: I'm not querying the status to see whether the sequencer is initialized. I just need to get the physical synchronizer id which the status endpoint has.

In theory the status endpoint could report as uninitialized which is still an OK status code but obviously that shoudl not happen after we call `initializeFromOnboardingState`

nicu-da: Missed the use of the return, makes sense, thanks

PR #1096, opened 2025-06-11
Implement DeleteCorruptAcsSnapshotTrigger

rautenrieth-da:
Second attempt at making deleting corrupt snapshots faster. This time we accept that this is a very long procedure and move it to a trigger.

The new trigger must complete before update history backfilling starts, because otherwise we could backfill import updates before finishing cleaning up corrupt snapshots, and the remaining corrupt snapshots would appear as valid.


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -2173,11 +2163,15 @@ object UpdateHistory {
   )
 
   case class State(
-      historyId: Option[Long]
+      historyId: Option[Long],
+      corruptSnapshotsDeleted: Boolean,
```
rautenrieth-da, line 2219: We could also remove this flag and have `ScanHistoryBackfillingTrigger` independently check for corrupt snapshots, at the cost of calling `migrationsWithCorruptSnapshots` more often.

`migrationsWithCorruptSnapshots` executes 3 independent SQL queries, each of them taking <10ms to run on CILR.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -152,6 +153,18 @@ class AcsSnapshotStore(
     }
   }
 
+  def deleteSnapshot(
+      snapshot: AcsSnapshot
+  )(implicit
+      tc: TraceContext
+  ): Future[Unit] = {
+    val statement = DBIOAction.seq(
+      sqlu"""delete from acs_snapshot where snapshot_record_time = ${snapshot.snapshotRecordTime}""",
```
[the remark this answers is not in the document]
rautenrieth-da: `snapshot_record_time` is the primary key in this table

ray-roestenburg-da (approved):
LGTM, thanks!

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/AcsSnapshotStoreTest.scala
OriolMunoz-da, line 1: for the sake of completeness, can we check the case where more than one migration id is corrupted? (Set.size > 1)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanHistoryBackfillingTrigger.scala
```diff
@@ -90,6 +90,9 @@ class ScanHistoryBackfillingTrigger(
     if (!store.updateHistory.isReady) {
       logger.debug("UpdateHistory is not yet ready")
       Future.successful(Seq.empty)
+    } else if (!store.updateHistory.corruptAcsSnapshotsDeleted) {
+      logger.debug("There may be corrupt ACS snapshots that need to be deleted")
+      Future.successful(Seq.empty)
```
OriolMunoz-da: why is this condition not part of the AcsSnapshotTrigger too?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/DeleteCorruptAcsSnapshotTrigger.scala
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  override protected def completeTask(task: DeleteCorruptAcsSnapshotTrigger.Task)(implicit
+      tc: TraceContext
+  ): Future[TaskOutcome] = task match {
+    case DeleteCorruptAcsSnapshotTrigger.Task(migrationId) =>
+      for {
+        lastSnapshotO <- store.lookupSnapshotBefore(migrationId, CantonTimestamp.MaxValue)
+        lastSnapshot = lastSnapshotO.getOrElse(
+          throw new RuntimeException("Task should never become stale")
```
OriolMunoz-da, line 65: it shouldn't? (I might be misremembering how triggers & staleness checks work)
say `store.lookupSnapshotBefore(migrationId, CantonTimestamp.MaxValue)` is successfully committed by postgres but there's a network error on the response, then the task fails and is checked for staleness, it says it's not stale but actually is because the delete went through, the app just didn't see it

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/DeleteCorruptAcsSnapshotTrigger.scala
```diff
@@ -0,0 +1,96 @@
...
+        )
+        _ <- store.deleteSnapshot(lastSnapshot)
+      } yield {
+        historyMetrics.CorruptAcsSnapshots.count.inc()
+        historyMetrics.CorruptAcsSnapshots.latestRecordTime.updateValue(
```
OriolMunoz-da, line 70: nit: this is not the latestRecordTime but the earliest... but I think I keep having issues with that :smile:

OriolMunoz-da (approved):
thanks!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanHistoryBackfillingTrigger.scala
rautenrieth-da: `AcsSnapshotTrigger` only does work when `ScanHistoryBackfillingTrigger` has completed backfilling, and the backfilling only does work when the corrupt snapshots are deleted.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/DeleteCorruptAcsSnapshotTrigger.scala
rautenrieth-da: Good point, will fix

docs/src/release_notes.rst
```diff
@@ -20,6 +20,17 @@ Upcoming
     <https://github.com/hyperledger-labs/splice/blob/0.4.12/Vagrantfile>`_ for
     details.
 
+- Scan app
```
rautenrieth-da: @moritzkiefer-da @martinflorian-da Can you please check if this makes sense? I have intentionally used

> deleted snapshots will be recomputed eventually

because the process for deleting snapshots and the process for backfilling missing import updates are separate, and this PR only enables the first process. The intention is to enable import update backfilling (already implemented but behind a feature flag) on CILR later today and by default before the next release. In worst case however, we could end up with a release that deletes corrupt snapshots but does not backfill the fixed ones. I would like to avoid using more feature flags, because I'm worried about breaking existing apps with some combination of flags that I did not think through.

moritzkiefer-da: Given that we want to enable both at once on prod, why not put them behind the same feature flag as well? You can then set that feature flag on CILR to true and when you change the default to true you can add the release notes at the same time for both.

rautenrieth-da: As of the current state of this PR, all backfilling would be paused until the new trigger completes, so putting the new trigger behind a feature flag would turn off _all_ backfilling, unless I change how backfilling depends on the new feature flag (i.e., allow regular backfilling but not import update backfilling while deleting snapshots). It would probably work, but I am bit worried there was a deeper reason why it's the way it is right now.

rautenrieth-da: Nevermind, I have put the new trigger behind the same feature flag as import update backfilling.

PR #1104, opened 2025-06-12
Adapt k8s prometheus requests resources 

julientinguely-da:
fixes https://github.com/DACH-NY/cn-test-failures/issues/4454

See explanation [here](https://github.com/DACH-NY/cn-test-failures/issues/4454#issuecomment-2965908157). 

Based on [30 last days](https://grafana.test.global.canton.network.digitalasset.com/d/85a562078cdf77779eaa1add43ccec1e/kubernetes-compute-resources-namespace-pods?var-datasource=default&var-cluster=&var-namespace=observability&orgId=1&from=now-30d&to=now&timezone=utc&refresh=10s) for cilr, dev, testnet.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/infra/src/observability.ts
+++ b/cluster/pulumi/infra/src/observability.ts
@@ -11,7 +11,8 @@ import { Input } from '@pulumi/pulumi';
 import {
   CLUSTER_BASENAME,
   CLUSTER_HOSTNAME,
-  CLUSTER_NAME, clusterProdLike,
+  CLUSTER_NAME,
+  clusterProdLike,
   COMETBFT_RETAIN_BLOCKS,
   ENABLE_COMETBFT_PRUNING,
   GCP_PROJECT,


cluster/pulumi/infra/src/observability.ts
```diff
@@ -278,8 +278,8 @@ export function configureObservability(dependsOn: pulumi.Resource[] = []): pulum
             retentionSize: infraConfig.prometheus.retentionSize,
             resources: {
               requests: {
-                memory: clusterProdLike ? (!clusterIsResetPeriodically ? '24Gi' : '6Gi') : '4Gi',
-                cpu: clusterProdLike ? (!clusterIsResetPeriodically ? '4' : '2') : '1',
+                memory: `${2 + (dsoSize - 1) * 0.5}Gi` ,
```
nicu-da: I am confused, why are we looking at the dso size?

julientinguely-da: It seems to be a good variable in the function because cilr requires more resources than testnet when looking at the [k8s resources dashboards](https://grafana.test.global.canton.network.digitalasset.com/d/85a562078cdf77779eaa1add43ccec1e/kubernetes-compute-resources-namespace-pods?var-datasource=default&var-cluster=&var-namespace=observability&orgId=1&from=now-30d&to=now&timezone=utc&refresh=10s.). CILR has 16 SVs, devnet 5 SVs, testnet 1 SV. Might be off, but then not really sure what value to set here. I don't think the actual function makes sense either way.

nicu-da: This is highly surprising behavior and I would definitely avoid using it in a such a fashion. 

Checking our scratches we seem to use 2gb so lets drop the 4gb to 2, the 6 to 3 and the 24 to 8.
Cpu wise we should also just halve all of them.

julientinguely-da: ok thanks, I will take your suggestion

PR #1119, opened 2025-06-13
Faster getPreviousMigrationId

rautenrieth-da:
See https://github.com/DACH-NY/cn-test-failures/issues/4500#issuecomment-2969255996

This query is currently overloading our database disk I/O.

Before:
```
QUERY PLAN
Result  (cost=0.74..0.75 rows=1 width=4) (actual time=38651.881..38651.884 rows=1 loops=1)
  Buffers: shared hit=5552 read=11529
  I/O Timings: read=25803.673
  InitPlan 1 (returns $0)
    ->  Limit  (cost=0.57..0.74 rows=1 width=4) (actual time=38651.860..38651.873 rows=1 loops=1)
          Buffers: shared hit=5552 read=11529
          I/O Timings: read=25803.673
          ->  Index Only Scan Backward using updt_hist_tran_hi_mi_rt_di on update_history_transactions  (cost=0.57..10995011.69 rows=64397161 width=4) (actual time=38651.858..38651.859 rows=1 loops=1)
                Index Cond: ((history_id = 1) AND (migration_id IS NOT NULL) AND (migration_id < 6))
                Heap Fetches: 0
                Buffers: shared hit=5552 read=11529
                I/O Timings: read=25803.673
Planning:
  Buffers: shared hit=215 read=6
  I/O Timings: read=20.340
Planning Time: 22.524 ms
Execution Time: 38652.007 ms
```

After:
```
QUERY PLAN
Aggregate  (cost=80.99..81.00 rows=1 width=4) (actual time=0.228..0.230 rows=1 loops=1)
  Buffers: shared hit=36
  CTE t
    ->  Recursive Union  (cost=0.57..78.72 rows=101 width=4) (actual time=0.066..0.216 rows=7 loops=1)
          Buffers: shared hit=36
          ->  Limit  (cost=0.57..0.74 rows=1 width=4) (actual time=0.065..0.065 rows=1 loops=1)
                Buffers: shared hit=5
                ->  Index Only Scan using updt_hist_tran_hi_mi_rt_di on update_history_transactions update_history_transactions_1  (cost=0.57..10820178.64 rows=64222991 width=4) (actual time=0.064..0.064 rows=1 loops=1)
                      Index Cond: ((history_id = 1) AND (migration_id < 6))
                      Heap Fetches: 0
                      Buffers: shared hit=5
          ->  WorkTable Scan on t t_1  (cost=0.00..7.60 rows=10 width=4) (actual time=0.020..0.020 rows=1 loops=7)
                Filter: (migration_id IS NOT NULL)
                Rows Removed by Filter: 0
                Buffers: shared hit=31
                SubPlan 1
                  ->  Limit  (cost=0.57..0.74 rows=1 width=4) (actual time=0.021..0.021 rows=1 loops=6)
                        Buffers: shared hit=31
                        ->  Index Only Scan using updt_hist_tran_hi_mi_rt_di on update_history_transactions  (cost=0.57..62412.73 rows=362596 width=4) (actual time=0.020..0.020 rows=1 loops=6)
                              Index Cond: ((history_id = 1) AND (migration_id < 6) AND (migration_id > t_1.migration_id))
                              Heap Fetches: 0
                              Buffers: shared hit=31
  ->  CTE Scan on t  (cost=0.00..2.02 rows=100 width=4) (actual time=0.069..0.222 rows=6 loops=1)
        Filter: (migration_id IS NOT NULL)
        Rows Removed by Filter: 1
        Buffers: shared hit=36
Planning:
  Buffers: shared hit=213
Planning Time: 2.026 ms
Execution Time: 0.609 ms
```

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
@@ -1726,10 +1726,20 @@ class UpdateHistory(
   ): Future[Option[Long]] = {
     def previousId(table: String) = {
       storage.query(
+        // The following is equivalent to:
+        //     select max(migration_id)
+        //     from #$table
+        //     where history_id = $historyId and migration_id < $migrationId
+        // but uses a recursive CTE to implement an index skip scan.
         sql"""
-             select max(migration_id)
-             from #$table
-             where history_id = $historyId and migration_id < $migrationId
+          WITH RECURSIVE t AS (
+             (SELECT migration_id FROM #$table where history_id = 1 and migration_id < $migrationId ORDER BY migration_id LIMIT 1)
+             UNION ALL
+             SELECT (SELECT migration_id FROM #$table WHERE history_id = 1 and migration_id < $migrationId and migration_id > t.migration_id ORDER BY migration_id LIMIT 1)
+             FROM t
+             WHERE t.migration_id IS NOT NULL
+             )
+          SELECT MAX(migration_id) FROM t WHERE migration_id IS NOT NULL;
            """.as[Option[Long]].head,
         s"getPreviousMigrationId.$table",
       )
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
@@ -1730,12 +1730,12 @@ class UpdateHistory(
         //     select max(migration_id)
         //     from #$table
         //     where history_id = $historyId and migration_id < $migrationId
-        // but uses a recursive CTE to implement an index skip scan.
+        // but uses a recursive CTE to avoid a backwards-index-scan which attempts to read most of the table.
         sql"""
           WITH RECURSIVE t AS (
-             (SELECT migration_id FROM #$table where history_id = 1 and migration_id < $migrationId ORDER BY migration_id LIMIT 1)
+             (SELECT migration_id FROM #$table where history_id = $historyId and migration_id < $migrationId ORDER BY migration_id LIMIT 1)
              UNION ALL
-             SELECT (SELECT migration_id FROM #$table WHERE history_id = 1 and migration_id < $migrationId and migration_id > t.migration_id ORDER BY migration_id LIMIT 1)
+             SELECT (SELECT migration_id FROM #$table WHERE history_id = $historyId and migration_id < $migrationId and migration_id > t.migration_id ORDER BY migration_id LIMIT 1)
              FROM t
              WHERE t.migration_id IS NOT NULL
              )


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1726,10 +1726,20 @@ class UpdateHistory(
...
+        //     where history_id = $historyId and migration_id < $migrationId
+        // but uses a recursive CTE to implement an index skip scan.
         sql"""
-             select max(migration_id)
-             from #$table
-             where history_id = $historyId and migration_id < $migrationId
+          WITH RECURSIVE t AS (
+             (SELECT migration_id FROM #$table where history_id = 1 and migration_id < $migrationId ORDER BY migration_id LIMIT 1)
+             UNION ALL
+             SELECT (SELECT migration_id FROM #$table WHERE history_id = 1 and migration_id < $migrationId and migration_id > t.migration_id ORDER BY migration_id LIMIT 1)
```
OriolMunoz-da: two things:
- history_id is hardcoded to 1
- isn't this query the same, without recursive CTE?:
``` 
select migration_id
from #$table
where history_id = $historyId and migration_id < $migrationId
ORDER BY migration_id DESC
LIMIT 1
```
? I'll give it a try following the instructions in the GH issue

OriolMunoz-da: image
yeah that seems pretty good :smile:

OriolMunoz-da: image
the plan is roughly the same, I don't understand how the cost is so high

OriolMunoz-da: okay I am now convinced the rec CTE is necessary (I probably went through the same cycle of non-believe you must've gone through :upside_down_face: ). I will update your PR to fix the hardcoded history_id

OriolMunoz-da:
sorry for the force-push, I had to sign the commit...

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
OriolMunoz-da: Also for learning purposes, this happens due to what the last notes in [pg docs](https://www.postgresql.org/docs/current/indexes-ordering.html) mention:

>  But it might be that the application frequently needs to use ORDER BY x ASC, y DESC. There is no way to get that ordering from a plain index, but it is possible if the index is defined as (x ASC, y DESC) or (x DESC, y ASC).

We wouldn't have problems with the backwards index if our index looked like:
`CREATE INDEX testindex1 on scan_sv_13.update_history_transactions(history_id, migration_id);`
image

or possibly (history_id, migration_id DESC, record_time, domain_id)... alas, that ship has sailed

ray-roestenburg-da: why has that ship sailed? we don't want to add another index?

OriolMunoz-da: it will take very long given the size of the tables, which (in our current setup) requires partners' approval cause rewards are lost...

ray-roestenburg-da: yeah this is the `flyway-on-startup` issue, we might want to take some time thinking about how we can improve that so that some things can take place in parallel.. might also be tricky. As discussed this is a good solution though.

Issue #575, opened 2025-06-05
failing to init the sv1 sequencer breaks the node


canton-network-da:
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#18312). Original author: @nicu-da ]

If this call fails https://github.com/DACH-NY/canton-network-node/blob/main/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/sv1/SV1Initializer.scala#L474

sv app getting errors from the participant

```
2025-03-13T15:34:49.119Z [⋮] INFO - o.l.s.a.a.c.ApiClientRequestLogger:SV=sv (bb1a98f0251a740594133ee2003f2458-bootstrapDomain-57cf4a3291caace8-f36c96f83a03f20f) - Request (tid:1a2eab7dd9ff31a6b65d86385dfaf3ba) com.digitalasset.canton.topology.admin.v30.TopologyManagerWriteService/Authorize to participant-0:5002: failed with ALREADY_EXISTS/TOPOLOGY_MAPPING_ALREADY_EXISTS(10,1a2eab7d): A matching topology mapping authorized with the same keys already exists in this state 
2025-03-13T15:34:49.150Z [⋮] INFO - o.l.s.s.o.s.SV1Initializer:SV=sv (bb1a98f0251a740594133ee2003f2458-init_sequencer-3db69a5b684cedb9-5f5d2e3249daae9b) - The operation 'Check whether sequencer is initialized and establish it if needed' failed with a retryable error (full stack trace omitted):
category=Some(InvalidGivenCurrentSystemStateResourceExists)
ErrorInfoDetail(TOPOLOGY_MAPPING_ALREADY_EXISTS,HashMap(participant -> participant, existing -> DecentralizedNamespaceDefinition(12206b4943af..., PositiveNumeric(1), Set(12202f8a1d06...)), category -> 10, tid -> 1a2eab7dd9ff31a6b65d86385dfaf3ba, keys -> Set(12202f8a1d06...)))
RequestInfoDetail(1a2eab7dd9ff31a6b65d86385dfaf3ba) 
[...]

PR #1131, opened 2025-06-16
[ci] Ensure proposeInitialMapping is idempotent

nicu-da:
Fixes https://github.com/hyperledger-labs/splice/issues/575

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
@@ -4,7 +4,7 @@
 package org.lfdecentralizedtrust.splice.environment
 
 import cats.data.{EitherT, OptionT}
-import cats.implicits.catsSyntaxOptionId
+import cats.implicits.{catsSyntaxApplicativeId, catsSyntaxOptionId}
 import cats.syntax.either.*
 import com.daml.nonempty.NonEmpty
 import com.digitalasset.canton.admin.api.client.commands.TopologyAdminCommands.Init.GetIdResult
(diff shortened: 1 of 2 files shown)


cluster/pulumi/sv-runbook/pulumiDown.ts
```diff
@@ -13,7 +13,7 @@ async function runRunbookDown() {
   operations.push(downOperation(svRunbookStack, abortController));
   const cantonStacks = runSvCantonForSvs(
     ['sv'],
-    'up',
+    'down',
```
nicu-da, line 16: drive-by fix

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
```diff
@@ -596,18 +598,38 @@ abstract class TopologyAdminConnection(
...
+      ),
+    ).flatMap(existing =>
+      existing
+        .find(_.mapping == mapping)
+        .flatMap(_.selectMapping[M])
+        .fold({
+          logger.info(
+            s"Proposing initial mapping for ${mapping.code} with serial 1: $mapping"
+          )
+          proposeMapping(
```
moritzkiefer-da, line 619: does this make sense to run if the `find(_.mapping == mapping)` above fails? Isn't that guaranteed to always fail then?

nicu-da: I assume you're referring to the `.flatMap(_.selectMapping[M])` call? That's just to make the compiler happy, we need to return `M`

moritzkiefer-da: Consider the case where you have an existing mapping but it's a different one than the one you are trying to submit `find(_.mapping == mapping)` will give you nothing in that case. But then you hit the first `fold` part and try to submit something with serial 1.

nicu-da: Ah right, that's intentional, it would've failed the same way before practically, and considering that this should handle only the initial mapping with serial one then it would require some manual intervention to fix it.

moritzkiefer-da: failing makes sense but why not fail with a clear error that the mapping is different instead of the somewhat confusing error that you get back from propose.

nicu-da: Good point, handling it

PR #1135, opened 2025-06-16
[ci] Track active_parties in scan aggregation for better performance.

### LINKED ISSUE #732: Scan aggregator makes a sequential scan over round_party_totals
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#14176). Original author: @moritzkiefer-da ]

This is clearly not great since that table will grow forever.

image

Have not looked at query plans or whatever.

ray-roestenburg-da:
Fixes #732 

- [x] Add a test with an empty `active_parties` table and a non-empty `round_party_totals`, verifying that  `initiate_active_parties` works an subsequent total and cumulative are correct.
- [x] Add a test where not all parties are active in all rounds. 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V040__active_parties.sql
+++ b/apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V040__active_parties.sql
@@ -1,0 +1,25 @@
+create table active_parties
+(
+    store_id          int not null,
+    party             text not null,
+    closed_round      bigint not null,
+    primary key       (store_id, party)
+);
+
+create index active_parties_sid_p_cr_idx
+    on active_parties (store_id, party, closed_round);
+
+create or replace procedure initialize_active_parties()
+language plpgsql
+as $$
+begin
+    if not exists (select 1 from active_parties limit 1) then
+        insert into active_parties (store_id, party, closed_round)
+        select store_id, party, max(closed_round)
+        from round_party_totals
+        group by store_id, party;
+    end if;
+end
+$$;
+
+grant execute on procedure initialize_active_parties() to public;
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
@@ -590,7 +590,7 @@ final class ScanAggregator(
     val previousRoundTotals = previousRoundTotalsO.getOrElse(
       RoundTotals(closedRound = -1L)
     )
-    def doQuery = {
+    def doQuery() = {
       sql"""
       with new_totals_per_entry as(
         select    round,
@@ -704,7 +704,7 @@ final class ScanAggregator(
     logger.info(
       s"""Aggregating round_totals for rounds up to including round ${lastClosedRound}$extraMsg."""
     )
-    doQuery
+    doQuery()
   }
 
   /** Aggregates the RoundPartyTotals in the round_party_totals table,
@@ -716,18 +716,28 @@ final class ScanAggregator(
   )(implicit
       traceContext: TraceContext
   ): DBIOAction[Unit, NoStream, Effect.Write] = {
-    def doQuery = {
-      /* The round_party_totals will only contain rows for parties which have balance change or have been rewarded or purchased extra traffic in rounds.
-       * Sum per round and party, where the party is the rewarded_party or extra_traffic_validator,
-       * since the most recently recorded rounds in round_party_totals, up until including the last closed round.
-       * Gets the previous cumulative sums for each party, for the most recent round that registered activity for that party
-       * (parties can be missing in rounds, since they did not get rewarded, or did not by traffic).
-       * Calculate cumulative sums left joining with possibly existing previous cumulative sums per party, keeping track of the most recent round per party.
-       * The left join with sums over all rounds grouped by parties ensures that all parties balance changes, rewards and purchases are included in the next result,
-       * and that it is safe to start the aggregation from any previously aggregated round, even if some parties are missing in some rounds.
-       * finally inserting the sums and cumulative sums in the round_party_totals table.
-       */
-      sql"""
+
+    /** Initialization of active_parties is done in an idempotent stored proc (Effect.Write).
+      * This is done outside of flyway migration so that it does not impact scan app startup time.
+      */
+    def initializeActiveParties(): DBIOAction[Unit, NoStream, Effect.Write] = {
+      sqlu"""call initialize_active_parties()"""
+        .andThen(DBIO.successful(()))
+    }
+
+    /** The round_party_totals will only contain rows for parties which have balance change or have been rewarded or purchased extra traffic in rounds.
+      * Sum per round and party, where the party is the rewarded_party or extra_traffic_validator,
+      * since the most recently recorded rounds in round_party_totals, up until including the last closed round.
+      * Gets the previous cumulative sums for each party from the active_parties table, which keeps track of the most recent round that registered activity for that party
+      * (parties can be missing in rounds, since they did not get rewarded, or did not buy traffic).
+      * Calculate cumulative sums left joining with possibly existing previous cumulative sums per party, keeping track of the most recent round per party.
+      * The left join with sums over all rounds grouped by parties ensures that all parties balance changes, rewards and purchases are included in the next result,
+      * and that it is safe to start the aggregation from any previously aggregated round, even if some parties are missing in some rounds.
+      * finally inserting the sums and cumulative sums in the round_party_totals table.
+      */
+    def aggregate(): DBIOAction[Unit, NoStream, Effect.Write] = {
+      sqlu"""
+      create temp table temp_cumulative_totals on commit drop as
       with previously_aggregated as(
         select coalesce(max(closed_round), -1) as last_closed_round from round_party_totals where store_id = $roundTotalsStoreId
       ),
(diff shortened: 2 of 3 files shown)


apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V040__active_parties.sql
```diff
@@ -0,0 +1,25 @@
...
+);
+
+create index active_parties_sid_p_cr_idx
+    on active_parties (store_id, party, closed_round);
+
+create or replace procedure initialize_active_parties()
+language plpgsql
+as $$
+begin
+    if not exists (select 1 from active_parties limit 1) then
```
ray-roestenburg-da, line 16: This will show as seq scan but is one row only since it stops immediately so no concern.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
```diff
@@ -716,18 +716,28 @@ final class ScanAggregator(
...
-       * and that it is safe to start the aggregation from any previously aggregated round, even if some parties are missing in some rounds.
-       * finally inserting the sums and cumulative sums in the round_party_totals table.
-       */
-      sql"""
+
+    /** Initialization of active_parties is done in an idempotent stored proc (Effect.Write).
+      * This is done outside of flyway migration so that it does not impact scan app startup time.
+      */
+    def initializeActiveParties(): DBIOAction[Unit, NoStream, Effect.Write] = {
+      sqlu"""call initialize_active_parties()"""
```
moritzkiefer-da, line 724: how fast is it on CILR scale data?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
```diff
@@ -818,39 +828,66 @@ final class ScanAggregator(
...
+        select  rpt.party as party,
+                coalesce(cumulative_app_rewards, 0) as prev_cumulative_app_rewards,
+                coalesce(cumulative_validator_rewards, 0) as prev_cumulative_validator_rewards,
+                coalesce(cumulative_traffic_purchased, 0) as prev_cumulative_traffic_purchased,
+                coalesce(cumulative_traffic_purchased_cc_spent, 0) as prev_cumulative_traffic_purchased_cc_spent,
+                coalesce(cumulative_traffic_num_purchases, 0) as prev_cumulative_traffic_num_purchases,
+                coalesce(cumulative_change_to_initial_amount_as_of_round_zero, 0) as prev_cumulative_change_to_initial_amount_as_of_round_zero,
+                coalesce(cumulative_change_to_holding_fees_rate, 0) as prev_cumulative_change_to_holding_fees_rate
+        from    round_party_totals rpt
+        join    active_parties ap
```
moritzkiefer-da, line 841: I have a hard time understanding why this join is correct. What about this scenario:
1. Party is active in round 1, it has a total of app rewards (or whatever else, doesn't really matter) of 42. This means we have an entry for that party in round 1 in `active_parties`.
2. Party is not active in round 2. This means we have no entry for that party in round 2 in `active_parties`
3. Party is active again in round 3.
4. Now you aggregate from round 2 to round 3. Doesn't this mean that the join here will not include any entry for that party so you consider the previous total 0 instead of 42?

ray-roestenburg-da: Rows in active_parties are mostly updated, unless new parties are active that have never been seen in the txlog.
There is only one row per party in the table (not considering the store_id)
1. yes. 
2. yes, but the `round_party_totals` also does not have it (not aggregated yet), it has an entry for that party for round 1 (and maybe one for round 0, who knows), and this is joined, and seen as the state before aggregation.
3. the active_parties only has one entry per party. so you can't be in the state where there is a round 1 and round 3 here (the most recent round it was 'active' in). If the party was active in round 3, before we aggregate, active_parties is already reflecting that, (and so is round_party_totals as part of the previous aggregation run) or it will only update that the next time as previous totals.
4. Depending on the data, aggregation is done for 0 .. many rounds at a time, we do a running total 'sum partition over' for the data for the new rounds, compared to the totals for the last known round per party
5. the aggregation is not per round, but for all rounds between previously aggregated, and what is read from tx log, up to 'last closed round'

moritzkiefer-da: Let's say aggregation happens each round so yes I'm aware we can skip over rounds but if we're fast enough we don't.

In the example I mentioned above, what is the state of `active_parties` after aggregation of round 2? Does the party still have an entry with round 1? Do we update it to round 2 even though it has not been active in round 2? If the latter, I think I don't understand what "active" means.

ray-roestenburg-da: > In the example I mentioned above, what is the state of active_parties after aggregation of round 2? Does the party still have an entry with round 1?

party has entry with round 1 in active_parties.

> Do we update it to round 2 even though it has not been active in round 2? 

no

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
ray-roestenburg-da: On a scan clone from CILR it completed in  ~48 seconds.

ray-roestenburg-da: (when it needed to initialize of course, once initialized the time it takes is negligible)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
moritzkiefer-da: ok I think I get it now so previous_totals is not the totals of the round we're aggregating from, it's the previous per partyt which can be different per party.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
```diff
@@ -818,39 +828,66 @@ final class ScanAggregator(
...
+                sum(nt.traffic_purchased) over (partition by nt.party order by round) + coalesce(pt.prev_cumulative_traffic_purchased,0) as cumulative_traffic_purchased,
+                sum(nt.traffic_purchased_cc_spent) over (partition by nt.party order by round) + coalesce(pt.prev_cumulative_traffic_purchased_cc_spent,0) as cumulative_traffic_purchased_cc_spent,
+                sum(nt.traffic_num_purchases) over (partition by nt.party order by round) + coalesce(pt.prev_cumulative_traffic_num_purchases,0) as cumulative_traffic_num_purchases,
+                sum(nt.change_to_initial_amount_as_of_round_zero) over (partition by nt.party order by round) + coalesce(pt.prev_cumulative_change_to_initial_amount_as_of_round_zero,0) as cumulative_change_to_initial_amount_as_of_round_zero,
+                sum(nt.change_to_holding_fees_rate) over (partition by nt.party order by round) + coalesce(pt.prev_cumulative_change_to_holding_fees_rate,0) as cumulative_change_to_holding_fees_rate
+      from      new_totals nt
+      left join previous_totals pt
+      on        nt.party = pt.party;
+
+      insert into active_parties (
```
moritzkiefer-da: We have an insert here and then an update below. Why do we need both? Aren't you just doing an upsert i.e. insert the max round for each party from temp_cumulative_totals and `on conflict do update round = max_round`?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanAggregatorTest.scala
```diff
@@ -390,126 +390,112 @@ class ScanAggregatorTest
...
+            store,
+            lastRound,
+            lastRound.toInt,
+          )
+          assertActiveParties(store, nrParties, lastRound)
+        }
+      }
+    }
+
+    "Aggregate round_party_totals where some parties are not always taking part" in {
```
moritzkiefer-da:
```suggestion
    "Aggregate round_party_totals where some parties are not active in all rounds" in {
```

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanAggregatorTest.scala
```diff
@@ -390,126 +390,112 @@ class ScanAggregatorTest
...
+      cleanup()
+      val (aggr, store) = mkAggregator(dsoParty).futureValue
+      val lastRound = 5L
+      val firstParties = Seq(1, 2)
+      val firstRounds = Seq(0L, 1L, 2L)
+      val lastParties = Seq(3, 4, 5)
+      val lastRounds = Seq(3L, 4L, 5L)
+      for {
+        first <- generateRoundPartyTotalsRange(store, firstRounds, firstParties)
+        last <- generateRoundPartyTotalsRange(store, lastRounds, lastParties)
```
moritzkiefer-da, line 473: can we make sure to cover the scenario where a party is not active for some rounds but is then active again later? afaik you only cover the scenario here where they are either not at all active at the start or stop activity completely after firstRounds but not where there are gaps

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
moritzkiefer-da: sounds good, thanks

moritzkiefer-da (approved):
Thanks! I think I convinced myself now that it works but the tests don't seem as detailed as they should be imho

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanAggregatorTest.scala
ray-roestenburg-da: Sounds good, will do.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
ray-roestenburg-da: I read around and convinced myself that it was faster to insert select only what is not there. Most of the time it's updates and conflicting on insert most of the time seemed like a bad idea. But on another note I see now that I have a strange leftover here! this insert should not be there, only the next one. Will fix.

ray-roestenburg-da: But maybe I am wrong about this assumption and upserts might be better, I'll check this again, thanks for spotting this.

ray-roestenburg-da: It might be better to do an upsert, postgres is supposedly fast at it, I was probably overworried about conflicting on inserts a lot.. I'll change it.

ray-roestenburg-da: Using upsert now.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanAggregatorTest.scala
ray-roestenburg-da: @moritzkiefer-da Added a test for this scenario.

Issue #479, opened 2025-06-05
Make JWKS connection timeouts configurable


canton-network-da:
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#19645). Original author: @moritzkiefer-da ]

https://support.digitalasset.com/s/case/500Pq00000ZvqH3IAJ/mainnet-canton-validator-temporarily-becomes-unavailable looks like JWKS requests get stuck. It's tricky to say much about what exactly is going on.

But the usual client-side solution for requests getting stuck is configuring timeouts.

This is configurable on the JWK provider here https://github.com/auth0/jwks-rsa-java/blob/d877bd7e74510d905ded29468d8b8b7b6f9703ef/src/main/java/com/auth0/jwk/JwkProviderBuilder.java#L152

We should also compare to the Canton implementation here https://github.com/DACH-NY/canton/blob/main/community/ledger/ledger-api-core/src/main/scala/com/digitalasset/canton/ledger/api/auth/CachedJwtVerifierLoader.scala.

This does configure default timeouts. It also has a custom cache instead of relying on the one in the auth0 library. We probably want to ask Canton what led to that decision and see whether we want to just reuse what Canton has instead of rolling this ourselves.

PR #1146, opened 2025-06-17
Make JWKS connection timeouts configurable

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/479

Decided NOT to use the canton version and instead kept the existing cache, I don't see any reason why we'd replace it.

### DIFF
--- a/cluster/images/sv-app/app.conf
+++ b/cluster/images/sv-app/app.conf
@@ -57,6 +57,8 @@ canton {
         algorithm = "rs-256"
         audience = ${SPLICE_APP_SV_AUTH_AUDIENCE}
         jwks-url = ${SPLICE_APP_SV_AUTH_JWKS_URL}
+        connection-timeout = ${?SPLICE_APP_VALIDATOR_AUTH_JWKS_CONNECTION_TIMEOUT}
+        read-timeout = ${?SPLICE_APP_VALIDATOR_AUTH_JWKS_READ_TIMEOUT}
       }
       domains {
         global {


OriolMunoz-da:
Should this be mentioned in release notes?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/auth/AuthConfig.scala
```diff
@@ -18,6 +20,8 @@ object AuthConfig {
   case class Rs256(
       audience: String,
       jwksUrl: URL,
+      connectionTimeout: NonNegativeDuration = NonNegativeDuration.ofSeconds(5),
```
moritzkiefer-da: is this the same default we had before?

OriolMunoz-da: afaict there was no default, the library had it to null

moritzkiefer-da: wfm, do the new defaults match the Canton defaults?

OriolMunoz-da: they do now :upside_down_face:

.pre-commit-config.yaml
```diff
@@ -87,3 +87,4 @@ repos:
      - id: rstcheck
        name: Check sphinx format
        additional_dependencies: ['rstcheck[sphinx]']
+       args: [--report-level, WARNING]
```
OriolMunoz-da, line 90: image

cluster/helm/splice-sv-node/templates/sv.yaml
```diff
@@ -108,6 +108,14 @@ spec:
               value: {{ .Values.auth.audience | quote }}
             - name: SPLICE_APP_SV_AUTH_JWKS_URL
               value: {{ .Values.auth.jwksUrl | quote }}
+            {{ if .Values.auth.connectionTimeout }}
```
moritzkiefer-da: should this be something like jwksConnectionTimeout or jwks.connectionTimeout? Otherwise it's pretty non-obvious what you're timeouting on. Same for the other one.

moritzkiefer-da: you made the mistake of adding this so now I'm obviously gonna ask for a helm unittest :) same for the others

moritzkiefer-da: I also suggest to manually deploy with this once.

apps/app/src/pack/examples/sv-helm/validator-values.yaml
```diff
@@ -26,6 +26,10 @@ auth:
   # replace OIDC_AUTHORITY_URL with your provider's OIDC URL
   jwksUrl: "https://OIDC_AUTHORITY_URL/.well-known/jwks.json"
 
+  # optionally, reconfigure the timeouts:
```
moritzkiefer-da:
```suggestion
  # optionally, reconfigure the timeouts used when querying the JWKS endpoint
```

moritzkiefer-da: same above

cluster/helm/splice-sv-node/templates/sv.yaml
OriolMunoz-da: > helm unittest

do we have them anywhere? (so I have an example, never wrote one)

moritzkiefer-da: https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#helm-checks

PR #1150, opened 2025-06-17
Cache SV node state based queries

### LINKED ISSUE #1141: Cache list of scans and sequencers
Currently, our CILR test cluster is experiencing heavy load on the app database, caused by excessive ACS queries.

It's difficult to tell what exactly is causing the load, but it looks like it's mostly fetching lists of scans and sequencers. Some pointers:
- `DomainConnector.getSequencerConnectionsFromScan`
- `BftScanConnection.Bft.getPeerScansFromStore`
- `HttpScanHandler.listDsoSequencers`
- `ScanStore.listFromSvNodeStates`
- `ScanStore.listDsoScans`

Implement some cache for this information instead of fetching it from the database every time. Note that the list of scans and sequencers is changing quickly while the network is starting up (especially after a HSM or in tests where we start many SV nodes at the same time), so make sure the cache is invalidated properly.

Note:

We have `CachingScanConnection` which caches `AnsRules` contracts for an unlimited time, and invalidates the cache any time a command submission fails with a `LOCAL_VERDICT_INACTIVE_CONTRACTS` error.
This approach won't work for caching `DsoRules` in the scan app, but we could create a similar cache that is hooked up to `DbMultiDomainAcsStore` and invalidated when the store sees the archive event for a cached contract. See also the existing `trait DsoRulesStore`.

moritzkiefer-da:
[ci]

fixes #1141

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/resources/include/scans/_scan.conf
+++ b/apps/app/src/test/resources/include/scans/_scan.conf
@@ -13,5 +13,5 @@
     ledger-api.auth-config.type = "none"
   }
   # Reduce in tests as SVs change relatively frequently here.
-  sv-node-state-cache-ttl = 1s
+  cache.sv-node-state-ttl = 1s
 }
--- a/apps/app/src/test/resources/include/scans/_scan.conf
+++ b/apps/app/src/test/resources/include/scans/_scan.conf
@@ -12,4 +12,6 @@
   participant-client {
     ledger-api.auth-config.type = "none"
   }
+  # Reduce in tests as SVs change relatively frequently here.
+  sv-node-state-cache-ttl = 1s
 }
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStoreMetrics.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStoreMetrics.scala
@@ -36,7 +36,8 @@ class DbScanStoreMetrics(metricsFactory: LabeledMetricsFactory) extends AutoClos
       -1L,
     )(MetricsContext.Empty)
 
-  val cache = new CacheMetrics(prefix :+ "cache", metricsFactory)
+  val totalAmuletBalanceCache = new CacheMetrics(prefix :+ "total-amulet-balance-cache", metricsFactory)
+  val svNodeStateCache = new CacheMetrics(prefix :+ "sv-node-state-cache", metricsFactory)
 
   val history = new HistoryMetrics(metricsFactory)(MetricsContext.Empty)
 
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStoreMetrics.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStoreMetrics.scala
@@ -36,8 +36,9 @@ class DbScanStoreMetrics(metricsFactory: LabeledMetricsFactory) extends AutoClos
       -1L,
     )(MetricsContext.Empty)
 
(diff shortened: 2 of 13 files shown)


moritzkiefer-da:
I don't have a great idea for an automatic test here. I did manually verify that we get some hits through metrics. If someone has a better idea, I can try adding something.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStoreMetrics.scala
```diff
@@ -36,7 +36,9 @@ class DbScanStoreMetrics(metricsFactory: LabeledMetricsFactory) extends AutoClos
       -1L,
     )(MetricsContext.Empty)
 
-  val cache = new CacheMetrics(prefix :+ "cache", metricsFactory)
```
moritzkiefer-da, line 39: this rename is technically breaking but the metrics didn't even work before so seems fine

canton/community/base/src/main/scala/com/digitalasset/canton/caching/ScaffeineCache.scala
```diff
@@ -29,7 +29,9 @@ object ScaffeineCache {
       context: String,
   ): TunnelledAsyncCache[K, V] = {
     val contextWithLoggerName = s"${tracedLogger.underlying.getName}:$context"
-    val asyncCache = cache.buildAsync[K, V]()
+    val cacheWithMetrics =
```
moritzkiefer-da, line 32: the metrics didn't work at all. There is an InstallMetrics call below but that only registers the size and weight gauge but misses the actually useful hits and misses metrics. Will upstream to Canton

apps/app/src/test/resources/include/scans/_scan.conf
```diff
@@ -12,4 +12,6 @@
   participant-client {
     ledger-api.auth-config.type = "none"
   }
+  # Reduce in tests as SVs change relatively frequently here.
+  sv-node-state-cache-ttl = 1s
```
moritzkiefer-da: we could make the cache more clever and automatically invalidate when we ingest new contracts. Doesn't really seem worth the complexity compared to a time based cache here. The query is cheap and we don't care about extremely fast updates.

nicu-da: nit: would it make sense to create an extra layer called `cache` and then include `state-cache-ttl` in it?

nicu-da (approved):
Thank you!

ray-roestenburg-da (approved):
LGTM thanks!

apps/app/src/test/resources/include/scans/_scan.conf
moritzkiefer-da: done

apps/app/src/test/resources/include/scans/_scan.conf
```diff
@@ -12,4 +12,6 @@
   participant-client {
     ledger-api.auth-config.type = "none"
   }
+  # Reduce in tests as SVs change relatively frequently here.
+  cache.sv-node-state-ttl = 1s
```
rautenrieth-da, line 16: For my own education, why overwrite it here and not in `ConfigTransforms.defaults`, like we do with the polling interval?

moritzkiefer-da: It's a test only config file, I don't see a lot of value in overwriting it from a config transform instead. That just seems to add complexity. Do you see a benefit?

rautenrieth-da: No, I don't think it matters. I only noticed that we overwrite the trigger polling interval (from default 30sec to 1sec) in a config transform, and overwriting the TTL serves a similar purpose. So I thought there might be some reason to it, like the config file and config transform applying to different tests.

PR #1151, opened 2025-06-17
Avoid refreshing BftScanConnection too often when less than f scans are failing

OriolMunoz-da:
Fixes too many refreshes happening even if BFT guarantees are still ensured (see https://github.com/hyperledger-labs/splice/issues/1141#issuecomment-2979555709)

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
@@ -132,28 +132,14 @@ class BftScanConnection(
           retryProvider.timeouts,
           "refresh_scan_list",
         )({ tc =>
-          // This retry makes sure any partial or complete failures are immediately retried with a backoff.
+          // refresh will throw if we're in a state where there's no BFT guarantees, in which case
+          // this will retry faster than the regular `bft.scansRefreshInterval`
           FutureUnlessShutdown.outcomeF(
             retryProvider.retry(
               RetryFor.LongRunningAutomation,
               "refresh_scan_list",
               "refresh_scan_list",
-              bft.refresh(this)(tc).flatMap { connections =>
-                val failed = connections.failed
-                val total = connections.totalNumber
-                val f = connections.f
-                if (connections.failed > connections.f)
-                  Future.failed(
-                    io.grpc.Status.UNAVAILABLE
-                      .withDescription(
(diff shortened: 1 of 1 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
OriolMunoz-da, line 1: the alternative would be to drop the whole retry mechanism, but the problem then is that it's only refreshed every ~10m (which is a good-enough default otherwise)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -139,10 +139,17 @@ class BftScanConnection(
               "refresh_scan_list",
               "refresh_scan_list",
               bft.refresh(this)(tc).flatMap { connections =>
-                if (connections.failed > 0)
+                val failed = connections.failed
+                val total = connections.totalNumber
+                val f = connections.f
+                if (connections.failed > connections.f)
```
moritzkiefer-da: Hm should this be f or 2f+1? I thought we want 2f+1 working connections so we can read from them and tolerate f failures.

OriolMunoz-da: actually that was stupid, `refresh` already does the check:
``` 
val defaultCallConfig = BftCallConfig.default(connections)
            // Most but not all calls will use the default config.
            // Fail early if there are not enough Scans for the default config
            if (!defaultCallConfig.enoughAvailableScans) {
              throw io.grpc.Status.FAILED_PRECONDITION
                .withDescription(
                  s"There are not enough Scans to satisfy f=${connections.f}. Will be retried. State: $newState"
                )
                .asRuntimeException()
            } else {
              connections
            }
```

OriolMunoz-da: fix'd

OriolMunoz-da: (where also the condition is: `connections.size >= targetSuccess`, i.e. `connections.size >= f +1`)

rautenrieth-da (approved):
Thanks!

Would it make sense to add a test, something like the existing `"periodically refresh the list of scans"` in BftScanConnectionTest?

Issue #1160, opened 2025-06-18
BftScanConnection.refresh can get stuck in a retry loop


OriolMunoz-da:
For a validator using BFT scan reads:

- Some initial state is valid
- refresh
- Scan list is bft-fetched
- Somehow, not enough scans work (so there are no BFT guarantees)
- refresh again
- fetching the scan list fails because there are not enough available scans
- refresh again
- fails again
...

PR #1162, opened 2025-06-18
Fix BFT refresh retry loop

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/1160

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
@@ -884,75 +884,85 @@ object BftScanConnection {
     )(implicit tc: TraceContext): Future[ScanConnections] = {
       val currentState @ BftState(currentScanConnections, currentFailed) =
         currentScanConnectionsRef.get()
-      val currentScans = (currentScanConnections.keys ++ currentFailed.keys).toSet
       logger.info(s"Started refreshing scan list from $currentState")
-      getScans(connection).flatMap { scansInDsoRules =>
-        val newScans = scansInDsoRules.filter(scan => !currentScans.contains(scan.publicUrl))
-        val removedScans = currentScans.filter(url => !scansInDsoRules.exists(_.publicUrl == url))
-        if (scansInDsoRules.isEmpty) {
-          // This is expected on app init, and is retried when building the BftScanConnection
-          Future.failed(
-            io.grpc.Status.FAILED_PRECONDITION
-              .withDescription(
-                s"Scan list in DsoRules is empty. Last known list: $currentState"
-              )
-              .asRuntimeException()
-          )
-        } else if (newScans.isEmpty && removedScans.isEmpty && currentFailed.isEmpty) {
-          logger.debug("Not updating scan list as there are no changes.")
-          Future.successful(currentState.scanConnections)
+
+      for {
+        // if the previous state had too many failed scans, we cannot fetch the new list of scans.
+        // thus, we retry all failed connections first.
+        (retriedScansFailedConnections, retriedScansSuccessfulConnections) <- attemptConnections(
+          currentFailed.map { case (uri, (_, svName)) =>
+            DsoScan(uri, svName)
+          }.toSeq
+        )
+        retriedCurrentState = BftState(
+          currentScanConnections ++ retriedScansSuccessfulConnections,
+          retriedScansFailedConnections.toMap,
+        )
+        // this state will be used to fetch the scans in the next call
+        _ = currentScanConnectionsRef.set(retriedCurrentState)
+        // these will be BFT-read, failing if there's no consensus
+        scansInDsoRules <- getScans(connection)
+        newState <- computeNewState(retriedCurrentState, scansInDsoRules)
+      } yield {
+        currentScanConnectionsRef.set(newState)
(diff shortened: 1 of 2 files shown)


apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
```diff
@@ -278,18 +285,85 @@ class BftScanConnectionTest
...
+      clock.advance(Duration.ofSeconds(refreshSeconds + 1))
+      // even after advancing it shouldn't refresh yet, as that's less than refreshSeconds
+      clock.advance(Duration.ofSeconds(1))
+      clock.advance(Duration.ofSeconds(1))
+      clock.advance(Duration.ofSeconds(1))
+      clock.advance(Duration.ofSeconds(1))
 
       // eventually the refresh goes through and the second connection is used
       eventually() {
+        refreshCalled.intValue() should be(1)
```
OriolMunoz-da, line 305: mostly tests the previous pr: https://github.com/hyperledger-labs/splice/pull/1151

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -884,75 +884,85 @@ object BftScanConnection {
...
-          logger.debug("Not updating scan list as there are no changes.")
-          Future.successful(currentState.scanConnections)
-        } else {
-          for {
-            (newScansFailedConnections, newScansSuccessfulConnections) <- attemptConnections(
-              newScans
+
+      for {
+        // if the previous state had too many failed scans, we cannot fetch the new list of scans.
+        // thus, we retry all failed connections first.
```
rautenrieth-da, line 891: This will catch the case when connections go from "failed" to "successful" between calls to `refresh()`.

What about the other way round? If a scan becomes unavailable, will this class ever move it from `openConnections` to `failedConnections` in the state? There is `SingleScanConnection.getStatus` for checking the health status of the remote.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -884,75 +884,85 @@ object BftScanConnection {
...
-              throw io.grpc.Status.FAILED_PRECONDITION
-                .withDescription(
-                  s"There are not enough Scans to satisfy f=${connections.f}. Will be retried. State: $newState"
-                )
-                .asRuntimeException()
-            } else {
-              connections
+            .asRuntimeException()
+        )
+      } else if (newScans.isEmpty && removedScans.isEmpty && currentFailed.isEmpty) {
```
rautenrieth-da: Why is `currentFailed` included in this check?

And why do we have this check in the first place? `attemptConnections(newScans)` and `removedScans.foreach` are no-ops if the sequences are empty, is it just to have better log output?

If it's about logging, then it could be interesting to log the values of `removedScans`, `newScansSuccessfulConnections`, and `newScansFailedConnections`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
OriolMunoz-da: even if all scans have an open connection, when running `getScans` they still need to reach consensus, so the behavior will effectively be the same

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
OriolMunoz-da: > Why is currentFailed included in this check?

leftover, cood gatch!

but as you say, the check doesn't make much sense, I'll remove it in favor of logging

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
rautenrieth-da: If a scan becomes unavailable and we still leave it in `openConnections`, won't it cause `BftScanConnection` to sometimes pick it for bft calls? Wouldn't it be better to move temporarily unreachable scans to `failedConnections`?

OriolMunoz-da: isn't then the problem where a temporary fluke on a scan will be marked as failed until the next refresh?

> Wouldn't it be better to move temporarily unreachable scans to failedConnections?

or how do you understand "temporarily"?

rautenrieth-da: I was thinking about "sv runbook is registered in dsoRules, but has been unhealthy for 4 days in a row" kind of temporary. But maybe that is not realistic, as on mainnet we would just offboard the SV.

rautenrieth-da (approved):

moritzkiefer-da (approved):
nice thanks

Issue #537, opened 2025-06-05
`Specified different active version and CHARTS_VERSION` in cncluster reset


canton-network-da:
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#18916). Original author: @isegall-da ]

## What is this about?

For `cncluster apply_operator` we no longer need to define CHARTS_VERSION in .envrc.vars, we only edit config.yaml (see cluster/README). But then, `cncluster reset` fails with e.g. `"Specified different active version and CHARTS_VERSION: { type: 'remote', version: '0.3.19' } - { type: 'local' }"`.

Clearly one can then do `CHARTS_VERSION=0.3.19 cncluster reset`, or we can go back to adding the CHARTS_VERSION to .envrc.vars just to make `cncluster reset` happy, but ideally we'd just fix `down` to not complain about versions.

## How important is this and why?

The workaround is fairly simple, so not top priority.

PR #1218, opened 2025-06-25
[static] Remove charts version and override version

nicu-da:
Got fed up with setting the same version in multiple places again and again.

also fixes https://github.com/hyperledger-labs/splice/issues/537

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


cluster/pulumi/common/src/config/migrationSchema.ts
```diff
@@ -1,34 +1,29 @@
...
 import { z } from 'zod';
 
-import { CHARTS_VERSION, CnChartVersion, parsedVersion } from '../artifacts';
+import { CnChartVersion, parsedVersion } from '../artifacts';
 import { spliceEnvConfig } from './envConfig';
 
 export const defaultActiveMigration = {
   id: 0,
-  version: CHARTS_VERSION,
+  version: 'local',
```
moritzkiefer-da, line 10: why is this change ok? Because I can always overwrite it with a config file? Does that work if I want to deploy local pulumi code but remote charts?

moritzkiefer-da (approved):
nice, probably worth a hdm test

cluster/pulumi/common/src/config/migrationSchema.ts
nicu-da: This is basically just the default for running locally with `cncluster`, or deploying in CI the latest version.
The pulumi code isn't influenced by this right? Only which charts are used. So you can always just set an active version and it will use that version of the charts

moritzkiefer-da: hm right should work I guess

PR #1221, opened 2025-06-25
[wip] Fix txlog parser blowing up on expired amulet as input

### LINKED ISSUE #1215: Wallet txlog parsers blow up on CreateTransferOfferInstruction when an expired amulet is used as the input
The problem seems to be this line here https://github.com/hyperledger-labs/splice/blob/a350d25da730e0734df8d1bf13ef030ce79f9f57/apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/UserWalletTxLogParser.scala#L589

We make the assumption that `parseTrees` returns a single event but never check for it. Usually we do indeed only get the transfer but we can also include the unlock. We probably want a `mergeBalanceChangesIntoTransfer` in there or something.

We should check both wallet and scan for similar patterns that may be problematic. Might just wanna remove the `setEventId` in favor of having an event id passed into `meregBalanceChangesIntoTransfer`. It seems like calling that on its own is essentially always a bad idea.

OriolMunoz-da:
Fixes #1215

TODO: let CI blow up
TODO: check scan

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletTxLogTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletTxLogTimeBasedIntegrationTest.scala
@@ -15,12 +15,18 @@ import org.lfdecentralizedtrust.splice.util.{
 }
 import org.lfdecentralizedtrust.splice.wallet.store.{
   BalanceChangeTxLogEntry,
+  PartyAndAmount,
   TransferTxLogEntry,
+  TxLogEntry,
   TxLogEntry as walletLogEntry,
 }
 import com.digitalasset.canton.HasExecutionContext
+import com.digitalasset.canton.data.CantonTimestamp
+import org.lfdecentralizedtrust.splice.codegen.java.splice.api.token.transferinstructionv1.TransferInstruction
+import org.lfdecentralizedtrust.splice.http.v0.definitions.TransferInstructionResultOutput.members
 
 import java.time.Duration
+import java.util.UUID
 
 class WalletTxLogTimeBasedIntegrationTest
     extends IntegrationTestWithSharedEnvironment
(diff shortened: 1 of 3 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletTxLogTimeBasedIntegrationTest.scala
```diff
@@ -337,5 +344,144 @@ class WalletTxLogTimeBasedIntegrationTest
...
+                logEntry.subtype.value shouldBe TxLogEntry.TransferTransactionSubtype.CreateTokenStandardTransferInstruction.toProto
+                logEntry.transferInstructionCid shouldBe offer.contractId
+                logEntry.transferInstructionReceiver shouldBe bobParty.toProtoPrimitive
+                logEntry.transferInstructionAmount shouldBe Some(BigDecimal(transferAmount))
+                logEntry.description shouldBe "transfer offer description"
+                // No balance is transferred here so receivers is empty
+                logEntry.receivers shouldBe empty
+                val expectedExtraLock = BigDecimal(transferAmount - lockAmount)
+                logEntry.sender.value.amount should beWithin(
+                  -expectedExtraLock * 1.25,
```
OriolMunoz-da, line 438: fees seem very large, open for alternatives

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/UserWalletTxLogParser.scala
```diff
@@ -1013,32 +1015,15 @@ object UserWalletTxLogParser {
...
     /** Given a parsing state where the parser has encountered exactly one transfer and zero or more balance changes,
       * returns a parsing state where all the balance changes have been merged into the transfer event.
       *
       * This is useful for app payments where the payment collection first unlocks locked amulets and immediately uses
       * them for a transfer. In this case, we only want to display one balance change for the user.
       */
     def mergeBalanceChangesIntoTransfer(
-        transactionSubtype: TransferTransactionSubtype
+        transactionSubtype: TransferTransactionSubtype,
+        transferEventIdOverride: Option[String],
```
OriolMunoz-da, line 1032: I hate that it's an Option, but some cases I really don't know what to assign it to

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletTxLogTimeBasedIntegrationTest.scala
```diff
@@ -337,5 +344,144 @@ class WalletTxLogTimeBasedIntegrationTest
...
+              // transfer instruction
+              bobTxs should have size 1
+              (aliceTxs, bobTxs)
+            }
+          }
+
+          checkTxHistory(
+            aliceWalletClient,
+            Seq(
+              // create transfer instruction
```
moritzkiefer-da, line 427: should we check that the transaction really includes the unlock? e.g. fetch the update id from scan or the ledger aPI and check that it contains the OwnerExpireLock exercise.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/UserWalletTxLogParser.scala
```diff
@@ -954,12 +955,13 @@ object UserWalletTxLogParser {
...
     def setBalanceChangeSubtype(
-        transactionSubtype: BalanceChangeTransactionSubtype
+        transactionSubtype: BalanceChangeTransactionSubtype,
+        eventId: String,
     ): State = {
       State(
         entries = entries.map {
           case b: BalanceChangeTxLogEntry =>
-            b.copy(subtype = Some(transactionSubtype.toProto))
+            b.copy(subtype = Some(transactionSubtype.toProto), eventId = eventId)
```
moritzkiefer-da, line 970: why is this ok? This is still mapping over a list and assigning the same event id to all of them. Don't you have the same issue there potentially? The safe options only seem to be:
1. You only have one event.
2. You merge events

PR #1224, opened 2025-06-25
Update SvSvStoreTest.scala

OriolMunoz:
This is just a test that PRs from non-contributors run the right code.


OriolMunoz-da:
And [here](https://github.com/hyperledger-labs/splice/actions/runs/15879591095/job/44776215851#step:4:3879) we have proof that the right code is ~running~ being compiled.

cc @moritzkiefer-da @isegall-da

PR #1225, opened 2025-06-25
Add sql index initialization trigger

rautenrieth-da:
Fixes https://github.com/hyperledger-labs/splice/issues/1242

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/automation/ValidatorAutomationService.scala
```diff
@@ -240,6 +243,13 @@ class ValidatorAutomationService(
       )
     }
   }
+
+  registerTrigger(
+    SqlIndexInitializationTrigger(
```
rautenrieth-da, line 247: Some apps run multiple automation services. I have picked one automation service in each app to be responsible for running initialization triggers. This is a bit error prone, but I didn't want to introduce yet another kind of automation service.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/InitializationTrigger.scala
```diff
@@ -0,0 +1,142 @@
...
+
+  override def closeAsync(): Seq[AsyncOrSyncCloseable] =
+    Seq(
+      SyncCloseable(
+        "promise blocking trigger body",
+        startedPromise.trySuccess(UnlessShutdown.AbortedDueToShutdown).discard,
+      ),
+      AsyncCloseable(
+        "initialization trigger body",
+        finishedPromise.future,
```
ray-roestenburg-da: If the finishedPromise is never completed, this will never close?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/InitializationTrigger.scala
```diff
@@ -0,0 +1,142 @@
...
+  def complete()(implicit traceContext: TraceContext): FutureUnlessShutdown[Unit]
+
+  implicit private val loggingContext: ErrorLoggingContext =
+    ErrorLoggingContext.fromTracedLogger(logger)(TraceContext.empty)
+
+  private val stateRef =
+    new AtomicReference[InitializationTriggerState](InitializationTriggerState.Initial)
+  private val startedPromise = Promise[UnlessShutdown[Unit]]()
+  private val finishedPromise = Promise[UnlessShutdown[Unit]]()
+
```
ray-roestenburg-da: nitpick, I would probably just add `val started= startedPromise.future` and `val finished = finishedPromise.future` here and then use those in the rest of the code code (probably wrapper in FutureUnlessShutdown since you are using that)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -0,0 +1,316 @@
...
+    Try(config.config.getString("properties.currentSchema")).toOption.getOrElse("public")
+
+  private def listIndexes(
+      schemaName: String
+  )(implicit traceContext: TraceContext): FutureUnlessShutdown[Map[String, IndexData]] = {
+    storage.query(
+      sql"""
+        select indexname, tablename, indexdef
+        from pg_indexes
+        where schemaname = $schemaName
```
OriolMunoz-da: you should be able to use `current_schema` too, so you don't need to get it from config and such: https://stackoverflow.com/questions/69837274/postgres-get-current-schema-name

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -0,0 +1,316 @@
...
+
+  private def completePostgres(
+      schemaName: String
+  )(implicit traceContext: TraceContext): FutureUnlessShutdown[Unit] = {
+    for {
+      actual <- listIndexes(schemaName)
+      unexpected = actual.keySet -- expectedIndexes.keySet
+      _ = if (unexpected.nonEmpty) {
+        logger.debug(s"Found SQL indexes: $actual")
+        logger.warn(s"Found unexpected SQL indexes: ${unexpected.mkString(", ")}. ")
```
OriolMunoz-da: why do we care to check for those? I'd have assumed any index not defined in this file was going to be flyway anyway

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  private def completePostgres(
+      schemaName: String
+  )(implicit traceContext: TraceContext): FutureUnlessShutdown[Unit] = {
+    for {
+      actual <- listIndexes(schemaName)
+      unexpected = actual.keySet -- expectedIndexes.keySet
+      _ = if (unexpected.nonEmpty) {
+        logger.debug(s"Found SQL indexes: $actual")
```
OriolMunoz-da: nit: shouldn't this be outside the if?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -0,0 +1,316 @@
...
+        logger.warn(s"Found unexpected SQL indexes: ${unexpected.mkString(", ")}. ")
+      }
+      _ <- MonadUtil.sequentialTraverse(expectedIndexes.toSeq) {
+        case (indexName, IndexAction.ManagedByFlyway) =>
```
OriolMunoz-da: likewise, why do we care to validate these?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -0,0 +1,316 @@
...
+          if (!actual.contains(indexName)) {
+            logger.warn(
+              s"Index $indexName is expected to have been created by Flyway, but was not found"
+            )
+          }
+          FutureUnlessShutdown.unit
+        case (indexName, IndexAction.Drop) =>
```
OriolMunoz-da: isn't `drop index` quick to execute? Do we need to do that in this trigger?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/InitializationTrigger.scala
ray-roestenburg-da: (I'm not sure why you don't also try to complete this future with (UnlessShutdown.AbortedDueToShutdown) here)

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/automation/ValidatorAutomationService.scala
ray-roestenburg-da: Error prone in the sense of that we could by mistake add it to many for the same app?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -0,0 +1,316 @@
...
+          storage.update(sqlu"drop index concurrently if exists #$indexName", indexName)
+        case (indexName, IndexAction.Create(action)) =>
+          if (!actual.contains(indexName)) {
+            logger.info(s"Creating index $indexName")
+            storage
+              .update(action, indexName)
+              .thereafter(_ => logger.info(s"Finished creating index $indexName"))
+          } else {
+            FutureUnlessShutdown.unit
+          }
```
OriolMunoz-da: there's a problem: this is not sufficient
`CREATE INDEX CONCURRENTLY` can create invalid indexes that will *NOT* be used, but that are still returned when you list indexes. See [here](https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY):

> The recommended recovery method in such cases is to drop the index and try again to perform CREATE INDEX CONCURRENTLY. 

Also I like [this SO answer](https://stackoverflow.com/questions/77236198/postgres-is-there-an-exhaustive-list-of-causes-that-make-create-index-concurre)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -0,0 +1,316 @@
...
+    /** Drop this index if it exists. */
+    final case object Drop extends IndexAction
+
+    /** Create this index if it does not exist. */
+    final case class Create(action: DBIOAction[?, NoStream, Effect.Write & Effect.Transactional])
+        extends IndexAction
+  }
+
+  /** All indexes created by Flyway after applying all migrations in `apps/common/src/main/resources/db/migration/canton-network/postgres/stable` */
+  val flywayIndexes: Map[String, IndexAction] = Map(
```
ray-roestenburg-da: Would it not be possible to parse the .sql files from flyway to get these names? (or get it from the flyway migration table or something like that) This is a lot of extra admin I would like to avoid, easy to forget.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -0,0 +1,316 @@
...
+  /** Indexes managed by this trigger class */
+  val customIndexes: Map[String, IndexAction] = Map(
+    "updt_hist_crea_hi_mi_ci_import_updates" -> IndexAction.Create(sqlu"""
+        create index concurrently if not exists updt_hist_crea_hi_mi_ci_import_updates
+        on update_history_creates (history_id, migration_id, contract_id)
+        where record_time = -62135596800000000
+         """)
+  )
+
+  val defaultExpectedIndexes: Map[String, IndexAction] = flywayIndexes ++ customIndexes
```
ray-roestenburg-da: how do you verify that the custom index is already covered by the flyway indexes (index exists on the same columns)

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/automation/ValidatorAutomationService.scala
ray-roestenburg-da: I guess another solution would be to create a InitAutomationService, have that only register the init trigger, and add that service to all apps? (I guess that would be overkill, since the automationservice does more than just manage triggers, so yeah I think this is fine)

ray-roestenburg-da (commented):
Not completely convinced by the `IndexAction.ManagedByFlyway` solution yet. if you add an index to the flyway configuration but forget to add it here, it fails in runtime? (maybe I'm getting that wrong)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  /** Indexes managed by this trigger class */
+  val defaultIndexActions: List[IndexAction] = List(
+    IndexAction.Create(
+      indexName = "updt_hist_crea_hi_mi_ci_import_updates",
+      createAction = sqlu"""
+        create index concurrently if not exists updt_hist_crea_hi_mi_ci_import_updates
+        on update_history_creates (history_id, migration_id, contract_id)
+        where record_time = -62135596800000000
```
OriolMunoz-da:
```suggestion
        where record_time = ${CantonTimestamp.MinValue}
```

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTriggerStoreTest.scala
```diff
@@ -0,0 +1,283 @@
...
+          .failOnShutdown
+        indexNamesBefore <- listIndexNames()
+        _ = indexNamesBefore should contain("test_index")
+        _ <- runTriggerUntilAllTasksDone(trigger)
+        indexNamesAfter <- listIndexNames()
+        _ = indexNamesAfter should not contain ("test_index")
+      } yield succeed
+    }
+
+    "do not create an index if it already exists" in {
```
OriolMunoz-da, line 107: this is only enforced because of the `if not exists` in the statements.
Likewise, invalid detection only works because of `concurrently`.
Should we check that both are _always_ specified in the statements?

rautenrieth-da: Good point!
`if not exists` is kind of enforced by `DbStorageIdempotency`, which will retry the create statement and would result in the test failing.
I don't think we need to always enforce `concurrently`, perhaps you have a good reason to create an index in a blocking way.
I will still add a test that checks both properties for the default indexes.

OriolMunoz-da: >  I don't think we need to always enforce concurrently, perhaps you have a good reason to create an index in a blocking way.

I don't think there's any reason:
- if it's fast, run it on flyway
- if it's not, without `concurrently` postgres locks the table against writes so you'll have downtime until the index is done - therefore the trigger should always use `concurrently`

PR #1244, opened 2025-06-26
transfer participants extraction for scan bigquery

stephencompall-DA:
SQL to retrieve all unique parties that were party to a transfer from ledger beginning up to record time, from Scan BigQuery. We only look at receivers because being a receiver at one point is the only way to become a sender.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


stephencompall-DA:
> I might not have full context here but reward minting is a self-transfer where the only output is your sender change and you are not listed explicitly as a receiver. Aren't you missing that?

It depends on what we mean by "transfer participant". Also, can those be incorporated simply by including the `.transfer.sender` in the union from the same choice arguments?

apps/app/src/test/resources/transfer-participants.sql
```diff
@@ -0,0 +1,27 @@
...
+         AND e.template_id_entity_name = 'AmuletRules'
+         AND (e.migration_id < migration_id
+           OR (e.migration_id = migration_id
+               AND e.record_time <= UNIX_MICROS(as_of_record_time)))
+       LIMIT 100) -- TODO (DACH-NY/canton-network-internal#703) remove limit for full test
+  SELECT DISTINCT JSON_VALUE(TransferOutput,
+                             -- .receiver
+                             '$.record.fields[0].value.party')
+  FROM daml_TransferOutput_jsons
+       INNER JOIN daml_TransferOutput_jsons.TransferOutput_array AS TransferOutput;
```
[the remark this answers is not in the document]
stephencompall-DA: The syntax is tricky, I have to think about it a bit more.

apps/app/src/test/resources/transfer-participants.sql
```diff
@@ -0,0 +1,27 @@
...
+         AND e.template_id_entity_name = 'AmuletRules'
+         AND (e.migration_id < migration_id
+           OR (e.migration_id = migration_id
+               AND e.record_time <= UNIX_MICROS(as_of_record_time)))
+       LIMIT 100) -- TODO (DACH-NY/canton-network-internal#703) remove limit for full test
+  SELECT DISTINCT JSON_VALUE(TransferOutput,
+                             -- .receiver
+                             '$.record.fields[0].value.party')
+  FROM daml_TransferOutput_jsons
+       INNER JOIN UNNEST(daml_TransferOutput_jsons.TransferOutput_array) AS TransferOutput;
```
stephencompall-DA: Pulled from [the docs here](https://cloud.google.com/bigquery/docs/arrays#querying_array-type_fields_in_a_struct) (this is JSON, not a struct, but a similar idea applies)

apps/app/src/test/resources/transfer-participants.sql
```diff
@@ -0,0 +1,27 @@
...
+                               -- .transfer.outputs[]
+                               '$.record.fields[0].value.record.fields[3].value.list.elements')
+              TransferOutput_array
+       FROM mainnet_da2_scan.scan_sv_1_update_history_exercises e
+       WHERE e.choice = 'AmuletRules_Transfer'
+         AND e.template_id_module_name = 'Splice.AmuletRules'
+         AND e.template_id_entity_name = 'AmuletRules'
+         AND (e.migration_id < migration_id
+           OR (e.migration_id = migration_id
+               AND e.record_time <= UNIX_MICROS(as_of_record_time)))
```
isegall-da: Does this buy us anything given that the request was "all", and we're anyway putting "<= now" ?

apps/app/src/test/resources/transfer-participants.sql
```diff
@@ -0,0 +1,27 @@
...
+       LIMIT 100) -- TODO (DACH-NY/canton-network-internal#703) remove limit for full test
+  SELECT DISTINCT JSON_VALUE(TransferOutput,
+                             -- .receiver
+                             '$.record.fields[0].value.party') party_id
+  FROM daml_TransferOutput_jsons src
+       INNER JOIN UNNEST(src.TransferOutput_array) AS TransferOutput;
```
isegall-da: Not sure I understand this INNER JOIN, but if it works then fine...

apps/app/src/test/resources/transfer-participants.sql
```diff
@@ -0,0 +1,27 @@
...
+                               '$.record.fields[0].value.record.fields[3].value.list.elements')
+              TransferOutput_array
+       FROM mainnet_da2_scan.scan_sv_1_update_history_exercises e
+       WHERE e.choice = 'AmuletRules_Transfer'
+         AND e.template_id_module_name = 'Splice.AmuletRules'
+         AND e.template_id_entity_name = 'AmuletRules'
+         AND (e.migration_id < migration_id
+           OR (e.migration_id = migration_id
+               AND e.record_time <= UNIX_MICROS(as_of_record_time)))
+       LIMIT 100) -- TODO (DACH-NY/canton-network-internal#703) remove limit for full test
```
isegall-da: Mostly for my own education, I ran this now with LIMIT 10, it returned 5 party IDs. That's because I'm applying the limit before the DISTINCT, right? So I'm selecting 10 transfers, then taking their recipients, and there's 5 unique recipients there. Correct?

isegall-da: Yeah, removed the DISTINCT and reran, indeed it returned 11 (why 11 and not 10?!) party IDs, with duplications.

apps/app/src/test/resources/transfer-participants.sql
isegall-da: I think I'd appreciate a walkthrough of this query, would be glad to better understand it (with anticipation that I'll need to write such queries myself in the not-too-distant future). Let's chat on Monday.

moritzkiefer-da:
> It depends on what we mean by "transfer participant". Also, can those be incorporated simply by including the .transfer.sender in the union from the same choice arguments?

Yes just including transfer.sender sounds fine to me.

apps/app/src/test/resources/transfer-participants.sql
stephencompall-DA: A couple things:

1. As this is concurrent with ingestion, it's best to choose a time where ingestion has likely stabilized, so that there's no chance of partial data.
2. If you refactor or alter the query later, you can compare results knowing it's against the same set of transactions.

apps/app/src/test/resources/transfer-participants.sql
stephencompall-DA: The Scala analogue to this SQL is `.take(10).flatten.distinct`; the result could be empty or have a million items, the 10 doesn't really constrain you in that way.

isegall-da:
> > We only look at receivers because being a receiver at one point is the only way to become a sender.
> 
> I might not have full context here but reward minting is a self-transfer where the only output is your sender change and you are not listed explicitly as a receiver. Aren't you missing that?

@moritzkiefer-da I'm missing something here. How come you're not listed as the received? It's a self-transfer, so you are the receiver, aren't you? Or is it implemented as a "send to nobody", so all outputs are change to the sender?

apps/app/src/test/resources/transfer-participants.sql
stephencompall-DA: [Longer explanation of the underlying syntax.](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#correlated_join)

ray-roestenburg-da (approved):

apps/app/src/test/resources/transfer-participants.sql
```diff
@@ -0,0 +1,40 @@
...
+  RETURNS array<string> AS (ARRAY(
+    SELECT JSON_VALUE(TransferOutput,
+                      -- .receiver
+                      '$.record.fields[0].value.party')
+    FROM UNNEST(TransferOutput_array) AS TransferOutput));
+
+WITH daml_Transfer_jsons AS (
+       SELECT JSON_VALUE(e.argument,
+                         -- .transfer.sender
+                         '$.record.fields[0].value.record.fields[0].value.party')
```
ray-roestenburg-da, line 53: is it possible to create functions for these kind of 'paths', or is that too far? probably easy to use for others.

stephencompall-DA: I'm not sure, it depends on what functionality is available in temp function signatures.

ray-roestenburg-da: maybe have a look?

stephencompall-DA: I've changed total-supply to include this (as well as this file); what do you think?

ray-roestenburg-da: I like it! I think it gives a good basis for reuse and ease of use.

PR #1266, opened 2025-06-30
Fix docker-compose docs re migrating to auth

isegall-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeValidatorFrontendIntegrationTest.scala
```diff
@@ -276,20 +276,23 @@ class DockerComposeValidatorFrontendIntegrationTest
           "participant"
         )
       }
-    }
-  }
 
-  "docker-compose based validator with auth works" in { _ =>
-    val validatorUserPassword = sys.env(s"COMPOSE_VALIDATOR_WEB_UI_PASSWORD")
+      clue("Stop the validator (without wiping its data)") {
```
isegall-da, line 289: The change here is simply moving the auth validator from being a separate test to running as part of the same test after just stopping the non-auth validator.

I thought of testing the migration of non-admin users, but it seems quite non-trivial to get a token for the admin endpoint, so I concluded it's not worth it. Happy for pushback if you disagree (+ ideas on how to obtain a token for the endpoint)

moritzkiefer-da: > but it seems quite non-trivial to get a token for the admin endpoint

is it not just making the oauth2 client credentials request? which is just a single curl command

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeValidatorFrontendIntegrationTest.scala
```diff
@@ -276,20 +276,23 @@ class DockerComposeValidatorFrontendIntegrationTest
...
 
-    withComposeValidator(
-      extraClue = "with auth",
-      startFlags = Seq("-a"),
-      extraEnv = Seq(
-        "GCP_CLUSTER_BASENAME" -> "cidaily" // Any cluster should work, as long as its UI auth0 apps were created with the localhost callback URLs
-      ),
-    ) {
       withFrontEnd("frontend") { implicit webDriver =>
+        val validatorUserPassword = sys.env(s"COMPOSE_VALIDATOR_WEB_UI_PASSWORD")
```
moritzkiefer-da, line 304: where do we test that anything actually gets preserved? I would have expected that we check the balance or something like this

moritzkiefer-da (approved):
thx

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeValidatorFrontendIntegrationTest.scala
isegall-da: Isn't that one of those that are capped in Auth0 thus the caches that we keep in k8s secrets?

isegall-da: Might be the trigger to try out mock-oauth which I wanted to do at some point anyway...

moritzkiefer-da: I thought we raised all those caps? The validator itself also makes the same request no? So at worst you're turning 1 into 2.

isegall-da: We did, I just don't remember if we made them infinite or just higher, and we lost access to seeing the quotas ourselves. I can probably ask helpsesk

isegall-da: Checked with helpdesk. Quota is 10k, and we had months where we exceeded 5k, so doubling that would be bad. I did test manually that re-onboarding the users works, so considering that good enough for now.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeValidatorFrontendIntegrationTest.scala
isegall-da: Added

PR #1273, opened 2025-07-01
daml: rate limit casting votes

meiersi-da:
Both for votes on a VoteRequest and for updating an SVs AmuletPriceVote.

Addresses Quanstamp CC-1.

@moritzkiefer-da : given that we anyways have a Daml upgrade on `main`, I suggest that just merge this one directly.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


docs/src/release_notes.rst
```diff
@@ -28,6 +41,14 @@ Upcoming
     walletPayments     0.1.9
     ================== =======
 
+    Additionally, addressing CC-1 requires updating the ``DsoRules`` configuration to enforce
+    a vote cooldown time of 1 minute by setting:
+
+    .. code-block:: none
+
+      voteCooldownTime.microseconds = 60000000
```
meiersi-da: @moritzkiefer-da : I haven't yet validated that this is the exact way how a reltime value is displayed. If you know by heart that this is right, then I'll happily take your word for it. Otherwise, I'll start a local frontend instance to double-check.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -854,7 +865,9 @@ template DsoRules with
         amuletPrice : Decimal
       controller sv
       do
-        void $ fetchAndArchive (ForOwner with dso; owner = sv) voteCid
+        pastVote <- fetchAndArchive (ForOwner with dso; owner = sv) voteCid
+        -- rate limit, so that an SV cannot block round advancement by changing their price vote too often
+        enforceCooldown ("voteCooldownTime for " <> partyToText sv) config.voteCooldownTime (Some pastVote.lastUpdatedAt)
```
meiersi-da: @fayi-da (FYI: @ray-roestenburg-da ) : this change and the one above imply that submitting a second vote or price update within the cooldown time will lead to a server error. Questions given that you've recently worked on the SV UI:

1. Do you know whether the exception message will be shown to the user?
2. How difficult would it be for you to change the UI to disable the submission button until the cooldown is over?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -402,6 +402,8 @@ data Vote = Vote with
       -- ^ Whether the responder accepted the request to execute the action or not.
     reason : Reason
       -- ^ The reason for voting this way.
+    optCastAt : Optional Time
```
moritzkiefer-da, line 404: I'm generally a bit nervous about the model of having optional fields that need an actual vote to be active for two reasons:

1. It adds extra overhead for SVs to manage the votes.
2. It prevents downgrading. Having a sensible default means you usually don't need to set the field for a while so you don't have an issue here. But otherwise you need to allow quite a bit of time until you actually set the field. I guess this is less bad here than it was for featuredAppActivityMarkerAmount as app devs don't rely on DsoRules. 

But I'd still like to find a consistent rule that makes life easy for SVs, app devs and avoids surprise.
I'd say you already vote on the Daml model change. So having a default in the code seems fine and avoids all the extra overhead and potential issues for app devs.

docs/src/release_notes.rst
moritzkiefer-da: that's correct

moritzkiefer-da (approved):
nice thank you! One question/suggestion on default values in Daml based on some of the experiences around featured app markers. Given that this one is only DsoRulse I'm also ok with merging as is but would be good to align on the general principle here.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
meiersi-da: Good points. Happy to switch to turn it on by default and set it to 1 minute *provided* you agree with that default value, do you?

moritzkiefer-da: Yes 1min sounds like a good default.

meiersi-da:
> @moritzkiefer-da : just noticed that several tests are failing, likely due to violating the vote cooldown. I'll wait for you to confirm that you agree with the mechanism I've chosen before fixing the tests.

OK. Proceeding, as I've just seen your review.

Issue #829, opened 2025-06-05
Switch to relying on package names instead of module names


canton-network-da:
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#9197). Original author: @cocreature ]

With https://docs.google.com/document/d/1oKH9VQsvYy2YpBG3QGhM11cB-AujuWgsLa7rb3bRIw4/edit#heading=h.j1o9vy5fqmrz we don't need to worry about module names as much anymore and can instead operate at a package level.

That said, for mainnet launch we can assume that module names are unique as well so this could be tackled afterwards.

canton-network-da:
[ from @cocreature ]: blocked on https://github.com/DACH-NY/canton/pull/17500#pullrequestreview-1905820521 which we should hopefully get with the Canton fork bump next week.

canton-network-da:
[ from @moritzkiefer-da ]: Yes

canton-network-da:
[ from @moritzkiefer-da ]: Hopefully we can fix it by removing the pkgidresolver but we need to do _something_

PR #1277, opened 2025-07-01
[wip] Support interface ingestion in stores

OriolMunoz-da:
TODO: store interface view compute errors
TODO: more tests
TODO: TODOs in the code

letting CI tell me what did I break until now


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
```diff
@@ -271,33 +283,26 @@ object MultiDomainAcsStore {
   }
 
   /** Static specification of a set of create events in scope for ingestion into an MultiDomainAcsStore. */
-  trait ContractFilter[R <: AcsRowData] {
```
OriolMunoz-da, line 274: all deleted methods were unused (and I didn't feel like maintaining them)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
```diff
@@ -315,72 +320,92 @@ object MultiDomainAcsStore {
...
-      templateFilters.contains(PackageQualifiedName(templateCompanion.getTemplateIdWithPackageId))
-
-    override def mightContain(identifier: Identifier): Boolean = {
-      templateFiltersWithoutPackageNames.contains(QualifiedName(identifier))
+      lazy val successfulInterfaces = ev.getInterfaceViews.asScala.filter { case (identifier, _) =>
+        interfaceFiltersWithoutPackageNames.get(QualifiedName(identifier)).exists(_.evPredicate(ev))
+      }
+      lazy val failedInterfaces = ev.getFailedInterfaceViews.asScala.filter {
+        case (identifier, _) =>
+          // for failed interfaces we cannot check the `evPredicate` yet we still must store the row
```
OriolMunoz-da: important for review

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsJdbcTypes.scala
```diff
@@ -309,8 +313,26 @@ object AcsJdbcTypes {
   // this output relies on a *big* invariant: comparing the raw JSON data
   // with the built-in SQL operators <, >, &c, yields equal results to
   // comparing the same data in a data-aware way. That's why we *must* use
-  // numbers-as-numbers in this codec, and why ISO-8601 strings
+  // numbers-as-numbers in both codecs, and why ISO-8601 strings
   // for dates and timestamps are so important.
+
+  private val javaapiDamlRecordCodec = {
```
OriolMunoz-da, line 319: I hate that I had to resurrect this but the one below can't be used with non-codegen DamlRecord

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -730,6 +741,50 @@ final class DbMultiDomainAcsStore[TXE](
...
+      companionClass: ContractCompanion[C, ICid, View],
+      tc: TraceContext,
+  ): Future[Seq[Contract[ICid, View]]] = waitUntilAcsIngested {
+    val interfaceId = companionClass.typeId(companion)
+    val opName = s"listInterfaceViews:${interfaceId.getEntityName}"
+    for {
+      // assumption: most rows will satisfy `AND interface_view IS NOT NULL`, so we don't need to index for it
+      rows <- storage.query(
+        sql"""
+             SELECT contract_id, interface_view, acs.created_at, acs.created_event_blob
```
OriolMunoz-da, line 754: TODO: review query plan with more data than the one in tests (but this shouldn't block reviews)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/JavaDecodeUtil.scala
```diff
@@ -24,8 +24,15 @@ object JavaDecodeUtil {
...
-      Some(companion.fromCreatedEvent(event))
+    if (
+      event.getInterfaceViews
+        .keySet()
+        .asScala
+        .map(QualifiedName(_))
+        .contains(QualifiedName(companion.getTemplateIdWithPackageId))
+    ) {
+      val result = Some(companion.fromCreatedEvent(event))
+      result
```
OriolMunoz-da: same as lines above, wasn't used before iirc

daml/dars.lock
```diff
@@ -50,6 +50,7 @@ splice-dso-governance 0.1.9 9ee83bfd872f91e659b8a8439c5b4eaf240bcf6f19698f884d7d
 splice-dso-governance-test 0.1.15 732d179b72be5861cbff64bf06ece3e13b4cdf4363a963482b7f558e81eafef4
 splice-token-standard-test 1.0.1 d2025d5f0048285367011f0beb76cc3e6f9c4f3a54d3e152d8e7337408454014
 splice-token-test-dummy-holding 0.0.1 1cd171c6c42ab46dc9cf12d80c6111369e00cea5cdf054924b4f26ce94b1ef5b
+splice-token-test-dummy-holding 0.0.2 4f40fb033ef3db89623642c1b494e846097fa32af138b3864a63aa15937a323d
```
OriolMunoz-da, line 56: I thought this wouldn't have been necessary, but I got breakage otherwise :shrug:

token-standard/examples/splice-token-test-dummy-holding/daml/Splice/Api/Token/Test/DummyTwoInterfaces.daml
```diff
@@ -0,0 +1,51 @@
...
+
+module Splice.Api.Token.Test.DummyTwoInterfaces where
+
+import Splice.Api.Token.MetadataV1 qualified as Api.Token.MetadataV1
+import Splice.Api.Token.HoldingV1 qualified as Api.Token.HoldingV1
+import Splice.Api.Token.AllocationV1 qualified as Api.Token.AllocationV1
+import Splice.Api.Token.AllocationRequestV1 qualified as Api.Token.AllocationRequestV1
+import DA.TextMap qualified as TextMap
+
+template DummyTwoInterfaces
```
OriolMunoz-da, line 12: I know this is nonsense, but I really wanted something implementing two interfaces and this seemed easiest

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
```diff
@@ -315,72 +320,92 @@ object MultiDomainAcsStore {
...
-        ev: CreatedEvent
-    ): Option[Contract[?, ?]] = {
-      for {
-        templateFilter <- templateFiltersWithoutPackageNames.get(QualifiedName(ev.getTemplateId))
-        contract <- templateFilter.decodeFromCreatedEvent(ev)
-      } yield contract
+    override def mightContain(
+        templateId: Identifier,
+        implementedInterfaces: Seq[Identifier],
+    ): Boolean = {
```
OriolMunoz-da: important for review: applicable only for exercisedevents

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
```diff
@@ -0,0 +1,30 @@
...
+    -- the index will allow queryiing `WHERE interface_id_qualified_name = some_qualified_interface_id ORDER BY acs_event_number`.
+    -- At time of writing, interfaces cannot really be upgraded (therefore the qualified name would change on every upgrade).
+    -- However, this might change in the future, so we include the package_id in the primary key.
+    -- package_id is in the last position because it's not expected to be used in queries.
+    PRIMARY KEY (interface_id_qualified_name, acs_event_number, interface_id_package_id)
+);
+
+-- to make deletion of an ACS entry by efficient:
+-- some_acs_store will delete the entry and cascade by primary key to this table
+CREATE INDEX ON interface_views_template (acs_event_number);
```
moritzkiefer-da: nit: name the index. unnamed indices are annoying to delete

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsRowData.scala
```diff
@@ -3,16 +3,53 @@
...
+    override val payload: Json = AcsJdbcTypes.payloadJsonFromDefinedDataType(contract.payload)
+    override val createdEventBlob: ByteString = contract.createdEventBlob
+    override val createdAt: Instant = contract.createdAt
+  }
+
+  /** Meant to be used when an interface filter matches, but no template filters do.
+    * Therefore, it can only contain the "bare minimum" fields, but cannot include index columns.
+    * If both an interface filter and a template filter match, then the template filter's AcsRowData should be used instead.
+    */
+  case class AcsRowDataFromInterface(
```
moritzkiefer-da, line 40: what is used when both an interface and a template match?

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStoreTest.scala
```diff
@@ -336,8 +344,62 @@ class DbMultiDomainAcsStoreTest
...
+          .newBuilder()
+          .setCode(n * 100)
+          .setMessage(s"test entry num $n")
+          .build()
+        contract -> failedView
+      }
+
+      for {
+        _ <- initWithAcs()
+        _ <- MonadUtil.sequentialTraverse(contractsToFailedViews) { case (contract, failedView) =>
```
moritzkiefer-da, line 370: how about checking something with both a failed and a successful view?

moritzkiefer-da: probably both:
one contract with a failed and successful view for different interfaces and
multiple contracts in one transaction with failed and successful views for the same interface

daml/dars.lock
moritzkiefer-da: related to https://github.com/hyperledger-labs/splice/issues/1304 to some degree

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
moritzkiefer-da: Why is it ok not to filter here? Couldn't we end up in a situation where you ingest a non-stakeholder contract for which you will never see the archive?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
moritzkiefer-da: same question as below

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsRowData.scala
OriolMunoz-da: it's there?

>     * If both an interface filter and a template filter match, then the template filter's AcsRowData should be used instead.

moritzkiefer-da: ah thanks missed that/saw it after writing that comment

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
```diff
@@ -0,0 +1,30 @@
+CREATE TABLE interface_views_template
```
rautenrieth-da: nit: almost all migration scripts use lower case for SQL statements

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
```diff
@@ -0,0 +1,30 @@
...
+(
+    acs_event_number            bigint not null REFERENCES acs_store_template (event_number) ON DELETE CASCADE,
```
rautenrieth-da: This creates an auto-named constraint, please create an explicitly named one. See `V037__denormalize_update_history.sql` where we had to remove and restore foreign key constraints.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
```diff
@@ -0,0 +1,30 @@
...
+    interface_id_package_id     text   not null,
+    interface_id_qualified_name text   not null,
+
+    -- the DamlRecord corresponding to the interface's view (if available)
+    interface_view              jsonb,
+    -- if the view computation failed, this will contain the error for debugging purposes
+    view_compute_error          jsonb,
+    CONSTRAINT has_view_or_error CHECK ((interface_view is null and view_compute_error is not null) OR
```
rautenrieth-da: Are constraint names global? If yes, consider prefixing it with the table name like we do for indexes.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
```diff
@@ -0,0 +1,30 @@
...
+                                        (view_compute_error is null and interface_view is not null)),
+
+    -- the index will allow queryiing `WHERE interface_id_qualified_name = some_qualified_interface_id ORDER BY acs_event_number`.
+    -- At time of writing, interfaces cannot really be upgraded (therefore the qualified name would change on every upgrade).
+    -- However, this might change in the future, so we include the package_id in the primary key.
+    -- package_id is in the last position because it's not expected to be used in queries.
+    PRIMARY KEY (interface_id_qualified_name, acs_event_number, interface_id_package_id)
```
rautenrieth-da: Not sure if we'll ever need to reference it, but `V037__denormalize_update_history.sql` I even explicitly named primary keys. Better safe than sorry.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/db/WalletTables.scala
```diff
@@ -86,4 +86,5 @@ object WalletTables extends AcsTables {
   val acsTableName: String = "user_wallet_acs_store"
   val externalPartyAcsTableName: String = "external_party_wallet_acs_store"
   val txLogTableName: String = "user_wallet_txlog_store"
+  val internfaceViewsTableName: String = "user_wallet_acs_interface_views"
```
rautenrieth-da:
```suggestion
  val interfaceViewsTableName: String = "user_wallet_acs_interface_views"
```

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
ray-roestenburg-da: +1

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
OriolMunoz-da: I have updated the signature of the method to `shouldArchive`, as that's its purpose and updated the docs.
Let me know if this answers your comment.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
OriolMunoz-da: We have to ingest the failed interface views for debugging purposes. Those do not contain the view, so I cannot use the `evPredicate`. I can only ingest them if the interface matches. Those are filtered out at query time (because they obviously cannot be parsed as Contract). They should be deleted by the shouldArchive... but that's untested, so will do.
I will also update the comment(s) to mention all of this

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
OriolMunoz-da: they aren't (otherwise the template `including all` wouldn't work):
image

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
OriolMunoz-da: F, no, that's a problem:
- Amulet implements Holding
- an incoming Amulet comes where  `payload.dsoParty != theRealDsoParty` with a failed Holding interface
- listInterfaceViews(Holding) will not list it :ok: 
- listContracts(Amulet) will list it, because that don't do post-filtering :warning: 

which means I need to rethink how to handle failed interface views

OriolMunoz-da: @moritzkiefer-da cc @meiersi-da (since you mentioned persisting the interface views)
the easiest solution seems to be to put them in a separate table with all information of the contract (as opposed to the main acs/interface_view tables). On archival (as per [shouldArchive](https://github.com/hyperledger-labs/splice/blob/f6f51d2843d0051245b21bc8bac1dda9cc3a03c8/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala#L299)) any entries will be deleted.

Any obvious objections?

moritzkiefer-da: I don't think that solves the problem. You still have the problem that you might ingest failed interface views of non-stakeholder contracts for which you never see the archive on the transaction stream so shouldArchive will never get called and you stop leaking storage.

I think at a minimum you would need to filter by stakeholder which are exposed in the CreatedEvent through the signatories and observers arguments. That way you can guarantee that you see the ArchivedEvent for that contract as well and won't keep it around forever..

The UI can then display them if needed.

That said, I'm wondering if we really gain all that much by storing them over just logging a warning on ingestion when we see it. It seems like you never expect this to happen in practice so investing in having the UI display it properly and everything seems a fair amount of effort for fairly little gain. @meiersi-da wdyt? Is there a better argument for doing that work other than ensuring the UI can display them as failed instead of just omitting them?

meiersi-da: Interesting indeed. I didn't foresee the challenges wrt filtering. My motivation for storing them was purely debugging, and the ability for the store to report whether it is has a complete set of interface views.

It seems that checking the `acs_delta` flag on created events as per the [design](https://docs.google.com/document/d/1I5w3qVhvXzZDHeiV6ui4XvT5GvYDcPXKDYealjsao1o/edit?tab=t.0#heading=h.j1o9vy5fqmrz) would solve the stale interface view computation errors problem.

That said: just logging the ingestion failure also seems an option. I would though log it at `ERROR` level, as the system really is broken; and fixing it should not be skipped over. Did you think `ERROR` or `WARN` @moritzkiefer-da ?

moritzkiefer-da: Definitely log at ERROR level

moritzkiefer-da: Given that my recommendation would be we don't store them at all and only log at ERROR level. You can query the participant if you need more info for debugging. That means we cannot expose it in the UI but that seems ok. And it seems like a fairly clean extension to add it later if we do discover failing views are common enough to jutify that.

OriolMunoz-da: should we continue ingestion or halt it completely when we encounter such an interface? The reason I'm asking is that then the operator of the app will have to reset the acs/interface tables and whatever else depends on them

moritzkiefer-da: Log an error and continue, halting it completely seems too annoying. I'd rather have a missing contract than have reward collection not working because of a broken interface view somehwere.

meiersi-da: Agreed. Don't stop ingestion. And log at ERROR level.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: https://github.com/hyperledger-labs/splice/pull/1277#issuecomment-3036906994

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
```diff
@@ -0,0 +1,27 @@
...
+(
+    acs_event_number            bigint not null,
+    interface_id_package_id     text   not null,
+    interface_id_qualified_name text   not null,
+
+    -- the DamlRecord corresponding to the interface's view
+    interface_view              jsonb not null,
+
+    -- the index will allow querying `WHERE interface_id_qualified_name = some_qualified_interface_id ORDER BY acs_event_number`.
+    -- At time of writing, interfaces cannot really be upgraded (therefore the qualified name would change on every upgrade).
```
meiersi-da:
```suggestion
    -- At time of writing, interfaces cannot be upgraded using SCU.
    -- The only option is to introduce new interface definitions, 
    -- and thus the qualified name will change on every upgrade.
```

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
```diff
@@ -0,0 +1,27 @@
...
+    -- However, this might change in the future, so we include the package_id in the primary key.
```
meiersi-da: Does the qualified_name include the package-name?

If interfaces become upgradeable, then it is going to be based on allowing to change the definition in the same package name.

So, it's not clear why you'd have to include the `package_id`.

I fear a bit that including the package-id makes the index unnecessarily more expensive.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
```diff
@@ -0,0 +1,27 @@
...
+    -- package_id is in the last position because it's not expected to be used in queries.
+    constraint interface_views_template_pkey primary key (interface_id_qualified_name, acs_event_number, interface_id_package_id),
+    constraint interface_views_template_acs_evt_num_fkey foreign key (acs_event_number) references acs_store_template (event_number) on delete cascade
+);
+
+-- to make deletion of an ACS entry by efficient:
```
meiersi-da: by efficient... missing text?

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
```diff
@@ -0,0 +1,27 @@
...
+-- some_acs_store will delete the entry and cascade by primary key to this table
+create index interface_views_template_aen on interface_views_template (acs_event_number);
```
meiersi-da, line 21: So a foreign key constraint does no automatically define a corresponding index?

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
```diff
@@ -0,0 +1,27 @@
+create table interface_views_template
+(
+    acs_event_number            bigint not null,
+    interface_id_package_id     text   not null,
+    interface_id_qualified_name text   not null,
```
meiersi-da, line 9: add comment stating that this is the qualified name of the definition of the interface. Also clarify whether it includes the package-name or not.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
```diff
@@ -315,72 +319,96 @@ object MultiDomainAcsStore {
...
-  ) extends ContractFilter[R] {
+      interfaceFilters: Map[
+        PackageQualifiedName,
+        InterfaceFilter[?, ?, ?, IR],
+      ],
+  ) extends ContractFilter[R, IR] {
+
+    private val templateFiltersWithoutPackageNames = filtersWithoutPackageNames(templateFilters)
+
+    private val interfaceFiltersWithoutPackageNames = filtersWithoutPackageNames(interfaceFilters)
```
meiersi-da, line 368: That one seems slightly risky to me, as we don't control the packages that implement interfaces, and they could in principle implement another interface with the same module and interface name, but defined in a different package from the one we care about. 

For interfaces we actually do care about the `package_id` and I'd suggest to just match on it on ingestion as well.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
```diff
@@ -315,72 +319,96 @@ object MultiDomainAcsStore {
...
+      lazy val interfaceViews = ev.getInterfaceViews.asScala.filter { case (identifier, _) =>
+        interfaceFiltersWithoutPackageNames.get(QualifiedName(identifier)).exists(_.evPredicate(ev))
+      }
+      lazy val failedInterfaces = ev.getFailedInterfaceViews.asScala.filter {
+        case (identifier, _) =>
+          interfaceFiltersWithoutPackageNames.contains(QualifiedName(identifier))
+      }
+      if (failedInterfaces.nonEmpty) {
+        elc.error(
+          s"Found failed interface views that match an interface id in a filter: $failedInterfaces. " +
```
meiersi-da: Do we have a better name for the filter available here? E.g simple class name?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
```diff
@@ -315,72 +319,96 @@ object MultiDomainAcsStore {
...
+            s"This might be a bug in the daml definition of the interface's view."
```
meiersi-da: Please print the error here; and advise to "resolve the error, and if required, reingest the data as per the documentation."

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
meiersi-da: Actually, reading more of the code I realize that we don't have pacakage names available on interface views (https://github.com/DACH-NY/canton/blob/7175b78a52c1080ea79b6e106a87b1b774266e65/community/ledger-api/src/main/protobuf/com/daml/ledger/api/v2/event.proto#L133-L152), and interface definitions really must be identified with their package-id qualified name.

So, I'd actually turn the index around and filter first by package-id. Thereby allowing the index to apply "suffix truncation" for the internal pages: https://github.com/postgres/postgres/blob/master/src/backend/access/nbtree/README#L823

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
```diff
@@ -0,0 +1,27 @@
...
+
+-- to make deletion of an ACS entry by efficient:
+-- some_acs_store will delete the entry and cascade by primary key to this table
+create index interface_views_template_aen on interface_views_template (acs_event_number);
+
+create table user_wallet_acs_interface_views
+(
+    like interface_views_template including all,
+
+    constraint user_wallet_acs_interface_views_acs_evt_num_fkey foreign key (acs_event_number) references user_wallet_acs_store (event_number) on delete cascade
```
meiersi-da, line 27: Question: can we use index columns here as well?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -730,6 +739,50 @@ final class DbMultiDomainAcsStore[TXE](
...
+            interfaceId,
+            contractId,
+            viewJson,
+            ByteString.copyFrom(createdEventBlob),
+            createdAt.toInstant,
+          )
+          .fold(
+            err =>
+              throw new IllegalStateException(
+                s"Stored a contract that cannot be decoded as interface view $interfaceId: $err"
```
meiersi-da: Let's make debugging easier by also printing a unique identifier for the data that could not be decoded? Or potentially even the data itself?

meiersi-da (approved):
Thanks a lot. Looks like we got some nice interface support with this one!

The main two points that I have are:
- use package-id qualified names for identifying interfaces
- ensure error messages are clear wrt how to handle them, and contain enough information to make debugging possible or even better easy :)

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
OriolMunoz-da: it does not - mariaDB does, but postgres doesn't: https://stackoverflow.com/questions/970562/postgres-and-indexes-on-foreign-keys-and-primary-keys

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
moritzkiefer-da: we could, just don't need any for now

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
OriolMunoz-da: > So, I'd actually turn the index around and filter first by package-id. Thereby allowing the index to apply "suffix truncation" for the internal pages: https://github.com/postgres/postgres/blob/master/src/backend/access/nbtree/README#L823

The queries are done
```
where interface_id_qualified_name = ? order by acs_event_number
```

so if I put the package id first, the index won't be used 

> Does the qualified_name include the package-name?

it does not:

```
/** To support upgrading templates are addressed
  * using their qualified name, ignoring the package id.
  */
final case class QualifiedName(moduleName: String, entityName: String)
```

> So, it's not clear why you'd have to include the package_id.

I think you're right and we might be able to just drop it from the PK? @moritzkiefer-da I know I discussed the package_id stuff with you, so you might know better.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
OriolMunoz-da: > **For interfaces** we actually do care about the package_id and I'd suggest to just match on it on ingestion as well.

why are they different from templates?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
OriolMunoz-da: the error is included: `failedInterfaces` is a Map[Identifier, Status]. I'll rename the variable / add typing to make it clear

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
OriolMunoz-da: no, but we should get all info required afaict, see https://github.com/hyperledger-labs/splice/pull/1277#discussion_r2189215412

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
OriolMunoz-da: we can - there's just no need for them atm afaict for AllocationRequests, as they'll just be listed

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
moritzkiefer-da: So it seems there are two things here:

1. It looks like we never pulled through the change to filter by package names everywhere and only filter by module names in a lot of places. While we should fix this at some point, it seems like interfaces don't really change that. We still control all packages that define interfaces and templates we care about so we can ensure that there are no collisions based on module names.
2. What do we filter for for interfaces? The two options seem to be non-upgradeable (package id, modulename, interface name) and upgradeable (packagename, modulename interface name). I'd view (modulename, interface name) as a hacky version of the upgradeable one. If we can do the upgradeable one directly without introducing issues that seems preferable to me.

moritzkiefer-da: Hm, thinking about this again, maybe we do have an issue for templates. What happens in this case:

1. Someone defines a template `Splice.Amulet:Amulet` in a different package that implements AllocationRequest.
2. We ingest that.
3. We then try to list all contracts listing `Splice.Amulet`. I suspect we might try to include the template and then fail decoding it.

So we definitely need to ensure we include package names for templates. For interfaces it still seems less important for now so I wouldn't make the fact that the events don't include a package name a hard blocker that forces us to filter by package id everywhere. That said, maybe optimizing for making it upgradeable later is also not worth it and just including the package id for now is the better tradeoff.

moritzkiefer-da (approved):
Nice thank you! Simon already did the detailed review and I agree with the review comments

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
OriolMunoz-da: resolving to move conversation to the thread above

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
meiersi-da: OK. Can you also switch to `show` so they are more readable?

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V041__interface_views.sql
moritzkiefer-da: just in case anyone else reads this thread:

The template issue is not new, fix is tracked in #829. We'll prioritize it after finishing the token standard UI for allocation requests.

For interfaces we'll filter by package id and include that in the index.

moritzkiefer-da (approved):
package id changes lgtm thanks!

moritzkiefer-da:
@meiersi-da Oriol pointed out yesterday that we have an issue because we don't upload `splice-api-token-allocation-request-v1` yet by default. While you might think that doesn't matter as any interface implementation will upload it anyway, the GetUpdates request will fail if we include that in the filter before it has been uploaded.


It seems there are two options:

1. Add it to the package config as an optional field and upload it based on that probably defaulting to `None` uploading the current version in this case (otherwise we have the whole downgrading issues again) although that also feels somewhat dodgy. That works but it's a fair amount of work and another set of votes to coordinate among SVs.
2. Always upload it. This actually does not seem that unreasonable to me. Interfaces are not upgradeable so we don't gain anything by being able to coordinate the version other than the initial upload. An interface itself also does nothing only interface implementations do so it also seems ok from a vetting perspective.

I'm leaning towards 2. wdyt?

PR #1297, opened 2025-07-02
Fail hard if migration ID incremented without restoring from migration dump

martinflorian-da:
Fixes https://github.com/DACH-NY/canton-network-internal/issues/730

I gave up on writing a proper test for this (see https://github.com/hyperledger-labs/splice/pull/1297/commits/19c9a1ed1c219b61ef7843de4342bdf6cbb286d5). It seems like there is no easy way to catch runtime errors we throw during app init without crashing all of sbt. (Please shout if you can teach me a way!) I did run my broken test locally though and can confirm that the init failed in the expected way.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -2174,6 +2174,32 @@ class UpdateHistory(
...
+      ec: ExecutionContext,
+      closeContext: CloseContext,
+      tc: TraceContext,
+  ): Future[Option[Long]] = {
+    storage match {
+      case storage: DbStorage =>
+        for {
+          queryResult <- storage.query(
+            sql"""
+        select max(migration_id)
```
moritzkiefer-da: What's the query plan for that? It's not obvious to me that this does not do a linear scan which we should avoid.

I would hope that you can instead query the last_ingested_offset table. For that one a linear scan would be less of an issue.

martinflorian-da: > What's the query plan for that?

It is terrible, thanks for making me check...

> I would hope that you can instead query the last_ingested_offset table

Nice idea, will try that now! If it doesn't work I might be able to speed up the query here by creating another index.

moritzkiefer-da: > Nice idea, will try that now! If it doesn't work I might be able to speed up the query here by creating another index.

indices are not free on a table of that size. I'd be cautious adding one just for a single query.

martinflorian-da: OK it seems to work now with `select max(migration_id) from update_history_last_ingested_offsets`, which also does a linear scan but that table is supposed to stay very small AFAIU.

moritzkiefer-da: Should be ok, we have one per migration id per local party. And it's a query that we only run on startup. And the plan is to remove the "per party" part to scale to more parties.

That said, for this one I'm also less worried about adding an index. Happy either way.

martinflorian-da: Ah I see... I still prefer not to overengineer this; if we create an index we might forget to remove it later when we get rid of all migration ID things.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -2174,6 +2174,31 @@ class UpdateHistory(
 }
 
 object UpdateHistory {
+
+  // Separate method so we can use this without a full UpdateHistory instance.
+  def getHighestKnownMigrationId(
```
moritzkiefer-da, line 2181: add a comment that we deliberately do not filter by store id because we run this too early in the setup where we don't have easy access to that.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/SvApp.scala
```diff
@@ -160,19 +160,28 @@ class SvApp(
...
-                SvCantonIdentifierConfig.default(config)
-              )
-              ParticipantInitializer.ensureParticipantInitializedWithExpectedId(
-                cantonIdentifierConfig.participant,
-                participantAdminConnection,
-                config.participantBootstrappingDump,
-                loggerFactory,
-                retryProvider,
-              )
+              UpdateHistory.getHighestKnownMigrationId(storage).flatMap {
```
moritzkiefer-da, line 163: nit: logic seems a bit duplicated across sv and validator app

moritzkiefer-da (approved):
thanks!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
martinflorian-da: I deliberately don't filter by anything, to avoid having to think which store is the most relevant one :)

moritzkiefer-da: Yes but let's document it please

martinflorian-da: ofc will do!

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/SvApp.scala
martinflorian-da: it is but extracting this into a helper seemed more confusing than spelling it out and accepting the single duplication...

PR #1319, opened 2025-07-03
[ci] Add caching to most of the store read methods for scan

nicu-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
@@ -62,7 +62,7 @@ final case class ScanCacheConfig(
       ttl = NonNegativeFiniteDuration.ofSeconds(30),
       maxSize = 100,
     ),
-    totalAmuletBalanceCache: CacheConfig = CacheConfig(
+    totalAmuletBalance: CacheConfig = CacheConfig(
       ttl = NonNegativeFiniteDuration.ofMinutes(2),
       maxSize = 1000,
     ),
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
@@ -58,7 +58,79 @@ case class ScanAppBackendConfig(
 }
 
 final case class ScanCacheConfig(
-    svNodeStateTtl: NonNegativeFiniteDuration = NonNegativeFiniteDuration.ofSeconds(30)
+    svNodeState: CacheConfig = CacheConfig(
+      ttl = NonNegativeFiniteDuration.ofSeconds(30),
+      maxSize = 100,
+    ),
+    totalAmuletBalanceCache: CacheConfig = CacheConfig(
+      ttl = NonNegativeFiniteDuration.ofMinutes(2),
+      maxSize = 1000,
+    ),
+    amuletRules: CacheConfig = CacheConfig(
+      ttl = NonNegativeFiniteDuration.ofSeconds(30),
+      maxSize = 1,
+    ),
+    ansRules: CacheConfig = CacheConfig(
+      ttl = NonNegativeFiniteDuration.ofSeconds(30),
+      maxSize = 1,
+    ),
+    totalRewardsCollected: CacheConfig = CacheConfig(
+      ttl = NonNegativeFiniteDuration.ofMinutes(2),
+      maxSize = 1,
+    ),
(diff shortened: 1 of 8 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
```diff
@@ -58,7 +58,79 @@ case class ScanAppBackendConfig(
 }
 
 final case class ScanCacheConfig(
```
nicu-da, line 60: will probs merge some of these but currently want to see the output for the metrics and see which one needs some different configs

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
```diff
@@ -58,7 +58,79 @@ case class ScanAppBackendConfig(
...
-    svNodeStateTtl: NonNegativeFiniteDuration = NonNegativeFiniteDuration.ofSeconds(30)
+    svNodeState: CacheConfig = CacheConfig(
+      ttl = NonNegativeFiniteDuration.ofSeconds(30),
+      maxSize = 100,
+    ),
+    totalAmuletBalanceCache: CacheConfig = CacheConfig(
```
moritzkiefer-da: did you do the math how much memory we use if all of those caches are filled? Some are single numbers but it's not clear if some are more complex and need more traffic.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/CachingScanStore.scala
```diff
@@ -67,108 +62,169 @@ class CachingScanStore(
...
-  private val totalAmuletBalanceCache
-      : ScaffeineCache.TracedAsyncLoadingCache[Future, CacheKey, CacheValue] = {
-    ScaffeineCache.buildTracedAsync[Future, DbScanStore.CacheKey, DbScanStore.CacheValue](
-      Scaffeine()
-        .maximumSize(1000),
-      implicit tc => key => store.getTotalAmuletBalance(key),
-      metrics = Some(storeMetrics.totalAmuletBalanceCache),
-    )(logger, "amuletBalanceCache")
-  }
+  private val cacheOfCaches = scala.collection.concurrent
```
moritzkiefer-da, line 65: is there a good argument for initializing this lazily? it seems like ventually you'll end up with all caches existing once you get enough requests so just starting with that might give more predictable behavior

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
```diff
@@ -58,7 +58,79 @@ case class ScanAppBackendConfig(
...
+      ttl = NonNegativeFiniteDuration.ofMinutes(1),
+      maxSize = 1000,
+    ),
+    walletBalance: CacheConfig = CacheConfig(
+      ttl = NonNegativeFiniteDuration.ofMinutes(1),
+      maxSize = 2000,
+    ),
+    amuletConfigForRound: CacheConfig = CacheConfig(
+      ttl = NonNegativeFiniteDuration.ofMinutes(1),
+      maxSize = 1000,
```
moritzkiefer-da, line 91: what's the rule for which have 1k and which have 2k?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/CachingScanStore.scala
nicu-da: not really, only reason why I went this way is because i saw an existing cache that didn't show up in devnet at all in the metrics so wanted to still keep it that way.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
nicu-da: just eye balling it, if it's a key that would have a higher cardinality then 2k, if it's some lower cardinality (like for example rounds) then it's 1000, though I will most likely re-adjust after this gets to devnet and lower/increase some of them

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
nicu-da: no, i went with the dumbest solution for now, we can also assign weights to the cache entries and set max weight based on weight but didn't see anything that would be really memory costly at a first glance. will keep an eye on the metrics though

PR #1337, opened 2025-07-04
Disable getLastImportUpdateId when feature flag is off

rautenrieth-da:


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1713,8 +1713,9 @@ class UpdateHistory(
   def getLastImportUpdateId(
       migrationId: Long
   )(implicit tc: TraceContext): Future[Option[String]] = {
-    storage.query(
-      sql"""
+    if (enableImportUpdateBackfill) {
```
rautenrieth-da, line 1716: The last import update id is used for import update backfilling, the trigger for which is already disabled by the feature flag.

But this query is also used for `migrationInfo`, which is used by regular backfilling. In case import update backfilling is disabled by a feature flag, we just return invalid data (`None`) as the data won't be used.

PR #1342, opened 2025-07-07
Enable bft sequencer connections for SVs

moritzkiefer-da:
[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


cluster/helm/splice-validator/values.schema.json
```diff
@@ -42,11 +42,8 @@
         }
       },
       "else": {
-        "required": ["decentralizedSynchronizerUrl"],
+        "required": [],
```
moritzkiefer-da, line 45: We should improve the UX here a bit. I'm thinking of a `disableSvValidatorBftSequencerConnection: Boolean: false` flag. We enforce that:
1. This is off if you don't have an sv validator.
2. If it is off, you must not set decentralizedSynchronizerUrl and you must set useSequencerConnectionsFromScan.
3. If it is on, you must set both.

I'm leaning towards doing this as a separate change so we can merge this and the Canton upgrade and start testing the actual BFT connections extensively while I work on this in parallel. Definitely will get it into the release though.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvDsoAutomationService.scala
```diff
@@ -394,18 +394,20 @@ class SvDsoAutomationService(
...
-    registerTrigger(
-      new LocalSequencerConnectionsTrigger(
-        triggerContext,
-        participantAdminConnection,
-        internalClientConfig.decentralizedSynchronizerAlias,
-        dsoStore,
-        internalClientConfig.sequencerInternalConfig,
-        config.participantClient.sequencerRequestAmplification,
-        config.domainMigrationId,
+  if (!config.bftSequencerConnection) {
```
nicu-da, line 397: Wondering if it still makes sense to keep this here, can't we do the same thing from the validator app already and thus move all the logic to the validator app?

nicu-da (approved):
Thanks, looks simpler than expected! :shipit:

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvDsoAutomationService.scala
moritzkiefer-da: It's not the same as for the validator because of timing: The LocalSequencerConnectionsTrigger will wait until the sequencer is actually fully working. The validator in single-SV config will try to connect to it immediately which could be before the sequencer is fully working and you might blow up on tombstones or some crap.

I would suggest to just keep it as is and try to delete the single-SV config once we have enough confidence that this one works.

PR #1349, opened 2025-07-07
[ci] Fix tx log backfilling metrics by closing gauges

nicu-da:
fixes https://github.com/hyperledger-labs/splice/issues/1347

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/HistoryMetrics.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/HistoryMetrics.scala
@@ -11,7 +11,7 @@ import com.digitalasset.canton.data.CantonTimestamp
 
 class HistoryMetrics(metricsFactory: LabeledMetricsFactory)(implicit
     metricsContext: MetricsContext
-) {
+) extends AutoCloseable {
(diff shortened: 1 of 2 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/HistoryMetrics.scala
```diff
@@ -164,4 +164,13 @@ class HistoryMetrics(metricsFactory: LabeledMetricsFactory)(implicit
     )(metricsContext)
 
   }
+
+  override def close(): Unit = {
+    UpdateHistoryBackfilling.completed.close()
```
rautenrieth-da, line 169: How does this interact with lazy vals? If the metric was not used until now, it will be created here and immediately closed? What's the benefit of using `lazy` then?

rautenrieth-da (approved):
Thanks!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/HistoryMetrics.scala
nicu-da: Doesn't matter that it will be created then, that should be safe as long as we close it. We just don't want to report it if we don't use it all.

PR #1356, opened 2025-07-07
CIP-0058 (Python script) - Add unclaimed_sv_rewards.py script

jose-velasco-ieu:
# unclaimed_sv_rewards.py script

## 📜 DESCRIPTION

Scans the transaction log over a given time range and reports statistics on `SvRewardCoupon` contracts — categorized as **claimed**, **expired**, or **unclaimed** — for a specific beneficiary. It calculates the corresponding **reward amounts** using data from `IssuingMiningRound` (claimed rewards) and `ClosedMiningRound` (expired rewards) contracts.

This script helps identify **unclaimed SV reward coupons** due to expired or unexercised coupons within the specified window. To ensure all relevant mining rounds are considered (especially those created shortly after coupon creation), a **grace period** is applied — defaulting to **60 minutes**.

---

## 🧾 INPUTS

| Argument | Description |
|----------|-------------|
| `scan_url` | Address of the Splice Scan server - **required** |
| `--loglevel` | Log level (e.g., DEBUG, INFO) - **default: INFO** |
| `--page-size` | Number of transactions to fetch per request - **default: 100** |
| `--grace-period-for-mining-rounds-in-minutes` | Minutes to extend `end-record-time` to ensure all relevant `IssuingMiningRound`s are included - **default: 60** |
| `--cache-file-path` | File path to save application state to. If the file exists, processing will resume from the persisted state. Otherwise, processing will start from begin-record-time provided.  |
| `--rebuild-cache` | Force the cache to be rebuilt from scratch. |
| `--beneficiary` | The party for which unclaimed rewards are calculated - **required** |
| `--begin-migration-id` | The migration ID active at `begin-record-time` - **required** |
| `--begin-record-time` | Start of the time range (exclusive) to consider `SvRewardCoupon` creation (ISO format) - **required** |
| `--end-record-time` | End of the time range (inclusive) to consider `SvRewardCoupon` creation (ISO format) - **required** |
| `--weight` | Weight of sv coupon rewards to consider- **required** |
| `--already-minted-weight` | Weight already minted for the time range provided- **required** |

---

## 🔄 EVENT COLLECTION AND PROCESSING

The script processes:

- `SvRewardCoupon` **creates** between `begin-record-time` and `end-record-time`
- `SvRewardCoupon` **exercises** (for the coupons above) between `begin-record-time` and `end-record-time + grace-period`
- `IssuingMiningRound` **creates** between `begin-record-time` and `end-record-time + grace-period`
- `ClosedMiningRound` **creates** between `begin-record-time` and `end-record-time + grace-period`

### Events collected

- `SvRewardCoupon` create  
- `SvRewardCoupon_DsoExpire` exercise  
- `SvRewardCoupon_ArchiveAsBeneficiary` exercise
- `IssuingMiningRound` create  
- `IssuingMiningRound` archive  
- `ClosedMiningRound` create  
- `ClosedMiningRound` archive  

### Event processing in detail

- Created `IssuingMiningRound`s are stored in two dictionaries:  (1) `active_issuing_rounds_cid_to_round_number`, keyed by **contractId** to **round number**, and (2) `active_issuing_rounds`, keyed by **round number** to **round info**. This information is later used to calculate reward amounts when rewards are claimed via the `SvRewardCoupon_ArchiveAsBeneficiary` exercise.
- Created `ClosedMiningRound`s are stored in two dictionaries:  (1) `active_closed_rounds_cid_to_round_number`, keyed by **contractId** to **round number**, and (2) `active_closed_rounds`, keyed by **round number** to **round info**. This information is later used to calculate reward amounts when rewards are expired via the `SvRewardCoupon_ArchiveAsBeneficiary` exercise.
- For each `SvRewardCoupon` created in the range, an entry is added to a `active_rewards` dictionary.
- For each `SvRewardCoupon_DsoExpire`, the corresponding amount is computed using the round number, the coupon’s weight, the weight requested, the already-minted-weight provided, and the `issuance_per_sv_reward` stored in the `active_issuing_rounds` dictionary. Additionally, a counter called `reward_expired_count` is incremented, and an accumulator called `reward_expired_total_amount` is updated with the computed amount. The final step is to remove the processed reward from `active_rewards`.
- For each `SvRewardCoupon_ArchiveAsBeneficiary`, the corresponding amount is computed using the round number, the coupon’s weight, the weight requested, the already-minted-weight provided and the `issuance_per_sv_reward` stored in the `active_closed_rounds` dictionary. Additionally, a counter called `reward_claimed_count` is incremented, and an accumulator called `reward_claimed_total_amount` is updated with the computed amount. The final step is to remove the processed reward from `active_rewards`.
- For each `Archive` on `IssuingMiningRound`, the corresponding round is removed from the `active_issuing_rounds_cid_to_round_number` and `active_issuing_rounds` dictionaries.
- For each `Archive` on `IssuingMiningRound`, the corresponding round is removed from the `active_issuing_rounds_cid_to_round_number` and `active_issuing_rounds` dictionaries.

> **Note:** Exercises that occur *after* `end-record-time + grace-period` will not affect state, leaving the coupon as unclaimed. This should not happen unless reward expiration logic failed.

#### Weight validation 
`--weight` is validated against `--already-minted-weight` and the ***weight of every coupon*** in the time period: Ensures that the provided weight meets the following condition for all coupons within the given time period:

```
weight (input) <= coupon_weight - already_minted_weight (input)
```
If the condition above is not met, the computed amount will correspond to the following available weight: 
```
available_weight = max(0, coupon_weight - already_minted_weight)
```

### SvRewardCoupon processing diagram

CIP-0058-script-1 (1)

---

## 🧾 RESULT 

The results are logged. It include:

- `reward_expired_count`
- `reward_expired_total_amount`
- `reward_claimed_count` *(expected to be 0)*
- `reward_claimed_total_amount` *(expected to be 0)*
- `reward_unclaimed_count` *(expected to be 0)*

> **Note:** The script asserts that ***reward_unclaimed_count*** is 0, and fails otherwise. 

---

## 🧠 Cache Mechanism for Resumable Execution

This script supports **resumable execution** through an optional cache mechanism that persists the application state between runs. This is useful when scanning long time ranges that might be interrupted due to network failures or API timeouts.

### How it works

- Use the `--cache-file-path` argument to enable caching. The script will save its internal state — including pagination position, processed rewards, and mining rounds — to the specified JSON file.
- On the next run, if the cache file exists, the script will attempt to **resume scanning from where it left off**, unless:
  - You pass the `--rebuild-cache` flag to start fresh, or
  - The input parameters that uniquely define a run differ from those used when the cache was created. These inputs are considered **relevant for determining the identity of a scan**, and include:
    - `--beneficiary`
    - `--begin-record-time`
    - `--end-record-time`
    - `--begin-migration-id`
    - `--weight`
    - `--already-minted-weight`

  If any of these values have changed, the cached state is considered invalid and will be ignored automatically.

### When to use

- Recommended for long scans across multiple hours of transaction history.
- Especially helpful in unreliable environments where intermittent failures may occur.

### Related options

- `--cache-file-path <file>` – enables caching to the given file.
- `--rebuild-cache` – bypasses any existing cache and starts from scratch.

---

### Using default values
```bash
python3 scripts/scan-txlog/unclaimed_sv_rewards.py https://scan.sv-1.global.canton.network.sync.global \
      --cache-file-path 'unclaimed_sv_rewards_cache.json' \
      --beneficiary 'IntellectEU-ghost-1::1220f5cf298f609f538b46a2a0347d299028ddca7ea008aaed65e0ca862db974c5e2' \
      --begin-migration-id 3 \
      --begin-record-time '2025-07-20T10:30:00Z' \
      --end-record-time '2025-07-20T11:30:00Z' \
      --weight 5000 \
      --already-minted-weight 0
```

### Overriding default values
```bash
python3 scripts/scan-txlog/unclaimed_sv_rewards.py https://scan.sv-1.global.canton.network.sync.global \
  --loglevel DEBUG \
  --page-size 200 \
  --grace-period-for-mining-rounds-in-minutes 70 \
  --cache-file-path 'unclaimed_sv_rewards_cache.json' \
  --rebuild-cache \
  --beneficiary 'IntellectEU-ghost-1::1220f5cf298f609f538b46a2a0347d299028ddca7ea008aaed65e0ca862db974c5e2' \
  --begin-migration-id 3 \
  --begin-record-time '2025-07-20T10:30:00Z' \
  --end-record-time '2025-07-20T11:30:00Z' \
  --weight 5000 \
  --already-minted-weight 0
```

### ⏱ Timeline

```
   t0                   t1                 t2                     t3
   | m₁ (w=10K)         | m₂ (w=10K)       | m₃ (w=20K)           |
---------------------------------------------------------------------------------
   |         40K        |        30K       |         20K          | 0 ...
```

### 🧩 Steps

- At **t0**:
  - GSF sets `Ghost-party-A` weight = 40K

- At **t1** (Milestone 1, m₁ is completed):
  - GSF sets weight = 30K
  - GSF runs the script with (single execution):
    - t0 to t1
    ```
    --begin-record-time=t0 --end-record-time=t1 --weight=10000 --already-minted-weight=0 
    ```
  - A vote is initiated for **m₁**

- At **t2** (Milestone 2, m₂ is completed):
  - GSF sets weight = 20K
  - GSF runs the script with (2 executions):
    - t0 to t1
    ```
    --begin-record-time=t0 --end-record-time=t1 --weight=10000 --already-minted-weight=10000  
    ```
    - t1 to t2
    ```
    --begin-record-time=t1 --end-record-time=t2 --weight=10000 --already-minted-weight=0  
    ```
  - A vote is initiated for **m₂**

- At **t3** (Milestone 3, m₃ is completed):
  - GSF sets weight = 0 (party removed)
  - GSF runs the script with (3 executions): 
    - t0 to t1
    ```
    --begin-record-time=t0 --end-record-time=t1 --weight=20000 --already-minted-weight=20000  
    ```
    - t1 to t2
    ```
    --begin-record-time=t1 --end-record-time=t2 --weight=20000 --already-minted-weight=10000
    ```
    - t2 to t3
    ```
    --begin-record-time=t2 --end-record-time=t3 --weight=20000 --already-minted-weight=0  
    ```
  - A vote is initiated for **m₃**
  
### IMPORTANT NOTES:  
- ***GSF*** should only configure the ***Ghost Party-A*** to the lower weight after the SV starts direct minting at their newly earned weight.  

- ***Weights in CIPs vs script***: In the CIPs, you’ll see weights like 0.5, 1, 2, and so on. These weights translate to 5K, 10K, and 20K in script terms (i.e., CIP-weight × 10K).

---

## ⏱️ TIMING

It was reported by one of the users that the script took ***1 hour and 40 minutes*** to run ***for a period of 591 hours (24 days)***.

Scenario:
- Period: 591 hours (24 days)
- Timing: 1hr and 40 mins
- Internet speed: Download: 7712.84 Mbit/s / Upload: 1094.13 Mbit/s 

```
python3 unclaimed_sv_rewards.py https://scan.sv-1.global.canton.network.sync.global/ \
  --beneficiary 'XXX' \
  --begin-migration-id 2 \
  --begin-record-time '2025-06-11T07:00:00Z' \
  --end-record-time '2025-07-05T22:00:00Z' \
  --loglevel=DEBUG \
  --weight=10000 \
  --already-minted-weight=0
...
...
INFO: global:End run. (5985.56 sec., 3757557 transaction(s), 37576 Scan API call(s))
DEBUG: global:active_mining_rounds count: 3
DEBUG: global:active_closed_rounds count: 0
INFO:global:reward_expired_count = 3423
INFO:global:reward_expired_total_amount = 3869192.6550120000
INFO:global:reward_claimed_count = 0
INFO: global:reward_claimed_total_amount = 0.0000000000
INFO:global:reward_unclaimed_count = 0
```
> **Note:**: Extrapolating over a 6-month period, it would take approximately 12 hours and 10 minutes with a similar internet connection (assuming each month has 30 days) .

---

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/scripts/scan-txlog/unclaimed_sv_rewards.py
+++ b/scripts/scan-txlog/unclaimed_sv_rewards.py
@@ -121,7 +121,7 @@ def _log_uncaught_exceptions():
 class TemplateQualifiedNames:
     sv_reward_coupon = "Splice.Amulet:SvRewardCoupon"
     issuing_mining_round = "Splice.Round:IssuingMiningRound"
-    closing_mining_round = "Splice.Round:ClosedMiningRound"
+    closed_mining_round = "Splice.Round:ClosedMiningRound"
 
 
 @dataclass
@@ -257,8 +257,8 @@ class LfValue:
     def get_issuing_mining_round_issuance_per_sv_reward(self):
         return self.__get_record_field("issuancePerSvRewardCoupon").__get_numeric()
 
-     # template ClosingMiningRound -> issuancePerSvRewardCoupon
-    def get_closing_mining_round_issuance_per_sv_reward(self):
+     # template ClosedMiningRound -> issuancePerSvRewardCoupon
+    def get_closed_mining_round_issuance_per_sv_reward(self):
         return self.__get_record_field("issuancePerSvRewardCoupon").__get_numeric()
 
     # template OpenMiningRound -> round
@@ -441,7 +441,7 @@ class IssuingRound:
     issuance_per_sv_reward: DamlDecimal
 
 @dataclass
-class ClosingRound:
+class ClosedRound:
     round: int
     record_time: datetime
     issuance_per_sv_reward: DamlDecimal
@@ -449,11 +449,16 @@ class ClosingRound:
 @dataclass
 class State:
     beneficiary: str
+    # Maps CIDs of active SvRewardCoupons to their reward information
     active_rewards: dict[str, Reward]
+    # Maps CIDs of active issuing rounds to their round numbers
+    active_issuing_rounds_cid_to_round_number: dict[str, int]
+    # Maps round numbers of active issuing rounds to their round information
     active_issuing_rounds: dict[int, IssuingRound]
-    active_issuing_rounds_contracts: dict[str, int]
-    active_closing_rounds: dict[int, ClosingRound]
-    active_closing_rounds_contracts: dict[str, int]
+    # Maps CIDs of active closed rounds to their round numbers
+    active_closed_rounds_cid_to_round_number: dict[str, int]
+    # Maps round numbers of active closed rounds to their round information
+    active_closed_rounds: dict[int, ClosedRound]
     begin_record_time: datetime
     end_record_time: datetime
     grace_period_for_mining_rounds_in_minutes: datetime
@@ -482,9 +487,9 @@ class State:
             beneficiary=args.beneficiary,
             active_rewards={},
             active_issuing_rounds={},
-            active_issuing_rounds_contracts={},
-            active_closing_rounds={},
-            active_closing_rounds_contracts={},
+            active_issuing_rounds_cid_to_round_number={},
+            active_closed_rounds={},
+            active_closed_rounds_cid_to_round_number={},
             begin_record_time=begin_record_time,
             end_record_time=end_record_time,
             grace_period_for_mining_rounds_in_minutes = grace_period_for_mining_rounds_in_minutes,
@@ -535,18 +540,18 @@ class State:
                     issuance_per_sv_reward=event.payload.get_issuing_mining_round_issuance_per_sv_reward(),
                 )
                 LOG.debug(f"Adding issuing round {issuing_round} to active_issuing_rounds")
-                self.active_issuing_rounds_contracts[event.contract_id] = round_number
+                self.active_issuing_rounds_cid_to_round_number[event.contract_id] = round_number
                 self.active_issuing_rounds[round_number] = issuing_round
-            case TemplateQualifiedNames.closing_mining_round:
+            case TemplateQualifiedNames.closed_mining_round:
                 round_number = event.payload.get_closed_mining_round_round()
-                closing_round = ClosingRound(
+                closed_round = ClosedRound(
                     round=round_number,
                     record_time=transaction.record_time,
-                    issuance_per_sv_reward=event.payload.get_closing_mining_round_issuance_per_sv_reward(),
+                    issuance_per_sv_reward=event.payload.get_closed_mining_round_issuance_per_sv_reward(),
                 )
-                LOG.debug(f"Adding closing round {closing_round} to active_closing_rounds")
-                self.active_closing_rounds_contracts[event.contract_id] = round_number
-                self.active_closing_rounds[round_number] = closing_round
+                LOG.debug(f"Adding closed round {closed_round} to active_closed_rounds")
+                self.active_closed_rounds_cid_to_round_number[event.contract_id] = round_number
+                self.active_closed_rounds[round_number] = closed_round
 
 
     def process_exercised_event(self, transaction: TransactionTree, event: ExercisedEvent):
@@ -567,21 +572,21 @@ class State:
             case TemplateQualifiedNames.issuing_mining_round:
                 match event.choice_name:
                     case "Archive":
-                        match self.active_issuing_rounds_contracts.pop(event.contract_id, None):
+                        match self.active_issuing_rounds_cid_to_round_number.pop(event.contract_id, None):
                             case None:
                                 pass
                             case round_number:
                                 self.active_issuing_rounds.pop(round_number)
                     case _:
                         self.process_events(transaction, event.child_event_ids)
-            case TemplateQualifiedNames.closing_mining_round:
+            case TemplateQualifiedNames.closed_mining_round:
                 match event.choice_name:
                     case "Archive":
-                        match self.active_closing_rounds_contracts.pop(event.contract_id, None):
+                        match self.active_closed_rounds_cid_to_round_number.pop(event.contract_id, None):
                             case None:
                                 pass
                             case round_number:
-                                self.active_closing_rounds.pop(round_number)
+                                self.active_closed_rounds.pop(round_number)
                     case _:
                         self.process_events(transaction, event.child_event_ids)
             case _:
@@ -597,34 +602,38 @@ class State:
             case None:
                 pass
             case reward:
-                issuing = self.active_issuing_rounds.get(reward.round)
-                closing = self.active_closing_rounds.get(reward.round)
-                match (issuing, closing):
-                    # If the mining round for a reward is not found, it means that a `SvRewardCoupon` was
-                    # created, but the corresponding mining round was created after the end-record-time + grace period.
-                    # This should not happen, as the additional grace period should be sufficient to ensure
-                    # we capture the complete set of mining rounds for the collected rewards.
-                    case (None, None):
-                        self._fail(
-                            f"Fatal: missing issuing round {reward.round} for reward {reward.contract_id}\n"
-                            f"Consider increase input: grace-period-for-mining-rounds-in-minutes.\n"
-                            f"Currently it is set to {self.grace_period_for_mining_rounds_in_minutes}"
-                        )
-                    case (mining_round_info, None) | (_, mining_round_info):
-                        amount = reward.weight * mining_round_info.issuance_per_sv_reward
-                        LOG.debug(
-                            f"Updating {status} summary with amount {amount}, corresponding to contract {event.contract_id}"
-                        )
-                        match status:
-                            case RewardStatus.EXPIRED:
+                match status:
+                    case RewardStatus.EXPIRED:
+                        match self.active_closed_rounds.get(reward.round):
+                            case None:
+                                self._fail_with_missing_round(reward)
+                            case mining_round_info:
+                                amount = reward.weight * mining_round_info.issuance_per_sv_reward
+                                LOG.debug(
+                                    f"Updating expired summary with amount {amount}, corresponding to contract {event.contract_id}"
+                                )
                                 self.reward_summary.reward_expired_count += 1
                                 self.reward_summary.reward_expired_total_amount += amount
-                            case RewardStatus.CLAIMED:
+                    case RewardStatus.CLAIMED:
+                        match self.active_issuing_rounds.get(reward.round):
+                            case None:
+                                self._fail_with_missing_round(reward)
+                            case mining_round_info:
+                                amount = reward.weight * mining_round_info.issuance_per_sv_reward
+                                LOG.debug(
+                                    f"Updating claimed summary with amount {amount}, corresponding to contract {event.contract_id}"
+                                )
                                 self.reward_summary.reward_claimed_count += 1
                                 self.reward_summary.reward_claimed_total_amount += amount
 
         self.process_events(transaction, event.child_event_ids)
 
+    def _fail_with_missing_round(self, reward):
+        self._fail(
+            f"Fatal: missing round {reward.round} for reward {reward.contract_id}\n"
+            f"Consider increase input: grace-period-for-mining-rounds-in-minutes.\n"
+            f"Currently it is set to {self.grace_period_for_mining_rounds_in_minutes}"
+        )
 
     def _fail(self, message, cause=None):
         LOG.error(message)
@@ -677,7 +686,7 @@ async def main():
         f"active_mining_rounds count: {len(app_state.active_issuing_rounds)}"
     )
     LOG.debug(
-        f"active_closing_rounds count: {len(app_state.active_closing_rounds)}"
+        f"active_closed_rounds count: {len(app_state.active_closed_rounds)}"
     )
 
     reward_summary = app_state.reward_summary


scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -0,0 +1,706 @@
...
+
+    LOG.info(f"reward_expired_count = {summary.reward_expired_count}")
+    LOG.info(f"reward_expired_total_amount = {summary.reward_expired_total_amount.decimal:.10f}")
+    LOG.info(f"reward_claimed_count = {summary.reward_claimed_count}")
+    LOG.info(f"reward_claimed_total_amount = {summary.reward_claimed_total_amount.decimal:.10f}")
+    LOG.info(f"reward_unclaimed_count = {summary.reward_unclaimed_count}")
+    LOG.info(f"reward_unclaimed_total_ammount = {summary.reward_unclaimed_total_ammount.decimal:.10f}")
+
+
+    if args.report_output:
```
moritzkiefer-da: What do we need a csv export for? If you think of the intended usage it seems like you get one number that SVs then have to use in their SV UI. So a machine readable output seems not all that useful.

scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -0,0 +1,706 @@
...
+                reward = Reward(
+                    round=event.payload.get_sv_reward_coupon_round(),
+                    weight=event.payload.get_sv_reward_coupon_weight(),
+                    sv=event.payload.get_sv_reward_coupon_sv(),
+                    beneficiary=event.payload.get_sv_reward_coupon_beneficiary(),
+                    contract_id=event.contract_id,
+                )
+                if (
+                    reward.beneficiary == self.beneficiary
+                    and transaction.record_time <= self.create_sv_reward_end_record_time
```
moritzkiefer-da, line 527: why do we need this filter? Don't we already have the time filter in `should_process`?

scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -0,0 +1,706 @@
...
+class IssuingRound:
+    round: int
+    record_time: datetime
+    issuance_per_sv_reward: DamlDecimal
+
+@dataclass
+class State:
+    beneficiary: str
+    rewards: dict[str, Reward]
+    issuing_rounds: dict[int, IssuingRound]
```
moritzkiefer-da: I'm a bit nervous about the approach of storing all rewards and issuing rounds you encounter here. That just leaks increasing amounts of memory. We need to expect that we need to run this script over a period of months.

Do we need to track this? It seems like we only care about:
1. Currently active activity records and rounds. That's constant space.
2. The accumulated totals of minted and expired rewards. That's constant space

scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -0,0 +1,706 @@
...
+                round_number = event.payload.get_issuing_mining_round_round()
+                issuing_round = IssuingRound(
+                    round=round_number,
+                    record_time=transaction.record_time,
+                    issuance_per_sv_reward=event.payload.get_issuing_mining_round_issuance_per_sv_reward(),
+                )
+                self.issuing_rounds[round_number] = issuing_round
+                LOG.debug(f"Adding issuing round {issuing_round} to issuing_rounds")
+
+    def process_exercised_event(self, transaction: TransactionTree, event: ExercisedEvent):
```
moritzkiefer-da, line 557: We should check and error out if we see another consuming choice on SvRewardCoupon. There shouldn't be any at this point but that can change and that error will then force us to update the script instead of silently reporting wrong results.

scripts/scan-txlog/unclaimed_sv_rewards.py
jose-velasco-ieu: I agree. I will remove it.

scripts/scan-txlog/unclaimed_sv_rewards.py
jose-velasco-ieu: The key point is that `self.create_sv_reward_end_record_time `is not the same as `self.end_record_time`.

In `should_process`, we're using `self.end_record_time`, which is computed as `args.end_record_time + grace_period`. This ensures we’ve accounted for all mining rounds relevant to rewards near the `args.end_record_time`.

```
def should_process(self, tx: dict):
        return datetime.fromisoformat(tx["record_time"]) < self.end_record_time

```

```
end_record_time = (
            datetime.fromisoformat(args.end_record_time)
            + timedelta(minutes=grace_period_for_mining_rounds_in_minutes)
        )

```

On the other hand, when handling reward creation events (i.e., `TemplateQualifiedNames.sv_reward_coupon`), we use `self.create_sv_reward_end_record_time`, which is set to exactly `args.end_record_time`.

scripts/scan-txlog/unclaimed_sv_rewards.py
jose-velasco-ieu: I see your point.
I’ll refactor the script to update the summary on the fly and only keep the active rounds and active rewards in the state.

scripts/scan-txlog/unclaimed_sv_rewards.py
moritzkiefer-da: ah right makes sense thx

scripts/scan-txlog/unclaimed_sv_rewards.py
jose-velasco-ieu: done

scripts/scan-txlog/unclaimed_sv_rewards.py
jose-velasco-ieu: done

scripts/scan-txlog/unclaimed_sv_rewards.py
jose-velasco-ieu: done

jose-velasco-ieu:
> Nice work thanks! This overall looks correct, left a few comments with suggestions.
> 
> We should also add a test. Probably easiest to write a dedicated integration tests that cycles through some rounds and expires and mints some rewards and then call the script at the end and check that the results match what we expect.

@moritzkiefer-da 
Would `SvExpiredRewardsCollectionTimeBasedIntegrationTest` be a good starting point, in your opinion?

moritzkiefer-da:
First example I could find is here https://github.com/hyperledger-labs/splice/blob/d301064076080fde4b25652b528ddb10a052f157/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletManualRoundsIntegrationTest.scala#L73

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
```diff
@@ -0,0 +1,231 @@
...
+        logger.debug(s"openMiningRounds: $openMiningRounds")
+      }
+
+      clue("Pause sv rewards generation") {
+        receiveSvRewardCouponTrigger.pause().futureValue
+      }
+
+      val svRewardCoupons = sv1WalletClient.listSvRewardCoupons()
+      val svRewardCouponsCount = svRewardCoupons.length
+      svRewardCouponsCount should not be 0
```
moritzkiefer-da:
```suggestion
      svRewardCoupons should not be empty
```

moritzkiefer-da: any reason why we cannot assert on the exact length? It seems like this should be deterministic no?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
```diff
@@ -0,0 +1,231 @@
...
+      logger.debug(s"SvRewardCoupons count: $svRewardCouponsCount")
+
+      // Expire
+      ///////////
+
+      clue("Resume trigger to expire some coupons") {
+        expireRewardCouponsTrigger.resume()
```
moritzkiefer-da: use `actAndCheck` instead of `clue` + `eventually`

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
```diff
@@ -0,0 +1,231 @@
...
+        eventuallySucceeds(timeUntilSuccess = 60.seconds) {
```
moritzkiefer-da: I don't understand why you need an `eventuallySucceeds`. Just a normal `eventually` should be fine (but becomes redundant with actAndCheck)

moritzkiefer-da: 60 seconds seems forever, why is this so slow?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
```diff
@@ -0,0 +1,231 @@
...
+          sv1WalletClient.listSvRewardCoupons().length should be < svRewardCouponsCount
```
moritzkiefer-da: `.length should be` is an anti pattern usually because you then get fairly useless errors. Use `should have size …` which will showy ou the full list if it fails rather than just the length.

moritzkiefer-da: Waiting just for the number of reward coupons to go down seems a bit weird to me and potentially make the test non-deterministic. Why can we not wait until we reach an exact amount of coupons?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
```diff
@@ -0,0 +1,231 @@
...
+      val svRewardCouponsAfterExpiryCount = svRewardCouponsAfterExpiry.length
+      val svRewardCouponsExpired = svRewardCoupons.diff(svRewardCouponsAfterExpiry)
+      val svRewardCouponsExpiredCount = svRewardCouponsExpired.length
+      logger.debug(s"SvRewardCoupons expired count: $svRewardCouponsExpiredCount")
+
+      // Claim
+      //////////
+
+      clue("Resume trigger to claim some coupons") {
+        sv1CollectRewardsAndMergeAmuletsTrigger.resume()
```
moritzkiefer-da: same as above:
1. Use `actAndCheck`
2. 60s seems too long
3. Use `have size` instead of `.length should be`
4. Can we switch to an exact assertion on the number of coupons instead of just that it goes down?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
```diff
@@ -0,0 +1,231 @@
...
+          sv1CollectRewardsAndMergeAmuletsTrigger.pause().futureValue
+        }
+      }
+
+      val svRewardCouponsAfterClaiming = sv1WalletClient.listSvRewardCoupons()
+      val svRewardCouponsClaimed = svRewardCouponsAfterExpiry.diff(svRewardCouponsAfterClaiming)
+      val svRewardCouponsClaimedCount = svRewardCouponsClaimed.length
+      val svRewardCouponsUnclaimedCount =
+        svRewardCouponsCount - svRewardCouponsExpiredCount - svRewardCouponsClaimedCount
+      logger.debug(s"SvRewardCoupons claimed count: $svRewardCouponsClaimedCount")
```
moritzkiefer-da: all these count logs feel a bit weird. Consider maybe logging the rounds of the coupons or something like that.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
```diff
@@ -0,0 +1,231 @@
...
+              _.contains(s"reward_expired_total_amount = $rewardExpiredTotalAmount")
+            ) shouldBe true
+          }
+          withClue(s"Expected: reward_claimed_count = $svRewardCouponsClaimedCount") {
+            readLines.exists(
+              _.contains(s"reward_claimed_count = $svRewardCouponsClaimedCount")
+            ) shouldBe true
+          }
+          withClue(s"Expected: reward_claimed_total_amount = $rewardClaimedTotalAmount") {
+            readLines.exists(_.contains("reward_claimed_total_amount =")) shouldBe true
```
moritzkiefer-da: did you mean to include `rewardClaimedTotalAmount` in that? Otherwise I don't understand what you're testing

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
```diff
@@ -0,0 +1,231 @@
...
+                "--end-record-time",
+                endRecordTime.toString,
+              )
+            )
+            .!(errorProcessor)
+
+          assert(exitCode == 0, s"Script exited with code $exitCode")
+          readLines.filter(_.startsWith("ERROR:")) shouldBe empty
+          withClue(s"Expected: reward_expired_count = $svRewardCouponsExpiredCount") {
+            readLines.exists(
```
moritzkiefer-da: you probably want something like

```
forExactly(1, readLines) {
  _ should include("s"reward_expired …")
}
```
That actually gives you a change to debug failures. Currently a failure will just be something like "false was not true" which is not particularly helpful.

scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -0,0 +1,644 @@
...
+class IssuingRound:
+    round: int
+    record_time: datetime
+    issuance_per_sv_reward: DamlDecimal
+
+@dataclass
+class State:
+    beneficiary: str
+    active_rewards: dict[str, Reward]
+    active_issuing_rounds: dict[int, IssuingRound]
```
moritzkiefer-da, line 457: It looks to me like you don't actually track active issuing rounds here. You can see that when you remove it: On the consuming exercise on the coupon which does not archive the round. And the other way around: When the issuing mining round does get archived and turned into a ClosedMiningRound you don't actually archive the issuing round. I would suggest we actually track active issuing and closing mining rounds. Otherwise it becomes somewhat tricky to follow what we do and extensions like tracking multiple beneficiaries would fail because you remove the round after the first one.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
jose-velasco-ieu: What's the difference between `eventuallySucceeds` and `eventually`?

jose-velasco-ieu: I set it to 60 seconds because the test occasionally fails with the default value.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
jose-velasco-ieu: yes

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
moritzkiefer-da: sounds to me like your machine might be significantly slower than usual. The default should be plenty.

I suggest keeping it at the default for now. If you want to make your life easier, I suggest creating a separate PR that just multiples each eventually by a given factor from your envrc.vars or something like that. That way you could just add a factor of 2 to all eventuallies to compensate for your slow machine.

moritzkiefer-da: `eventually` just waits for an assertion to succeed. `eventuallySucceeds` assumes you have a command (not an assertion) that actually fails for a while and suppresses the logs from the failure.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
jose-velasco-ieu: The issue is that for example the number of expired contracts changes depending on how the execution is performed, which is not good.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
jose-velasco-ieu: done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
jose-velasco-ieu: done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
jose-velasco-ieu: done

jose-velasco-ieu: done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
jose-velasco-ieu: done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
jose-velasco-ieu: done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
jose-velasco-ieu: done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
jose-velasco-ieu: done

jose-velasco-ieu: done

scripts/scan-txlog/unclaimed_sv_rewards.py
jose-velasco-ieu: done

scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -0,0 +1,697 @@
...
+    def handle_sv_reward_coupon_exercise(self, transaction, event, status):
+        # Only handle SvRewardCoupons created within the configured time range for the given beneficiary
+        match self.active_rewards.pop(event.contract_id, None):
+            # None means either:
+            # - The reward was created outside the time range, or
+            # - The exercise event is for a different beneficiary
+            case None:
+                pass
+            case reward:
+                issuing = self.active_issuing_rounds.get(reward.round)
```
moritzkiefer-da: can we move the round lookup into the `match status` part? It seems like then you know which one you need: For claimed you lookup the issuing round, for expired you lookup the closed round.

scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -0,0 +1,697 @@
...
+    round: int
+    record_time: datetime
+    issuance_per_sv_reward: DamlDecimal
+
+@dataclass
+class State:
+    beneficiary: str
+    active_rewards: dict[str, Reward]
+    active_issuing_rounds: dict[int, IssuingRound]
+    active_issuing_rounds_contracts: dict[str, int]
```
moritzkiefer-da: rename to omething like `active_issuing_rounds_cid_to_round_number` or something and maybe add a comment. Currently it's pretty non-obvious from reading this here what you are tracking

scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -0,0 +1,697 @@
...
+class ClosingRound:
+    round: int
+    record_time: datetime
+    issuance_per_sv_reward: DamlDecimal
+
+@dataclass
+class State:
+    beneficiary: str
+    active_rewards: dict[str, Reward]
+    active_issuing_rounds: dict[int, IssuingRound]
```
moritzkiefer-da, line 457: add a comment that the key in the map is the round number, same for issuing

scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -0,0 +1,697 @@
...
+
+        LOG.error("Uncaught exception", exc_info=(exc_type, exc_value, exc_traceback))
+
+    sys.excepthook = handle_exception
+
+
+class TemplateQualifiedNames:
+    sv_reward_coupon = "Splice.Amulet:SvRewardCoupon"
+    issuing_mining_round = "Splice.Round:IssuingMiningRound"
+    closing_mining_round = "Splice.Round:ClosedMiningRound"
```
moritzkiefer-da: typo: it's closedmininground not closingmininground

moritzkiefer-da (approved):
nice thanks

scripts/scan-txlog/unclaimed_sv_rewards.py
jose-velasco-ieu: done

scripts/scan-txlog/unclaimed_sv_rewards.py
jose-velasco-ieu: done

scripts/scan-txlog/unclaimed_sv_rewards.py
jose-velasco-ieu: done

scripts/scan-txlog/unclaimed_sv_rewards.py
jose-velasco-ieu: done

waynecollier-da:
@moritzkiefer-da Any further comments on this PR? Is the script ready to use?

moritzkiefer-da:
> Is the script ready to use?

imho yes

PR #1360, opened 2025-07-07
Define new runner type for CircleCI-based deployments and wire up SAs for it

martinflorian-da:
Part of https://github.com/DACH-NY/canton-network-internal/issues/798 (let's discuss there)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/circleci/src/index.ts
+++ b/cluster/pulumi/circleci/src/index.ts
@@ -18,6 +18,9 @@ const circleCiNamespace = new Namespace('circleci-runner', {
   },
 });
 
+// Below logic assumes that the cluster has workload identity enabled.
+// You can enforce this via `cncluster cluster_enable_workload_identity`.
+
 // Create a GCP SA for CCI (CCI SA) and a k8s service account (CCI KSA) that is connected to the GCP SA.
 const cciGcpServiceAccount = new gcp.serviceaccount.Account('cci-deployments-gcp-sa', {
   accountId: 'cci-deployments-gcp-sa',


cluster/pulumi/circleci/src/index.ts
```diff
@@ -16,6 +17,31 @@ const circleCiNamespace = new Namespace('circleci-runner', {
     name: 'circleci-runner',
   },
 });
+
+// Create a GCP SA for CCI (CCI SA) and a k8s service account (CCI KSA) that is connected to the GCP SA.
+const cciGcpServiceAccount = new gcp.serviceaccount.Account('cci-deployments-gcp-sa', {
```
isegall-da, line 25: I presume we don't envision to ever have more than one cluster for runners for CI, right? (My initial thought was that this belongs better in the gcp or gcp-project stack, but that's a bit more work as you then need to export from there and import here)

martinflorian-da: >  gcp or gcp-project stack

Hmm aren't those supposed to be for general cluster-related setup? This SA here is going to be unique to the CCI cluster and unique in general.

>  we don't envision to ever have more than one cluster for runners for CI

Yeah that sounds reasonable to me...

isegall-da: > Hmm aren't those supposed to be for general cluster-related setup? This SA here is going to be unique to the CCI cluster and unique in general.

That's one way to think of the split. 
Another would be "configuring the project" vs "configuring the cluster".

nicu-da: I think we also should be fairly granular with the service accounts, so we should rather just create new ones instead of re-using

nicu-da (approved):
Thanks, LGTM!

How do we ensure workload identity is enabled in the cluster?

cluster/pulumi/circleci/src/index.ts
isegall-da: Fair. You'll have naming conflicts, but happy to resolve that if we ever do need more than one instance of this.

martinflorian-da: > You'll have naming conflicts

Ahh... hours later I have caught up with what you guys mean :sweat_smile: Yes I see the problem now but it sill seems reasonable to me to keep this encapsulated to the circleci stack, at least for now. Easy enough to fix if we ever want to but (I find that) these and the upcoming changes are already complex enough as they are...

isegall-da: Yup, I tend to agree.

PR #1378, opened 2025-07-08
[ci] Add http server metrics to scan

nicu-da:
fixes https://github.com/hyperledger-labs/splice/issues/1377

running a cluster deployment to build an actual dashboard

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/ScanApp.scala
```diff
@@ -291,22 +291,45 @@ class ScanApp(
             requestLogger(traceContext) {
               HttpErrorHandler(loggerFactory)(traceContext) {
                 concat(
-                  ScanResource.routes(scanHandler, _ => provide(traceContext)),
+                  ScanResource.routes(
+                    scanHandler,
+                    operation => {
+                      nodeMetrics.httpServerMetrics
```
moritzkiefer-da: is there no way to inject this a level higher like we do for the request logger so you don't need to push it into each individual route? It seems like the requestlogger does essentially the same kind of tracking of requests so I don't see why we couldn't create metrics from that.

nicu-da: Not without losing the operation label, which is essential for us. 
The request logger can work around it because it doesn't care about that label, but instead we add it by creating a new span in the actual implementation of each operation (we could actually lift that one up one layer so we don't have to do it everywhere).

moritzkiefer-da: ah I see thanks, maybe add a comment somewhere

nicu-da:
image
dashboard

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/ScanApp.scala
```diff
@@ -291,22 +291,45 @@ class ScanApp(
             requestLogger(traceContext) {
               HttpErrorHandler(loggerFactory)(traceContext) {
```
moritzkiefer-da: Should we move the error handler below the metrics? it feels like ideally the status code produced by that is included in the metric

moritzkiefer-da (approved):
nice thank you!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/ScanApp.scala
nicu-da: Ugh, good catch, let me see how I can do that

nicu-da: moved the error handler in the route implementation, based on its implementation it should make no difference. 
I think we will still miss 404s that match no route but we should just see them with status unhandled

PR #1385, opened 2025-07-08
[static] Fix istiod helm chart values

nicu-da:
the api was changed, not sure with what it was replaced, there's no release notes, i hate everything

fixes https://github.com/DACH-NY/cn-test-failures/issues/4890

will keep testing out with cilr 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/expected/infra/expected.json
+++ b/cluster/expected/infra/expected.json
@@ -1842,6 +1842,13 @@
           "defaultHttpRetryPolicy": {
             "attempts": 0
           },
+          "defaultProviders": {
+            "metrics": [
+              {
+                "name": "prometheus"
+              }
+            ]
+          },
           "enablePrometheusMerge": false
         },
         "pilot": {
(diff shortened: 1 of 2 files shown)


cluster/expected/infra/expected.json
```diff
@@ -1842,6 +1842,13 @@
           "defaultHttpRetryPolicy": {
             "attempts": 0
           },
+          "defaultProviders": {
```
nicu-da, line 1845: added this as i found it in an issue and hope to help with something. but istio docs are bad, reallly really bad

nicu-da:
> > no but we fixed the metrics, though they worked nice on a scratch they seem to not scale on cilr, so trying to fix that
> 
> I think the part that confused me is "the api was changed". Why does that matter if we didn't upgrade? Did we accidentally just use old format even though we were already on the new version?

Ah, yes that's correct..the only docs I found were for older versions and it seems the api was recently changed. Most useful things are found in issues so it's harder to figure out for which version are they actually valid for.

PR #1391, opened 2025-07-08
Fix allocation timestamps in the FE

OriolMunoz-da:
successful preflight: https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/19121/workflows/7151d802-82f7-4d9e-aa24-7ac15d08595c

### DIFF
--- a/apps/wallet/frontend/src/components/CreateAllocation.tsx
+++ b/apps/wallet/frontend/src/components/CreateAllocation.tsx
@@ -127,7 +127,12 @@ const CreateAllocation: React.FC = () => {
                 })
               }
             />
-            <Typography variant="h6">Requested at</Typography>
+            {/*For these timestamp fields: a date picker doesn't work because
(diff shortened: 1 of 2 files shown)


apps/wallet/frontend/src/components/CreateAllocation.tsx
```diff
@@ -130,51 +127,44 @@ const CreateAllocation: React.FC = () => {
                 })
               }
             />
+            <Typography variant="h6">Requested at</Typography>
+            <TextField
```
moritzkiefer-da, line 136: Why do we want a text field instead of a date picker? Because you need it to be more precise? Which timezone do you enter it in?

moritzkiefer-da: probably just add a comment on this

moritzkiefer-da (approved):
thanks

Issue #1099, opened 2025-06-12
show allocation requests that require the wallet party to allocate Amulet from the expected DSO


OriolMunoz-da:
must show the filled and unfilled requests (i.e., show whether the wallet party already created a matching allocation)

Issue #1100, opened 2025-06-12
allow accepting an allocation => this should open the allocation creation dialog/page with the details prefilled


OriolMunoz-da:

hrischuk-da:
HI @OriolMunoz-da.   I was wondering when you were expecting to conclude this?

PR #1415, opened 2025-07-09
Show allocation requests in FE

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/1099

see https://github.com/hyperledger-labs/splice/pull/1415#issuecomment-3057864939 for screenshots


meiersi-da:
Looking at the screenshot, what happens if the same sender is listed for multiple legs? What happens if not all of the legs require AMT?

apps/wallet/frontend/src/components/ListAllocationRequests.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  const transferLegId = Object.keys(transferLegs).find(
+    transferLegId => transferLegs[transferLegId].sender === userParty
+  );
+  const transferLeg = transferLegs[transferLegId || ''];
+  if (!transferLeg) return null;
+
+  const { amount, receiver } = transferLeg;
+  const { allocateBefore, settlementRef, executor } = settlement;
+  const id = [settlementRef.id, settlementRef.cid].filter(x => !!x).join(' - ');
```
meiersi-da: This join has some security issues, as an attacker could also inline `-<some-other-contract-id>` in their `id` field.

I'd suggest to just make it two separate labelled text fields.

apps/wallet/frontend/src/components/ListAllocationRequests.tsx
```diff
@@ -0,0 +1,107 @@
...
+};
+
+const AllocationRequestDisplay: React.FC<{
+  request: Contract<AllocationRequest>;
+  userParty: string;
+}> = ({ request, userParty }) => {
+  const { transferLegs, settlement } = request.payload;
+
+  const transferLegId = Object.keys(transferLegs).find(
+    transferLegId => transferLegs[transferLegId].sender === userParty
```
meiersi-da: Supporting a single transfer-leg only won't work e.g. for splitwell, where one settlement requires the sender to send money to all the counter-parties that they owe money to.

I'd suggest the following display for allocation requests:
1. list settlementinfo at the top, in a way that makes it easy to identify the trade being settled.
2. list all transfer legs in a table below the settlement info, one column per record field.
3. add one extra column to this table where you show the actions available on this leg:
  - if there's no matching allocation: (accept) button
  - if there is a mtaching allocation: "allocated <date>" (withdraw) button to withdraw the allocation

apps/wallet/frontend/src/components/ListAllocationRequests.tsx
```diff
@@ -0,0 +1,107 @@
...
+            <Typography variant="body2">{dayjs(allocateBefore).fromNow(true)} left</Typography>
+          </Stack>
+          {/*TODO (#1100): uncomment and implement callback*/}
+          {/*<Button variant="pill" size="small" className="transfer-offer-accept">*/}
+          {/*  Accept*/}
+          {/*</Button>*/}
+          {/*TODO (#1413): uncomment and implement callback*/}
+          {/*<Button variant="pill" color="warning" size="small" className="transfer-offer-accept">*/}
+          {/*  Reject*/}
+          {/*</Button>*/}
```
meiersi-da: left-over code?

apps/wallet/frontend/src/components/ListAllocationRequests.tsx
meiersi-da: Note that showing all legs also makes the request easier to identify and trust, as one will see the counter-parties' legs as well.

meiersi-da (commented):
Thanks a lot. The page looks right, but as per the comments the display of an allocation request needs some refinement.

apps/wallet/frontend/src/components/ListAllocationRequests.tsx
OriolMunoz-da: not quite (though I should change the classnames): they will be used for the accept/reject of allocations so I kept them there
also useful every time I update the design

OriolMunoz-da:
latest without buttons
image
latest with buttons
image

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AllocationsFrontendIntegrationTest.scala
```diff
@@ -27,19 +27,34 @@ import java.time.format.DateTimeFormatter
...
 class AllocationsFrontendIntegrationTest
     extends FrontendIntegrationTestWithSharedEnvironment("alice")
     with WalletTestUtil
     with WalletFrontendTestUtil
-    with FrontendLoginUtil {
+    with FrontendLoginUtil
+    with TokenStandardTest {
+
+  private val tokenStandardTestDarPath =
+    "token-standard/splice-token-standard-test/.daml/dist/splice-token-standard-test-current.dar"
```
moritzkiefer-da: this probably needs to be adjusted to match https://github.com/hyperledger-labs/splice/pull/1428/files

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AllocationsFrontendIntegrationTest.scala
```diff
@@ -172,7 +178,72 @@ class AllocationsFrontendIntegrationTest
...
+        bobParty,
+        bobTransferAmount,
+        venueParty,
+      )
+
+      withFrontEnd("alice") { implicit webDriver =>
+        browseToAliceWallet(aliceDamlUser)
+        browseToAllocationsPage()
+
+        clue("check that the allocation request is shown") {
```
moritzkiefer-da, line 213: shouldn't you check that you can accept it? The usual guideline I give is that you want one integration test per write request but the read side is usually fine to test through vitest. This seems to do the opposite. You are testing the read side in an integration test but you don't test the write side.

apps/wallet/frontend/src/components/ListAllocationRequests.tsx
```diff
@@ -0,0 +1,281 @@
...
+            <Typography variant="body2">
+              Allocate before: <DateWithDurationDisplay datetime={allocateBefore} enableDuration />
+            </Typography>
+            <Typography variant="body2">
+              Settle before: <DateWithDurationDisplay datetime={settleBefore} enableDuration />
+            </Typography>
+            {Object.keys(requestMeta.values).length > 0 ? (
+              <>
+                <Typography variant="h5">Request Meta</Typography>
+                <MetaDisplay meta={requestMeta.values} id={`${request.contractId}-settlement`} />
```
moritzkiefer-da: I think I don't understand the point of your `id` field here. It's not a globally unique id as you seem to reuse the same for request metadata and settlement metadata. And you only use it in the `key` field which does not have to be globally unique. So why do we need it at all?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AllocationsFrontendIntegrationTest.scala
moritzkiefer-da: ah looking below you didn't actually implement the allocation

OriolMunoz-da: that's next https://github.com/hyperledger-labs/splice/issues/1100

moritzkiefer-da: makes sense :+1:

PR #1423, opened 2025-07-10
[ci] Set topology delay to 250ms

nicu-da:
Should improve network throughput

Relates to https://github.com/DACH-NY/canton-network-node/issues/6055

Assumption is that validFrom = sequenced + change delay (effective), lets see what the tests say 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/sv1/SV1Initializer.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/sv1/SV1Initializer.scala
@@ -408,8 +408,7 @@ class SV1Initializer(
         )
         val initialValues = DynamicSynchronizerParameters.initialValues(clock, ProtocolVersion.v33)
         val values = initialValues.tryUpdate(
-          // TODO(DACH-NY/canton-network-node#6055) Consider increasing topology change delay again
-          topologyChangeDelay = NonNegativeFiniteDuration.tryOfMillis(0),
+          topologyChangeDelay = NonNegativeFiniteDuration.tryOfMillis(250),
           trafficControlParameters = Some(initialTrafficControlParameters),
           reconciliationInterval =
             PositiveSeconds.fromConfig(SvUtil.defaultAcsCommitmentReconciliationInterval),
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvHandler.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvHandler.scala
@@ -574,7 +574,7 @@ class HttpSvHandler(
                 case _ => None
               } match {
               case Some(activeMapping) =>
-                activeMapping.base.sequenced
+                activeMapping.base.validFrom
               case None =>
                 throw Status.NOT_FOUND
                   .withDescription(


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/sv1/SV1Initializer.scala
```diff
@@ -408,8 +408,7 @@ class SV1Initializer(
         )
         val initialValues = DynamicSynchronizerParameters.initialValues(clock, ProtocolVersion.v33)
         val values = initialValues.tryUpdate(
-          // TODO(DACH-NY/canton-network-node#6055) Consider increasing topology change delay again
-          topologyChangeDelay = NonNegativeFiniteDuration.tryOfMillis(0),
+          topologyChangeDelay = NonNegativeFiniteDuration.tryOfMillis(250),
```
moritzkiefer-da: that won't actually work on prod clusters will it? I think you need to add a config parameter to the app for the desired topology change delay and then set it from the reconcile domain parameters trigger

nicu-da: Yup, just wanted to make sure first that it doesn't blow up too bad.
Was planning on just hardcoding to 250ms in the reconcile but good point on making it configurable.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/EnvironmentDefinition.scala
```diff
@@ -178,6 +178,10 @@ case class EnvironmentDefinition(
             ),
           )
         }
+        participants(env).foreach { p =>
+          logger.info(s"Ensuring vetting topology is effective for ${p.name}")(TraceContext.empty)
+          p.topology.synchronisation.await_idle()
```
moritzkiefer-da, line 183: what exactly does that check? I'm always a bit skeptical of the Canton synchronization macros

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/LocalSynchronizerNode.scala
```diff
@@ -249,6 +250,27 @@ final class LocalSynchronizerNode(
...
             }
+            state
+          },
+        logger,
+      )
+      _ <- retryProvider.waitUntil(
+        RetryFor.WaitingOnInitDependency,
+        "mediator_topology_transaction_active",
+        "Mediator onboard topology transaction is active",
+        participantAdminConnection
```
moritzkiefer-da, line 261: do we still need this given that it didn't fix the issue? It's not obvious to me why synchronizing on this helps much

moritzkiefer-da: also not clear to me if synchronizing on the participant is even correct here

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -261,6 +261,8 @@ case class SvAppBackendConfig(
       NonNegativeFiniteDuration.ofHours(24),
     // Defaults to 48h as it must be at least 2x preparationTimeRecordtimeTolerance
     mediatorDeduplicationTimeout: NonNegativeFiniteDuration = NonNegativeFiniteDuration.ofHours(48),
+    topologyChangeDelayDuration: NonNegativeFiniteDuration =
+      NonNegativeFiniteDuration.ofMillis(250),
```
moritzkiefer-da, line 265: are we confident enough in this to already change the default? I think I'd be slightly more comfortable overwriting only in tests and on clusters for now until we've confirmed it doesn't break something. Of  course we can revert before the next release but that's easy to forget especially if someone else cuts the release.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/EnvironmentDefinition.scala
nicu-da: It mentioned some topology queues, it did fix the issue we had so it seems to sort of work

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/LocalSynchronizerNode.scala
nicu-da: We could try it without, though not sure how lucky we were with the timing.
In theory we need to sync on the topology state, right? Can't connect the mediator before the sequencer sees it as active.
Isn't that participant call returned the time of the last even from the sequencer? Not sure how else to check actual sequencer time to ensure the topology transaction is effective

moritzkiefer-da: >  Isn't that participant call returned the time of the last even from the sequencer? Not sure how else to check actual sequencer time to ensure the topology transaction is effective

It's the wrong sequencer though right? I think at this point your participant is not yet connected to your own sequencer.

nicu-da: Ugh, right..I guess checking local time in that case should be sufficient

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
nicu-da: sure, let me set it to 0 and set it to 250 in the test files.
will add support to the pulumi code and start enabling it on clusters next week

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/EnvironmentDefinition.scala
moritzkiefer-da: alright thanks

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
nicu-da: Thinking more I'm not sure it makes sense, it either goes through our CI clusters ok and then I will deploy it to CILR by eod, or we revert in between. If we deploy to CILR and it's unstable we can patch it afterwards to add the support of configuring it to 0.

Issue is that this has to be the new default for everyone to apply it.

moritzkiefer-da: > Issue is that this has to be the new default for everyone to apply it.

I was thinking of changing the default if we're confident it works and otherwise just keeping it at zero on dev/test/mainnet. But fair enough, we can also go for the revert option.

PR #1433, opened 2025-07-10
Fix the server doesn't have a resource type 'stack'

OriolMunoz-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/4968


nicu-da (approved):

julientinguely-da (approved):
thanks

PR #1441, opened 2025-07-10
Attempt to log more information from rejected promises & failed requests

OriolMunoz-da:
An attempt at better debugging https://github.com/DACH-NY/cn-test-failures/issues/4946

### DIFF
--- a/apps/common/frontend/utils/src/OpenAPILoggingMiddleware.ts
+++ b/apps/common/frontend/utils/src/OpenAPILoggingMiddleware.ts
@@ -55,35 +55,44 @@ export class OpenAPILoggingMiddleware<
     const traceparent = context.headers['traceparent'];
 
     // Unfortunately we don't have the URL
-    return context.getBodyAsAny().then(body => {
-      console.debug(
-        `${this.name} got response with status code`,
-        context.httpStatusCode,
-        'and body:',
-        body && JSON.stringify(body).substring(0, MAX_BODY_LENGTH_TO_LOG),
-        `[${traceparent}]`
-      );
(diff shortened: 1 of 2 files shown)


apps/common/frontend/utils/src/OpenAPILoggingMiddleware.ts
```diff
@@ -55,35 +55,44 @@ export class OpenAPILoggingMiddleware<
...
+        };
+      },
+      err => {
+        console.error(
+          `Got an error when getting response body (traceparent: [${traceparent}]:`,
+          err
+        );
+        throw err;
+      }
+    );
```
OriolMunoz-da, line 96: I doubt this accomplishes much, but it's worth a try I guess

apps/common/frontend/utils/src/OpenAPILoggingMiddleware.ts
```diff
@@ -55,35 +55,44 @@ export class OpenAPILoggingMiddleware<
...
-              return Promise.resolve(body);
-            }
-          },
-        },
-      };
-    });
+        };
+      },
+      err => {
+        console.error(
```
moritzkiefer-da, line 90: do we hit this for every 404 for cns?

OriolMunoz-da: no (CI would prove me wrong otherwise): this is only for the case where `context.getBodyAsAny()` is a failed promise for mysterious but important reasons

moritzkiefer-da: sounds good thanks

fayi-da (approved):

Issue #1408, opened 2025-07-09
Ensure that SV app can start without its own sequencer being healthy


moritzkiefer-da:
Currently fails with 
```
2025-07-09T14:30:53.570Z [⋮] INFO - o.l.s.s.NodeInitializer:SV=sv (473a902f970d40ec75b3b94130918b45-node_id-a55789cdf6eb5f5b-c592abaf8d3af97f) - The operation 'Canton Sequencer Admin API answers the getId request' failed with a retryable error (full stack trace omitted): UNAVAILABLE: no healthy upstream
statusCode=UNAVAILABLE 
2025-07-09T14:30:53.571Z [⋮] DEBUG - o.l.s.s.NodeInitializer:SV=sv (473a902f970d40ec75b3b94130918b45-node_id-a55789cdf6eb5f5b-c592abaf8d3af97f) - Retrying on same error kind transient error (request infinite retries) for StatusRuntimeException/UNAVAILABLE: no healthy upstream 
2025-07-09T14:30:53.571Z [⋮] INFO - o.l.s.s.NodeInitializer:SV=sv (473a902f970d40ec75b3b94130918b45-node_id-a55789cdf6eb5f5b-c592abaf8d3af97f) - The operation 'Canton Sequencer Admin API answers the getId request' has failed with an exception. Retrying after a number of 10 failures, and after 322 milliseconds.  
```

There may be more issues

PR #1480, opened 2025-07-15
Support starting SV app without a working sequencer

moritzkiefer-da:
Will do the helm/pulumi changes next.

part of #1408

[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
@@ -267,6 +267,10 @@ case class SvAppBackendConfig(
     expectedTaskDuration: Long = 5000, // milliseconds
     expiredRewardCouponBatchSize: Int = 100,
     bftSequencerConnection: Boolean = true,
+    // Skip synchronizer initialization and synchronizer config reconciliation.
+    // Can be safely set to true for an SV that has completed onboarding unless you
+    // 1. try to reset one of your sequencers or mediators
+    // 2. change sequencer URLs that need to get published externally.
     skipSynchronizerInitialization: Boolean = false,
 ) extends SpliceBackendConfig {
   override val nodeTypeName: String = "SV"


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -267,6 +267,7 @@ case class SvAppBackendConfig(
     expectedTaskDuration: Long = 5000, // milliseconds
     expiredRewardCouponBatchSize: Int = 100,
     bftSequencerConnection: Boolean = true,
+    skipSynchronizerInitialization: Boolean = false,
```
moritzkiefer-da, line 274: I tried to build an option that just handles the sequencer but mediator and sequencer are too closely coupled so this seems like the better option.


I think eventually we might also want to change the default to true. But I think for the next release having an option that SVs need to turn on is good enough.

martinflorian-da, line 274: I briefly contemplated whether this shouldn't be `skipSynchronizerNodeInitialization` instead, but I guess this way it also covers the "sv-1" case better.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
```diff
@@ -349,50 +349,75 @@ class JoiningNodeInitializer(
...
+                decentralizedSynchronizer,
+              )
+              _ = dsoAutomationService.registerPostUnlimitedTrafficTriggers()
+            } yield ()
+          }
+        } else {
+          if (!skipTrafficReconciliationTriggers) {
+            dsoAutomationService.registerTrafficReconciliationTriggers()
+          }
+          checkTrafficReconciliationTriggersRegistered(dsoAutomationService)
```
martinflorian-da, line 399: Should this be in an else?

And more generally - it seems like `skipTrafficReconciliationTriggers` should never be true if `skipSynchronizerInitialization` is true so how about just failing hard if this happens?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
```diff
@@ -349,50 +349,75 @@ class JoiningNodeInitializer(
...
+              config.scan,
             )
-          // For domain migrations, the traffic triggers have already been registered earlier and so we skip that step here.
-          _ = if (!skipTrafficReconciliationTriggers)
-            dsoAutomationService.registerTrafficReconciliationTriggers()
-          _ <- localSynchronizerNode.initializeLocalMediatorIfRequired(
-            decentralizedSynchronizer
+        } else {
+          logger.info(
+            "Skipping synchronizer config reconciliation as skipSynchronizerInitialization is enabled"
```
martinflorian-da, line 417: Hmm so this is a bit surprising... we're not only skipping initialization but we're also not updating the synchronizer-related config if needed.

How about adding a comment around the config option that it needs to get set to `false` every time you changed something in your sequencer/mediator config that needs to get updated on ledger; this might be implied in nearly 100% of the cases relevant in practice (replacing sequencer/mediator with a fresh one...) but perhaps better to be explicit?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
moritzkiefer-da: I prefer as little deviation between the two config options as possible. Any deviation will just lead to surprises. So I'd rather just keep the logic as is to match how we register the triggers if they are enabled.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
moritzkiefer-da: added a comment

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
martinflorian-da: Ah I see now that this what we're doing for the "initialize synchronizer" case as well. Hm I think I'd still replace the if with an assert(false), to make our expectation explicit and so that in case we get a surprise here we are more likely to hear about it. But no strong opinion.

moritzkiefer-da: if you don't feel strongly about it I'll keep it as is, just seems less to think about for me to say "it's the same as the other code path" than to reason about which combination of options maybe could be excluded

PR #1489, opened 2025-07-15
Add allocation request accept button

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/1100

"Design" unchanged from https://github.com/hyperledger-labs/splice/pull/1415#issuecomment-3057864939

### DIFF
--- a/apps/wallet/frontend/src/components/ListAllocationRequests.tsx
+++ b/apps/wallet/frontend/src/components/ListAllocationRequests.tsx
@@ -21,7 +21,7 @@ import {
   DisableConditionally,
   Loading,
 } from '@lfdecentralizedtrust/splice-common-frontend';
-import { AllocationRequest } from '@daml.js/splice-api-token-transfer-instruction/lib/Splice/Api/Token/AllocationRequestV1/module';
+import { AllocationRequest } from '@daml.js/splice-api-token-allocation-request/lib/Splice/Api/Token/AllocationRequestV1/module';
 import { Contract } from '@lfdecentralizedtrust/splice-common-frontend-utils';
 import { usePrimaryParty } from '../hooks';
 import dayjs from 'dayjs';
--- a/apps/wallet/frontend/src/hooks/useTokenStandardAllocationRequests.ts
+++ b/apps/wallet/frontend/src/hooks/useTokenStandardAllocationRequests.ts
@@ -4,7 +4,7 @@ import { Contract } from '@lfdecentralizedtrust/splice-common-frontend-utils';
 import { useQuery, UseQueryResult } from '@tanstack/react-query';
 
 import { useWalletClient } from '../contexts/WalletServiceContext';
-import { AllocationRequest } from '@daml.js/splice-api-token-transfer-instruction/lib/Splice/Api/Token/AllocationRequestV1/module';
+import { AllocationRequest } from '@daml.js/splice-api-token-allocation-request/lib/Splice/Api/Token/AllocationRequestV1/module';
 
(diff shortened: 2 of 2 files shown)


apps/wallet/frontend/package.json
```diff
@@ -14,7 +14,9 @@
     ]
   },
   "dependencies": {
-    "@daml.js/splice-api-token-transfer-instruction": "file:../../common/frontend/daml.js/splice-api-token-allocation-request-v1-1.0.0",
+    "@daml.js/splice-api-token-metadata": "file:../../common/frontend/daml.js/splice-api-token-metadata-v1-1.0.0",
+    "@daml.js/splice-api-token-allocation": "file:../../common/frontend/daml.js/splice-api-token-allocation-v1-1.0.0",
+    "@daml.js/splice-api-token-allocation-request": "file:../../common/frontend/daml.js/splice-api-token-allocation-request-v1-1.0.0",
```
OriolMunoz-da, line 19: without these, the version must be hardcoded to 1.0.0 in the imports (and happens to work, which i hate)

apps/wallet/frontend/src/components/ListAllocationRequests.tsx
```diff
@@ -211,34 +225,65 @@ const TransferLegsDisplay: React.FC<{
...
+      );
+      return await createAllocation(payload);
+    },
+    onSuccess: () => {},
+    onError: error => {
+      console.error('Failed to submit allocation', error);
+    },
+  });
+
+  if (!actionAllowed) return null;
```
moritzkiefer-da, line 258: just returning null doesn't seem very helpful. Two questions:

1. We don't ever display the button if the action is not allowed do we?
2. So if that's the case, can we just log a console error here at least?

apps/wallet/frontend/src/components/ListAllocationRequests.tsx
```diff
@@ -211,34 +225,65 @@ const TransferLegsDisplay: React.FC<{
...
+  // TODO (#1413): show the withdraw button and implement the callback, instead of showing nothing
+  if (alreadyAccepted) return null;
```
moritzkiefer-da, line 260: nit: move at the start of the function, doing a lot of stuff only to then return null seems pointless

moritzkiefer-da (approved):
looks reasonable to me thank you!

apps/wallet/frontend/src/components/ListAllocationRequests.tsx
OriolMunoz-da: > We don't ever display the button if the action is not allowed do we?

correct

> So if that's the case, can we just log a console error here at least?

ultimately I'm moving the `null` somewhere: either the parent checks that the button should be rendered, or this component itself does (no strong feelings one way or the other)

note that this is because we want to show all legs of an allocation. From my discussion with @meiersi-da :
image

apps/wallet/frontend/src/components/ListAllocationRequests.tsx
OriolMunoz-da: I don't think we can:
all of the stuff being done is either used for these two variables or is the `useMutation`, which is a hook and I cannot use it _after_ the `if` check because that breaks the (annoying) condition that react hooks [must not be called after a conditional return statement](https://react.dev/warnings/invalid-hook-call-warning#breaking-rules-of-hooks)

moritzkiefer-da: ah ofc forgot about the react hook conditions

PR #1509, opened 2025-07-16
Fix synchronizer connection in domain data restorer

moritzkiefer-da:
fixes https://github.com/DACH-NY/canton-network-internal/issues/900

This is the reason why it only fails in that one ciperiodic test. Everything else properly sets the named argument.

[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/migration/DomainDataRestorer.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/migration/DomainDataRestorer.scala
@@ -42,7 +42,7 @@ class DomainDataRestorer(
         manualConnect = true,
         initializeFromTrustedSynchronizer = true,
         timeTracker = SynchronizerTimeTrackerConfig(
-          timeTrackerMinObservationDuration
+          minObservationDuration = timeTrackerMinObservationDuration
         ),
       )
       // We rely on the calls here being idempotent


moritzkiefer-da:
> I am confused, why did this fail only with the topology change delay. Did that actually require a larger observation latency and we were breaking it because we were setting it to that low value from defaultminObservationDuration.

Without the topology change delay when you see the transaction submitted it goes through immediately and all is good.

With a topology change delay, for the topology snapshot to update, the node needs to observe a sequencing time >= the effective time. The way this work is through time proofs. But this made the time proofs very infrequent.

This is also partially why I'm starting to be a bit nervous about this on mainnet: It could be that we might add a _lot_ of time proofs and actually make things worse. Easy to see on CILR though.

gerolf-da:
`minObservationLatency` says that if you request a timeproof for time `t`, the `SynchronizerTimeTracker` will actually request a tick only if it hasn't seen a sequenced timestamp after `t + minObservationLatency`. Basically, the idea is that on a busy-ish ledger, you don't actually need to send timeproofs, because you've observed a timestamp from some other event. However, when the node does not receive any other events, it artificially delays the topology transaction from becoming effective from the topology client's point of view. In other words, the approximate snapshot returns a snapshot without the topology transaction for longer, in this case up to 1 minute longer. The default value is 250ms btw, so this is quite a significant increase of that delay.

PR #1529, opened 2025-07-17
docs, fix: Move splice-info notes to unreleased

giner:
This is a follow up fix for https://github.com/hyperledger-labs/splice/pull/332


moritzkiefer-da:
@giner missing signed off message

PR #1565, opened 2025-07-17
Add configuration around validator & participant init containers

sgatewood-da:
## Background

- We have been using [fluxCD patches](https://fluxcd.io/flux/components/helm/helmreleases/#post-renderers) to avoid upstreaming configuration we need
  - We have seen the error of our ways and want to upstream things instead 🙂
- This PR addresses an issue we have with the `pg-init` containers:
  - we actually need a _different_ init container to run the cloudSQL proxy
  - the pg-init container that exists is also not desired because we create/bootstrap our DBs separately (and the `cnadmin` hardcoded user doesn’t work for us anyway)

## Approach

- Added `.Values.persistence.enablePgInitContainer` to allow us to disable the existing pg-init container
- Added `.Values.extraInitContainers` to allow us to specify extra init containers (i.e. the CloudSQL proxy)

## Note on scope

- This PR is just scoped to **validators** (participant & validator deployments) for ease of review, lower risk, and to avoid breaking my timebox
  - I think our partners want this on all deployments though, so I will follow up with similar changes for the rest (perhaps in a few PRs as well)

## Manual testing

This might be a little hard for anyone to follow, but fwiw I [smoke tested](https://github.com/DACH-NY/canton-network-utilities/pull/5290/commits) these changes in our repo with our actual helm values (just templating tho)

I think this in combination with the helm unittests and the `/cluster_test` should give us confidence in this PR.

#### Cluster Testing
- [ ] ~If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.~

#### PR Guidelines
- [x] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines

### DIFF
--- a/docs/src/validator_operator/validator_helm.rst
+++ b/docs/src/validator_operator/validator_helm.rst
@@ -704,7 +704,7 @@ Refer to the Canton documentation for more details on participant pruning:
     :end-before: PARTICIPANT_PRUNING_SCHEDULE_END
 
 Configuring init containers
--------------------
+---------------------------
 
 If you need to configure init containers on the participant or validator deployments, you can use the following helm
 values for ``splice-participant`` or ``splice-validator``:


sgatewood-da:
(I'm starting work on helm tests now 👍 )

sgatewood-da:
(still TODO: tests on the validator)

nix/daml2js.nix
```diff
@@ -6,15 +6,15 @@ stdenv.mkDerivation rec {
   version = sources.version;
   src = if use_enterprise then
       builtins.fetchurl {
-        url = "https://digitalasset.jfrog.io/artifactory/assembly/daml/${sources.tooling_sdk_version}/daml-sdk-${sources.tooling_sdk_version}-${if stdenv.isDarwin then "macos" else "linux-intel"}.tar.gz";
+        url = "https://digitalasset.jfrog.io/artifactory/assembly/daml/${sources.tooling_sdk_version}/daml-sdk-${sources.tooling_sdk_version}-${if stdenv.isDarwin then "macos-x86_64" else "linux-intel"}.tar.gz";
```
martinflorian-da: Did you mean to commit that? Can we make it a dedicated PR if so? It's unclear to me what the impact here, also as I'm not on MacOS, and traditionally nix changes are more likely than others to cause problems. A dedicated PR makes it easier to revert....

cluster/helm/.gitignore
```diff
@@ -5,3 +5,4 @@ values.yaml
 /.app-charts
 /.image-digests
 LICENSE
+.debug
```
martinflorian-da: Fair, thanks :)

How about we put it in the root `.gitignore`? I frequently run `helm unittest` from there and it then creates the `.debug` folder at the root.

cluster/helm/splice-participant/templates/participant.yaml
```diff
@@ -123,7 +123,9 @@ spec:
             {{- end }}
           {{- end }}
         {{- end }}
+      {{- if or (.Values.persistence.enablePgInitContainer) (.Values.extraInitContainers) }}
       initContainers:
+        {{- if .Values.persistence.enablePgInitContainer }}
```
martinflorian-da, line 128: I think what we really want is a utility function in `splice-util-lib` that reduces the code duplication around init container for all our helm charts...

@sgatewood-da Did you consider this option? (Not a loaded question I promise! I'm considering whether we should open a follow-up tech debt issue for that + adding this feature to all our helm charts.)

martinflorian-da: Ah apologies for not reading the PR description with sufficient attention :)

> I think our partners want this on all deployments though, so I will follow up with similar changes for the rest (perhaps in a few PRs as well)

...I'd definitely nudge you to try to extract the init containers logic into one or more helper methods then. IMO it's enough to only test the result in 1 or 2 sets of Helm tests then, as long as we use the same helper methods everywhere.

cluster/helm/.gitignore
sgatewood-da: moved to https://github.com/hyperledger-labs/splice/pull/1612 (and used the root gitignore) ✅

nix/daml2js.nix
sgatewood-da: 👍 moved to https://github.com/hyperledger-labs/splice/pull/1611

cluster/helm/splice-participant/templates/participant.yaml
sgatewood-da: Agreed that DRY is better for this one since it'll be used in a lot of deployments

Just to check, are you open to us doing that as a followup when we start to generalize this to the other charts ([rule of 3](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)))?

I'm just always looking for de-risking -- my team likes to pull the plug on non-critical things that take more than 1 day to do ⚰️

sgatewood-da: such a change might be good to do as a NOP as well -- i.e. make sure the before/after manifest is identical when moving to the general solution 🙂

martinflorian-da: > are you open to us doing that as a followup when we start to generalize this to the other charts

If adding support for validators and participants quickly would be of noticeable benefit to validator operators then yes :+1: ...and since you went through the trouble of opening this PR I'm guessing that this is the case?

You still need to convince CI though, that is not negotiable :upside_down_face:

sgatewood-da: > If adding support for validators and participants quickly would be of noticeable benefit to validator operators then yes 👍 

This is indeed the case (at least for **our team** as a validator operator 🙂). All other work in this vein from my end is just community service for our SV partners who seemed to want the same thing

I never mentioned -- this is actually not a change for DA SV-1, but rather for the CNU validator deployments

sgatewood-da:
@martinflorian-da 404 for me 🥲

image

sgatewood-da:
Ok I think I have fixed the headers for real now: 
image

sgatewood-da:
^^ nvm I have to add one

PR #1566, opened 2025-07-17
pulumi, info, fix: Do not prepend "md5" suffix twice

giner:
The suffix is prepended by the helm chart so we don't need to do this again.


giner:
I've rebased the branch to reword the commit messages, no code changes

PR #1586, opened 2025-07-18
Backend to allow rejecting an AllocationRequest

OriolMunoz-da:
Part of #1413


apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStoreTest.scala
```diff
@@ -1452,6 +1452,65 @@ abstract class MultiDomainAcsStoreTest[
...
+          ),
+        )
+        _ <- d2.create(
+          contract2,
+          implementedInterfaces = Map(
+            holdingv1.Holding.INTERFACE_ID_WITH_PACKAGE_ID -> view2.toValue
+          ),
+        )
+        result1 <- store.findInterfaceViewByContractId(
+          holdingv1.Holding.INTERFACE
```
moritzkiefer-da, line 1484: tbc this is only in tests right? In prod we can always query for amulet or do we actually have to query for holding somewhere?

moritzkiefer-da: looks like we only use it for allocrequest so all good

moritzkiefer-da (approved):
Nice thanks!

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStoreTest.scala
OriolMunoz-da: we do use this method in prod: https://github.com/hyperledger-labs/splice/pull/1586/files#diff-364146aeae5784cbde9c444a8c21d4ef2b2183a1c8072acf3dcfe0451b65c3f4R1310
but only for allocrequest, yes

PR #1462, opened 2025-07-11
CIP-0058 (UI) Add CreateUnallocatedUnclaimedActivityRecord vote 

jose-velasco-ieu:
## Summary

This PR introduces support for the `SRARC_CreateUnallocatedUnclaimedActivityRecord` vote request action. It includes the following updates:

- Added a new form component (`CreateUnallocatedUnclaimedActivityRecord`) with fields for beneficiary, amount, reason, and expiration date.
- Pre-fills the expiration date to be one day after the "Vote Request Effective At" timestamp.
- Validates the `amount` input to ensure it's a positive number greater than zero.
- Ensures conditional rendering of form sections based on the selected action.
- Added a new test case to the `SvFrontendIntegrationTest` suite to verify rendering and interaction of the new form.

These changes enable Super Validators (SVs) to create vote requests for unclaimed activity records with appropriate validation and review flows.

> Note: The new test case in `SvFrontendIntegrationTest` has not been run locally due to memory limitations on my machine. However, I’m confident it will pass as it is straightforward and largely copied from an existing test. I’m actively working on resolving the issue with my local setup.

cip-0058-ui-1
cip-0058-ui-2_2
cip-0058-ui-3
cip-0058-ui-4_2
cip-0058-ui-5
cip-0058-ui-6_2

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [X] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [X] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #1587, opened 2025-07-18
CIP-0058 (Script - 2)  Add weight as input + test for invalid weight

jose-velasco-ieu:
## 🧾 Summary

This PR improves the `unclaimed_sv_rewards.py` script to support milestone-based reward accounting by allowing fine-grained weight input and enforcing correct validation.

The script summarizes **claimed**, **expired**, and **unclaimed** minting rewards for a given **beneficiary** within a specified time range and **weight**.

This pull request is an extension of [#1462](https://github.com/hyperledger-labs/splice/pull/1462).

---

## ✅ Key Improvements

- **Validates `--weight` input**: Ensures the provided weight is not greater than any eligible reward coupon’s weight.
- **Fails early with clear messaging**: Invalid weight causes an immediate and informative failure.
- **Fix**: Removed duplicated error logging (errors were previously logged in both `_fail` and via `sys.excepthook`).

---

### ⏱ Timeline

```
   t0                    t1                 t2                      t3
   |   m₁ (w=1)          |   m₂ (w=1)       |   m₃ (w=2)            |
---------------------------------------------------------------------------------
   |444444444444444444444|333333333333333333|22222222222222222222222|00000...
```

### 🧩 Steps

- At **t0**:
  - GSF sets `Ghost-party-A` weight = 4

- At **t1** (m₁ is completed):
  - GSF sets weight = 3
  - GSF runs the script with:
    ```
    --begin-record-time=t0 --end-record-time=t1 --weight=1
    ```
  - A vote is initiated for **m₁**

- At **t2** (m₂ is completed):
  - GSF sets weight = 2
  - GSF runs the script with:
    ```
    --begin-record-time=t0 --end-record-time=t2 --weight=1
    ```
  - A vote is initiated for **m₂**

- At **t3** (m₃ is completed):
  - GSF sets weight = 0 (party removed)
  - GSF runs the script with:
    ```
    --begin-record-time=t0 --end-record-time=t3 --weight=2
    ```
  - A vote is initiated for **m₃**

This enables rewards to be split according to milestone weights and transitions over time.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/scripts/scan-txlog/unclaimed_sv_rewards.py
+++ b/scripts/scan-txlog/unclaimed_sv_rewards.py
@@ -5,7 +5,7 @@
 
 """
 Summarizes claimed, expired, and unclaimed minting rewards for a given beneficiary
-within a specified time range, based on SvRewardCoupon activity.
+within a specified time range and weight, based on SvRewardCoupon activity.
 """
 
 import aiohttp
@@ -102,6 +102,12 @@ def _parse_cli_args() -> argparse.Namespace:
     parser.add_argument(
         "--end-record-time",
         help="End of the record time range to consider SvRewardCoupon creation. Expected in ISO format: 2025-07-01T12:30:00Z",
+        required=True,
+    )
+    parser.add_argument(
+        "--weight",
+        type=int,
+        help="Weight of sv coupon rewards to consider",
         required=True,
     )
     return parser.parse_args()
@@ -464,6 +470,7 @@ class State:
     grace_period_for_mining_rounds_in_minutes: datetime
     create_sv_reward_end_record_time: datetime
     pagination_key: PaginationKey
+    weight: int
     reward_summary: RewardSummary
 
     @classmethod
@@ -495,6 +502,7 @@ class State:
             grace_period_for_mining_rounds_in_minutes = grace_period_for_mining_rounds_in_minutes,
             create_sv_reward_end_record_time = datetime.fromisoformat(args.end_record_time),
             pagination_key=pagination_key,
+            weight = args.weight,
             reward_summary = reward_summary,
         )
 
@@ -608,7 +616,8 @@ class State:
                             case None:
                                 self._fail_with_missing_round(reward)
                             case mining_round_info:
-                                amount = reward.weight * mining_round_info.issuance_per_sv_reward
+                                self._verify_weight(reward)
+                                amount = self.weight * mining_round_info.issuance_per_sv_reward
                                 LOG.debug(
                                     f"Updating expired summary with amount {amount}, corresponding to contract {event.contract_id}"
                                 )
@@ -619,7 +628,8 @@ class State:
                             case None:
                                 self._fail_with_missing_round(reward)
                             case mining_round_info:
-                                amount = reward.weight * mining_round_info.issuance_per_sv_reward
+                                self._verify_weight(reward)
+                                amount = self.weight * mining_round_info.issuance_per_sv_reward
                                 LOG.debug(
                                     f"Updating claimed summary with amount {amount}, corresponding to contract {event.contract_id}"
                                 )
@@ -627,6 +637,11 @@ class State:
                                 self.reward_summary.reward_claimed_total_amount += amount
 
         self.process_events(transaction, event.child_event_ids)
+
+    def _verify_weight(self, reward):
+        if self.weight > reward.weight:
+            msg = f"Invalid weight input: {self.weight} must be less than or equal to {reward.weight}"
+            self._fail(msg)
 
     def _fail_with_missing_round(self, reward):
         self._fail(
@@ -636,7 +651,6 @@ class State:
         )
 
     def _fail(self, message, cause=None):
-        LOG.error(message)
         raise Exception(
             f"Stopping script (error: {message})"
         ) from cause
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedSvRewardsScriptIntegrationTest.scala
@@ -71,6 +71,11 @@ class UnclaimedSvRewardsScriptIntegrationTest
       val svRewardCouponsClaimedCount = 1L
       val svRewardCouponsUnclaimedCount = 2L
 
+      val sv1Name = sv1Backend.config.onboarding.map(_.name).getOrElse(fail("sv1 name not found"))
+      val sv1TotalWeight: Long =
+        sv1Backend.config.rewardWeightBpsOf(sv1Name).getOrElse(fail("sv1 weight not found"))
+      sv1TotalWeight should be > 0L
+
       // Some rewards gets created before now
       val beginRecordTime = Instant.now().minus(10, ChronoUnit.MINUTES)
 
@@ -90,7 +95,7 @@ class UnclaimedSvRewardsScriptIntegrationTest
         "All reward coupons got created",
         _ => {
           val svRewardCoupons = sv1WalletClient.listSvRewardCoupons()
-          svRewardCoupons should have size (svRewardCouponsCount)
+          svRewardCoupons should have size svRewardCouponsCount
           svRewardCoupons
         },
       )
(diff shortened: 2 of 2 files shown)


scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -608,7 +616,8 @@ def handle_sv_reward_coupon_exercise(self, transaction, event, status):
                             case None:
                                 self._fail_with_missing_round(reward)
                             case mining_round_info:
-                                amount = reward.weight * mining_round_info.issuance_per_sv_reward
+                                self._verify_weight(reward)
+                                amount = self.weight * mining_round_info.issuance_per_sv_reward
```
moritzkiefer-da: This seems dangerous to me. Consider this:

Rewards get issued for weight 10.

You then run the script with weight 7 (let's assume this was a milestone or something). This succeeeds.

You then rerun it with script weight 6. This will also succeed.

But clearly the totals don't match up.

Now you can argue that this will never happen in practice but it doesn't seem that unlikely to me. Just a misconfiguration in the beneficiaries could result in it. Or you've changed the weight sometime in the time range deliberately but you started running the script before that when the weight was not effective yet.

jose-velasco-ieu: I see your point. However, if the user runs the script with invalid parameters, the other SVs can reject the vote based on the number provided by the user requesting the vote. I don’t see a way to fully protect against erroneous or malicious inputs.
cc: @waynecollier-da

moritzkiefer-da: > I see your point. However, if the user runs the script with invalid parameters, the other SVs can reject the vote based on the number provided by the user requesting the vote

Sure but I have no idea how I as an SV can detect that the parameters are invalid so how should I know to reject it?


I think the problem with the weight option you implemented is that it assumes you can run independent queries with different weights whereas the safe option is to figure out how much you want to mint for that and then only look at the _remainder_ for your next query. But with the script I don't know what the remainder was so I have to rerun it normally which might lead to me double counting some portion of rewards.

The only case where your approach is safe is if the weight stayed constant all the time. So maybe one option is that you add a flag for the expected total weight and assert on that each time you see one created. That guards against mistakes.

It seems like the cleanest option would be to actually use different parties for different milestones but I guess we didn't do that?

jose-velasco-ieu: > It seems like the cleanest option would be to actually use different parties for different milestones but I guess we didn't do that?

Unfortunately, a single "ghost-xxx" party is used for all the milestones. The way GSF reduces the weight after each completed milestone is as follows:

```
   t0                    t1                 t2                      t3
   |   m₁ (w=1)          |   m₂ (w=1)       |   m₃ (w=2)            |
---------------------------------------------------------------------------------
   |444444444444444444444|333333333333333333|22222222222222222222222|00000...
```
So the weight does not stay constant all the time.

cc: @waynecollier-da

moritzkiefer-da: Ok so how about this:

My concern as I explained above is that you might double count stuff twice and I don't see a good way for me as an operator to prevent this with just the weight option. What if we add an `already-minted-weight` option that you subtract first and then only the remainder can be consider for the current `weight` option? As an operator I can reasonably be expected to set `already-minted-weight` option and if it is set, it is always subtracted first so I'm not gonna double count anything.

jose-velasco-ieu: @moritzkiefer-da 
I'm not sure if I'm following you. 

> subtract first 

Do you mean subtract from the processing coupon weight?
Correct me if I'm wrong. The check would like that: 

`coupon_weight - already_minted_weight (input) >= weight (input)`

```
   t0                    t1                 t2                      t3
   |   m₁ (w=1)          |   m₂ (w=1)       |   m₃ (w=2)            |
---------------------------------------------------------------------------------
   |444444444444444444444|333333333333333333|22222222222222222222222|00000...
```
For example in this scenario: 

At t1:
`-begin-record-time=t0 --end-record-time=t1 --weight=1 --already-minted-weight=0`
Between t0 and t1: 4 - 0 >= 1 ---> OK 

At t2:
`-begin-record-time=t0 --end-record-time=t2 --weight=1 --already-minted-weight=1`
Between t0 and t1: 4 - 1 >= 1 ---> OK 
Between t1 and t2: 3 - 1 >= 1 ---> OK 

However at t3:
`-begin-record-time=t0 --end-record-time=t3 --weight=2 --already-minted-weight=2`
Between t0 and t1: 4 - 2 >= 2 ---> OK
Between t1 and t2: 3 - 2 >= 2 ---> FAIL
Between t1 and t2: 2 - 2 >= 2 ---> FAIL

I think we are still missing something, because the already-minted-weight depends on the time period. 
Or maybe I didn't get your idea.

moritzkiefer-da: > I think we are still missing something, because the already-minted-weight depends on the time period.

Can't I just run it with different already minted weights across different time periods?

jose-velasco-ieu: Sure. 
Do you agree with this?
@moritzkiefer-da @waynecollier-da 

```
   t0                    t1                 t2                      t3
   |   m₁ (w=1)          |   m₂ (w=1)       |   m₃ (w=2)            |
---------------------------------------------------------------------------------
   |444444444444444444444|333333333333333333|22222222222222222222222|00000...
```


**CHECK:** In each time period, `weight (input) <= coupon_weight - already_minted_weight (input)`

- At **t0**:
  - GSF sets `Ghost-party-A` weight = 4

- At **t1** (Milestone 1, m₁ is completed):
  - GSF sets weight = 3
  - GSF runs the script with (single execution):
    - t0 to t1
    ```
    --begin-record-time=t0 --end-record-time=t1 --weight=1 --already-minted-weight=0  # 1 <= 4 - 0  (OK)
    ```
  - A vote is initiated for **m₁**

- At **t2** (Milestone 2, m₂ is completed):
  - GSF sets weight = 2
  - GSF runs the script with (2 executions):
    - t0 to t1
    ```
    --begin-record-time=t0 --end-record-time=t1 --weight=1 --already-minted-weight=1  # 1 <= 4 - 1  (OK)
    ```
    - t1 to t2
    ```
    --begin-record-time=t1 --end-record-time=t2 --weight=1 --already-minted-weight=0  # 1 <= 3 - 0  (OK)
    ```
  - A vote is initiated for **m₂**

- At **t3** (Milestone 3, m₃ is completed):
  - GSF sets weight = 0 (party removed)
  - GSF runs the script with (3 executions): 
    - t0 to t1
    ```
    --begin-record-time=t0 --end-record-time=t1 --weight=2 --already-minted-weight=2  # 2 <= 4 - 2  (OK)
    ```
    - t1 to t2
    ```
    --begin-record-time=t1 --end-record-time=t2 --weight=2 --already-minted-weight=0  # 2 <= 3 - 1  (OK)
    ```
    - t2 to t3
    ```
    --begin-record-time=t2 --end-record-time=t3 --weight=2 --already-minted-weight=0  # 2 <= 2 - 0  (OK)
    ```
  - A vote is initiated for **m₃**

moritzkiefer-da: that sounds ok to be. Do the way you have set the times for this here currently work out such that you don't have any gaps or double count something? iirc you add some slack around the end time.

jose-velasco-ieu: I add a grace period to process coupons created  near `--end-record-time.` and claimed and expired later. 
On the other hand I think my execution example needs to add maybe 1 sec to --begin-record-time to avoid double counting. 

For instance (m3):

GSF runs the script with (3 executions):
- t0 to t1
```
--begin-record-time=t0 --end-record-time=t1 --weight=2 --already-minted-weight=2  # 2 <= 4 - 2  (OK)
```
- t1 to t2
```
--begin-record-time=t1+1sec --end-record-time=t2 --weight=2 --already-minted-weight=0  # 2 <= 3 - 1  (OK)
```
- t2 to t3
```
--begin-record-time=t2+1sec --end-record-time=t3 --weight=2 --already-minted-weight=0  # 2 <= 2 - 0  (OK)
```

jose-velasco-ieu: If you agree I can push the changes with the approach above.

jose-velasco-ieu: pushed

moritzkiefer-da: One second is not gapless. Transactions can be closer together than 1s.

The record time in the Scan request is exclusive (just like the offset on the ledger API).

So I don't see why you need it.

If you query from begin record time (t0, t1) you will look at all coupons created with record time > t0 and <= t1. So I think this is already fine.

I think the one thing we are missing is to ensure that the grace period that you wait for is long enough.

An easy option for that seems to be to add an assertion that the `active_rewards` at the end are empty. If they're not you didn't run for long enough which could for example happen if there was an issue that resulted in rounds not advancing.

 This also seems to point at being able to remove that config option and just always read until it is empty but I'm fine with just the assertion for now.

jose-velasco-ieu: done

scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -628,6 +652,12 @@ def handle_sv_reward_coupon_exercise(self, transaction, event, status):
 
         self.process_events(transaction, event.child_event_ids)
 
+    def _verify_weight(self, reward):
+        available_weight = reward.weight - self.already_minted_weight
+        if self.weight > available_weight:
+            msg = f"Invalid weight input: {self.weight} must be less than or equal to {available_weight}"
```
moritzkiefer-da: don't you still want to continue in that case, log a warning and just clamp it to zero? Consider the case where someone misconfigured their beneficiaries for a brief amount of time. They just got fewer rewards then but totally breaking doesn't seem useful.

jose-velasco-ieu: yeah, makes sense

jose-velasco-ieu: done

scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -628,6 +650,23 @@ def handle_sv_reward_coupon_exercise(self, transaction, event, status):
 
         self.process_events(transaction, event.child_event_ids)
 
+    def _calculate_amount(self, reward, mining_round_info):
+        return (
+            self.weight * mining_round_info.issuance_per_sv_reward
```
moritzkiefer-da: This still doesn't look right to me. Let's say I have weight 2 still available but I want weight 3. I would expect you just use weight 2 then and log a warning.

jose-velasco-ieu: I had understood that I had to clamp it to zero.
I see your point. Let me change it.

moritzkiefer-da: yeah sorry I had phrased it poorly

jose-velasco-ieu: No worries 
Done

scripts/scan-txlog/unclaimed_sv_rewards.py
```diff
@@ -651,21 +651,18 @@ def handle_sv_reward_coupon_exercise(self, transaction, event, status):
...
-        return (
-            self.weight * mining_round_info.issuance_per_sv_reward
-            if self._is_weight_ok(reward)
-            else 0
-        )
+        return self._calculate_weight(reward) * mining_round_info.issuance_per_sv_reward
 
-    def _is_weight_ok(self, reward):
+    def _calculate_weight(self, reward):
         available_weight = reward.weight - self.already_minted_weight
```
moritzkiefer-da: Can you guarantee that this never goes negative? I think that's what I meant with clamp to zero earlier.

jose-velasco-ieu: done

PR #1627, opened 2025-07-22
Sketch: Delegated reward minting

moritzkiefer-da:
Intended to be used as basis for a CIP. Needs more polish and testing before it can be merged.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


daml/splice-amulet/daml/Splice/MintingDelegation.daml
```diff
@@ -0,0 +1,112 @@
...
+
+    nonconsuming choice MintingDelegation_Mint : TransferResult
+      with
+        inputs : [TransferInput]
+        context : PaymentTransferContext
+      controller delegate
+      do assertWithinDeadline "expiresAt" expiresAt
+         -- require "Minting is unfeatured" (context.context.featuredAppRightCid == None)
+         -- TODO: do we want to allow amulet inputs? Seems like other than contention there is no real downside and merging
+         -- usually is still reasonable.
```
meiersi-da: Contention with creating a transfer with a time-consuming external signing process seems problematic. That said: it's probably more of a configuration problem. 

I'd expect the following strategy to work quite well:
1. If the owner has at most one holding, then the minter does not merge the holdings.
2. If the owner has more than one holding, the minter merges the rewards with the smallest holding.

This way the owner keeps control over all but their smallest holding.

meiersi-da: So yes, I'd leave the input contracts here, so that such a strategy and others can be implemented. Merging by default seems good for network scalability.

daml/splice-amulet/daml/Splice/MintingDelegation.daml
```diff
@@ -0,0 +1,112 @@
...
+         exercise context.amuletRules AmuletRules_Transfer with
+           transfer = Transfer with
+             sender = beneficiary
+             provider = beneficiary -- TODO: consider whether the provider should be the delegate, perhaps irrelevant if fees go to zero
```
meiersi-da: Yeah, I think that would be a good option. Is there any downside to it?

daml/splice-amulet/daml/Splice/MintingDelegation.daml
```diff
@@ -0,0 +1,112 @@
...
+
+    -- We do not charge a fee for MiningDelegation but we make them expire
+    -- and force paying traffic to renew them to protect against attacks.
+    choice MintingDelegation_Renew : ContractId MintingDelegation
+      with
+        renewalDuration : RelTime
+      controller delegate
+      do require ("renewalDuration " <> show renewalDuration <> " must be positive") (renewalDuration > seconds 0)
+         -- TODO: Consider making maximum renewalDuration configurable
+         require ("renewalDuration must not be more than 30 days " <> show renewalDuration <> " must be positive") (renewalDuration <= days 30)
```
meiersi-da: Should we make this a config parameter? E.g. in `AmuletRules`?

daml/splice-amulet/daml/Splice/MintingDelegation.daml
```diff
@@ -0,0 +1,112 @@
...
+         -- TODO: Consider enforcing that this is "close" to expiry, e.g., not more than 10 days before expiry
```
meiersi-da: Yes good idea. I'd probably go for half of the maximal renewal duration.

daml/splice-amulet/daml/Splice/MintingDelegation.daml
```diff
@@ -0,0 +1,112 @@
...
+      controller p
+      do require ("Controller " <> show p <> " must be one of the stakeholders" <> show (stakeholder this)) (p `elem` stakeholder this)
+         assertDeadlineExceeded "expiresAt" expiresAt
+         pure ()
+
+    -- We do not charge a fee for MiningDelegation but we make them expire
+    -- and force paying traffic to renew them to protect against attacks.
+    choice MintingDelegation_Renew : ContractId MintingDelegation
+      with
+        renewalDuration : RelTime
```
meiersi-da: Any reason to not just always add the maximal renewal duration?

daml/splice-dso-governance-test/daml/Splice/Scripts/TestMintingDelegation.daml
```diff
@@ -0,0 +1,76 @@
...
+    inputs = [InputValidatorLivenessActivityRecord livenessActivityRecordCid]
+    context = PaymentTransferContext with
+      context
+      amuletRules
+
+  [(_, amulet)] <- query @Amulet beneficiary
+
+  amulet.owner === beneficiary
+
+  pure ()
```
meiersi-da, line 77:
```suggestion
  pure ()
  
-- FIXME: test renewal, and the other choices
```

daml/splice-amulet/daml/Splice/ExternalPartyAmuletRules.daml
```diff
@@ -63,6 +65,24 @@ template ExternalPartyAmuletRules
...
 
+
+    nonconsuming choice ExternalPartyAmuletRules_CreateMintingDelegationProposal : ExternalPartyAmuletRules_CreateMintingDelegationProposalResult
+      with
+        beneficiary : Party
+        delegate : Party
+        expiresAt : Time
+      controller beneficiary
+      do assertWithinDeadline "expiresAt" expiresAt
+         -- TODO: Consider making maximum expiration configurable
```
meiersi-da: Yes. Added a like comment below.

Any particular reason why this choice is on `ExternalPartyRules`?

daml/splice-amulet/daml/Splice/MintingDelegation.daml
```diff
@@ -0,0 +1,112 @@
...
+      controller beneficiary
+      do pure ()
+
+    choice MintingDelegation_Expire : ()
+      with
+        p : Party
+      controller p
+      do require ("Controller " <> show p <> " must be one of the stakeholders" <> show (stakeholder this)) (p `elem` stakeholder this)
+         assertDeadlineExceeded "expiresAt" expiresAt
+         pure ()
```
meiersi-da, line 100: Might be nice to share the code for these choices.

Perhaps a boolean field `accepted : Bool` would be OK? It goes against our default modeling though. So happy to leave it.

daml/splice-amulet/daml/Splice/MintingDelegation.daml
moritzkiefer-da: no will switch

daml/splice-amulet/daml/Splice/MintingDelegation.daml
moritzkiefer-da: > Should we make this a config parameter? E.g. in AmuletRules?

So the problem with doing that is that we then reintroduce contention on AmuletRules config changes which iirc was the reason we added ExternalPartyAmuletRules in the first place as it doesn't work well for external parties and signing delays. I guess we do have the same problem for token standard transfers so maybe we can just accept it? Does feel a bit unfortunate though, we have a config change roughly every week atm.

meiersi-da: Don't we expect the delegate to be able to act every 10' at least? So they can deal well with this kind of contention can't they?

moritzkiefer-da: Ah sorry I think I commented on the wrong choice. I meant the one for creating the proposal which the beneficiary exercises. For the other ones you're right, this is not an issue.
 
So I guess what we can do is that the proposal just has a hardcoded expiry of a week or something like that and then the expiry of the actual minting delegation is determined through amulet rules.

meiersi-da: Yes, making the proposal expire after a week seems like a good idea.

moritzkiefer-da: Hm hardcoding a week doesn't work. For that I would need `getTime` access to determine the absolute time or I would need the time primitives that are relative to contract creation time.

So I'll go for hardcoding a maximum of a week but let the user chose the absolute time.

moritzkiefer-da:
Currently likely still missing:

Add a flag to `ValidatorLicense` whether it is a validator operator or not. Because it seems like the obvious first question we get once we support this for all parties is "how can I figure out who is a validator operator". It seems ok that this flag is just enforced through the SV app backend/UI when granting the license.

daml/splice-amulet/daml/Splice/MintingDelegation.daml
```diff
@@ -0,0 +1,118 @@
...
+
+template MintingDelegation
+  with
+    dso : Party
+    beneficiary : Party
+    delegate : Party
+    expiresAt : Time
```
meiersi-da, line 57:
```suggestion
    expiresAt : Time
    amuletMergeLimit : Int
      -- ^ The number of amulet contracts to keep after auto-merging; i.e.,
      -- auto-merge contracts once there are strictly more than this number of contracts.  
```

daml/splice-amulet/daml/Splice/MintingDelegation.daml
```diff
@@ -0,0 +1,118 @@
...
+import Splice.AmuletRules
+import Splice.Round
+import Splice.Schedule
+import Splice.Types
+import Splice.Util
+import Splice.ValidatorLicense
+
+template MintingDelegationProposal
+  with
+    dso : Party
```
meiersi-da, line 15:
```suggestion
```

@dfordivam : the DSO party is actually not required on this and the other contracts; and we'll really want to remove it to avoid bloating the ACS of the DSO; and the amount of automation it needs to run to collect expired contracts.

Can you please change the code here to move the `MintingDelegation` into the `splice-wallet` and out of `splice-amulet`. It should be a bilateral contract only.

dfordivam: Ok, I will make the change

PR #1629, opened 2025-07-22
cluster, helm, splice-info: Configuration tuning

giner:
- Hardening: Do not expose Nginx version
- Enable gzipping of responses
- Allow localdev as a value for "network" (useful for local development)

### DIFF
--- a/cluster/helm/splice-info/values.schema.json
+++ b/cluster/helm/splice-info/values.schema.json
@@ -62,7 +62,7 @@
       "properties": {
         "network": {
           "type": "string",
-          "enum": ["dev", "test", "main"]
+          "enum": ["localdev", "dev", "test", "main"]
         },
         "configDigest": {
           "type": "object",
--- a/cluster/helm/splice-info/templates/configmap-config.yaml
+++ b/cluster/helm/splice-info/templates/configmap-config.yaml
@@ -8,7 +8,12 @@ metadata:
   namespace: {{ .Release.Namespace }}
 data:
   staticfile.conf: |
+    server_tokens off;
+
     add_header Cache-Control no-cache;
+
+    gzip on;
+    gzip_types {{ .Values.contentType }};
 
     types {
         {{ .Values.contentType }} html;


cluster/helm/splice-info/values.schema.json
```diff
@@ -62,7 +62,7 @@
       "properties": {
         "network": {
           "type": "string",
-          "enum": ["dev", "test", "main"]
+          "enum": ["localdev", "dev", "test", "main"]
```
martinflorian-da: Why not make this a simple string so that people can configure whatever network name they like?

giner: The only reason is to enforce consistency and avoid having dev/test/main along with devnet/testnet/mainnet. Either of these these would be fine. Do you have a case when it doesn't work? Should we change this?

martinflorian-da: > The only reason is to enforce consistency and avoid having dev/test/main along with devnet/testnet/mainnet

OK, that sounds like a good reason really!

> Do you have a case when it doesn't work? Should we change this?

We didn't yet get to deploying this ourselves yet but I suspect we might run into problems with all our testing clusters (`ci*`, `scratch*`)... Happy to change this myself then as needed to support our own deployment logic.

giner: add "scratch" for now

> Happy to change this myself then as needed to support our own deployment logic.

sound good, thank you

giner:
> Thanks! LMK if you still want to make changes; if not I'll merge towards the end of my day.

it's ready to be merged, thank you

cluster/helm/splice-info/values.schema.json
martinflorian-da: > add "scratch" for now

Hm we probably won't use that though; if we do want to set this more correctly we'll probably want `scratcha`, `scratchb` etc.. But again, that's easy enough for me to squeeze in a PR as well once I get to fixing our own deployments of this.

giner: yeah, the check would need to be turned into regex I guess but that can be done later

PR #1649, opened 2025-07-23
Pulumi: fix granting of KMS participant-related IAM roles

martinflorian-da:
~TBH I'm still not fully sure what the problem was but ~this simpler version works on a manual test now. Curiously pulumi was completely messed up on CILR and I had to refresh for it to realize that the IAM mappings were actually missing.

So a plausible hypothesis (thanks @moritzkiefer-da) is that by adding the `parent:` option to IAMMappings I caused a recreate of the mapping (under the parent) followed by a delete of the mapping without parent... but since both map to the same thing in GCP it just killed the actual mapping on GCP.

Some drive-by tweaks while trying to get splice expected files to be more useful (failed at that).

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/common/src/serviceAccount.ts
+++ b/cluster/pulumi/common/src/serviceAccount.ts
@@ -35,9 +35,7 @@ export class GcpServiceAccount extends pulumi.ComponentResource {
     this.name = account.name;
 
     const iamMappings = roles.map(role =>
-      _addRoleToGcpServiceAccount(name, account.project, account.member, role, {
-        dependsOn: [account],
-      })
+      _addRoleToGcpServiceAccount(name, account.project, account.member, role, opts)
     );
     this.registerOutputs({ account, iamMappings });
   }


cluster/deployment/config.yaml
```diff
@@ -5,9 +5,6 @@ svs:
       appsLogLevel: DEBUG
       cantonLogLevel: DEBUG
       cometbftLogLevel: debug
-#  sv-1:
-#    cometbft:
-#      snapshotName: "nicu-test-cometbft-cloning"
```
moritzkiefer-da, line 10: why remove this completely?

cluster/pulumi/common/src/serviceAccount.ts
```diff
@@ -35,10 +35,7 @@ export class GcpServiceAccount extends pulumi.ComponentResource {
     this.name = account.name;
 
     const iamMappings = roles.map(role =>
-      _addRoleToGcpServiceAccount(name, account.project, account.member, role, {
-        parent: this,
-        dependsOn: [account],
```
moritzkiefer-da, line 40: why is removing the dependsOn ok?

moritzkiefer-da (approved):
thx

cluster/pulumi/common/src/serviceAccount.ts
```diff
@@ -35,10 +35,7 @@ export class GcpServiceAccount extends pulumi.ComponentResource {
     this.name = account.name;
 
     const iamMappings = roles.map(role =>
-      _addRoleToGcpServiceAccount(name, account.project, account.member, role, {
-        parent: this,
```
martinflorian-da, line 39: Removing this is the actual fix to avoid this problem on the next cluster we update; the rest is just because I decided I don't want to try to be smart anymore.

cluster/deployment/config.yaml
martinflorian-da: It's commented out and has "test" in it :shrug: Happy to add it back in if you think serves a purpose

moritzkiefer-da: oh I missed that it was commented out, all good then

cluster/pulumi/common/src/serviceAccount.ts
martinflorian-da: Empirically it works without it (it was added in my "refactoring PR").

I think GCP doesn't care if the SA were granting this to actually exists.

martinflorian-da: ...and the one additional error I found in that PR was that I'm ignoring the `opts` here; patching the `opts` to add only the `dependsOn` didn't seem worth the complexity given that I don't expect the pulumi dependency to do anything meaningful here.

PR #1652, opened 2025-07-23
Add app rate limiting - included in scan http apis

nicu-da:
part of https://github.com/hyperledger-labs/splice/issues/1343

Decided to not use canton's rate limiter because it uses an atomic ref which has high contention on concurrent access. the guava implementation should be battle tested 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanIntegrationTest.scala
@@ -758,7 +758,7 @@ class ScanIntegrationTest extends IntegrationTest with WalletTestUtil with TimeT
       )
     }
 
-    loggerFactory.suppressErrors {
+    loggerFactory.suppressWarningsAndErrors {
       // ignore timeout failures
       Try {
         doCall()


apps/common/src/test/scala/org/lfdecentralizedtrust/splice/util/SpliceRateLimiterTest.scala
```diff
@@ -0,0 +1,125 @@
...
+
+    "reject requests that are over the limit" in {
+      val (rateLimitMetrics, rateLimiter) = newRateLimiter
+
+      val results = runThroughRateLimiter(rateLimiter, 11)
+
+      val (accepted, rejected) = results.partition(identity)
+
+      // estimate for running 9 seconds
+      accepted.length should (be > 88 and be < 92)
```
martinflorian-da:
```suggestion
      accepted.length should (be > 85 and be < 95)
```

More buffer to guard against flakiness?

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/util/SpliceRateLimiterTest.scala
```diff
@@ -0,0 +1,125 @@
...
+          "limiter",
+          "test",
+        ),
+      ) should be(rejected.length)
+
+    }
+
+  }
+
+  private def runThroughRateLimiter(rateLimiter: SpliceRateLimiter, runRate: Int) = {
```
martinflorian-da:
```suggestion
  private def runThroughRateLimiter(rateLimiter: SpliceRateLimiter, runsPerSecond: Int) = {
```

nit

cluster/images/scan-app/app.conf
```diff
@@ -47,6 +47,17 @@ canton {
         name-service-name = ${SPLICE_APP_UI_NAME_SERVICE_NAME}
         name-service-name-acronym = ${SPLICE_APP_UI_NAME_SERVICE_NAME_ACRONYM}
       }
+      parameters {
```
martinflorian-da, line 50: Not sure if we want a release notes entry; might be nice info but not a must IMO.

apps/app/src/test/resources/include/scans/_scan.conf
```diff
@@ -14,6 +14,19 @@
     ledger-api.auth-config.type = "none"
   }
 
+  parameters {
+    rate-limiting {
+      default {
+        rate-per-second = 200
+      }
+      rate-limiters {
+        force-acs-snapshot-now {
```
moritzkiefer-da: Did you mix up endpoints here? Force acs snapshot now is a test only endpoint only enabled by `enableForcedAcsSnapshots`. A dedicated rate limit on that doesn't seem all that useful.

I assume you actually meant the ACS endpoint used for recover from key?

moritzkiefer-da: also as a side note: If we do have endpoint specific rate limits, how about we use the actual endpoint path as the config option here?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/HttpRateLimiter.scala
```diff
@@ -0,0 +1,43 @@
...
+      ),
+    )
+
+    import org.apache.pekko.http.scaladsl.server.Directives.*
+
+    if (rateLimiter.markRun()) {
+      pass
+    } else {
+      complete(
+        StatusCodes.TooManyRequests,
```
martinflorian-da: Will we get some form of log message on the server side when we've hit the rate limit? Or just the metrics?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceParametersConfig.scala
```diff
@@ -10,10 +10,12 @@ import com.digitalasset.canton.config.{
   SessionSigningKeysConfig,
   WatchdogConfig,
 }
+import org.lfdecentralizedtrust.splice.util.SpliceRateLimitConfig
 
 final case class SpliceParametersConfig(
     batching: BatchingConfig = BatchingConfig(),
     caching: CachingConfigs = CachingConfigs(),
+    rateLimiting: RateLimitersConfig = RateLimitersConfig(SpliceRateLimitConfig(200), Map.empty),
```
moritzkiefer-da: What is the default based on? Do we have metrics on mainnet now?

apps/app/src/test/resources/include/scans/_scan.conf
nicu-da: I just used a random endpoint that has a low probability to be used in another test. This is mostly to test the rate limiter in the integration test. 
In the actual conf files used by the release I added the endpoint that is used to download the acs snapshot.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceParametersConfig.scala
nicu-da: I used devnet to check actual use. But the most popular endpoint is still well under 10tps/second. 
I've added a pretty high limit to only guard against clients that might go "wild" with the idea of tuning it in the future based on issues we observe.

apps/app/src/test/resources/include/scans/_scan.conf
nicu-da: I've chosen to use the openapi operation name because it's also more in line with the metrics. 
The issue with using the paths is that in certain cases we might need a regex to match it and that seems like a complication that might not add a lot of benefits. But definitely open for it in the future if we observe a need for it.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/HttpRateLimiter.scala
nicu-da: Depends if you have api logging enabled.

apps/app/src/test/resources/include/scans/_scan.conf
moritzkiefer-da: > I just used a random endpoint that has a low probability to be used in another test

How about we overwrite it in the test then? That seems more sensible than have a random endpoint as the default here.

moritzkiefer-da: wfm

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceParametersConfig.scala
moritzkiefer-da: is 0.4.5 the one that enables the metrics? wondering when we can check mainnet

nicu-da:
> > Decided to not use canton's rate limiter because it uses an atomic ref which has high contention on concurrent access. the guava implementation should be battle tested
> 
> Do you know why Canton decided to go for the atomic ref? I'm a bit nervous about introducing yet another rate limiting implementation. I think Canton internally already has several ….

I haven't checked why, we can definitely use the canton implementation (I think there's only one for basic rate limiting), though I also don't see lots of benefits. It should be pretty straightforward to just change the implementation for the SpliceRateLimiter to use the canton RateLimiter.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceParametersConfig.scala
nicu-da: Yes, that's the one enabling both our http server metrics and the istio metrics. I can check the istio access log logs, can probs extract something from there.

nicu-da: the request rate for all the endpoints is at ~35/second in mainnet

apps/app/src/test/resources/include/scans/_scan.conf
nicu-da: done, sync-ed the configs

PR #1654, opened 2025-07-23
Return record_time and update_id in TS CLI transfer

### LINKED ISSUE #908: Return record_time and update_id in TS CLI transfer
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#18610). Original author: @OriolMunoz-da ]

This requires usage of the completions API, which is why it wasn't done as part of DACH-NY/canton-network-node#18522

OriolMunoz-da:
Fixes #908

### DIFF
--- a/token-standard/README.md
+++ b/token-standard/README.md
@@ -218,8 +218,11 @@ npm run cli -- transfer \
 
 Output:
 
-If the output is `{}`, the transfer was successful.
-TODO (#908): record_time and update_id will be added to the output.
+The output will include ` "status": "success" `, when the transfer was successful.
+Additionally, it will also include:
+- `updateId`: the update id identifying the transfer in the ledger.
+- `recordTime`: the time in the ledger at which the transfer happened.
+- `synchronizerId`: the synchronizer id in which the transfer was executed.
 
 
 #### Accept transfer instruction
@@ -264,5 +267,8 @@ from the output of the `list-transfer-instructions` command.
 
 Output:
 
-If the output is `{ "status": "success" }`, the acceptance of the transfer instruction was successful.
-TODO (#908): record_time and update_id will be added to the output.
+The output will include ` "status": "success" `, when the instruction acceptance  was successful.
+Additionally, it will also include:
+- `updateId`: the update id identifying the instruction acceptance in the ledger.
+- `recordTime`: the time in the ledger at which the instruction acceptance happened.
+- `synchronizerId`: the synchronizer id in which the instruction acceptance was executed.
--- a/token-standard/README.md
+++ b/token-standard/README.md
@@ -224,6 +224,9 @@ Additionally, it will also include:
 - `recordTime`: the time in the ledger at which the transfer happened.
 - `synchronizerId`: the synchronizer id in which the transfer was executed.
 
+Any error will be logged with  "Failed to execute transfer:", followed by error details.
+These can happen due to any intermediate request failing, or timing out.
+
 
 #### Accept transfer instruction
 
@@ -272,3 +275,6 @@ Additionally, it will also include:
 - `updateId`: the update id identifying the instruction acceptance in the ledger.
 - `recordTime`: the time in the ledger at which the instruction acceptance happened.
 - `synchronizerId`: the synchronizer id in which the instruction acceptance was executed.
+
+Any error will be logged with  "Failed to accept transfer instruction:", followed by error details.
+These can happen due to any intermediate request failing, or timing out.
--- a/token-standard/cli/src/apis/ledger-api-utils.ts
+++ b/token-standard/cli/src/apis/ledger-api-utils.ts
@@ -387,6 +387,11 @@ async function awaitCompletion(
   });
 
   if (wantedCompletion) {
+    if (wantedCompletion.completionResponse.Completion.value.status) {
+      throw new Error(
+        `Command failed with status: ${wantedCompletion.completionResponse.Completion.value.status}`,
+      );
+    }
     return {
       synchronizerId:
         wantedCompletion.completionResponse.Completion.value.synchronizerTime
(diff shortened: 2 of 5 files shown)


token-standard/cli/src/commands/acceptTransferInstruction.ts
```diff
@@ -61,8 +61,7 @@ export async function acceptTransferInstruction(
     publicKey,
     privateKey,
   );
-  // TODO (#908): this should include record_time and update_id
-  const result = { status: "success" };
+  const result = { ...completion, status: "success" };
```
moritzkiefer-da: I'm confused why can you always return "success"? What if the completion is a failed completion?

token-standard/cli/src/apis/ledger-api-utils.ts
```diff
@@ -322,3 +335,70 @@ function signTransaction(
...
+  synchronizerId?: string;
+  recordTime?: string;
+}
+
+/**
+ * Polls the completions endpoint until
+ * the completion with the given (userId, commandId, submissionId) is returned.
+ * Then returns the updateId, synchronizerId and recordTime of that completion.
+ */
+async function awaitCompletion(
```
moritzkiefer-da, line 359: should there be a timeout or something on this after 2min maybe? (with 1min being the local participant timeout where you would definitely expect _something_ but your connection with the partipant might die or the participant could lose its synchronizer connection so timing out on the client side still seems reasonable)

token-standard/cli/src/commands/acceptTransferInstruction.ts
OriolMunoz-da: what's a failed completion?

moritzkiefer-da: discussed in slack

token-standard/cli/src/apis/ledger-api-utils.ts
```diff
@@ -322,3 +341,98 @@ function signTransaction(
...
+    return (
+      completion.value.userId === userId &&
+      completion.value.commandId === commandId &&
+      completion.value.submissionId === submissionId
+    );
+  });
+
+  if (wantedCompletion) {
+    const status = wantedCompletion.completionResponse.Completion.value.status;
+    if (status && status.code) {
```
moritzkiefer-da:
```suggestion
    if (status && status.code !== 0) {
```
maybe a bit clearer here?

moritzkiefer-da (approved):
nice thanks

token-standard/cli/src/apis/ledger-api-utils.ts
```diff
@@ -239,15 +242,31 @@ export async function submitExerciseCommand(
...
+    ledgerClient,
+    ledgerEnd.offset,
+    partyId,
+    userId,
+    commandId,
+    submissionId,
+  );
+  return promiseWithTimeout(
+    completionPromise,
+    60_000 * 2, // 2m
```
meiersi-da: Consider setting to 30s, as this is for interactive use. Waiting 2minutes seems very long.

Also consider telling users 'getting completion for command submission with command-id <command-id> and submission-id <submission-id>.' so they can query themselves for the status.

token-standard/cli/src/apis/ledger-api-utils.ts
```diff
@@ -322,3 +341,98 @@ function signTransaction(
...
+ * Polls the completions endpoint until
+ * the completion with the given (userId, commandId, submissionId) is returned.
+ * Then returns the updateId, synchronizerId and recordTime of that completion.
+ */
+async function awaitCompletion(
+  ledgerClient: LedgerJsonApi,
+  ledgerEnd: number,
+  partyId: string,
+  userId: string,
+  commandId: string,
```
meiersi-da, line 364: @moritzkiefer-da : given that we are not using command dedup here, we should probably only use the submissionId and leave the command-id unset. This way we do not confuse users with the overly specific check that compares both submission-id and command-id. do you agree?

token-standard/README.md
```diff
@@ -218,8 +218,11 @@ npm run cli -- transfer \
 
 Output:
 
-If the output is `{}`, the transfer was successful.
-TODO (#908): record_time and update_id will be added to the output.
+The output will include ` "status": "success" `, when the transfer was successful.
```
meiersi-da, line 221: What happens in case of an error? do you include the gRPC status code and message from the completion?

meiersi-da: Consider distinguishing an error that is reported from a timeout to observe the completion.

token-standard/cli/src/apis/ledger-api-utils.ts
moritzkiefer-da: We usually set command ids to a random UUID and not leave it unset. I thought you could reuse submission ids across different command ids so just matching on submission ids is unreliable?

token-standard/cli/src/apis/ledger-api-utils.ts
OriolMunoz-da: chose 2m from https://github.com/hyperledger-labs/splice/pull/1654#discussion_r2225370118
cc @moritzkiefer-da

moritzkiefer-da: Waiting a bit more than 30s so you at least can get a proper error in case of unavailable counter parties sounds like a good idea to me.

moritzkiefer-da: I'm ok with 45s though instead of 2min

token-standard/cli/src/apis/ledger-api-utils.ts
meiersi-da: Ah OK. Yeah, I guess it's fine then. What I was concerned about is how to handle retries for the same intent with command dedup. There I'd expect that we only match on the command-id for finding a successful submission; and use the precise submission-id match for seeing whether our current submission definitely failed. 

So all good. Resolving.

PR #1666, opened 2025-07-23
Bump timeout for getAcsSnapshot endpoint

OriolMunoz-da:
As seen in mainnet. Adds support for other endpoints (but only scan)

Tested by:
```
"getAcsSnapshot respects custom timeout" in  { implicit env =>
    sv1ScanBackend.getAcsSnapshot(dsoParty, None)
  }
```
and setting the timeout to 1ms.


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/ScanApp.scala
```diff
@@ -291,11 +291,24 @@ class ScanApp(
...
+                  val base = httpErrorHandler.directive(traceContext).tflatMap { _ =>
                     provide(traceContext)
                   }
-                )
+                  (httpService, config.customTimeouts.get(operation)) match {
+                    // custom HTTP timeouts
+                    case ("scan", Some(customTimeout)) =>
+                      withRequestTimeout(
+                        customTimeout.duration,
+                        httpErrorHandler.timeoutHandler(customTimeout.duration, _),
```
OriolMunoz-da, line 304: also tried:
```
            def buildRouteForOperation(operation: String, httpService: String) = {
              nodeMetrics.httpServerMetrics
                .withMetrics(httpService)(operation)
                .tflatMap { _ =>
                  val httpErrorHandler = new HttpErrorHandler(loggerFactory)
                  val base = httpErrorHandler.directive(traceContext).tflatMap { _ =>
                    provide(traceContext)
                  }
                  (httpService, config.customTimeouts.get(operation)) match {
                    // custom HTTP timeouts
                    case ("scan", Some(customTimeout)) =>
                      withRequestTimeout(
                        customTimeout.duration,
                        httpErrorHandler.timeoutHandler(customTimeout.duration, _),
                      ).tflatMap { _ =>
                        base
                      }
                    case _ =>
                      base
                  }
                }
            }
```
BUT for reasons I don't understand, it does timeout after 1ms (see description in the issue) but still logs that it timed out after 38s. This works, but ping me if you know what akka shenanigans is causing that..

moritzkiefer-da (approved):
nice thanks

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
```diff
@@ -48,6 +48,8 @@ case class ScanAppBackendConfig(
     txLogBackfillBatchSize: Int = 100,
     bftSequencers: Seq[BftSequencerConfig] = Seq.empty,
     cache: ScanCacheConfig = ScanCacheConfig(),
+    customTimeouts: Map[String, NonNegativeFiniteDuration] =
```
nicu-da: Can we move this to the parameters block? That's easy to then extend to all the apps.
It would also be in line with the rate limiting config I'm adding in https://github.com/hyperledger-labs/splice/pull/1652

PR #1670, opened 2025-07-24
cluster, images, fix: Make empty value for SPLICE_MALLOC_ARENA_MAX work

giner:
Use the default only if SPLICE_MALLOC_ARENA_MAX is not set. Previously, setting SPLICE_MALLOC_ARENA_MAX to an empty value would still result in the default (which is 2) being used.

Note: this change is no-op unless your configuration have environment variable SPLICE_MALLOC_ARENA_MAX set to `''`


moritzkiefer-da (approved):
thanks

PR #1677, opened 2025-07-24
Switch load-tester to token-standard transfers

moritzkiefer-da:
There are two main differences:

1. There is no status endpoint. This makes retries a bit harder. We could in theory make some of this work directly through the JSON API. But it seems like if we need this something is already a bit sketchy and treating that as a failed transaction doesn't seem too bad.
2. There is no complete stage. If accept goes through it's done, if it fails it fails.

This likely needs some adjustment on the target rate to keep the same number of confirmation requests as before.

My plan is to merge this and the reduction in SV status reports and then callibrate CILR to have the same as before.

[ci]

fixes #1449

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/load-tester/config.sample.json
+++ b/load-tester/config.sample.json
@@ -6,6 +6,7 @@
       "walletBaseUrl": "$K6_WALLET_BASE_URL",
       "auth": {
         "kind": "oauth",
+        "audience": "$OIDC_AUTHORITY_VALIDATOR_AUDIENCE",
         "oauthDomain": "$K6_OAUTH_DOMAIN",
         "oauthClientId": "$K6_OAUTH_CLIENT_ID",
         "usersPassword": "$K6_USERS_PASSWORD",
--- a/load-tester/src/client/validator/models.ts
+++ b/load-tester/src/client/validator/models.ts
@@ -10,53 +10,17 @@ const contract = z.object({
   created_at: z.string(),
 });
 
-const transferOfferPayload = z.object({
-  trackingId: z.string(),
-  description: z.string(),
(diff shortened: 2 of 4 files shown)


load-tester/config.sample.json
```diff
@@ -6,6 +6,7 @@
       "walletBaseUrl": "$K6_WALLET_BASE_URL",
       "auth": {
         "kind": "oauth",
+        "audience": "$OIDC_AUTHORITY_VALIDATOR_AUDIENCE",
```
moritzkiefer-da, line 9: this was just missing, only matters when running locally but that's convenient for iterating.

load-tester/src/client/validator/models.ts
```diff
@@ -10,54 +10,18 @@ const contract = z.object({
   created_at: z.string(),
 });
 
-const transferOfferPayload = z.object({
```
moritzkiefer-da, line 13: I see no reason to decode any of that. All we care about is the contract id.

load-tester/src/client/validator/models.ts
```diff
@@ -10,54 +10,18 @@ const contract = z.object({
...
+  output: z.object({
+    transfer_instruction_cid: z.string(),
+  }),
 });
 
 export type CreateTransferOfferResponse = z.infer<typeof createTransferOfferResponse>;
 
-const partyAndAmount = z.any();
-
-const listTransactionsItem = z.object({
```
moritzkiefer-da, line 40: no idea why we had that, we never look at transactions.

PR #1687, opened 2025-07-24
Improve ACS snapshot query performance

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/1685

Query plan on CILR:
before:
image

after:
image

as mentioned, it no longer needs to fetch and join the entire snapshot, but only those rows that match.

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
@@ -218,6 +218,12 @@ class AcsSnapshotStore(
                   where row_id between $begin and $end
                """ ++ partyIdsFilter ++ templatesFilter ++ sql"""
                   group by create_id
+                  order by row_id asc
+                  -- this CTE already will contain all snapshot rows (filtered by party id and template, if necessary).
+                  -- They just need to be joined with u_h_creates.
+                  -- Applying the limit later yields a worse query plan, where it requires fetching all snapshots first.
+                  -- See #1685
+                  limit ${sqlLimit(limit)}
                )
                select
                  snapshot.row_id,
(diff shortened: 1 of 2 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -218,6 +218,12 @@ class AcsSnapshotStore(
                   where row_id between $begin and $end
                """ ++ partyIdsFilter ++ templatesFilter ++ sql"""
                   group by create_id
+                  order by row_id asc
+                  -- this CTE already will contain all snapshot rows (filtered by party id and template, if necessary).
+                  -- They just need to be joined with u_h_creates.
+                  -- Applying the limit later yields a worse query plan, where it requires fetching all snapshots first.
+                  -- See #1685
+                  limit ${sqlLimit(limit)}
```
moritzkiefer-da, line 226: do we still also need the limit below? It seems like that should be redundant now?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/AcsSnapshotStoreTest.scala
```diff
@@ -449,12 +453,60 @@ class AcsSnapshotStoreTest
...
           }
           _ <- store.insertNewSnapshot(None, DefaultMigrationId, timestamp1)
-          result <- queryRecursive(store, None, Vector.empty)
+          result <- queryRecursive(store, None, Vector.empty, Seq.empty, Seq.empty)
         } yield {
           result should be(contracts.map(_.contractId.contractId))
         }
       }
 
+      "paginate with filters by both party and template" in {
```
moritzkiefer-da, line 462: Can you explain what exactly the test coverage is that you're adding here? Just a bit hard to see without looking through all existing tests

moritzkiefer-da (approved):
thanks for the quick fix! That looked much simpler than I expected

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/AcsSnapshotStoreTest.scala
OriolMunoz-da: there was a paginate test, one filtering by party, and one by template - I just took the "better be safe than sorry" approach of testing all of them at once

PR #1694, opened 2025-07-25
Refactor traffic waste alert

rautenrieth-da:
Fixes https://github.com/hyperledger-labs/splice/issues/1213

Refactors the traffic alerts:
- Previously, we had one alert for "X kilobytes wasted over the last Y minutes". The thresholds were too low for high load clusters, and difficult to tweak.
- Now, there are two alerts:
  - One for "X kilobytes wasted over the last Y hours". The threshold for this should be high, and defaults to 100kb wasted over 4h (up from 1kb/5min)
  - One for "Median of X kilobytes wasted over the last Y days, when sampling the traffic waste in Z hour intervals". We take the median instead of just using another "X kilobytes wasted over the last Y days", because the latter is usually (when looking at historical data from the last 30 days) dominated by bursts in traffic waste. By taking the median, we filter out spikes and get an alert that is more independent from the previous one. This defaults to a median waste of 40kb over 4h (roughly equivalent to the previous default of 1kb/5min).
  - The alerts use longer time ranges. Previously, the default time range was 5min, now the default time range is 4h. That leads to longer response times, but results in less noise.


rautenrieth-da:
Closing in favor of https://github.com/hyperledger-labs/splice/issues/1632

PR #1701, opened 2025-07-28
Bump nix binary on gha

OriolMunoz-da:
Same as https://github.com/DACH-NY/canton-network-internal/pull/1075 but for GHA


moritzkiefer-da (approved):
what could go wrong

PR #1707, opened 2025-07-28
Fix AnsSubscriptionInitialPaymentTrigger

julientinguely-da:
fixes https://github.com/DACH-NY/cn-test-failures/issues/5151

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/AnsSubscriptionInitialPaymentTrigger.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/AnsSubscriptionInitialPaymentTrigger.scala
@@ -116,7 +116,7 @@ class AnsSubscriptionInitialPaymentTrigger(
                             // check if we confirmed the same initial payment for the same entry already
                             getExistingInitialPaymentConfirmation(
                               entryName,
-                              _ == payment.contractId,
+                              _ => true,
                             ).flatMap {
                               case Some(c) =>
                                 TaskSuccess(


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/AnsSubscriptionInitialPaymentTrigger.scala
```diff
@@ -113,15 +113,19 @@ class AnsSubscriptionInitialPaymentTrigger(
                               s"other initial payment collection has been confirmed for the same ans name ($entryName) with confirmation ${c}"
                             )
                           case None =>
-                            // check if we confirmed the same initial payment for the same entry already
+                            // check if a confirmation exists for this entryName
                             getExistingInitialPaymentConfirmation(
                               entryName,
-                              _ == payment.contractId,
```
julientinguely-da, line 119: It's what led to this morning's flake in which two entries where created concurrently. Note in the previous code, we did not filter for this reason I believe

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/AnsSubscriptionInitialPaymentTrigger.scala
```diff
@@ -113,15 +113,19 @@ class AnsSubscriptionInitialPaymentTrigger(
...
+                              _,
                             ).flatMap {
                               case Some(c) =>
-                                TaskSuccess(
-                                  s"skipping as collection of this initial payment has already been confirmed with confirmation ${c}"
-                                ).pure[Future]
+                                confirmRejectPayment(
```
martinflorian-da: I'm confused: How can this ever be needed?

- if we had confirmed the same payment previously (what this check was about), we don't need the rejection
- if we had confirmed the other payment, then one of these should hold:
  - we caught it in line 108 and rejected it
  - we did not catch it in line 108 due to timing; so we tried confirming an accept and that failed thanks to deduplication; on the next loop when then catch it in 108 and reject

PR #1715, opened 2025-07-29
Pulumi: Support creating arbitrary CRs via infra stack

martinflorian-da:
Part of https://github.com/DACH-NY/canton-network-internal/issues/1078

Tested the mock example in a cluster - works as expected.

The added flexibility seems more useful than building something just for blacklisting endpoints.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/deployment/mock/config.yaml
+++ b/cluster/deployment/mock/config.yaml
@@ -135,3 +135,12 @@ infra:
           - to:
               - operation:
                   paths: ["/api/sv/v0/devnet/onboard/validator/prepare"]
+    mock-cr:
+      apiVersion: mock.example.com/v1
+      kind: MockResource
+      metadata:
+        name: mock-resource
+        namespace: validator1
+      spec:
+        key: value
+        anotherKey: anotherValue
--- a/cluster/expected/infra/expected.json
+++ b/cluster/expected/infra/expected.json
@@ -2051,6 +2051,25 @@
     "custom": true,
     "id": "",
     "inputs": {
+      "apiVersion": "mock.example.com/v1",
+      "kind": "MockResource",
+      "metadata": {
+        "name": "mock-resource",
+        "namespace": "validator1"
+      },
+      "spec": {
+        "anotherKey": "anotherValue",
+        "key": "value"
+      }
+    },
+    "name": "mock-cr",
+    "provider": "",
+    "type": "kubernetes:mock.example.com/v1:MockResource"
+  },
+  {
+    "custom": true,
+    "id": "",
+    "inputs": {
       "displayName": "mock Slack Alert Notification Channel",
       "labels": {
         "channel_name": "#mock-slack-channel"


cluster/pulumi/infra/src/config.ts
```diff
@@ -72,6 +73,7 @@ console.error(
 
 export const infraConfig = fullConfig.infra;
 export const monitoringConfig = fullConfig.monitoring;
+export const extraCustomResourcesConfig = fullConfig.extraCustomResources;
```
nicu-da: shouldn't this be under the infra key?

martinflorian-da: I opted for a dedicated top-level key because we can add CRs about whatever we want here. We also use dedicated top-level keys for monitoring, even though it's also deployed from the infra stack.

Not married to that in any way, happy to change if you prefer.

nicu-da: > We also use dedicated top-level keys for monitoring, even though it's also deployed from the infra stack.

I think it makes more sense to have it under the infra key, the monitoring one is with the notion of using it cross projects but that's not the case here.

cluster/deployment/mock/config.yaml
```diff
@@ -117,3 +117,19 @@ svs:
       kms:
         keyRingId: sv-3_participant_mock
         locationId: us-central1
+extraCustomResources:
+  sv-1-deny-onboard-prepare-endpoint:
+    apiVersion: security.istio.io/v1
+    kind: AuthorizationPolicy
+    metadata:
+      name: deny-onboard-prepare-endpoint
+      namespace: sv-1
```
nicu-da: It's gonna be a bit repetitive though as we want to usually create it in multiple namespaces, any chance we can at least extract the namespaces on top of the resource definition and handle that in pulumi?

martinflorian-da: Can I convince us to invest in the extra complexity once we know we'll need it? So far we only ever needed this once. Worst case we'll have to manually copy paste this config a few times... and then we can still invest into better ergonomics.

nicu-da: Sure, though we would need to copy that 5 times for devnet and 16 times for cilr :see_no_evil:

martinflorian-da: I had an idea now that doesn't seem to add too much complexity: https://github.com/hyperledger-labs/splice/pull/1715/commits/3d506374ad0b45e1e4b722edf96f2d81aaef29da

cluster/pulumi/infra/src/config.ts
martinflorian-da: works for me, moved it

cluster/pulumi/infra/src/extraCustomResources.ts
```diff
@@ -0,0 +1,27 @@
...
+import * as k8s from '@pulumi/kubernetes';
+
+import { infraConfig } from './config';
+
+// Automatically duplicates CRs if multiple namespaces given
+export function installExtraCustomResources(): void {
+  const extraCrs = infraConfig.extraCustomResources;
+  Object.entries(extraCrs).forEach(([name, spec]) => {
+    if (spec.metadata && spec.metadata.namespace) {
+      spec.metadata.namespace.split(' ').forEach((ns: string) => {
```
moritzkiefer-da: any reason not to support a list of namespaces instead of string splitting?

moritzkiefer-da (approved):
Nice thank you!

cluster/pulumi/infra/src/extraCustomResources.ts
martinflorian-da: It seemed more concise and less confusing to write in the config.yaml... happy to change to a yaml list if you prefer, don't care that much about it

martinflorian-da: changing; I guess you have a point that it would be nicer

PR #1732, opened 2025-07-29
cluster, images: Add qemu-user-static to test runner

giner:
Add qemu-user-static to splice-test-docker-runner so that docker could use it for multi-platform builds


isegall-da:
@stas-sbi this we can close, right? Decided against using it.

PR #1742, opened 2025-07-30
Reset network from non-zero initial round

### LINKED ISSUE #1124: Tracking: Reset as of round
Before the coming TestNet reset, we need to support resetting it to a given round number > 0, so that we can reset it roughly to the same point in the mining curve that MainNet is on. 

(I thought we had an issue for this already, but can't find it)

julientinguely-da:
fixes #1124 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/NodeInitializerUtil.scala
```diff
@@ -257,6 +271,125 @@ trait NodeInitializerUtil extends NamedLogging with Spanning with SynchronizerNo
...
+                    val initialRound = migrationDump.participantUsers.users.collectFirst {
+                      case user if user.id == config.ledgerApiUser =>
+                        user.annotations.get(INITIAL_ROUND_USER_METADATA_KEY)
+                    }.flatten match {
+                      case None =>
+                        logger.info(
+                          s"Initial round not found in user's metadata dump as resetting networks with non-zero round " +
+                            s"is only possible from dsoGovernance=0.1.17, so setting it to 0 for good."
+                        )
+                        "0"
```
julientinguely-da, line 319: The fix: set it to 0 if it's not in user's metadata yet, which is anyway the case as no resets happened yet

julientinguely-da: succeeding hdm: https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/23790/workflows/4f2a0e5b-2018-495a-8c4b-d93837785b63

moritzkiefer-da:
@julientinguely-da could you squash the fixes compared to the original change into one commit? It's quite hard to see what you actually did to fix the issue.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/NodeInitializerUtil.scala
```diff
@@ -257,6 +271,129 @@ trait NodeInitializerUtil extends NamedLogging with Spanning with SynchronizerNo
...
+  ): Future[String] =
+    SvConnection(
+      sponsorConfig,
+      upgradesConfig,
+      retryProvider,
+      loggerFactory,
+    ).flatMap { svConnection =>
+      svConnection
+        .getDsoInfo()
+        // the sponsor might use the old api, in this case initial round has to be 0
```
moritzkiefer-da, line 391: I think I don't understand how this fixes migrations. Migrations are special in two ways:

1. The old participant does not have the metadata field set.
2. There is no connection to the sponsor

In particular you don't have JoinWithKey set but the dedicated migration config.

What I would expect to happen is that in this case we accept that the metadata field does not exist and default it to zero.

But your change only seems to affect the query to the sponsor not when we read user metadata How does that help?

julientinguely-da: > I think I don't understand how this fixes migrations.

The code you're commenting on fixes the code to pass AppUpgradeIntegrationTest in which sv2 upgrades before sv1. In this case sv1 still uses the old api and sv2 can't learn from the sponsor, so we set it to 0.

Note this issue wasn't caught last week by AppUgpradeIntegrationTest, but was on this PR

> What I would expect to happen is that in this case we accept that the metadata field does not exist and default it to zero.

yes, that's what we do [here](https://github.com/hyperledger-labs/splice/pull/1742/commits/54c1724ca421e5aa5ca0a7de91a15933f43f26a9)

moritzkiefer-da: ah I see thanks

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/migration/ParticipantUsersDataRestorer.scala
```diff
@@ -48,6 +48,7 @@ class ParticipantUsersDataRestorer(
       _ <- Future.sequence(users.map { user =>
         {
           logger.info(s"Restoring user ${user.id}")
+          logger.info(s"user's annotations: ${user.annotations}")
```
moritzkiefer-da: do we really need this as an info log? That seems a bit too much noise

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/NodeInitializerUtil.scala
```diff
@@ -257,6 +271,129 @@ trait NodeInitializerUtil extends NamedLogging with Spanning with SynchronizerNo
...
+                  case domainMigrationConfig: DomainMigration =>
+                    val migrationDump = loadDomainMigrationDump(domainMigrationConfig.dumpFilePath)
+                    val initialRound = migrationDump.participantUsers.users.collectFirst {
+                      case user if user.id == config.ledgerApiUser =>
+                        user.annotations.get(INITIAL_ROUND_USER_METADATA_KEY)
+                    }.flatten match {
+                      case None =>
+                        logger.info(
+                          s"Initial round not found in user's metadata dump as resetting networks with non-zero round " +
+                            s"is only possible from dsoGovernance=0.1.17, so setting it to 0 for good."
```
moritzkiefer-da:
```suggestion
                          s"Initial round not found in user's metadata dump, defaulting to 0"
```

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/NodeInitializerUtil.scala
```diff
@@ -257,6 +271,129 @@ trait NodeInitializerUtil extends NamedLogging with Spanning with SynchronizerNo
...
+                      case Some(rnd) =>
+                        logger.info(
+                          s"Setting the initial round to $rnd from migration user's metadata dump."
+                        )
+                        rnd
+                    }
+                    setInitialRound(connection, initialRound.toLong)
+                }
+              case None =>
+                logger.debug("No SV onboarding config was found, setting the initial round to 0.")
```
moritzkiefer-da, line 329: that looks wrong to me. Consider this:

1. We rebootstrap testnet with round 5000.
2. An SV joins with JoinWithKey
3. The SV now removes the onboarding config which should no longer be required
4. This seems to overwrite it with round 0.

You need to check metadata here.

julientinguely-da: Then it falls into the code in lines 290-292 and reads it from its own metadata.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/migration/ParticipantUsersDataRestorer.scala
julientinguely-da: removing

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/NodeInitializerUtil.scala
moritzkiefer-da: Ah I see thanks

moritzkiefer-da (approved):
Thanks for bearing with me while I tried to convince myself this safe now, looks good, thanks!

PR #1756, opened 2025-07-30
[ci] start test with paused MergeValidatorLicenseContractsTrigger

ray-roestenburg-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/4938

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvMergeDuplicatedValidatorLicenseIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvMergeDuplicatedValidatorLicenseIntegrationTest.scala
@@ -3,6 +3,10 @@ package org.lfdecentralizedtrust.splice.integration.tests
 import com.daml.ledger.javaapi.data.Identifier
 import com.digitalasset.canton.logging.SuppressionRule
 import org.lfdecentralizedtrust.splice.codegen.java.splice.validatorlicense.ValidatorLicense
+import org.lfdecentralizedtrust.splice.config.ConfigTransforms.{
+  ConfigurableApp,
+  updateAutomationConfig,
+}
 import org.lfdecentralizedtrust.splice.integration.EnvironmentDefinition
 import org.lfdecentralizedtrust.splice.sv.automation.delegatebased.MergeValidatorLicenseContractsTrigger
 import org.lfdecentralizedtrust.splice.util.TriggerTestUtil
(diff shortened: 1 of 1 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvMergeDuplicatedValidatorLicenseIntegrationTest.scala
```diff
@@ -46,43 +55,37 @@ class SvMergeDuplicatedValidatorLicenseIntegrationTest
...
-        forAll(_)(
-          _.warningMessage should include(
-            "has 2 Validator License contracts."
-          )
-        ),
-      )
-    }
+        }
+        // Pause to make sure we don't get more log messages.
+        pauseAllDsoDelegateTriggers[MergeValidatorLicenseContractsTrigger]
```
moritzkiefer-da, line 82: if pause is not synchronous (I haven't checked, definitely could be) don't you have the same problem where some logs might come in after this?

moritzkiefer-da: Hm or maybe not because once it merges it is happy and doesn't log anymore. But then I don't understand why you need to pause

ray-roestenburg-da: The pause is async, and in this case since it is based on SourceBasedTrigger, it just returns with a Future.successful. My assumption is that this pause happens too late, or that the trigger already asynchronously started pulling from the source, and has already gotten past the mapAsync that waits in the promise, which causes it to do the work anyway. I tried yesterday to change the trigger such that it returns when it is really paused, but that did not work out for most tests, and I realized later that this does not do much if the trigger basically already started earlier, which is why I changed the test now to start with a paused trigger (which runs the trigger with paused = true), so that it can never do any work. We should probably tell everyone to be careful with pausing triggers based on SourceBasedTrigger like this.

ray-roestenburg-da: And yes, I think you can theoretically get more logs, but as you say, once it merges it's happy. It's all not very ideal. I had a version of SourceBasedTrigger that buffers tasks if paused, and completes a promise once it has seen some new data (returned as future by pause()), so you know for sure it has been paused (but asynchronously something could have still moved after you call pause), but too many tests don't work with this, so I abandoned that, but some feedback on things really getting quiet is probably a good idea.

PR #1764, opened 2025-07-31
Handle HttpCommandException in BftScanConnection 

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/1757

### DIFF
--- a/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
+++ b/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
@@ -29,6 +29,7 @@ import org.lfdecentralizedtrust.splice.environment.{
   RetryProvider,
   SpliceLedgerClient,
 }
+import org.lfdecentralizedtrust.splice.http.v0.definitions.ErrorResponse
 import org.lfdecentralizedtrust.splice.scan.admin.api.client.BftScanConnection.Bft
 import org.lfdecentralizedtrust.splice.scan.admin.api.client.commands.HttpScanAppClient.{
   DomainScans,
@@ -224,7 +225,7 @@ class BftScanConnectionTest
   val notFoundCommandFailure = HttpCommandException(
     HttpRequest(),
     StatusCodes.NotFound,
-    "Whatever thing was not found",
+    HttpCommandException.ErrorResponseBody(ErrorResponse("Whatever thing was not found")),
   )
   val partyIdA = PartyId.tryFromProtoPrimitive("whatever::a")
   val partyIdB = PartyId.tryFromProtoPrimitive("whatever::b")
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
@@ -1,17 +1,29 @@
 package org.lfdecentralizedtrust.splice.integration.tests
 
-import com.digitalasset.canton.console.CommandFailure
+import com.digitalasset.canton.logging.SuppressionRule
+import com.digitalasset.canton.{HasActorSystem, HasExecutionContext}
+import org.apache.pekko.http.scaladsl.Http
+import org.apache.pekko.http.scaladsl.model.headers.{Authorization, OAuth2BearerToken}
+import org.lfdecentralizedtrust.splice.codegen.java.splice.api.token.transferinstructionv1.TransferInstruction
+import org.lfdecentralizedtrust.splice.http.v0.wallet as http
+import org.lfdecentralizedtrust.splice.http.v0.wallet.AcceptTokenStandardTransferResponse
 import org.lfdecentralizedtrust.splice.integration.EnvironmentDefinition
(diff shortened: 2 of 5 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -730,6 +731,13 @@ object BftScanConnection {
                 Future.successful(BftScanConnection.ExceptionFailureResponse(failure))
```
OriolMunoz-da, line 731: so I was wrong and we were actually handling HTTP failures, just that the exception type was wrong. Or maybe it's right sometimes. I cannot tell because this is exceptions

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -730,6 +731,13 @@ object BftScanConnection {
...
             }
           )
+      case Failure(unexpected: HttpCommandException) =>
+        Future.successful(
+          BftScanConnection.HttpFailureResponse(
+            unexpected.status,
+            Json.obj("message" -> Json.fromString(unexpected.message)),
```
OriolMunoz-da, line 738: this likely seems like a bad idea, as we're assuming that messages:
- are deterministic
- will not change between versions

but OTOH if we want to give anything useful beyond the status code to the user, we need to include it

thoughts?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
```diff
@@ -251,14 +257,24 @@ class BftScanConnectionTest
...
     "forward the failure if the agreement is a failure" in {
       val connections = getMockedConnections(n = 4)
-      connections.foreach(makeMockReturn(_, partyIdA))
       val bft = getBft(connections)
 
       connections.foreach(makeMockFail(_, notFoundFailure))
 
       for {
         failure <- bft.getDsoPartyId().failed
-      } yield failure should be(failure)
```
OriolMunoz-da, line 261: clearly i wasn't inspired when I wrote this

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
moritzkiefer-da: It doesn't seem too bad to me. The responses are often quite useful. So I definitely wouldn't omit them. 
The messages being deterministic doesn't seem too bad.

For version issues given that we usually have all SVs only on two versions don't you always have f+1 on the same version so it should be consistent?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
moritzkiefer-da: blame the reviewer (it probably was me)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
moritzkiefer-da, line 738: Are we messing with the response schema here? It seems like we turn the original response body into a json object with a message field?  For cases where we do define a 404 (or other failure) in the openapi schema isn't that then inconsistent? Intuitively I would have hoped we can preserve the actual body 1:1 without any transformations.

moritzkiefer-da (approved):
Nice thanks!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
OriolMunoz-da: > For version issues given that we usually have all SVs only on two versions don't you always have f+1 on the same version so it should be consistent?

yeah, true

OriolMunoz-da: yep, this is the offending code:
https://github.com/hyperledger-labs/splice/blob/6986124ee72a5bc2709269787396c7693604f7d9/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/admin/api/client/commands/HttpCommand.scala#L114
I can change it to include the `ErrorResponse` instead of the message as part of the case class fields (I think)

moritzkiefer-da: sounds like a good change to make

moritzkiefer-da: Maybe also extend your test to:
1. Query a single scan
2. Query the BFT scan
3. Check that status code and response bodies are the same.

moritzkiefer-da (approved):
nice thank you!

PR #1809, opened 2025-08-05
cluster, helm, info: Add option imagePullPolicy

giner:
Makes splice-info more consistent with other Helm charts.


cluster/helm/splice-info/values.schema.json
```diff
@@ -43,6 +43,10 @@
       "type": "string",
       "description": "The Docker image to use for Nginx. Defaults to 'nginx:latest' if not present."
     },
+    "imagePullPolicy": {
+      "type": "string",
+      "enum": ["Always", "IfNotPresent", "Never"]
```
martinflorian-da, line 48: Nice improvement! I wonder if I can nudge someone :tm: to improve this everywhere...

PR #1812, opened 2025-08-05
[static] Fix cometbft ingress port

nicu-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/infra/src/istio.ts
+++ b/cluster/pulumi/infra/src/istio.ts
@@ -224,13 +224,18 @@ function configureCometBFTGatewayService(
   const numSVs = dsoSize < 5 && isDevNet ? 5 : dsoSize;
 
   const cometBftIngressPorts = Array.from({ length: numMigrations }, (_, i) => i).flatMap(
-    migration =>
-      Array.from({ length: numSVs }, (_, node) => node).map(node =>
+    migration => {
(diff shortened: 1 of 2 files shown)


cluster/pulumi/infra/src/istio.ts
```diff
@@ -224,13 +224,18 @@ function configureCometBFTGatewayService(
...
-      Array.from({ length: numSVs }, (_, node) => node).map(node =>
+    migration => {
+      const res = Array.from({ length: numSVs }, (_, node) => node).map(node =>
         ingressPort(
           `cometbft-${migration}-${node + 1}-gw`,
           cometBFTExternalPort(migration, node + 1)
         )
-      )
+      );
+      if (!isMainNet) {
```
moritzkiefer-da, line 234: Can we add a comment that this is for the SV runbook? It seems rather confusing to tie it to mainnet.

I'm also not sure we're consistent in excluding sv runbook stuff on mainnet (outside of not deploying it ofc) so I somewhat wondering whether just always adding it isn't the better option at this stage (I think the clean option would be to tie it to the deploy_sv_runbook flag but we don't set that consistently when deploying infra iirc)

nicu-da: Done.
Indeed we shouldn't have tied it that way but jsut replicated what we already had

PR #1829, opened 2025-08-05
Make it easy to disable SV participant BFT sequencer connections via `config.yaml`

martinflorian-da:
Part of https://github.com/DACH-NY/canton-network-internal/issues/1169

~Doesn't support SV runbook; I wanted to make this work then got angry then realized we don't need it so timeboxing.~

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #1839, opened 2025-08-06
0.4.7 backport-in-spirit: Make it easy to disable SV participant BFT sequencer connections via `config.yaml` (#1829)

martinflorian-da:
backports #1829

[static]

### DIFF
--- a/cluster/deployment/mock/config.yaml
+++ b/cluster/deployment/mock/config.yaml
@@ -110,6 +110,9 @@ loadTester:
   iterationsPerMinute: 30
   minRate: 0.123
 svs:
+  default:
+    participant:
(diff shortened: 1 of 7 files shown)


cluster/deployment/mock/config.yaml
```diff
@@ -110,6 +110,9 @@ loadTester:
   iterationsPerMinute: 30
   minRate: 0.123
 svs:
+  default:
```
martinflorian-da, line 113: changed this to be extra sure we get what we want

martinflorian-da: as backporting was slightly more involved here

moritzkiefer-da (approved):

PR #1849, opened 2025-08-06
move some jobs to gh-hosted runners

isegall-da:
We don't seem to be gaining anything from self-hosted runners for this, and sometimes need to scale up a whole node just for this tiny job. Seems wasteful when GH runners are free and unlimited.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/.github/workflows/assign_fork_prs.yml
+++ b/.github/workflows/assign_fork_prs.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   assign:
-    runs-on: self-hosted-docker-tiny
+    runs-on: ubuntu-24.04
     if: github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
     steps:
       - name: Assign PR from fork
--- a/.github/workflows/cluster_tests.yml
+++ b/.github/workflows/cluster_tests.yml
@@ -17,7 +17,7 @@ on:
 
 jobs:
   trigger_cluster_test:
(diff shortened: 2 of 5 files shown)


nicu-da (approved):
> GH runners are free and unlimited.

we should just try switching the integration tests to them :trollface:

.github/workflows/assign_fork_prs.yml
```diff
@@ -6,7 +6,7 @@ on:
 
 jobs:
   assign:
-    runs-on: self-hosted-docker-tiny
+    runs-on: ubuntu-24.04
```
moritzkiefer-da, line 9: can we move the final result stuff? That one seems quite slow often and there seems to be no reason why it has to be on our runners. I suspect the hosted ones probably work better there.

isegall-da: Yes, I was thinking of that too. For now, I focused on stuff outside of the build workflow, but I expect that the final result indeed can and should be moved. Will do that separately.

isegall-da:
> > GH runners are free and unlimited.
> 
> we should just try switching the integration tests to them :trollface:

The serious answer is: they are fairly small (4 CPUs, 16 GB IIRC), caching is a mess if you try combining self-hosted and gh-hosted, and I suspect that there is a limit on number parallel runs (but haven't looked into that yet). But having said that, I definitely plan on looking into the option nevertheless.

PR #1852, opened 2025-08-06
[ci] Do not group expired reward coupons by party

nicu-da:
fixes https://github.com/hyperledger-labs/splice/issues/1841 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpireRewardCouponsTrigger.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpireRewardCouponsTrigger.scala
@@ -21,6 +21,7 @@ import com.digitalasset.canton.tracing.TraceContext
 import com.digitalasset.canton.util.ShowUtil.*
 import io.opentelemetry.api.trace.Tracer
 import org.apache.pekko.stream.Materializer
+import org.lfdecentralizedtrust.splice.store.{Limit, PageLimit}
 
 import java.util.Optional
 import scala.concurrent.{ExecutionContext, Future}
@@ -43,15 +44,15 @@ class ExpireRewardCouponsTrigger(
       tc: TraceContext
   ): Future[Seq[ExpiredRewardCouponsBatch]] = for {
     dsoRules <- store.getDsoRules()
-    numSvs = dsoRules.payload.svs.size()
     batches <- store
-      .getExpiredRewards(
+      .getExpiredCouponsInBatchesPerRoundAndCouponType(
         dsoRules.domain,
         context.config.enableExpireValidatorFaucet,
+        PageLimit.tryCreate(svTaskContext.delegatelessAutomationExpiredRewardCouponBatchSize),
       )
-      .map(seq =>
-        Random.shuffle(seq).take(svTaskContext.delegatelessAutomationExpiredRewardCouponBatchSize)
-      )
+      // We select at most parallelism batches per round as  processing more than that would most likely just hit contention
+      // If any work was done the trigger will run again anyway so it's safer to just requery the stores
+      .map(seq => Random.shuffle(seq).take(context.config.parallelism))
   } yield batches
 
   override protected def isStaleTask(expiredRewardsTask: ExpiredRewardCouponsBatch)(implicit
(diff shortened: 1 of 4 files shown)


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpireRewardCouponsTrigger.scala
```diff
@@ -43,15 +44,15 @@ class ExpireRewardCouponsTrigger(
...
-    numSvs = dsoRules.payload.svs.size()
     batches <- store
-      .getExpiredRewards(
+      .getExpiredCouponsInBatchesPerRoundAndCouponType(
         dsoRules.domain,
         context.config.enableExpireValidatorFaucet,
+        PageLimit.tryCreate(svTaskContext.delegatelessAutomationExpiredRewardCouponBatchSize),
       )
-      .map(seq =>
-        Random.shuffle(seq).take(svTaskContext.delegatelessAutomationExpiredRewardCouponBatchSize)
```
nicu-da, line 53: Not sure my thinking is wrong or this logic was wrong.

nicu-da:
@moritzkiefer-da Feel free to ignore the review request so far until we resolve https://github.com/hyperledger-labs/splice/issues/1841#issuecomment-3161368043

nicu-da:
@moritzkiefer-da Anything against proceed with this as a rough improv to actually handle batches and spin off a tech debt for https://github.com/hyperledger-labs/splice/issues/1841#issuecomment-3162625300 to reduce contention and preserve round ordering?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpireRewardCouponsTrigger.scala
moritzkiefer-da: duh yeah looks right

moritzkiefer-da:
> Anything against proceed with this as a rough improv to actually handle batches and spin off a tech debt for https://github.com/hyperledger-labs/splice/issues/1841#issuecomment-3162625300 to reduce contention and preserve round ordering?

How do we ensure that this does not make contention worse and is actually counter productive? The current logic doesn't batch much but with random selection and individual elements selected you're also reasonably unlikely to collide.

If you now switch to larger batches don't you risk that those batches overlap across SVs and you produce more contention?

nicu-da:
> > Anything against proceed with this as a rough improv to actually handle batches and spin off a tech debt for [#1841 (comment)](https://github.com/hyperledger-labs/splice/issues/1841#issuecomment-3162625300) to reduce contention and preserve round ordering?
> 
> How do we ensure that this does not make contention worse and is actually counter productive? The current logic doesn't batch much but with random selection and individual elements selected you're also reasonably unlikely to collide.
> 
> If you now switch to larger batches don't you risk that those batches overlap across SVs and you produce more contention?

It's batched per round and per type of coupon, with a limit of 100 coupons by default. 
Even if you do have contention I don't see how submitting each coupon individually helps compared to batching and just accepting some contention. In most scenarios you should end up with just a few submissions needed anyway, even with high contention I don't see how it can be worse than the current format. 

Also we have the random delay that should help with that.

moritzkiefer-da:
I think the danger is that you repeatedly retry on contention causing more load than you would otherwise no?

moritzkiefer-da:
> Good point, I do wonder if our retry strategy makes sense here, any reason why we don't stop the processing of a task if we get a contention error back?

You don't know if the thing you're contending with will succeed. It could still fail in the end.

> Further more contention would be rejected by the local participant right? So it's not that bad.

Is that always the case? When the transaction got committed definitely but what if it is in-flight? Can't you end up with it getting rejected in phase 4?

moritzkiefer-da:
I think you're probably right though that batching is fine  so let's try it :crossed_fingers:

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/SvDsoStore.scala
```diff
@@ -348,15 +348,19 @@ trait SvDsoStore
     ]]
   ]
 
-  final def getExpiredRewards(
+  /** Returns at most expired coupon batches per round and coupon type.
```
moritzkiefer-da, line 351:
```suggestion
  /** Returns at totalCouponsLimit most expired coupon batches per round and coupon type.
```
probably? otherwise I don't understand the sentence. The name sounds off though then as it's clearly not the total but the per round and coupon type max so maybe rename to batch size or something.

moritzkiefer-da: also is there any reason why we split it by coupon type? That seems a bit pointless

moritzkiefer-da (approved):
Nice thanks!

PR #1880, opened 2025-08-11
Randomize submission of vetting transactions to avoid load spikes

### LINKED ISSUE #1863: Add random delay to new vetting transactions
It turns out submitting 500 topology transactions at the same time is a questionable idea.

Should only apply to the future dated transactions not the initial ones when first onboarding a node.

moritzkiefer-da:
[ci]

I can't come up with a great integration test for that. I did test it on a scratchnet and it seemed to behave as expected.

Plan for ciupgrade is to set the wait to 1min so the behavior is at least observable but doesn't have any meaningful slowdown.

fixes #1863

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/resources/include/validators/_validator.conf
+++ b/apps/app/src/test/resources/include/validators/_validator.conf
@@ -25,4 +25,7 @@
   }
   # Ensure we get frequent acks to handle our aggressive pruning intervals in tests
   time-tracker-min-observation-duration = 10s
+  max-vetting-delay = 0s
+  # Don't want to sleep in tests
+  max-vetting-delay = 0s
 }
--- a/apps/app/src/test/resources/include/validators/_validator.conf
+++ b/apps/app/src/test/resources/include/validators/_validator.conf
@@ -25,7 +25,6 @@
   }
   # Ensure we get frequent acks to handle our aggressive pruning intervals in tests
   time-tracker-min-observation-duration = 10s
-  max-vetting-delay = 0s
   # Don't want to sleep in tests
   max-vetting-delay = 0s
 }
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
@@ -737,6 +737,7 @@ abstract class TopologyAdminConnection(
                         logger.info(
                           s"Waiting until $minSubmissionTime before submitting topology transaction"
                         )
+                        // This is a noop if minSubmissionTime is in the past.
                         clock.scheduleAt(_ => (), minSubmissionTime).onShutdown(())
                       }
                     sleep.flatMap { _ =>
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
@@ -41,7 +41,7 @@ import com.digitalasset.canton.logging.NamedLoggerFactory
 import com.digitalasset.canton.logging.pretty.PrettyUtil.*
 import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
 import com.digitalasset.canton.protocol.DynamicSynchronizerParameters
-import com.digitalasset.canton.time.FetchTimeResponse
+import com.digitalasset.canton.time.{Clock, FetchTimeResponse}
 import com.digitalasset.canton.topology.*
 import com.digitalasset.canton.topology.admin.grpc
 import com.digitalasset.canton.topology.admin.grpc.{
@@ -704,6 +704,7 @@ abstract class TopologyAdminConnection(
         RecreateOnAuthorizedStateChange.Recreate,
       forceChanges: ForceFlags = ForceFlags.none,
       waitForAuthorization: Boolean = true,
+      maxSubmissionDelay: Option[(Clock, NonNegativeFiniteDuration)] = None,
   )(implicit traceContext: TraceContext): Future[TopologyResult[M]] = {
     withSpan("establish_topology_mapping") { implicit traceContext => _ =>
       logger.info(s"Ensuring that $description")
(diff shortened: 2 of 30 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
```diff
@@ -723,17 +724,37 @@ abstract class TopologyAdminConnection(
...
-                      signed.mapping
+                    val sleep: Future[Unit] =
+                      maxSubmissionDelay.fold(Future.unit) { case (clock, maxSubmissionDelay) =>
+                        val delay: NonNegativeFiniteDuration =
+                          NonNegativeFiniteDuration.tryFromJavaDuration(
+                            java.time.Duration
+                              .ofNanos((math.random() * maxSubmissionDelay.unwrap.toNanos).toLong)
+                          )
+                        logger.info(s"Waiting for $delay before submitting topology transaction")
+                        retryProvider
```
moritzkiefer-da: This feels like it's a bit too deep inside the logic. The reason for that is that I'd like to only add the wait when we actually submit a transaction and not on each trigger run. The outer logic really goes down that far in each trigger run and relies on this being idempotent instead of checking somewhere higher up.

apps/app/src/test/resources/include/validators/_validator.conf
```diff
@@ -25,4 +25,7 @@
   }
   # Ensure we get frequent acks to handle our aggressive pruning intervals in tests
   time-tracker-min-observation-duration = 10s
+  max-vetting-delay = 0s
```
martinflorian-da: dup

cluster/deployment/scratchneta/.envrc.vars
```diff
@@ -39,3 +39,5 @@ export APPROVE_SV_RUNBOOK=true
 # TODO(DACH-NY/canton-network-node#14679): Remove
 export VALIDATOR_RUNBOOK_POSTGRES_PVC_SIZE="240Gi"
 export CACHE_DEV_DOCKER_REGISTRY=$DEV_DOCKER_REGISTRY
+
```
martinflorian-da: happens to the best of us but maybe rm again?

cluster/pulumi/common/src/config/networkWideConfig.ts
```diff
@@ -0,0 +1,17 @@
...
+  networkWide: z
+    .object({
+      maxVettingDelay: z.string().optional(),
+    })
+    .optional(),
+});
+
+export type NetworkWideConfig = z.infer<typeof NetworkWideConfigSchema>;
+
+export const networkWideConfig = NetworkWideConfigSchema.parse(clusterYamlConfig).networkWide;
```
martinflorian-da, line 17: Idea for a future refactoring: we could go with a hierarchy such as:

```
spliceNodes
- default
- validators
  - default
  - validator1
  - ...
- svs
  - default
  - sv-1
  - ...
```

Happy to create the issue if it looks realistic to you that we'll get to that in the next year... seems like some refactoring of our configs is painfully needed so might just do it nicely.

cluster/pulumi/common/src/config/networkWideConfig.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { clusterYamlConfig } from 'splice-pulumi-common/src/config/configLoader';
+import { z } from 'zod';
+
+// Network wide configuration expected to be shared by all nodes.
+export const NetworkWideConfigSchema = z.object({
+  networkWide: z
+    .object({
+      maxVettingDelay: z.string().optional(),
```
martinflorian-da, line 10: How about adding this to the mock `config.yaml` so we have more confidence the pulumi code path works going forward?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -327,6 +327,9 @@ case class SvAppBackendConfig(
     // 1. try to reset one of your sequencers or mediators
     // 2. change sequencer URLs that need to get published externally.
     skipSynchronizerInitialization: Boolean = false,
+    // The maximum delay before submitting a package vetting change. The actual delay will be chosen randomly to ensure that
```
martinflorian-da:
```suggestion
    // The maximum delay before submitting a package vetting change. The actual delay will be chosen randomly (uniformly distributed between 0 and the maximum delay) to ensure that
```

I wondered this when I read this. If you accept please also add to the validator config (and you probably want to fix newlines).

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
```diff
@@ -837,6 +837,7 @@ class ValidatorApp(
         config.sequencerRequestAmplificationPatience,
         config.contactPoint,
         initialSynchronizerTime,
+        config.maxVettingDelay,
```
martinflorian-da, line 840: Hm so what happens if we were offline for a while; the new DARs are active and we now still have to wait for up to an hour before we vet them.

I guess it's fair to say "well set this to 0 in the config then" - but then we might want to document this?

Maybe it would be easier if instead we could instead fix the logic here to find the time until the next DAR that we haven't yet vetted will be needed, and use the min of that time (0 if negative) and whatever we configured?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
martinflorian-da: Hm does this roll a new sleep duration then on each retry? So that we then need to wait for another "up to an hour"?

You might want to somehow ensure that we sleep only only once :thinking: (If you're already doing that I'm missing the elite Scala nuance here, sorry.)

That or (perhaps better, see also my other comment): use `latestVettingTime` as an internal argument everywhere, and recalculate the `maxSubmissionDelay` based on that whenever we need it.

moritzkiefer-da: > Hm does this roll a new sleep duration then on each retry? So that we then need to wait for another "up to an hour"?

It does. But that seems fine to me. The authorize call does not wait for the transaction to be sequenced. So failures are quite unlikely here.

moritzkiefer-da: let me see how hard it is to change it I guess anyway.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
moritzkiefer-da: fair let me change it

cluster/pulumi/common/src/config/networkWideConfig.ts
moritzkiefer-da: good idea will do

cluster/pulumi/common/src/config/networkWideConfig.ts
moritzkiefer-da: Yeah I think some refactoring makes sense. Not quite sure if that structure fully works but maybe worth a try.

martinflorian-da: I think I cursed about the current config structure frequently enough that I might at least create an issue for us to improve things at some point...

martinflorian-da: https://github.com/hyperledger-labs/splice/issues/1892 (no special prioritization so far)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/PackageVetting.scala
```diff
@@ -77,9 +96,15 @@ class PackageVetting(
...
     val vettingTimeSortedDars = vettingSchedule.toSeq.sortBy(_._1)
     logger.info(s"Vetting for schedule $vettingTimeSortedDars from amulet rules $schedule")
+
     MonadUtil
       .sequentialTraverse(vettingTimeSortedDars) { case (validFrom, packages) =>
-        vetPackages(domainId, packages.toSeq, Some(validFrom))
+        vetPackages(
+          domainId,
+          packages.toSeq,
+          Some(validFrom),
```
martinflorian-da, line 105: uf I missed that we're passing this anyway, so we could also have hidden the adjustment a bit deeper; not a strong opinion at all though, happy as is too

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
```diff
@@ -723,17 +732,26 @@ abstract class TopologyAdminConnection(
...
-                      logger.info(
-                        s"Submitted proposal ${signed.mapping} for $description, waiting until the proposal gets accepted"
-                      )
-                      signed.mapping
+                    val sleep: Future[Unit] =
+                      minSubmissionTimeO.fold(Future.unit) { case (clock, minSubmissionTime) =>
+                        logger.info(
+                          s"Waiting until $minSubmissionTime before submitting topology transaction"
+                        )
+                        clock.scheduleAt(_ => (), minSubmissionTime).onShutdown(())
```
martinflorian-da, line 741: nit: Maybe add a comment that this is a NOP if minsubmissiontime is in the past (I had ocd again and checked the implementation)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/PackageVetting.scala
moritzkiefer-da: >  so we could also have hidden the adjustment a bit deeper

We still can't do it in the place where we actually wait as that one is agnostic to the topology transaction. So this seemed better over adding a third place that now needs to do more than just pass delays along.

martinflorian-da: :+1: sounds good

PR #1884, opened 2025-08-11
Remove useless isHealthy checks

rautenrieth-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/5138

See https://github.com/DACH-NY/cn-test-failures/issues/5138#issuecomment-3174077531 for context


OriolMunoz-da (approved):
thanks!

PR #1886, opened 2025-08-11
Fix triggers possibly being unhealthy inbetween tasks

OriolMunoz-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/5138

See https://github.com/DACH-NY/cn-test-failures/issues/5138#issuecomment-3129233815 for explanation

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PollingTrigger.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PollingTrigger.scala
@@ -136,6 +136,8 @@ trait PollingTrigger extends Trigger with FlagCloseableAsync {
         ): Unit = LoggerUtil.logOnThrow {
 
           def loopWithDelay(newState: PollingTriggerState) = {
+            // Like below,
+            // we use updateAndGet even if it can be reapplied during contention because we know there's no contention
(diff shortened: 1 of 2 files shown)


rautenrieth-da (approved):
Thanks! The new `futureCompletionGracePeriod ` is essentially a limit on the overhead in `PollingTrigger` itself between trigger invocations. 

The default limit of 1 second should never be exceeded it in a healthy application. We might exceed the limit when the application is overloaded/deadlocked though, causing kubernetes to automatically restart them in that case? Not sure if this would happen in practice but still good to keep in mind.

Unrelated to this PR, the section

```scala
pollingLoopRef.updateAndGet(_.map(_.flatMap { state =>
              context.retryProvider
                .scheduleAfterUnlessShutdown(
                  Future.successful(pollingLoop(newState)),
                  context.pollingClock,
                  pollingInterval,
                  pollingJitter,
                )
                .onShutdown(())
                .map(_ => state)
            }))
```
is suspicious, the function passed to `updateAndGet` should be side effect free. It should not cause any actual issues because operations on `polllingLoopRef` are (AFAICT) never executed concurrently. I think we should move the comment
```scala
// We use updateAndGet even if it can be reapplied during contention because we know there's no contention
```
that appears futher down in the file to this place.

PR #1890, opened 2025-08-12
cluster: Build cross-platform Docker images natively

giner:
This change removes hacks for building cross-platform images to run native builds instead. This is cleaner and more reliable.


cluster/images/canton/Dockerfile
```diff
@@ -32,23 +30,8 @@ COPY target/LICENSE.txt .
...
-   && rm -rf /var/lib/apt/lists/*
-
-COPY --from=build /usr/bin/tini /usr/bin/tini
-COPY --from=build /usr/bin/screen /usr/bin/screen
-COPY --from=grpcurl /bin/grpcurl /bin/grpcurl
-COPY --from=build /app/ /app/
-
-WORKDIR /app
-
+# TODO: Explain purpose of LABEL in a comment or remove it if not needed
```
moritzkiefer-da: TODOs without an issue reference fail our CI. So you either need to remove the TODO or create an issue that you associate it with, I would suggest the former here. To answer the question in the TODO: These are standardized labels from https://github.com/opencontainers/image-spec/blob/main/annotations.md that are required by some image scanning tools.

moritzkiefer-da (commented):
Thanks! Looks reasonable to me. You'll need to fix the todo comment

cluster/images/canton/Dockerfile
giner: noted, thank you, I'll update it

giner: Open Containers annotations docs are pretty confusing and are probably not even correct
> If the image.**base**.name annotation is specified, the image.**base**.digest annotation SHOULD be the digest of the manifest referenced by the image.**ref**.name annotation.

moritzkiefer-da (approved):
passed tests, thanks!

cluster/images/canton/Dockerfile
```diff
@@ -1,22 +1,25 @@
...
-# xx provides tools to support easy cross-compilation from Dockerfiles, see https://github.com/tonistiigi/xx
-FROM --platform=$BUILDPLATFORM tonistiigi/xx@sha256:923441d7c25f1e2eb5789f82d987693c47b8ed987c4ab3b075d6ed2b5d6779a3 AS xx
-ARG BUILDPLATFORM
-ARG TARGETPLATFORM
+ARG BASE_IMAGE="eclipse-temurin:21-jdk-noble@sha256:5a65f334da5a91a66076735d78e3ae30483a2593ac108f830dcd59521f2535cd"
 
-FROM --platform=$BUILDPLATFORM debian:bookworm-slim@sha256:6ac2c08566499cc2415926653cf2ed7c3aedac445675a013cc09469c9e118fdd AS build
+FROM fullstorydev/grpcurl:v1.9.2@sha256:0c3796edc8b6bb1814da2f11a554996237c0dd98380fce182efb1dc44c556ac1 AS grpcurl
+
+FROM ${BASE_IMAGE}
```
moritzkiefer-da, line 8: This currently fails CI due to
```
cluster/images/canton/Dockerfile:8:FROM ${BASE_IMAGE}
ERROR: docker images must pin base images to a specific digest
```

Do we need the indirection through the arg or can we just inline it to make CI happy?

giner: > Do we need the indirection through the arg or can we just inline it to make CI happy?

In this case we would have to specify and maintain the same image/hash twice, one in FROM and one in LABEL. I've updated the check script to make it parse Dockerfile instead.

scripts/check-base-images.sh
```diff
@@ -7,9 +7,33 @@ set -euo pipefail
 
 source "${TOOLS_LIB}/libcli.source"
 
-# We exclude "base_version", which we use for base images that are built from this repo in the same version
+python_cli=(/usr/bin/env python3)
 
-if find cluster/images/ -name Dockerfile -exec grep -Hn "FROM" {} + | grep -v sha | grep -v base_version; then
-  _error "docker images must pin base images to a specific digest"
-fi
+# Use virtual environment for Python dependencies
```
moritzkiefer-da: Please don't, the dependencies should come from nix.

moritzkiefer-da: You should be able to add it in here https://github.com/hyperledger-labs/splice/blob/main/nix/shell.nix#L61

moritzkiefer-da (changes requested):

scripts/check-base-images.sh
giner: I see. I'll do but I won't be able to test locally, so I'll rely on CI.

giner: dockerfile-parse is there, so I'll just drop it

giner: done

giner:
rebased

PR #1895, opened 2025-08-12
Retry accept/reject transfer offer in wallet UI

OriolMunoz-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/5375

I chose 

> If that's too expensive, then in the test we should either retry the accept button

because it doesn't seem worth the effort to setup everything required for something that just manifests on (sometimes too fast) tests...
But shout if you disagree.

Also used the opportunity to use useMutation.

### DIFF
--- a/apps/wallet/frontend/src/components/TransferOffers.tsx
+++ b/apps/wallet/frontend/src/components/TransferOffers.tsx
@@ -140,6 +140,12 @@ interface TransferOfferProps {
   transferOffer: WalletTransferOffer;
 }
 
+// An accept or a reject can fail if for example the transfer offer is visible to the validator,
+// but not to f+1 Scans (which are required to get the accept/reject context from).
+// Or overall other transient errors.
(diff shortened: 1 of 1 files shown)


apps/wallet/frontend/src/components/TransferOffers.tsx
```diff
@@ -147,8 +155,29 @@ export const TransferOfferDisplay: React.FC<TransferOfferProps> = props => {
     acceptTokenStandardTransfer,
     rejectTokenStandardTransfer,
   } = useWalletClient();
-  const accept = offer.isTokenStandard ? acceptTokenStandardTransfer : acceptTransferOffer;
-  const reject = offer.isTokenStandard ? rejectTokenStandardTransfer : rejectTransferOffer;
+
+  const acceptMutation = useMutation({
```
fayi-da, line 159: Yaay mutation introduced 🕺

PR #1909, opened 2025-08-13
Clean up public ingress things and allow making SV info endpoint public

martinflorian-da:
[static]

Fixes https://github.com/DACH-NY/canton-network-internal/issues/933
Fixes https://github.com/DACH-NY/canton-network-internal/issues/360

The removed code paths were not used in a while and at least some of them look broken to me.

The approach of defining an extra gateway for public things is also quite complex; it seems like everything we might need can be achieved with istio by the same way we allow the info endpoint here: by adding `AuthorizationPolicy` CRs.

Tested on a scratch that I can reach the target host without VPN. (Other hosts still got denied as expected.)

### DIFF
--- a/build-tools/cncluster
+++ b/build-tools/cncluster
@@ -1999,7 +1999,7 @@ function _create_grafana_silence() {
     _info "Adding silence with the following: $startsAt, $endsAt, $createdBy, $comment"
 
     curl -sSL --fail-with-body \
-        "https://public.${GCP_CLUSTER_HOSTNAME}/grafana/api/alertmanager/grafana/api/v2/silences" \
+        "https://grafana.${GCP_CLUSTER_HOSTNAME}/api/alertmanager/grafana/api/v2/silences" \
         -H 'content-type: application/json' \
         --request POST \
         -H "x-non-jwt-auth: Bearer ${token}" \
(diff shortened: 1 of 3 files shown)


cluster/deployment/config.yaml
```diff
@@ -20,7 +20,7 @@ pulumiProjectConfig:
     # this for example lets you create the cloud sql instances without having deployments using them, and restoring them from other sources
     installDataOnly: false
     isExternalCluster: false
-    hasPublicDocs: false
+    hasPublicInfo: false
```
moritzkiefer-da, line 23: if we're willing to make it public on mainnet, any reason not to just always make it public?

cluster/pulumi/common/src/config/configSchema.ts
```diff
@@ -9,7 +9,7 @@ import { defaultActiveMigration, SynchronizerMigrationSchema } from './migration
 const PulumiProjectConfigSchema = z.object({
   installDataOnly: z.boolean(),
   isExternalCluster: z.boolean(),
-  hasPublicDocs: z.boolean(),
```
moritzkiefer-da, line 12: so was this just broken? It has been quite useful for 3.3 previews and I could see it being useful again in the future.

cluster/deployment/config.yaml
martinflorian-da: seemed odd on scratchnets and cilr; don't think we want anything public there

cluster/pulumi/common/src/config/configSchema.ts
martinflorian-da: I didn't test it but the authorization policy looked to me as if it couldn't work like that...

Re-introducing this is adding that config flag back in and adding a new `AuthorizationPolicy` object. Seems easy enough to do once we need it?

moritzkiefer-da: fair enough

martinflorian-da: A meta motivation here is that the end goal is that nearly everything is public; so doing all the stuff we did before this PR, with an extra IP address, extra URLs etc, seems just wrong in that context.

PR #1910, opened 2025-08-13
Update nix and add functions-framework dependency

rautenrieth-da:
`functions-framework` is used for [serverless](https://github.com/DACH-NY/canton-network-internal/blob/main/cluster/serverless/README.md?plain=1#L9)


rautenrieth-da:
Closing: upgrading nixkpgs requires an upgrade of nix, and we had issues with nix upgrades recently

PR #1916, opened 2025-08-14
Support a custom fluent bit instance to reduce logging overhead

moritzkiefer-da:
[static]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/cluster/src/fluentBit.ts
+++ b/cluster/pulumi/cluster/src/fluentBit.ts
@@ -1,105 +1,113 @@
 // Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 import * as k8s from '@pulumi/kubernetes';
-import { appsAffinityAndTolerations, CLUSTER_NAME, exactNamespace, GCP_REGION, infraAffinityAndTolerations } from 'splice-pulumi-common';
+import {
+  appsAffinityAndTolerations,
+  CLUSTER_NAME,
+  exactNamespace,
+  GCP_REGION,
+  infraAffinityAndTolerations,
+} from 'splice-pulumi-common';
 
-
-export function installFluentBit() {
+export function installFluentBit(): void {
   // GCP does not allow configuring the fluent-bit instance it deploys. So we disable GCP's fluent-bit instance for workloads and install our own that
   const fluentBit = exactNamespace('fluent-bit');
 
   const values = {
-    tolerations: infraAffinityAndTolerations.tolerations.concat(appsAffinityAndTolerations.tolerations),
-    config: {inputs: [
-      // Input config is roughly copied from the default GCP config available from `kubectl get configmap -n kube-system fluentbit-gke-config-v1.4.0 -o yaml`
-      "[Input]",
-      "    Name              tail",
-      "    Path              /var/log/containers/*.log",
-      "    Exclude_Path     /var/log/containers/*_kube-system_*.log,/var/log/containers/*_istio-system_*.log,/var/log/containers/*_knative-serving_*.log,/var/log/containers/*_gke-system_*.log,/var/log/containers/*_config-management-system_*.log,/var/log/containers/*_gmp-system_*.log,/var/log/containers/*_gke-managed-cim_*.log,/var/log/containers/*_gke-managed-volumepopulator_*.log,/var/log/containers/*_gke-managed-checkpointing_*.log,/var/log/containers/*_gke-managed-lustrecsi_*.log",
-      "    DB                /var/log/flb_kube.db",
-      "    Mem_Buf_Limit     100MB",
-      "    Skip_Long_Lines   On",
-      "    Refresh_Interval  1",
-      "    Buffer_Max_Size   20MB",
-      "    Read_from_Head True",
-      "    Tag kube.*"
-    ].join("\n"),
-    outputs: [
-      // Stack driver
-      "[Output]",
-      "    Name stackdriver",
-      "    Match kube.*",
-      // This ensures that non-JSON fields get parsed as textPayload
-      "    text_payload_key message",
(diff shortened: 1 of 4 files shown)


moritzkiefer-da:
Log example from a scratchnet run

image

And a non-json payload message

image

build-tools/cncluster
```diff
@@ -334,6 +334,14 @@ function _prompt_to_confirm() {
    USER_PROMPT_CONFIRMED=1
 }
 
+function _get_logging_components() {
```
moritzkiefer-da, line 337: This really should be pulumi. But we don't have any pulumi code that modifies our clusters so annoyingly the economic option is to continue doing what we did so far. Will double check if we have a tech debt already and if not create one.

cluster/deployment/scratchneta/.envrc.vars
```diff
@@ -39,3 +39,5 @@ export APPROVE_SV_RUNBOOK=true
 # TODO(DACH-NY/canton-network-node#14679): Remove
 export VALIDATOR_RUNBOOK_POSTGRES_PVC_SIZE="240Gi"
 export CACHE_DEV_DOCKER_REGISTRY=$DEV_DOCKER_REGISTRY
+
+export SELF_HOSTED_FLUENT_BIT=true
```
moritzkiefer-da, line 43: This is actually deliberate for once because I did apply the cluster config here already.

My overall plan is to gradually apply it on all internal clusters, leave it on CILR for at least a week and then probably propagate to dev/test/mainnet with another week of delay each.

cluster/pulumi/cluster/src/fluentBit.ts
```diff
@@ -0,0 +1,122 @@
...
+      "    Merge_Log_Key    log_processed",
+      "    Annotations      Off",
+      // Tweak the k8s labels to only keep the ones we care about
+      "[FILTER]",
+      "    Name    lua",
+      "    Match   kube.*",
+      "    script  /fluent-bit/scripts/k8s_labels.lua",
+      "    call    tweak_k8s_labels",
+    ].join("\n"),
+    customParsers: [
```
moritzkiefer-da: Random side note: I think having our own instance also allows us fixing the parsing of e.g. the gha runners where everything gets parsed as a warning even if they don't add json support. Whether we care enough to do so is another question.

cluster/pulumi/cluster/src/fluentBit.ts
```diff
@@ -0,0 +1,122 @@
+import * as k8s from '@pulumi/kubernetes';
+import { appsAffinityAndTolerations, CLUSTER_NAME, exactNamespace, GCP_REGION, infraAffinityAndTolerations } from 'splice-pulumi-common';
+
+
+export function installFluentBit() {
```
moritzkiefer-da: I considered whether this should go into infra instead. In the end, it seems like we are replacing the default cluster logging which should have been configured through cluster so this seems more appropriate.

build-tools/cncluster
martinflorian-da: > any pulumi code that modifies our clusters

Well we do, we just barely use it: https://github.com/hyperledger-labs/splice/tree/40fcc001b07a0e94222743304c2b14d478e53f8a/cluster/pulumi/cluster

> tech debt

Maybe something to attach to https://github.com/DACH-NY/canton-network-internal/issues/417 or create as a follow-up to it. (I'm OK for us to keep in cncluster for now.)

moritzkiefer-da: > Well we do, we just barely use it: https://github.com/hyperledger-labs/splice/tree/40fcc001b07a0e94222743304c2b14d478e53f8a/cluster/pulumi/cluster

That code does not modify the cluster itself. It updates things within the cluster. Updating the cluster itself requires importing it into pulumi which is a bit annoying but we should definitely do.

martinflorian-da: > importing it into pulumi

Yeah that is part of https://github.com/DACH-NY/canton-network-internal/issues/417

moritzkiefer-da: :+1:

cluster/pulumi/cluster/src/fluentBit.ts
```diff
@@ -0,0 +1,128 @@
...
+  // GCP does not allow configuring the fluent-bit instance it deploys. So we disable GCP's fluent-bit instance for workloads and install our own that
+  const fluentBit = exactNamespace('fluent-bit');
+
+  const values = {
+    tolerations: infraAffinityAndTolerations.tolerations.concat(appsAffinityAndTolerations.tolerations),
+    config: {inputs: [
+      // Input config is roughly copied from the default GCP config
```
martinflorian-da:
```suggestion
      // Input config is roughly copied from the default GCP config
```

Oh is that public? Could you add a link in case it changes and/or we need to understand what is going on here some day?

cluster/pulumi/cluster/src/fluentBit.ts
```diff
@@ -0,0 +1,128 @@
...
+      "    Merge_Log        Off",
+      "    Merge_Log_Key    log_processed",
+      "    Annotations      Off",
+      // Tweak the k8s labels to only keep the ones we care about
+      "[FILTER]",
+      "    Name    lua",
+      "    Match   kube.*",
+      "    script  /fluent-bit/scripts/k8s_labels.lua",
+      "    call    tweak_k8s_labels",
+    ].join("\n"),
```
martinflorian-da: I almost wonder if it wouldn't have been easier and nicer to create external files for the configs, so we can write them down normally. It seems like they are quite verbose and maybe we'll add more of those.

Happy to ignore for this PR.

moritzkiefer-da: hm fair, I'll keep it for now since I'm lazy but worth considering if we need to change it.

cluster/pulumi/cluster/src/fluentBit.ts
moritzkiefer-da: Added a pointer

PR #1872, opened 2025-08-07
Write how-to docs for token standard usage

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/1594

PR #1946, opened 2025-08-18
Reorganize token standard docs

OriolMunoz-da:
Implements https://github.com/hyperledger-labs/splice/pull/1872#discussion_r2279008650

There is (barely) any text change, it's all moved around

this is how the structure looks now:
image


docs/src/app_dev/token_standard/openapi/allocation_instruction.rst
OriolMunoz-da, line 1: having these in separate files (as opposed to a single big one as before) makes it easier to link to them (read: actually possible for me, because sphinx was driving me insane) - but also reduces clutter/overscrolling

PR #1955, opened 2025-08-19
Match package name on template filter

OriolMunoz-da:
Part of #829

This is the easy part. If no interfaces are involved, only ingestion has to change.
Interface-related changes to come later.

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
@@ -29,7 +29,6 @@ import org.lfdecentralizedtrust.splice.util.{
   Contract,
   ContractWithState,
   PackageQualifiedName,
-  QualifiedName,
   TemplateJsonDecoder,
 }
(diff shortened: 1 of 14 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
OriolMunoz-da, line 1: this is the main change, everything else is mostly test/refactor noise

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/QualifiedName.scala
```diff
@@ -18,13 +15,26 @@ final case class PackageQualifiedName(packageName: String, qualifiedName: Qualif
 }
 
 object PackageQualifiedName {
-  def apply(identifier: Identifier): PackageQualifiedName = {
-    val resource = DarResources
+  def findFromResources(identifier: Identifier): Option[PackageQualifiedName] = {
```
moritzkiefer-da: nit: maybe lookup instead of find? We seem to use that more often

PR #1962, opened 2025-08-19
Match package id on interface filter

OriolMunoz-da:
Part of #829

This already worked, but allows me to drop `interfaceFiltersWithoutPackageNames`.


moritzkiefer-da (approved):
Nice thanks!

PR #1976, opened 2025-08-20
Switch to relying on package names instead of module names

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/829

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStoreTest.scala
```diff
@@ -1601,23 +1556,94 @@ abstract class MultiDomainAcsStoreTest[
...
+        resultHolding.map(_.contractId.contractId) should contain theSameElementsAs Seq(
+          goodContract.contractId.contractId
+        )
+        resultAmulet.map(_.contractId.contractId) should contain theSameElementsAs Seq(
+          goodContract.contractId.contractId
+        )
+      }
+    }
+
+    "prevent against ingestion of same (moduleName, entityName) with different package name - via interface fallback" in {
```
OriolMunoz-da, line 1577: this test is the best summary for the reason of being of this PR

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V043__package_name_acs_stores.sql
```diff
@@ -0,0 +1,361 @@
+-- strictly not necessary, but lets us define the constraint as mandatory afterwards,
+-- and also makes index creation faster since there's no data to index
+truncate table acs_store_template cascade;
+truncate table user_wallet_acs_store cascade;
+truncate table external_party_wallet_acs_store cascade;
+truncate table validator_acs_store cascade;
+truncate table sv_acs_store cascade;
+truncate table dso_acs_store cascade;
+truncate table scan_acs_store cascade;
+truncate table splitwell_acs_store cascade;
```
OriolMunoz-da: double-check me that this is not a bad idea

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
```diff
@@ -175,22 +179,6 @@ trait MultiDomainAcsStore extends HasIngestionSink with AutoCloseable with Named
...
-    * contract ID and participant ID.
-    *
-    * The idea is that different apps making the same migration on different
-    * participants will split the work better, while preserving determinism of a
-    * specific running app for fault-tolerance.  For the former to happen, the
-    * position of a contract on one list must have no correlation with that on
-    * another list; that is why the contract ID by itself cannot be used by
-    * itself as the source of the sort key.
-    */
-  def listAssignedContractsNotOnDomainN(
```
OriolMunoz-da, line 184: unused except in a test (also deleted)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsQueries.scala
```diff
@@ -31,24 +31,33 @@ import io.grpc.Status
...
   /** @param tableName Must be SQL-safe, as it needs to be interpolated unsafely.
     *                  This is fine, as all calls to this method should use static string constants.
     */
-  protected def selectFromAcsTable(
+  protected def selectFromAcsTable[C, TCid <: ContractId[_], T](
       tableName: String,
       storeId: AcsStoreId,
       migrationId: Long,
-      where: SQLActionBuilder,
+      companion: C,
```
OriolMunoz-da, line 41: this should've been the case all along, instead of having `where ti_q_n = ?` on every query

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStoreTest.scala
```diff
@@ -837,51 +837,6 @@ abstract class MultiDomainAcsStoreTest[
       MultiDomainAcsStoreTest.generatedCoids.value.size should (be >= 900 and be <= 1000)
     }
 
-    "read assignment-mismatched contracts in a stable order" in {
```
OriolMunoz-da, line 840: tested `listAssignedContractsNotOnDomainN`, deleted

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V043__package_name_acs_stores.sql
OriolMunoz-da: note that all descriptors are bumped

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsQueries.scala
moritzkiefer-da: agreed, thanks for pulling through that simplification!

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V043__package_name_acs_stores.sql
```diff
@@ -0,0 +1,362 @@
+-- strictly not necessary, but lets us define the constraint as mandatory afterwards,
```
moritzkiefer-da: Do we know what performance we get for a full ACS backfill? I'm a bit nervous about this given that the ACS queries are known to not perform well without pruning and there is no pruning atm on SV participants.

How about on CILR you take one SV (maybe best to use the SV runbook given that it resets daily anyway so if it goes wrong we can just reset again) manually force  a reingestion by dropping the data from the table and last ingested offset?

If it does take too long, the easy short term option we have is to merge this into the 3.4 branch only. Afaict, there isn't an immediate need to have this be on 3.3.

rautenrieth-da: > How about on CILR you take one SV (maybe best to use the SV runbook given that it resets daily anyway so if it goes wrong we can just reset again) manually force a reingestion by dropping the data from the table and last ingested offset?

What is an "ACS backfill"? After truncating the ACS tables and bumping the store descriptor, the new ACS store will be initialized from the ACS endpoint of the ledger API, is that what you meant? Apart from the participant being slow to deliver the ACS, we are not batching SQL inserts when consuming the ACS
```scala
override def ingestAcs() = {
    ...
    // TODO (#989): batch inserts
}
```
and we load the whole ACS into one in-memory structure:
```scala
private def ingestAcsAndInFlight(offset: Long) = {
    ...
    // TODO(#863): stream contracts instead of ingesting them as a single Seq
```
so I agree that one test on CILR is a good idea.

moritzkiefer-da: > is that what you meant? 

yes

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V047__package_name_acs_stores.sql
```diff
@@ -0,0 +1,362 @@
+-- strictly not necessary, but lets us define the constraint as mandatory afterwards,
+-- and also makes index creation faster since there's no data to index
+truncate table store_last_ingested_offsets cascade;
+truncate table acs_store_template cascade;
```
rautenrieth-da, line 3: This table should be empty, but it doesn't hurt to truncate it.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V043__package_name_acs_stores.sql
```diff
@@ -0,0 +1,362 @@
...
+-- and also makes index creation faster since there's no data to index
+truncate table store_last_ingested_offsets cascade;
```
rautenrieth-da: I don't think you can truncate this table, it's also used by txlog stores, see https://github.com/hyperledger-labs/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala#L1080-L1086.

OriolMunoz-da: yep, you're right... I _technically_ don't have to truncate it since I'm bumping the ACS descriptor, though then the "last ingested offset" for those would be incorrect
Since anyway I need to figure out [how to safely* delete things on the SV runbook](https://github.com/hyperledger-labs/splice/pull/1976#discussion_r2299890457), I'll do the same here

*in the SV runbook, the descriptors don't change, so I can't leave last_ingested_offsets lingering after truncating

OriolMunoz-da: https://github.com/DACH-NY/canton-network-node/pull/18149

> They will always be identical in practice, except when one of the stores is reset by changing the descriptor.

yeah that makes it hard for me to tell what I'm supposed to delete from the last_ingested :cry:

OriolMunoz-da: actually nvm - I can just use the version numbers...

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/db/DbUserWalletStore.scala
```diff
@@ -83,9 +81,7 @@ class DbUserWalletStore(
         ),
       ),
       txLogStoreDescriptor = StoreDescriptor(
-        // Note that the V005__no_end_user_name_in_user_wallet_store.sql DB migration converts from version 1 descriptors
-        // to version 2 descriptors.
-        version = 2,
+        version = 3,
```
OriolMunoz-da: this is wrong, I shouldn't have touched the txlog descriptor

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V043__package_name_acs_stores.sql
OriolMunoz-da: so, given this:
image
and this:
```
txLogStoreDescriptor = StoreDescriptor(
        version = 1,
        name = "DbScanStore",
        party = key.dsoParty,
        participant = participantId,
        key = Map(
          "dsoParty" -> key.dsoParty.toProtoPrimitive
        ),
      )
```
```
txLogStoreDescriptor = StoreDescriptor(
        version = 2,
        name = "DbUserWalletStore",
        party = key.endUserParty,
        participant = participantId,
        key = Map(
          "endUserParty" -> key.endUserParty.toProtoPrimitive,
          "validatorParty" -> key.validatorParty.toProtoPrimitive,
          "dsoParty" -> key.dsoParty.toProtoPrimitive,
        ),
      ),
```

I believe I can:
```
delete
from store_last_ingested_offsets
where store_id in (select id
                   from store_descriptors
                   where descriptor ->> 'name' IN
                         -- these are ACS only, so we can delete everything
                         ('DbSplitwellStore', 'DbSvSvStore', 'DbValidatorStore', 'DbExternalPartyWalletStore')
                      -- these have txlog, preserve it. Version matches that of the txlog descriptor
                      OR (descriptor ->> 'name' = 'DbUserWalletStore' AND descriptor ->> 'version' != 2)
                      OR (descriptor ->> 'name' = 'DbScanStore' AND descriptor ->> 'version' != 1));
```
but I need an extra pair of eyes before i do anything stupid

rautenrieth-da: > I technically don't have to truncate it since I'm bumping the ACS descriptor, though then the "last ingested offset" for those would be incorrect

I don't understand this part, why can't we just leave the `store_last_ingested_offsets` table untouched? If you bump the store descriptor it should generate a new store id, and the existing row in `store_last_ingested_offsets` will simply become unused, no?

There is a special case where you bump the acs store descriptor such that it becomes identical with an already used txlog store descriptor. It looks like we don't handle this correctly 😱 , I have opened https://github.com/hyperledger-labs/splice/issues/2033.

If you're worried about leaving garbage, there is another table that references the acs store id (AFAICT it might not be empty, people can implement workflows that reassign non-dso contracts like transfer proposals):
```
incomplete_reassignments
```

Additionally, the following tables _also_ reference a store id, but they happen to only reference txlog stores
```
active_parties
round_totals
round_party_totals
txlog_backfilling_status
txlog_first_ingested_update
```

We have the general issue that bumping store descriptors leaves the old data. I opted for not cleaning up old data automatically, in order to have fallback data in case something goes wrong.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V047__package_name_acs_stores.sql
```diff
@@ -0,0 +1,362 @@
...
+
+-- obtained with (unfortunately not all indexes are named consistently, so some require manual intervention) and IJ formatting:
+-- select
+--     'DROP INDEX ' || (regexp_matches(indexdef, 'CREATE INDEX (.*) ON'))[1] || ';' || E'\n'
+--         'CREATE INDEX ' || replace((regexp_matches(indexdef, 'CREATE INDEX (.*) ON'))[1], 'tid', 'pn_tid') || ' ON ' || replace(replace(replace((regexp_matches(indexdef, 'CREATE INDEX (.*) ON (.*)'))[2], 'template_id_qualified_name', 'package_name, template_id_qualified_name'), 'public.', ''), 'USING btree', '') || ';' || E'\n'
+-- from pg_indexes
+-- where tablename like '%acs_store%'
+--   and indexdef like '%template_id_qualified_name%';
+DROP INDEX acs_store_template_sid_mid_tid_ce;
+CREATE INDEX acs_store_template_sid_mid_pn_tid_ce ON acs_store_template (store_id, migration_id, package_name,
```
rautenrieth-da, line 34: This is a lot of indexes being created, did you time it on CILR? If it's more than a minute we should probably mention it in the release notes. There's also [SqlIndexInitializationTrigger](https://github.com/hyperledger-labs/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala) for creating expensive indexes asynchronously, but having the apps start without fast ACS queries is probably asking for trouble.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsQueries.scala
```diff
@@ -326,12 +356,22 @@ object AcsQueries {
         companionClass: ContractCompanion[C, TCId, T],
         decoder: TemplateJsonDecoder,
     ): Contract[TCId, T] = {
+      // safety check: if the templates don't match,
```
rautenrieth-da:
```suggestion
      // safety check: if the packages don't match,
```
?

OriolMunoz-da: `// safety check: if the PackageQualifiedNames don't match,`

is most correct, will fix

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V043__package_name_acs_stores.sql
OriolMunoz-da: > I don't understand this part, why can't we just leave the store_last_ingested_offsets table untouched? If you bump the store descriptor it should generate a new store id, and the existing row in store_last_ingested_offsets will simply become unused, no?

my problem is with how to handle package_name in that case... either I set it to null/"INVALID" or I fill it up, but I can't do that from the existing data*

*I could fetch it from update_history_creates, but that's even less likely to perform properly

OriolMunoz-da: Then again, that invalid package_name won't be used after bumping the descriptor (old data, unused, whenWeGetToIt :tm: deleted), so maybe that's not so bad?

rautenrieth-da: > how to handle package_name in that case

That is not a column in `store_last_ingested_offsets`? What I meant is you can truncate the acs tables, but not the `store_last_ingested_offsets` table. Then you can do whatever you want to the schema of the acs tables, and will only end up with one garbage row in `store_descriptors` and a few garbage rows in `store_last_ingested_offsets` (one per migration).

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V047__package_name_acs_stores.sql
OriolMunoz-da: took less than a minute to create all indexes for a DSO store copy, which is the biggest one according to https://grafana.cilr.global.canton.network.digitalasset.com/d/dduss3xr5or28c/splice-store-acs-size?orgId=1&from=now-1h&to=now&timezone=UTC&var-namespace=sv-1&var-store_name=$__all
(total of 300k rows across all store ids)

a lot less, if I end up managing to truncate the tables

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V043__package_name_acs_stores.sql
OriolMunoz-da: > and will only end up with one garbage row in store_descriptors 

it's not just garbage, but invalid, as the offset is wrong and the ACS won't be reingested if we rolled back to it - it's not clear to me that that's fine

OriolMunoz-da:
timings for ACS reingestion:

a bit more than 2m for scan, [logs](https://console.cloud.google.com/logs/query;query=resource.labels.cluster_name%3D%22cn-cilrnet%22%0Aresource.labels.namespace_name%3D%22sv%22%0Aresource.labels.container_name%3D%22scan-app%22%0A%22JAVA_TOOL_OPTIONS%22%20%20OR%20jsonPayload.logger_name%3D%22o.l.s.s.d.DbMultiDomainAcsStore:scan%3Dscan-app%22%20OR%20%22UpdateIngestionService%22;pinnedLogId=2025-08-27T12:49:01.880Z%2F18ebbjnf7fzlxy;summaryFields=resource%252Flabels%252Fnamespace_name,jsonPayload%252Flogger_name:false:32:beginning;cursorTimestamp=2025-08-27T12:49:32.517Z;duration=PT15M?project=da-cn-ci-2&invt=Ab6mVg):
image

and SV app doesn't reingest, see https://github.com/hyperledger-labs/splice/pull/2041

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V043__package_name_acs_stores.sql
rautenrieth-da: Ah you're worried that the old data in the database would be corrupt, even if that data is not used? That's only a problem if we want to roll back the change to store descriptors relatively quickly, in order to use the old data before it becomes so outdated that it's useless? And since we're truncating the acs tables, there's no point doing a rollback, we'd have to restore from backup or do another store version bump, no?

OriolMunoz-da: copy pasting, this should be reasonable to do:
* truncate all ACS_store tables, don't touch any other table
* bump descriptors
* indexes will be quick to create cause I just truncated

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V043__package_name_acs_stores.sql
moritzkiefer-da: @OriolMunoz-da So given the very rapid increase in ACS size on mainnet I'm rather nervous about merging this atm. Can we hold off on this?

moritzkiefer-da: Or merge it into 3.4? There we will ingest the acs anyway so it's not making anything worse

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V047__package_name_acs_stores.sql
moritzkiefer-da: For the record: This now runs after the truncation so it is instant as pointed out in a comment

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/QualifiedName.scala
```diff
@@ -37,6 +37,15 @@ object PackageQualifiedName {
       QualifiedName(event.getTemplateId.getModuleName, event.getTemplateId.getEntityName),
     )
   }
+
+  def fromJavaCodegenCompanion(
```
moritzkiefer-da, line 41: wondering if we could limit visibility of `QualifiedName` to avoid accidental usages instead of `PackageQualifiedName`. Or maybe make it not comparable or something like that. Doesn't have to be in this PR.

PR #1978, opened 2025-08-20
Support running static tests on gh-hosted runners

isegall-da:
OK, so I convinced myself that this is actually safe as it is:
1. No access to secrets for the workflow that runs unreviewed because there's no `secrets: inherit` (and that definition is in `on: pull_request_target` so cannot be modified by the PR)
2. Unreviewed workflows do not run on self-hosted runners (and that definition is in `on: pull_request_target`, so ..)

Nevertheless, I added an extra safety mechanism where we don't checkout the full repo from the fork, only a subset of directories. It does not provide 100% protection from running scripts from the fork, but due to 1+2, that's acceptable.

Happy for a critical :eyes: on these claims.

test run: https://github.com/hyperledger-labs/splice/actions/runs/17100982593/job/48497214932?pr=1978

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


.github/workflows/test.yml
isegall-da, line 1: To be removed before merging, I'm letting the test run finish first

.github/workflows/build.static_tests.yml
```diff
@@ -4,22 +4,42 @@ on:
...
+        shell: bash
+        run: |
+          echo "Selectively checking out code under review"
+          git fetch origin ${{ inputs.commit_sha }} --depth=1
+          git checkout ${{ inputs.commit_sha }} -- \
+            apps/ \
+            cluster/ \
+            daml/ \
+            docs/ \
+            token-standard/
```
isegall-da: This is the part added for extra safety

martinflorian-da: not fully convinced yet that the extra complexity here is worth it (see slack)

.github/workflows/pr_non_contributors.yml
```diff
@@ -29,6 +23,27 @@ jobs:
...
+      actions: write # To cancel itself if not opted in
     needs: env_hold
     with:
       commit_sha: ${{ github.event.pull_request.head.sha }}
+
+  # Note: unapproved runs must not be granted access to secrets, and must not run on self-hosted runners
+  no_approval_static_tests:
+    name: Static Tests (No Approval Required)
+    uses: ./.github/workflows/build.static_tests.yml
+    if: github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
```
martinflorian-da, line 40: can we add something like `secrets: {}` to make it extra explicit that we don't want this to see any secrets?

isegall-da: (actionlint was unhappy with that initially, but I added an ignore to force it to let us add that)

PR #1981, opened 2025-08-20
Make workflow ids of import updates consistent

rautenrieth-da:
Fixes BFT consistency errors on CILR

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
@@ -25,6 +25,7 @@ import org.lfdecentralizedtrust.splice.environment.ledger.api.{
   TreeUpdate,
   TreeUpdateOrOffsetCheckpoint,
 }
+import org.lfdecentralizedtrust.splice.environment.ParticipantAdminConnection.IMPORT_ACS_WORKFLOW_ID_PREFIX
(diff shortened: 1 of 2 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1344,13 +1345,30 @@ class UpdateHistory(
...
   private def decodeImportTransaction(
       updateRow: SelectFromImportUpdates
   ): UpdateHistoryResponse = {
-    // The result should be consistent across SVs, so we generate a deterministic update id and event id.
     // We don't use any prefix so that we can use an index on contract ids when fetching import updates.
     val updateId = updateRow.contractId
     val eventNodeId = 0
+    // The prefix needs to be preserved, because we're relying on it to determine whether a given update
+    // was an import update.
+    val workflowId = s"${IMPORT_ACS_WORKFLOW_ID_PREFIX}-${updateId}"
```
rautenrieth-da, line 1367: Double checking: we don't need to worry about string length here, right?

moritzkiefer-da: Afaict the ledger API enforces a 255 character limit. It sounds like we're still below that right? Update ids have fixed length and our prefix isn't all that long.

moritzkiefer-da (approved):
nice thanks!

PR #1985, opened 2025-08-21
vagrant: Restart nix-daemon after mounting cache

giner:
- Makes the initial boot more predictable.
- Allows recovering after deleting the cache file without re-creating the VM. To recover run `vagrant up --provision`.


moritzkiefer-da (approved):
thanks!

PR #1997, opened 2025-08-21
Fix Token Standard CLI on transfer receivers

OriolMunoz-da:
Our current test suite didn't catch this because the only TransferIn happened via `TransferInstruction_Accept`, not transfer.

I'll improve the test suite later, removing bugs for CNU first...


moritzkiefer-da (approved):
thanks! duh, a normal single step transfer should work for testing here right?

OriolMunoz-da:
test case before and after the impl change:
image

PR #2006, opened 2025-08-25
Add missing CO_TransferPreapprovalSend case in UserWalletTxLogParser

OriolMunoz-da:
Don't you love that we can't have exhaustive pattern matching with Java classes?


rautenrieth-da (approved):
Thanks!

👍 for adding a test

OriolMunoz-da:
btw: this isn't breaking ingestion, but it becomes an unknowntxlogentry and that's ugly (plus logs error)

PR #2027, opened 2025-08-26
Add additional env vars to scan

OriolMunoz-da:


cluster/pulumi/common-sv/src/singleSvConfig.ts
```diff
@@ -37,6 +37,7 @@ const SingleSvConfigSchema = z
         cometbftExtraLogLevelFlags: z.string().optional(),
       })
       .optional(),
+    additionalEnvVars: z.array(EnvVarConfigSchema).default([]),
```
moritzkiefer-da: shouldn't this be under scan similar to the section in SvAppConfigSchema?

OriolMunoz-da: do you mean [this scan](https://github.com/hyperledger-labs/splice/blob/6858381e9149ddb87ecbde2c8be60b9f13977920/cluster/pulumi/common-sv/src/config.ts#L102) or a new section under SingleSvConfigSchema? I cannot tell why `scan~App~` is there and not here or viceversa :confused:

moritzkiefer-da: I mean a new section here makes more sense for now. We have an issue to cleanup that config structure mess (that ofc we will never get to)

moritzkiefer-da (approved):
thanks, maybe set it in the mock config to see that it gets applied properly?

Issue #863, opened 2025-06-05
Stream contracts instead of ingesting them as a single Seq


canton-network-da:
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#5534). Original author: @rautenrieth-da ]

`UpdateIngestionService.ingestAcsAndInFlight` currently loads the entire ACS into a single `Seq`, then sends them to the store for ingestion. This does not scale for large ACSs. 

Instead, pass the ACS stream from the ledger directly to the store and let the store handle streaming ingestion. Make sure to handle the case when the app crashes during the ACS ingestion.

canton-network-da:
[ from @meiersi-da ]: > Instead, pass the ACS stream from the ledger directly to the store and let the store handle streaming ingestion. Make sure to handle the case when the app crashes during the ACS ingestion.

So far we've kept the store passive and just exposed an ingestion sink. I'm not opposed to your proposal of having the store drive the ingestion by itself, but wanted to call out the difference. The benefit of using the ingestion sink is IMHO easier testability.

canton-network-da:
[ from @cocreature ]: Yeah I'd also lean towards having a method per stream element on the ingestion sink and drive it from the outside.

canton-network-da:
[ from @rautenrieth-da ]: What I was proposing was
```scala
  trait IngestionSink {
    def ingestAcs(acs: Stream[ActiveContract]): Future[Unit]
  }
```
If I understand correctly, you are proposing something like
```scala
  trait IngestionSink {
    def ingestAcsBegin(): Future[Unit]
    def ingestAcsChunk(acs: Seq[ActiveContract]): Future[Unit]
    def ingestAcsEnd(): Future[Unit]
  }
```
instead?

canton-network-da:
[ from @isegall-da ]: @rautenrieth-da @cocreature 
Still think it's a good idea? Moving to ~tech-debt~ tactical performance tuning for now, please close if you see fit.

canton-network-da:
[ from @cocreature ]: Tactical performance tuning makes sense. We currently make the assumption in a bunch of places that the ACS fits in memory (SV bootstrapping is another one). I think we can get away with that for early mainnet but we'll definitely need to address it at some point.

meiersi-da:
@moritzkiefer-da @nicu-da : do you have this one on your radar wrt the large ACS that we now have?

PR #2041, opened 2025-08-27
Stream ACS instead of updates from ledger end / latest pruned offset

OriolMunoz-da:
[HDM test](https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/51420/workflows/ebaa7cfc-3568-4d4d-a4f7-c6eb7f74282c)

Fixes https://github.com/hyperledger-labs/splice/issues/3047
Fixes https://github.com/hyperledger-labs/splice/issues/863

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
@@ -68,6 +68,11 @@ class UpdateIngestionService(
             _ = logger.debug(
               s"Starting ingestion from participant begin at $participantBegin"
             )
+            /** * This only applies to UpdateHistory. Does not load any ACS. Anything before the latest pruned offset:
+              * - Scan: it will be backfilled.
+              * - Wallet: won't miss any updates, because all updates visible to the wallet party only appear
+              *           after you onboard that party through the wallet app. Does not backfill.
+              */
             _ <- ingestionSink.ingestAcsStreamInBatches(
               Source.empty,
               participantBegin,
--- a/apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/StoreTestBase.scala
+++ b/apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/StoreTestBase.scala
@@ -961,6 +961,15 @@ abstract class StoreTestBase
     updateId = updateId,
   )
 
+  protected def acsImportEntryToActiveContract(entry: StoreTestBase.AcsImportEntry) = entry match {
+    case StoreTestBase.AcsImportEntry(contract, domain, counter, implementedInterfaces) =>
+      ActiveContract(
+        domain,
+        toCreatedEvent(contract, Seq(dsoParty), implementedInterfaces = implementedInterfaces),
+        counter,
+      )
+  }
+
   protected def acs(
       acs: Seq[StoreTestBase.AcsImportEntry] = Seq.empty,
       incompleteOut: Seq[StoreTestBase.AcsImportIncompleteEntry] = Seq.empty,
@@ -969,14 +978,7 @@ abstract class StoreTestBase
   )(implicit store: MultiDomainAcsStore): Future[Unit] = for {
     _ <- store.testIngestionSink.ingestAcs(
       acsOffset,
-      acs.map {
-        case StoreTestBase.AcsImportEntry(contract, domain, counter, implementedInterfaces) =>
-          ActiveContract(
-            domain,
-            toCreatedEvent(contract, Seq(dsoParty), implementedInterfaces = implementedInterfaces),
-            counter,
-          )
-      },
+      acs.map(acsImportEntryToActiveContract),
       incompleteOut.map {
         case StoreTestBase.AcsImportIncompleteEntry(
               c,
--- a/apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStoreTest.scala
+++ b/apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStoreTest.scala
@@ -4,7 +4,11 @@ import com.daml.ledger.javaapi.data.{DamlRecord, Identifier, OffsetCheckpoint}
 import com.digitalasset.daml.lf.data.Time.Timestamp
 import org.lfdecentralizedtrust.splice.codegen.java.splice.amulet.AppRewardCoupon
 import org.lfdecentralizedtrust.splice.environment.ParticipantAdminConnection.IMPORT_ACS_WORKFLOW_ID_PREFIX
-import org.lfdecentralizedtrust.splice.environment.{DarResources, RetryProvider}
+import org.lfdecentralizedtrust.splice.environment.{
+  BaseLedgerConnection,
+  DarResources,
+  RetryProvider,
+}
 import org.lfdecentralizedtrust.splice.environment.ledger.api.{
   TransactionTreeUpdate,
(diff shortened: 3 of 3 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1055,7 +1055,7 @@ final class DbMultiDomainAcsStore[TXE](
                 s"Initializing the ACS at an offset chosen by the ingestion service, and resuming ingestion from there."
             )
             initState(acsStoreId, Some(txLogStoreId), None)
-            IngestionStart.InitializeAcsAtLatestOffset
+            IngestionStart.InitializeAtParticipantBegin
```
rautenrieth-da: We could probably also leave this at `InitializeAcsAtLatestOffset`, as we have a process for backfilling the missing txlog entries. This could mean the ACS store could be faster to catch up, but I don't think it's worth the trouble with dealing with incomplete txlog stores at app startup.

rautenrieth-da (commented):
Previously, the choice whether to start ingesting at participant begin or ledger end was left to the IngestionService. The name `InitializeAcsAtLatestOffset` is a bit misleading, its comment describes better what it does.

Looking at the code, ALL ingestion services are set up to always initialize at participant begin (unless we override `ingestFromParticipantBegin` in some config?). This is a bit surprising to me, as:

1. AcsStores without a txlog can safely initialize at ledger end
2. I thought the wallet AcsStore can also initialize at ledger end. The store has a txlog, but we create the store when we onboard the user, and before that the user can't do anything relevant to the wallet store.

Point 1 would be addressed by your PR.

I'm not so sure about point 2. The assumption that the store is initialized at an offset before the first relevant update is a bit shaky: the store is created through a trigger, so there are delays, and what if they bypass the validator app and submit commands directly to the participant before onboarding completes. On the other hand, having to stream in all updates since participant begin every time we onboard a new user sounds wasteful?

Other than that, the change looks good to me. Let's make sure we remove all the `ingestFromParticipantBegin` and `ingestUpdateHistoryFromParticipantBegin` config parameters when they become unused.

OriolMunoz-da:
> I'm not so sure about point 2. The assumption that the store is initialized at an offset before the first relevant update is a bit shaky: the store is created through a trigger, so there are delays, and what if they bypass the validator app and submit commands directly to the participant before onboarding completes. On the other hand, having to stream in all updates since participant begin every time we onboard a new user sounds wasteful?

@rautenrieth-da wasn't that the behavior before we added the ingestFromParticipantBegin boolean flag? Which was fine afaiu

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: restored it to `InitializeAcsAtLatestOffset`, tests felt very strongly that that's the right choice

> This could mean the ACS store could be faster to catch up, but I don't think it's worth the trouble with dealing with incomplete txlog stores at app startup.

I would guess the former is better, because it can take a long time and we have `waitUntilAcsIngested` all over the place

OriolMunoz-da: fascinating, [that doesn't work](https://github.com/hyperledger-labs/splice/actions/runs/17463021537/job/49592033581?pr=2041)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
```diff
@@ -766,6 +766,10 @@ object MultiDomainAcsStore extends StoreErrors {
         */
       final case object InitializeAcsAtLatestOffset extends IngestionStart
 
+      /** Ingestion service should not ingest the ACS, and instead start from the participant's begin
+        */
+      final case object InitializeAtParticipantBegin extends IngestionStart
```
moritzkiefer-da: The comment and naming seems confusing:
1. It's not participant begin, it's latest pruned offset
2. Because it is latest pruned offset there can actually be a non-empty ACS that we need to ingest

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
```diff
@@ -61,30 +60,26 @@ class UpdateIngestionService(
...
           } yield offset
+        case IngestionStart.InitializeAtParticipantBegin =>
+          for {
+            participantBegin <- connection.latestPrunedOffset()
+            _ = logger.debug(
+              s"Starting ingestion from participant begin at $participantBegin"
+            )
+            _ <- ingestionSink
+              .ingestAcs(
+                participantBegin,
```
moritzkiefer-da: So looking at the code it seems like this was also broken before but I don't see how an empty sequence is correct if you actually pruned

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
```diff
@@ -61,30 +60,26 @@ class UpdateIngestionService(
           for {
             _ <- ingestAcsAndInFlight(offset)
           } yield offset
+        case IngestionStart.InitializeAtParticipantBegin =>
```
moritzkiefer-da: The distinction between "InitializeAcs" and "Initialize" doesn't really exist for the update history right? We just reuse the type so we inherit it?

docs/src/release_notes.rst
```diff
@@ -520,6 +520,11 @@
         Thereby making the ``AmuletRules`` contract smaller and saving traffic for transactions using it.
         This is motivated by `CIP-0078 CC Fee Removal <https://github.com/global-synchronizer-foundation/cips/blob/main/cip-0078/cip-0078.md>`__ .
 
+  - Splice apps
```
moritzkiefer-da: I probably would just omit this, I think this is gonna confuse people more than it will help

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
OriolMunoz-da: I'm not sure if you're pointing at something to be fixed here?
The code here only marks the ACS as ingested by not ingested anything... bit confusing, so I will fix it in https://github.com/hyperledger-labs/splice/pull/3863

moritzkiefer-da: So we intentionally miss created events here? Why is that ok?

moritzkiefer-da: ah I guess because we backfill in scan? We don't in the wallet but maybe it's still ok to miss them given how we use the update history. Feels kinda weird though that you can have an archive for which you did not see the create.

rautenrieth-da: My assumption was that wallet update history can't miss any updates, because all updates visible to the wallet party only appear after you onboard that party through the wallet app. It's probably a very shaky assumption.

moritzkiefer-da: hm ok, I think it's fine. Maybe let's just add a comment that we explicitly ignore them.

PR #2045, opened 2025-08-27
Fix log ignore

rautenrieth-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/5420

We added some log ignores in https://github.com/hyperledger-labs/splice/pull/1770, then fixed them in https://github.com/hyperledger-labs/splice/pull/1874, but they are still broken:
- The pattern doesn't escape `()` and doesn't match the warning
- The line `shutdown did not complete gracefully in allotted 3 seconds` wasn't ignored (it appeared at least in https://github.com/DACH-NY/cn-test-failures/issues/5420 before all the other warnings)

### DIFF
--- a/project/ignore-patterns/canton_network_test_log.ignore.txt
+++ b/project/ignore-patterns/canton_network_test_log.ignore.txt
@@ -120,8 +120,11 @@ Failed to connect to scan of .*SvDevNetReonboardingIntegrationTest
 # Log suppression doesn't seem to work here as the tests use a different logger
 GENERIC_CONFIG_ERROR.*SpliceConfigTest
 
+# shutdown can be slower than the timeouts
+shutdown did not complete gracefully in allotted 3 seconds
+
 # ExecutorServiceMetrics closes the gauges twice through shutdown and shutownNow in cases where shutdown times out
-.*InstrumentDescriptor.*name=daml.executor.* has called close() multiple times
+.*InstrumentDescriptor.*name=daml.executor.* has called close\(\) multiple times
 
 
 # Make sure to have a trailing newline


project/ignore-patterns/canton_network_test_log.ignore.txt
```diff
@@ -120,8 +120,11 @@ Failed to connect to scan of .*SvDevNetReonboardingIntegrationTest
 # Log suppression doesn't seem to work here as the tests use a different logger
 GENERIC_CONFIG_ERROR.*SpliceConfigTest
 
+# shutdown can be slower than the timeouts
+shutdown did not complete gracefully in allotted 3 seconds
```
nicu-da: didn't we fix these? Why do we need the ignore for this?

rautenrieth-da: It happened here 7 days ago: https://github.com/hyperledger-labs/splice/actions/runs/17093487633/job/48472441200#step:4:3958, and was the cause for the duplicate close() calls (if I understand your explanation at https://github.com/DACH-NY/cn-test-failures/issues/5161#issuecomment-3140054561).

Did we do anything in the past week to fix the shutdown being slow? It sounds to me like we should either ignore both lines touched in this PR, or none of them.

nicu-da: Fair enough, though I am wondering if we can increase that timeout instead. (the ignore for duplicate calls makes sense)

rautenrieth-da: There could be many different reasons why shutdown times out. For example, we investigated and fixed one previously [here](https://github.com/DACH-NY/canton-network-node/issues/9827). Upon second thought, I don't think it's a good idea to blanket ignore all shutdown timeouts, so I have removed the new ignore again. Ignoring the `duplicate close()` warnings is still a good idea because it focuses us on investigating the cause (something blocking shutdown) instead of a symptom.

nicu-da (approved):

PR #2046, opened 2025-08-27
Prevent AmuletRules_BuyMemberTraffic from causing flakes

OriolMunoz-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/5436


rautenrieth-da (approved):
Looks sensible, thanks!

Should we also check all other occurences of `.listTransactions` in `*Test.scala`? [This one](https://github.com/hyperledger-labs/splice/blob/main/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DisasterRecoveryIntegrationTest.scala#L529) for example looks like it could be affected as well.

OriolMunoz-da:
> [This one](https://github.com/hyperledger-labs/splice/blob/main/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DisasterRecoveryIntegrationTest.scala#L529) for example looks like it could be affected as well.

that's from scan

though there are some for some validator wallets where this can still happen, so adding it to those

PR #2047, opened 2025-08-27
Add docs about wasted traffic

### LINKED ISSUE #576: Document what `Wasted traffic cost` means; maybe how to monitor and what to do about it
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#18301). Original author: @martinflorian-da ]

Some bits from a discussion from Five North that (partially) sound like documentation to me:

> Am i right that this "Waste traffic" always happen, it is just seems way more from our node? Because when I search my own log, I see a lot for "Wasted traffic cost" from other nodes on my own log as well. But we're general 3-5x had more Wasted compare with other nodes.

> You're not entirely wrong; here is some nuance:
>
> - it happens quite frequently for mediators as they frequently (maybe even always; would have to check) confirm transactions that were already confirmed by a sufficient number of other mediators; so these extra confirmations above the threshold are essentially wasting traffic; for now we're just ignoring wasted traffic from mediators
> - the main reason we have an alert for this are validators; if they waste traffic they very often literally waste (their own) money; if this happens on a larger scale we might have a bug somewhere that we should address
> - for the past weeks we see it increasingly happening also for SV participants; on a smaller frequency and severity this is sadly unavoidable due to CometBFT quirks; your participant seems to be especially "wasteful" though, which based on our analysis above points to an issue in your sequencer and/or cometbft that you might want to address

> Is there a particular dashboard that monitor this ?

> `canton-network -> Synchronizer Fees (SV View)` has a panel for this at the bottom. However it aggregates over all participants on the synchronizer. We have an alert per-member and if you want to dig deeper on the history you can go to `Explore` on that panel and change the query. For example to this query to show you your participant:
>
> ```sum by (namespace, job, member) (rate(daml_sequencer_traffic_control_wasted_traffic_total{namespace=~"sv-1", member=~"PAR::Five.*"}[$__rate_interval]))```
>
> For validators there is `canton-network -> Synchronizer Fees (SV View)` where the same kind of info can be inferred from the `Rejected Event Traffic` panels for example.

We might also want to add pointers for debugging wasted traffic? For example for SVs, looking at `Sequencer Client Delay` on the `canton-network > Global Domain Catchup` dashboard.

martinflorian-da:
Fixes #576

image

### DIFF
--- a/docs/src/deployment/traffic.rst
+++ b/docs/src/deployment/traffic.rst
@@ -185,7 +185,7 @@ wasted traffic only occurs whenever a synchronizer event is rejected after seque
 Some level of wasted traffic is expected and unavoidable, due to factors such as:
 
 - Submission request amplification.
-  Participants that use BFT sequencer connections duplicate submission requests after a timeout to ensure speedy delivery in the face of nonresponsive sequencers;
+  Participants that use BFT sequencer connections retry submission requests after a timeout to ensure speedy delivery in the face of nonresponsive sequencers;
   if processing was simply slower than usual but the sequencer was not faulty, the duplicate request counts as wasted traffic.
 - Duplication of messages within the ordering layer, typically linked to transient networking issues or load spikes.
 - Duplication of submissions on the participant/app side, for example when catching up after restoring from a backup or after some crashes.
@@ -193,7 +193,7 @@ Some level of wasted traffic is expected and unavoidable, due to factors such as
 Validator perspective
 +++++++++++++++++++++
 
-Validator operators are encouraged to monitor the rate of failed submissions on their validators and investigate the causes of repeatedly failing submissions eagerly.
+Validator operators are encouraged to investigate the causes of repeatedly failing submissions.
 As stated above, not all failed submissions result in wasted traffic, and some wasted traffic is unavoidable.
 Attention is warranted, however, if the rate of wasted traffic increases significantly at some point in time.
 


docs/src/deployment/traffic.rst
```diff
@@ -171,3 +171,37 @@ the validator app will
...
 For configuring the built-in top-up automation, please refer to the :ref:`validator deployment guide <helm_validator_topup>`.
 Configuring alternative methods for buying traffic, e.g., using third-party services, exceeds the scope of this documentation.
+
+.. _traffic_wasted:
+
+Wasted traffic
+--------------
+
+`Wasted traffic` is defined as synchronizer events that have been sequenced but will not be delivered to their recipients.
+Wasted traffic is problematic for validators because of traffic fees:
```
moritzkiefer-da:
```suggestion
Wasted traffic can be problematic for validators because of traffic fees:
```

moritzkiefer-da: I don't want this to sound too scary.

docs/src/deployment/traffic.rst
```diff
@@ -171,3 +171,37 @@ the validator app will
...
+it means that :ref:`traffic <traffic_accounting>` has been charged for a message that was ultimately not delivered.
+Not all failed submissions result in wasted traffic:
+wasted traffic only occurs whenever a synchronizer event is rejected after sequencing but before delivery.
+
+Validator perspective
++++++++++++++++++++++
+
+Validator operators are encouraged to investigate failed submissions eagerly to avoid systemic causes for wasted traffic that are due
+to their individual configuration and/or the specific applications using their validators.
+The Splice distribution contains a :ref:`Grafana dashboard <metrics_grafana_dashboards>` about `Synchronizer Fees (validator view)` that can be helpful in addition to inspecting logs;
```
moritzkiefer-da: Are you sure that rejected traffic is the same as wasted traffic? I believe it is from a quick look at the code but not entirely sure if it doesn't also include the stuff that gets rejected before sequencing already. If you are not sure I'd double check with Thibault.

docs/src/deployment/traffic.rst
```diff
@@ -171,3 +171,37 @@ the validator app will
...
+`Wasted traffic` is defined as synchronizer events that have been sequenced but will not be delivered to their recipients.
+Wasted traffic is problematic for validators because of traffic fees:
+it means that :ref:`traffic <traffic_accounting>` has been charged for a message that was ultimately not delivered.
+Not all failed submissions result in wasted traffic:
+wasted traffic only occurs whenever a synchronizer event is rejected after sequencing but before delivery.
+
+Validator perspective
++++++++++++++++++++++
+
+Validator operators are encouraged to investigate failed submissions eagerly to avoid systemic causes for wasted traffic that are due
```
moritzkiefer-da: I think this sounds too scary. Investigating any single non-zero wasted traffic is a waste of time. You have BFT sequencer connections + request amplification. You also have an ordering layer that can drop messages (this does not change for dabft). You are gonna waste some amount of traffic. So I think we need to make that explicit and tell people that some amount of it is expected but if they see a sudden increase something is probably off.

docs/src/deployment/traffic.rst
```diff
@@ -171,3 +171,37 @@ the validator app will
...
+
+SV operators are encouraged to monitor wasted traffic across all synchronizer members,
+as reported for example by sequencer :ref:`metrics <metrics>`,
+to avoid cases where misconfiguration incurs excessive monetary losses for validators.
+The Splice distribution contains a :ref:`Grafana dashboard <metrics_grafana_dashboards>` about `Synchronizer Fees (SV view)` that can be helpful,
+as well as an alert definition that focuses on validator participants.
+
+Note that wasted traffic is less relevant for SVs themselves as SV components have unlimited traffic.
+Note also that SV mediators and sequencers waste traffic as part of their regular operation;
+they frequently use aggregate submissions where all composite submission requests beyond the aggregation threshold get discarded.
```
moritzkiefer-da: I never heard the term composite submission request. I know what you mean here but I think for someone not familiar with how sequencer message aggregation works this is just confusing.

docs/src/deployment/traffic.rst
```diff
@@ -171,3 +171,37 @@ the validator app will
...
+
+Validator operators are encouraged to investigate failed submissions eagerly to avoid systemic causes for wasted traffic that are due
+to their individual configuration and/or the specific applications using their validators.
+The Splice distribution contains a :ref:`Grafana dashboard <metrics_grafana_dashboards>` about `Synchronizer Fees (validator view)` that can be helpful in addition to inspecting logs;
+see, for example, the `Rejected Event Traffic` panel there.
+
+SV perspective
+++++++++++++++
+
+SV operators are encouraged to monitor wasted traffic across all synchronizer members,
```
moritzkiefer-da, line 208: similar comment as above: I think we need to make it very clear that some amount of waste is expected and only increases should be monitored and alerted on.

moritzkiefer-da (commented):
Thanks! I think overall it makes sense to document this but the current documentation sounds too scary imho. We need to make it clear that waste is expected due to network blibs, ordering layer unreliance, restarts, … and only increases should be something to worry about. Otherwise we will get a bunch of confused users that start worrying about every non-zero value which isn't helpful for anyone.

docs/src/deployment/traffic.rst
martinflorian-da: tbf I wasn't entirely sure, hence the vague wording, but looking at our own comments I am now pretty sure that this is what this panel is for: https://github.com/hyperledger-labs/splice/blob/fee65f529f888fd36c860961486b13d233c912f0/cluster/pulumi/infra/grafana-dashboards/canton-network/synchronizer-fees-validator.json#L475

docs/src/deployment/traffic.rst
martinflorian-da: > You also have an ordering layer that can drop messages (this does not change for dabft).

Trying to understand how this can lead to wasted traffic.... Once it's sequenced (precondition for counting it as wasted traffic), the ordering layer dropping messages shouldn't matter anymore?

moritzkiefer-da: Hm right a full drop doesn't matter. A delay or replay does though and we do observe that occasionally.

docs/src/deployment/traffic.rst
martinflorian-da: reworded more than that

docs/src/deployment/traffic.rst
martinflorian-da: changed all of it

docs/src/deployment/traffic.rst
```diff
@@ -171,3 +171,49 @@ the validator app will
...
+
+`Wasted traffic` is defined as synchronizer events that have been sequenced but will not be delivered to their recipients.
+For validators, which are subject to traffic fees,
+wasted traffic implies that :ref:`traffic <traffic_accounting>` has been charged for a message that was ultimately not delivered.
+Not all failed submissions result in wasted traffic:
+wasted traffic only occurs whenever a synchronizer event is rejected after sequencing but before delivery.
+Some level of wasted traffic is expected and unavoidable, due to factors such as:
+
+- Submission request amplification.
+  Participants that use BFT sequencer connections duplicate submission requests after a timeout to ensure speedy delivery in the face of nonresponsive sequencers;
```
moritzkiefer-da:
```suggestion
  Participants that use BFT sequencer connections retry submission requests after a timeout to ensure speedy delivery in the face of nonresponsive sequencers;
```
somehow sounds more intuitive to me than duplicate

docs/src/deployment/traffic.rst
```diff
@@ -171,3 +171,49 @@ the validator app will
...
+  if processing was simply slower than usual but the sequencer was not faulty, the duplicate request counts as wasted traffic.
+- Duplication of messages within the ordering layer, typically linked to transient networking issues or load spikes.
+- Duplication of submissions on the participant/app side, for example when catching up after restoring from a backup or after some crashes.
+
+Validator perspective
++++++++++++++++++++++
+
+Validator operators are encouraged to monitor the rate of failed submissions on their validators and investigate the causes of repeatedly failing submissions eagerly.
```
moritzkiefer-da: Do we document how they can do so? It feels a bit weird to talk about monitoring failed submissions but then we only tell people how to monitor wasted traffic after you explain that not all failed submisions are wasted traffic

docs/src/deployment/traffic.rst
```diff
@@ -171,3 +171,49 @@ the validator app will
...
+As stated above, not all failed submissions result in wasted traffic, and some wasted traffic is unavoidable.
+Attention is warranted, however, if the rate of wasted traffic increases significantly at some point in time.
+
+The Splice distribution contains a :ref:`Grafana dashboard <metrics_grafana_dashboards>` about `Synchronizer Fees (validator view)`,
+to assist in monitoring traffic-related metrics.
+The `Rejected Event Traffic` panel on this dashboard is especially relevant for determining the rate of wasted traffic.
+(Hover on the ⓘ symbols in panel headers for precise descriptions of the shown data.)
```
moritzkiefer-da, line 203: impressive unicode skills

moritzkiefer-da (approved):
very nice thank you! Thanks for taking the time to address my suggestions

docs/src/deployment/traffic.rst
martinflorian-da: I'm mainly impressed that it rendered as expected. It did!

docs/src/deployment/traffic.rst
martinflorian-da: :+1: accepting thanks

docs/src/deployment/traffic.rst
martinflorian-da: Hmpf so I was thinking of them monitoring it via what their apps do... Let me just remove the explicit mention of monitoring... not sure that I want to expand this a lot further right now.

martinflorian-da: I'm keeping that part

> Validator operators are encouraged to investigate the causes of repeatedly failing submissions.

...because it seems like sometimes stating obvious things is helpful...

PR #2053, opened 2025-08-28
Expose ledger API in docker compose setup

### LINKED ISSUE #2038: Expose ledger API in docker compose setups
Help for exposing the ledger api in our docker compose setup was requested [here](https://docs.dev.sync.global/app_dev/ledger_api/index.html). We should expose this by default or document how to.

moritzkiefer-da:
fixes #2038

[ci]

This keeps coming up every few days so rather than copy pasting the config snippet around let's just include it.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


isegall-da (approved):
Nice, thank you. Shouldn't we also add it to localnet?

moritzkiefer-da:
> Nice, thank you. Shouldn't we also add it to localnet?

localnet already has support

isegall-da:
So the main difference seems to be that in localnet we made an effort not to expose everything, but only `/docs/openapi` and `/v2`. Do you know what else is there and if there's a reason not to expose that "something else" too?

I tend to agree that otherwise, the proposal in this PR seems nicer. I guess that in localnet we can have both for a while, and label the existing deprecated, no?

Issue #565, opened 2025-06-05
Document manually creating an identity/DR dump


canton-network-da:
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#18462). Original author: @moritzkiefer-da ]

```
val keys = "[" + sequencer.keys.secret.list().filter(k => k.name.get.unwrap != "cometbft-governance-keys").map(key => s"{\"keyPair\": \"${Base64.getEncoder.encodeToString(sequencer.keys.secret.download(key.publicKey.fingerprint).toByteArray)}\", \"name\": \"${key.name.get.unwrap}\"}").mkString(",") + "]"
val authorizedStoreSnapshot = Base64.getEncoder.encodeToString(sequencer.topology.transactions.export_topology_snapshot(filterStore = "Authorized", timeQuery = TimeQuery.Range(from = None, until = None), filterMappings = Seq(
          TopologyMapping.Code.NamespaceDelegation,
          TopologyMapping.Code.OwnerToKeyMapping,
          TopologyMapping.Code.IdentifierDelegation,
          // only relevant for sequencers
          TopologyMapping.Code.VettedPackages,
        ), filterNamespace = sequencer.id.namespace.toProtoPrimitive).toByteArray)

```
for seq, med, par

Merge with data snapshot from other node using jq. Could also get data snapshots that way in principle.

Wondering if we should just ship this as a script that we test.

moritzkiefer-da:
@martinflorian-da For testing, how about we add a Canton script that produces the dump and add an integration test that checks that the dump produced by that is identical to the one from the validator app? That seems like a fairly cheap test to add and ensures that the instructions actually work.

moritzkiefer-da:
Yes agreed, I would not do anything for sequencers and mediators at this point.

PR #2054, opened 2025-08-28
Docs improvements: validator disaster recovery

martinflorian-da:
Fixes https://github.com/hyperledger-labs/splice/issues/565

Fixes https://github.com/hyperledger-labs/splice/issues/2048

### DIFF
--- a/apps/app/src/pack/examples/recovery/manual-identities-dump.sc
+++ b/apps/app/src/pack/examples/recovery/manual-identities-dump.sc
@@ -4,7 +4,8 @@ import java.util.Base64
 
 val id = participant.id.toProtoPrimitive
 
-val keys = "[" + participant.keys.secret.list().filter(k => k.name.get.unwrap != "cometbft-governance-keys").map(key => s"{\"keyPair\": \"${Base64.getEncoder.encodeToString( participant.keys.secret.download(key.publicKey.fingerprint).toByteArray)}\", \"name\": \"${key.name.get.unwrap}\"}") .mkString(",") + "]"
+// This line needs to be adapted if your participant stores keys in an external KMS
+val keys = "[" + participant.keys.secret.list().filter(k => k.name.get.unwrap != "cometbft-governance-keys").map(key => s"{\"keyPair\": \"${Base64.getEncoder.encodeToString(participant.keys.secret.download(key.publicKey.fingerprint).toByteArray)}\", \"name\": \"${key.name.get.unwrap}\"}") .mkString(",") + "]"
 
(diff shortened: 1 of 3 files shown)


apps/app/src/pack/examples/recovery/manual-identities-dump.sc
```diff
@@ -0,0 +1,16 @@
+import com.digitalasset.canton.topology.transaction.TopologyMapping
+import com.digitalasset.canton.topology.store.TimeQuery
+import java.util.Base64
+
+val id = participant.id.toProtoPrimitive
+
+val keys = "[" + participant.keys.secret.list().filter(k => k.name.get.unwrap != "cometbft-governance-keys").map(key => s"{\"keyPair\": \"${Base64.getEncoder.encodeToString( participant.keys.secret.download(key.publicKey.fingerprint).toByteArray)}\", \"name\": \"${key.name.get.unwrap}\"}") .mkString(",") + "]"
```
moritzkiefer-da: Does this work for KMS? It seems like it probably needs tweaking for that. I'm ok merging this PR without fixing it though but maybe let's open a separate issue then?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ParticipantPlaintextIdentitiesDumpIntegrationTest.scala
```diff
@@ -109,6 +117,29 @@ class ParticipantPlaintextIdentitiesIntegrationTest extends IntegrationTest with
...
+        )
+        NodeIdentitiesDump
+          .fromJsonFile(
+            dumpPath,
+            ParticipantId.tryFromProtoPrimitive,
+          )
+          .value
+      }
+
+      clue("Manually dumped identities match the ones dumped via the API") {
```
moritzkiefer-da, line 136: Any reason you don't just do a svParticipantDumpManual.copy(version = svParticipantDump.version) should be svParticipantDump?

Explicitly checking fields usually doesn't work well if we ever add more fields as we'll forget to update this.

apps/app/src/pack/examples/recovery/manual-identities-dump.sc
martinflorian-da: Pretty sure it doesn't work. I didn't do it immediately because I don't see evidence yet that anyone will need that soon.

And for the same reason I opted for not creating a follow-up issue. I don't see us prioritizing it unless someone asks for it, so we can also create the issue then? Not a strong opinion at all, happy to create that issue if this reasoning seems meh to you.

I do think that I should highlight that limitation in the docs text though, let me add that.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ParticipantPlaintextIdentitiesDumpIntegrationTest.scala
martinflorian-da: > Any reason you don't just do a svParticipantDumpManual.copy(version = svParticipantDump.version)

I just didn't think of that :) Thanks changing

apps/app/src/pack/examples/recovery/manual-identities-dump.sc
moritzkiefer-da: fair enough so just adding a comment to the script and the docs that it doesn't work for kms is probably good enough for now.

PR #2064, opened 2025-08-28
Fix createTransferPreapproval timing out in tests

OriolMunoz-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/5437

This wasn't because it didn't have funds, but because the validator took too long to accept the preapproval and the walletclient timed out.


OriolMunoz-da:
note: only showed up in one test, but I expect others have the same potential

PR #2072, opened 2025-08-29
Use mediatorReactionTimeout for pausing synchronizers on migrations

moritzkiefer-da:
Fixes https://github.com/DACH-NY/canton-network-internal/issues/699

[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/DomainParamsIngestionTrigger.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/DomainParamsIngestionTrigger.scala
@@ -4,9 +4,10 @@
 package org.lfdecentralizedtrust.splice.automation
 
 import org.lfdecentralizedtrust.splice.environment.ParticipantAdminConnection
-import org.lfdecentralizedtrust.splice.migration.SynchronizerParametersStateTopologyConnection
+import org.lfdecentralizedtrust.splice.environment.TopologyAdminConnection.TopologyTransactionType.AuthorizedState
 import org.lfdecentralizedtrust.splice.store.DomainParamsStore
 import com.digitalasset.canton.SynchronizerAlias
+import com.digitalasset.canton.topology.store.TopologyStoreId
 import com.digitalasset.canton.tracing.TraceContext
 import com.digitalasset.canton.util.ShowUtil.*
 import io.grpc.{Status, StatusRuntimeException}
@@ -26,10 +27,6 @@ final class DomainParamsIngestionTrigger(
       quiet = true,
     ) {
 
-  private val domainStateTopology = new SynchronizerParametersStateTopologyConnection(
-    participantAdminConnection
-  )
-
   override def completeTask(
       task: PeriodicTaskTrigger.PeriodicTask
   )(implicit tc: TraceContext): Future[TaskOutcome] =
@@ -39,9 +36,12 @@ final class DomainParamsIngestionTrigger(
         Future.successful(TaskNoop)
       case Failure(e) => Future.failed(e)
       case Success(synchronizerId) =>
-        domainStateTopology
-          .firstAuthorizedStateForTheLatestSynchronizerParametersState(synchronizerId)
-          .value
+        participantAdminConnection
+          .lookupSynchronizerParametersState(
+            TopologyStoreId.SynchronizerStore(synchronizerId),
+            synchronizerId,
+            AuthorizedState,
+          )
           .flatMap {
             case None => Future.successful(TaskNoop)
             case Some(params) =>
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
@@ -1231,7 +1231,7 @@ abstract class TopologyAdminConnection(
       })
   }
 
-  def listSynchronizerParametersState(
+  def listSynchronizerParametersStateHistory(
       synchronizerId: SynchronizerId,
       proposals: TopologyTransactionType = AuthorizedState,
   )(implicit tc: TraceContext): Future[Seq[TopologyResult[SynchronizerParametersState]]] = {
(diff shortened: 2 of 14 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/DomainParamsIngestionTrigger.scala
```diff
@@ -39,9 +36,12 @@ final class DomainParamsIngestionTrigger(
         Future.successful(TaskNoop)
       case Failure(e) => Future.failed(e)
       case Success(synchronizerId) =>
-        domainStateTopology
-          .firstAuthorizedStateForTheLatestSynchronizerParametersState(synchronizerId)
```
moritzkiefer-da, line 43: I have no idea why this trigger fetched the history before. This just seems confusing and wasteful.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SynchronizerMigrationUtil.scala
```diff
@@ -0,0 +1,51 @@
...
+import com.digitalasset.canton.topology.SynchronizerId
+import com.digitalasset.canton.topology.transaction.SynchronizerParametersState
+import com.digitalasset.canton.tracing.TraceContext
+import org.lfdecentralizedtrust.splice.environment.TopologyAdminConnection
+import org.lfdecentralizedtrust.splice.environment.TopologyAdminConnection.TopologyResult
+import scala.concurrent.Future
+
+final object SynchronizerMigrationUtil {
+  def synchronizerIsPaused(params: TopologyResult[SynchronizerParametersState]): Boolean =
+    params.mapping.parameters.mediatorReactionTimeout == NonNegativeFiniteDuration.Zero
```
moritzkiefer-da, line 20: I'm a bit unsure if we should check mediatorReactionTimeout _and_ confirmationRequestsMaxRate. I did that initially but then I could be in the case where synchronizerIsPaused was false and synchronizerIsUnpaused was also false which doesn't seem very useful. In the end for safety we really care about mediator reaction timeout and the rest is just UX.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
```diff
@@ -1231,7 +1231,7 @@ abstract class TopologyAdminConnection(
       })
   }
 
-  def listSynchronizerParametersState(
+  def listSynchronizerParametersStateHistory(
```
moritzkiefer-da, line 1234: The name was quite confusing. Usually we use list functions to give us all current entries not history.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SynchronizerMigrationUtil.scala
nicu-da: Is there any scenario in the future where we could set the mediator reaction timeout to zero and not confirmation requests? 

> synchronizerIsPaused was false and synchronizerIsUnpaused was also false

how would that happen though? We  set both fields at the same time

nicu-da (approved):
Nice, thank you!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SynchronizerMigrationUtil.scala
moritzkiefer-da: > Is there any scenario in the future where we could set the mediator reaction timeout to zero and not confirmation requests?

Don't see one.

> how would that happen though? We set both fields at the same time

It never should. But the functions here would allow for it which feeled a bit weird.

nicu-da: wfm

martinflorian-da: hm a bit awkward but makes sense; maybe add a short in-code comment here as well?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SynchronizerMigrationUtil.scala
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+  def ensureSynchronizerIsUnpaused(
+      topologyConnection: TopologyAdminConnection,
+      synchronizerId: SynchronizerId,
+  )(implicit tc: TraceContext): Future[TopologyResult[SynchronizerParametersState]] =
+    topologyConnection.ensureDomainParameters(
+      synchronizerId,
+      _.tryUpdate(
+        // TODO(DACH-NY/canton-network-node#8761) hard code unpaused parameters for now
```
martinflorian-da: maybe just remove that todo at this point? the issue is both closed and in the legacy repo

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/migration/DomainParametersStateTopologyConnection.scala
```diff
@@ -21,19 +24,47 @@ class SynchronizerParametersStateTopologyConnection(connection: TopologyAdminCon
...
+    pausedState: TopologyAdminConnection.TopologyResult[
+      SynchronizerParametersState
+    ],
+    lastUnpausedState: TopologyAdminConnection.TopologyResult[SynchronizerParametersState],
+) {
+  def exportTimestamp: Instant = {
+    // We only check this here as some places invoke `firstAuthorizedStateForTheLatestSynchronizerParametersState`
+    // to wait for the synchronizer to be paused so it is not always paused but when someone
+    // wants to export, it definitely must be paused.
+    require(SynchronizerMigrationUtil.synchronizerIsPaused(pausedState))
```
martinflorian-da: I find this confusing; why instantiate a `PausedSynchronizerState` if the synchronizer might actually not be paused in that state? Can we not refactor this a bit for clarity, for example with some sealed trait or something?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SynchronizerMigrationUtil.scala
martinflorian-da: (sorry started my review before your convo but got interrupted; reopened the thread because an in-code comment might be nice)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/migration/DomainParametersStateTopologyConnection.scala
moritzkiefer-da: Will change the name, I don't want to refactor it even further. The logic is already a mess and I'd rather focus on things to get rid of HDMs than cleanup logict hat has been like this for literally a year and worked fine.

martinflorian-da: Changing the name sounds great thanks

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SynchronizerMigrationUtil.scala
moritzkiefer-da: added

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SynchronizerMigrationUtil.scala
moritzkiefer-da: left the comment but removed the todo

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/migration/DomainParametersStateTopologyConnection.scala
moritzkiefer-da: done

PR #2087, opened 2025-09-01
Enable import update backfilling by default

rautenrieth-da:
Fixes https://github.com/hyperledger-labs/splice/issues/1164

Enables the feature flag by default and removes it from Pulumi and test code, but (as before with other feature flags) does not remove it from the app config or helm charts.


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanAutomationService.scala
```diff
@@ -89,7 +89,9 @@ class ScanAutomationService(
       snapshotStore,
       store.updateHistory,
       config.acsSnapshotPeriodHours,
-      config.updateHistoryBackfillEnabled,
+      // The acs snapshot trigger should not attempt to backfill snapshots unless the backfilling
+      // UpdateHistory is fully enabled and complete.
+      config.updateHistoryBackfillEnabled && config.updateHistoryBackfillImportUpdatesEnabled,
```
rautenrieth-da, line 94: This change protects against corrupt snapshots in case someone manages to turn the feature flag on and then off again.

rautenrieth-da:
> also should this have release notes?

Yes it should, I will amend the PR. 

> slightly confused on what we shipped in 0.4.13 then

Fixes to the backfilling implementation, but still behind a feature flag. This is the last change that will enable everything by default.

PR #2088, opened 2025-09-01
Fix npm-run-parallel on mac

rautenrieth-da:
npm interprets `--if-present` as the value for `--workspace`, resulting in `--workspace=--if-present`.

This causes `sbt formatFix` and `sbt pulumi/npmFix` to fail on macos.

Btw: still no `rust-parallel` in the current nix.


fayi-da (approved):
nice catch

PR #2092, opened 2025-09-02
Prevent content-runtime-updater from logging errors for every transient failure

OriolMunoz-da:
Part of https://github.com/hyperledger-labs/splice/issues/2091

With `LOG_TO_STDERR_AFTER_FAILURES=3`:
Sample output to stdout:
```
curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the webpage mentioned above.
Failed to fetch DSO from https://scan.sv-26.dev.global.canton.network.digitalasset.com (failure #9): curl: (60) SSL: no alternative certificate subject name matches target hostname 'scan.sv-26.dev.global.canton.network.digitalasset.com'
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the webpage mentioned above.
Failed to fetch DSO from https://scan.sv-26.dev.global.canton.network.digitalasset.com (failure #10): curl: (60) SSL: no alternative certificate subject name matches target hostname 'scan.sv-26.dev.global.canton.network.digitalasset.com'
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the webpage mentioned above.
```

Sample output to stderr:
```
ERROR: 3 consecutive failures fetching DSO from https://scan.sv-26.dev.global.canton.network.digitalasset.com
ERROR: 6 consecutive failures fetching DSO from https://scan.sv-26.dev.global.canton.network.digitalasset.com
ERROR: 9 consecutive failures fetching DSO from https://scan.sv-26.dev.global.canton.network.digitalasset.com
```

### DIFF
--- a/cluster/helm/splice-info/templates/deployment.yaml
+++ b/cluster/helm/splice-info/templates/deployment.yaml
@@ -62,7 +62,7 @@ spec:
 
             fail_count=0
             # prevent spamming alerts by logging only after the issues are persistent
-            LOG_TO_STDERR_AFTER_FAILURES=100
+            log_to_stderr_after_failures=${LOG_TO_STDERR_AFTER_FAILURES:-100}
             while true; do
               start_time=$(date +%s);
 
@@ -77,7 +77,7 @@ spec:
               else
                 fail_count=$((fail_count + 1))
                 echo "Failed to fetch DSO from $SCAN_URL (failure #$fail_count): $result"
-                if [ $((fail_count % LOG_TO_STDERR_AFTER_FAILURES)) -eq 0 ]; then
+                if [ $((fail_count % log_to_stderr_after_failures)) -eq 0 ]; then
                   echo "ERROR: $fail_count consecutive failures fetching DSO from $SCAN_URL" 1>&2
                 fi
               fi;
@@ -89,6 +89,10 @@ spec:
         env:
           - name: SCAN_URL
             value: {{ .Values.runtimeDetails.scanUrl | quote }}
+          {{ - if .Values.runtimeDetails.logToStderrAfterFailures }}
+          - name: LOG_TO_STDERR_AFTER_FAILURES
+            value: {{ .Values.runtimeDetails.logToStderrAfterFailures | quote }}
+          {{- end }}
         volumeMounts:
         - name: content-runtime
           mountPath: /usr/share/nginx/html/runtime
--- a/cluster/helm/splice-info/templates/deployment.yaml
+++ b/cluster/helm/splice-info/templates/deployment.yaml
@@ -48,7 +48,7 @@ spec:
         command: ["/bin/sh", "-c"]
         args:
           - |
-            set -eu;
+            set -u;
 
             period=60;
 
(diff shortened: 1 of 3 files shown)


cluster/helm/splice-info/templates/deployment.yaml
```diff
@@ -48,7 +48,7 @@ spec:
         command: ["/bin/sh", "-c"]
         args:
           - |
-            set -eu;
+            set -u;
```
moritzkiefer-da: I don't think I understand why you need to remove this. set -e has no effect in an `if` anyway. So which errors are you ignoring? Consider more selectively ignoring them with something like `|| true`

cluster/helm/splice-info/templates/deployment.yaml
```diff
@@ -60,24 +60,32 @@ spec:
 
             echo "{\"dso\": \"/$dso_json_path\"}" > "$runtime_index_file";
 
+            fail_count=0
+            # prevent spamming alerts by logging only after the issues are persistent
+            LOG_TO_STDERR_AFTER_FAILURES=100
```
moritzkiefer-da: how about letting this be overwritable with flag?

moritzkiefer-da (approved):
thanks!

cluster/helm/splice-info/templates/deployment.yaml
OriolMunoz-da: I don't anymore, fixed it and forgot to restore the flag... good catch, thanks!

cluster/helm/splice-info/templates/deployment.yaml
OriolMunoz-da: we already get the SCAN_URL env vars... so I could:
`log_to_stderr_after_failures=${LOG_TO_STDERR_AFTER_FAILURES:-100}`
(or whatever the syntax for defaults is)
wdyt?

moritzkiefer-da: sounds ok provided that env var is actually propagated cross the helm chart. Otherwise you might need to make it a helm var.

cluster/helm/splice-info/templates/deployment.yaml
```diff
@@ -60,27 +60,39 @@ spec:
 
             echo "{\"dso\": \"/$dso_json_path\"}" > "$runtime_index_file";
 
+            fail_count=0
+            # prevent spamming alerts by logging only after the issues are persistent
+            log_to_stderr_after_failures=${LOG_TO_STDERR_AFTER_FAILURES:-100}
```
moritzkiefer-da: consider moving the default to helm values instead. Currently it's not very clear what happens if you don't set it whereas if you slap it in values-templates you can probably just make it mandatory.

PR #2096, opened 2025-09-02
Fix info updater missing jq

OriolMunoz-da:
...by just not using jq

Fixes https://github.com/DACH-NY/cn-test-failures/issues/5469


moritzkiefer-da (approved):
thanks!

PR #2105, opened 2025-09-03
[ci] Apply the reward coupon batch limit to the actual contract not the rouds

nicu-da:
will create an issue to fix it properly and add tests for coverage

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
@@ -529,7 +529,10 @@ class DbSvDsoStore(
       } yield applyLimit(opName, totalCouponsLimit, result).map { case (round, batch) =>
         RoundBatch(
           round,
-          batch.map(cid => companionClass.toContractId(companion, cid.contractId)).toSeq,
+          batch
+            .take(totalCouponsLimit.limit)
+            .map(cid => companionClass.toContractId(companion, cid.contractId))
+            .toSeq,
         )
       }
     }


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
```diff
@@ -529,7 +529,10 @@ class DbSvDsoStore(
       } yield applyLimit(opName, totalCouponsLimit, result).map { case (round, batch) =>
         RoundBatch(
           round,
-          batch.map(cid => companionClass.toContractId(companion, cid.contractId)).toSeq,
+          batch
+            .take(totalCouponsLimit.limit)
```
moritzkiefer-da, line 533: Why not filter this out directly in SQL?

moritzkiefer-da: Also can we fix the limit that we do use in SQL? It seems off now. The name `totalCouponsLimit` also seems off, it seems like a `batchCouponsLimit` or something like that.

moritzkiefer-da (approved):
thanks!

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
nicu-da: >  Why not filter this out directly in SQL?

Would need a more complex query change and I just wanna fix this quickly for now to fix cilr. 

>  Also can we fix the limit that we do use in SQL? It seems off now. The name totalCouponsLimit also seems off, it seems like a batchCouponsLimit or something like that.

Fix it in what sense? I've created https://github.com/hyperledger-labs/splice/issues/2106 to fix it properly through sql and adjust the limits but this was broken from before so at least now we respect the actual batch limit.

moritzkiefer-da: > Would need a more complex query change and I just wanna fix this quickly for now to fix cilr.

wfm i guess it's not worse.

> Fix it in what sense? I've created https://github.com/hyperledger-labs/splice/issues/2106 to fix it properly through sql and adjust the limits but this was broken from before so at least now we respect the actual batch limit.

I think we really just need two  parameters. One for the number of batches and one for the batch size.

nicu-da: agree, all good if I just add that to https://github.com/hyperledger-labs/splice/issues/2106?

moritzkiefer-da: wfm

PR #2108, opened 2025-09-03
[ci] Add a simple way to ignore certain parties for expire rewards triggers

nicu-da:
most basic fix for https://github.com/DACH-NY/canton-network-internal/issues/1545

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/resources/include/svs/_sv.conf
+++ b/apps/app/src/test/resources/include/svs/_sv.conf
@@ -42,7 +42,7 @@
       "org.lfdecentralizedtrust.splice.sv.automation.delegatebased.ExpiredAnsSubscriptionTrigger"
     ]
     # example of configuration to ignore contracts for expiry
-    ignored-expired-rewards-contract-ids = [ "00653ace6cd7e7577576bab4b801daf7ee65e1b0a1a32e49bb04e1c6d8f8d6f16fca1112204721fb06877cab5f76ce7f2cb528d86d0c82064f1a17aaf56f9452fb2dd8de68" ]
+    ignored-expired-rewards-party-ids = [ "test-party::1220b3eeb21b02e14945e419c5d9e986ce8102171c50e1444010ab054e11eba262c9" ]
   }
   # Ensure we get frequent acks to handle our aggressive pruning intervals in tests
   time-tracker-min-observation-duration = 10s
--- a/apps/app/src/test/resources/include/svs/_sv.conf
+++ b/apps/app/src/test/resources/include/svs/_sv.conf
@@ -41,6 +41,8 @@
       "org.lfdecentralizedtrust.splice.sv.automation.delegatebased.ExpiredLockedAmuletTrigger",
       "org.lfdecentralizedtrust.splice.sv.automation.delegatebased.ExpiredAnsSubscriptionTrigger"
     ]
+    # example of configuration to ignore contracts for expiry
+    ignored-expired-rewards-contract-ids = [ "00653ace6cd7e7577576bab4b801daf7ee65e1b0a1a32e49bb04e1c6d8f8d6f16fca1112204721fb06877cab5f76ce7f2cb528d86d0c82064f1a17aaf56f9452fb2dd8de68" ]
   }
   # Ensure we get frequent acks to handle our aggressive pruning intervals in tests
   time-tracker-min-observation-duration = 10s
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
@@ -442,7 +442,7 @@ class DbSvDsoStore(
   override def listAppRewardCouponsGroupedByRound(
       domain: SynchronizerId,
       totalCouponsLimit: Limit,
-      ignoredContracts: Set[ContractId[?]],
+      ignoredParties: Set[PartyId],
   )(implicit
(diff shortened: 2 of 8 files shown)


apps/app/src/test/resources/include/svs/_sv.conf
```diff
@@ -41,6 +41,8 @@
       "org.lfdecentralizedtrust.splice.sv.automation.delegatebased.ExpiredLockedAmuletTrigger",
       "org.lfdecentralizedtrust.splice.sv.automation.delegatebased.ExpiredAnsSubscriptionTrigger"
     ]
+    # example of configuration to ignore contracts for expiry
```
moritzkiefer-da, line 44: what do you think of making the config a mapping from trigger name => list of contract ids? I suspect we'll need that for more triggers.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
```diff
@@ -526,15 +537,18 @@ class DbSvDsoStore(
...
-          round,
-          batch
-            .take(totalCouponsLimit.limit)
-            .map(cid => companionClass.toContractId(companion, cid.contractId))
-            .toSeq,
-        )
-      }
+      } yield applyLimit(opName, totalCouponsLimit, result)
+        .map { case (round, batch) =>
+          RoundBatch(
```
moritzkiefer-da, line 545: Doesn't this have exactly the problem I mentioned where you can end up fetching only an empty list why I wanted to move it into the DB queries? Picking some extreme values to make the example clearer:

Let's say totalCouponsLimit = 1. Your DB query fetches one batch. Let.s assume that one batch only contains ignored contracts, you then filter out the ignored contracts so now you have no batch. And you just repeat that over and over.

apps/app/src/test/resources/include/svs/_sv.conf
nicu-da: Do you have in mind a way to easily generalize this? The way we use contract ids in the input of the triggers is really trigger specific. The way I see it we would need to do a big refactoring of the triggers to make it generic.

moritzkiefer-da: I wouldn't yet make it generic, just setup the config format for it to make it easy to extend later.

moritzkiefer-da: if we do make it specific to expiry filtering by party might almost work better than by contract id.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
nicu-da: The limit is for the number of rounds we're fetching. Indeed this is a valid scenario. We could in theory then just increase the batch size and we should fix it as no new contracts that have this issue should be created. 

Let me just move the filtering into the query though, only issue I'm seeing is if we're somehow exceeding some limits of how many params can we send in the query.

moritzkiefer-da: > Let me just move the filtering into the query though, only issue I'm seeing is if we're somehow exceeding some limits of how many params can we send in the query.

Might be another good reason to make it per party rather than contract id, then you need to specify significantly fewer params.

nicu-da: That sounds like a better solution, easier to source the data as well. Let me just quickly change to ignore party

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
```diff
@@ -520,21 +531,25 @@ class DbSvDsoStore(
                   and assigned_domain = $domain
                   and reward_party is not null -- otherwise index is not used
                   and reward_round is not null -- otherwise index is not used
+                  and reward_party NOT IN (${ignoredParties
```
moritzkiefer-da, line 534: How does this work for rewards with multiple parties, e.g., an app reward coupon with a beneficiary?

moritzkiefer-da:
Let's add release notes for this

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
moritzkiefer-da: I guess it's ok but you have to overapproximate it so if you have a problematic beneficiary you can't specify that beneficiary but need to specify the provider (or the other way around didn't actually check).

nicu-da: we would probably need to specify all of them. sounds like ideally we would have both party and contract ignore but do you see scenarios where we would have that many parties to be a problem? 
i think we either go with this and try on devnet or postpone to the next release

moritzkiefer-da: should be fine to go with this

nicu-da:
@moritzkiefer-da added release notes

moritzkiefer-da (approved):
:shipit

PR #2111, opened 2025-09-03
[ci] One query for hasArchived.

ray-roestenburg-da:
Fixes #838. [Query plan on cilr clone](https://github.com/hyperledger-labs/splice/issues/838#issuecomment-3249563054)
Deleted [sv-1-cn-apps-pg-c8f4841-clone-ray](https://console.cloud.google.com/sql/instances/sv-1-cn-apps-pg-c8f4841-clone-ray/overview?project=da-cn-ci-2)
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
@@ -232,7 +232,7 @@ final class DbMultiDomainAcsStore[TXE](
   def containsArchived(ids: Seq[ContractId[?]])(implicit
       traceContext: TraceContext
   ): Future[Boolean] = waitUntilAcsIngested {
-    if (ids.isEmpty) Future.successful(true)
+    if (ids.isEmpty) Future.successful(false)
     else {
       val contractIds = inClause(ids)
       val expectedCount = ids.size
--- a/apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStoreTest.scala
+++ b/apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStoreTest.scala
@@ -543,7 +543,7 @@ class DbMultiDomainAcsStoreTest
           .containsArchived(Seq(coupon2.contractId, coupon3.contractId))
           .futureValue shouldBe true
         _ = store.containsArchived(Seq(coupon3.contractId)).futureValue shouldBe false
-        _ = store.containsArchived(Seq()).futureValue shouldBe true
+        _ = store.containsArchived(Seq()).futureValue shouldBe false
       } yield succeed
     }
   }


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -231,9 +231,27 @@ final class DbMultiDomainAcsStore[TXE](
...
+    else {
+      storage.query(
+        sql"""
+        select not exists (
+           select acs.contract_id
+           from #$acsTableName acs
+           where acs.store_id = $acsStoreId
+           and acs.migration_id = $domainMigrationId
+           and acs.contract_id in (#${ids
+            .map(id => lengthLimited(s"'${id.contractId}'"))
```
ray-roestenburg-da: This is not great, I'll check if there is a better way with Slick to do this instead of raw, manual interpolation.

ray-roestenburg-da: @nicu-da @rautenrieth-da  let me know how safe you think this is. Chances of injection seem quite low, since hasArchived accepts `Seq[ContractId[?]]`, so maybe it's fine?

stephencompall-DA: Can you use `inClause`?

https://github.com/hyperledger-labs/splice/blob/03bbae02019515d3b307f46a7213752001f0f59b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsQueries.scala#L259-L260

https://github.com/hyperledger-labs/splice/blob/a82fa0d82c19d12d195e40575525022160fbcfc6/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala#L217-L219

ray-roestenburg-da: Oh nice, I probably can! It is really safer though? I guess this binds properly through SetParameter?

stephencompall-DA: Exactly.

ray-roestenburg-da: Thanks @stephencompall-DA ! Fixed.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -231,9 +231,27 @@ final class DbMultiDomainAcsStore[TXE](
...
+
+      storage.query(
+        (sql"""
+        select not exists (
+           select acs.contract_id
+           from #$acsTableName acs
+           where acs.store_id = $acsStoreId
+           and acs.migration_id = $domainMigrationId
+           and acs.contract_id in """ ++ contractIds ++ sql"""
+         )
```
stephencompall-DA: I don't think the before and after logic match here. Given the contract IDs {x, y}, where x exists and y does not, the prior implementation returns true (sequence returns {defined, undefined}, isEmpty returns {false, true}, there exists such a value) whereas the SQL here returns false (there is a row such that contract ID (x) in {x, y}).

The scaladoc doesn't match either version

https://github.com/hyperledger-labs/splice/blob/5f34045aa841e6ba7ddf7b298e68955d641eb562/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala#L106

but based on the two usage points I think the prior behavior is probably what we want

https://github.com/hyperledger-labs/splice/blob/a0f64a297a7979075e51216be95c4cb8ebb3533e/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpireRewardCouponsTrigger.scala#L58-L61

https://github.com/hyperledger-labs/splice/blob/6da48e826c46b88dfb3e8e2bae639bc60d12d59e/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedRewardsTrigger.scala#L55-L58

We could match the prior implementation by doing `select count(*)` and comparing to `ids.size` instead.

ray-roestenburg-da: isnt it seq[option] -> option[seq] and then just checking the seq is empty? (on phone)

ray-roestenburg-da: I'll check in any case tomorrow, thanks

stephencompall-DA: No, we sequence the seq of future to future of seq; if we only checked the seq is empty there would be no point to the `sequence` because traversal never alters the spinal structure of the traversable functor in question (e.g. if you traverse a seq, the seq under the applicative has the same length; a traversed option does not change defined-ness; &c). Instead we search the seq for an undefined option.

ray-roestenburg-da: I thought lookup was Future[Option[...]], ok so from Seq[Future[Option] to Future[Seq[Option]], so it's just finding out if any of the contracts cannot be found? I did not expect that logic. will look deeper tomorrow

stephencompall-DA: Yes, and when we traverse we get a Future[Seq[Option[...]]], and we ask the Seq about the existence of a None. The equivalent to the SQL would be if we asked the Seq if _all items_ were None.

ray-roestenburg-da: yeah I got tripped up by the name I guess, it would have made more sense to me for a method `hasArchived(ids)` to only return true if actually all ids were archived / deleted. Now got to figure out what the intent is where it gets used in staleTask in the triggers.. does it really make sense that the trigger is stale if one or more of the ids has been archived (but not necessarily all)? The comment on the super method makes no sense to me as well. Thanks for checking this!

ray-roestenburg-da: Maybe I should have read that as `containsArchived`.. I'll rename it. it will return true if any of the ids has been archived (and I guess it makes sense for the trigger to be stale if at least one of the ids it wants to work on is already archived)

ray-roestenburg-da: @nicu-da would like a double check on this since the comment does not fit the method and the name likely has been changed over time.

moritzkiefer-da: We only use it in staleness check from a quick grep and there you do want to consider a task a stale if _any_ of the contracts is archived as your transaction will fail in that case.

ray-roestenburg-da: Thanks @moritzkiefer-da ! I've changed the query

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -229,11 +229,31 @@ final class DbMultiDomainAcsStore[TXE](
...
 
-  def hasArchived(ids: Seq[ContractId[?]])(implicit
-      traceContext: TraceContext
-  ): Future[Boolean] =
-    // TODO(#838): implement this as a single DB query
-    Future.sequence(ids.map(lookupContractStateById)).map(_.exists(_.isEmpty))
+  def containsArchived(ids: Seq[ContractId[?]])(implicit
+                                                traceContext: TraceContext
+  ): Future[Boolean] = waitUntilAcsIngested {
+    if (ids.isEmpty) Future.successful(true)
```
moritzkiefer-da: exists is false if the list is empty, now you return true. Why change this?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -229,11 +229,31 @@ final class DbMultiDomainAcsStore[TXE](
...
+    else {
+      val contractIds = inClause(ids)
+      val expectedCount = ids.size
+      storage.query(
+        (sql"""
+        select count(1) from (
+           select acs.contract_id
```
moritzkiefer-da:
```suggestion
           select count(1)
```

Does this not work? I don't understand why you need a nested select.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -229,11 +229,31 @@ final class DbMultiDomainAcsStore[TXE](
...
+           from #$acsTableName acs
+           where acs.store_id = $acsStoreId
+           and acs.migration_id = $domainMigrationId
+           and acs.contract_id in """ ++ contractIds ++ sql"""
```
moritzkiefer-da: this is not gonna work for very large numbers of contract ids. Have you double checked that the use sites are ok with this?

moritzkiefer-da (approved):
thanks, maybe double check the query plan the one you linked seems to from before you made some changes

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
ray-roestenburg-da: right, will fix

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
ray-roestenburg-da: yeah that will work too

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
ray-roestenburg-da: @nicu-da  mentioned that these were limited to 20-100, though maybe I misunderstood and he meant that we can limit them to that number (I think he patched it though). Reading about very large numbers of in arguments, it is still fine even at 1k-10k items in the in clause using postgres (of course the query string gets huge at some point). Another option is to use a temp table but I'm assuming we don't need that for now. If we remove the limit we'll have to do that, insert into a temp table without log, join on that with the acs table, seemed overkill for now. Let me know if you want that instead anyway already, or as follow up PR.

ray-roestenburg-da:
Updated query plan for `select count(1) ...`
```
[
  {
    "QUERY PLAN": "Aggregate  (cost=104.28..104.29 rows=1 width=8) (actual time=0.414..0.415 rows=1 loops=1)"
  },
  {
    "QUERY PLAN": "  ->  Index Only Scan using scan_acs_store_store_id_migration_id_contract_id_idx on scan_acs_store acs  (cost=0.55..104.28 rows=1 width=0) (actual time=0.237..0.397 rows=20 loops=1)"
  },
  {
    "QUERY PLAN": "        Index Cond: ((store_id = 1) AND (migration_id = 0) AND (contract_id = ANY ('{archived1,archived2,00002e5da0b7ad26639568e13fa197c0e4da7724cf71c3867e6e79f86fe9dd0ce6ca1012201c49974bd1ad76967454a046566ff1368f95c1d5076231a65b878288997a4ffc,000061607ecd1d3d74446191c89b13f919151f6d0349a7803ca9b1eef9c011d207ca101220efeda7862f80bfa09087d9aa9a365814b930d9db0338a54d46f12973dbd6b0dc,00008ac31b0f9304d41ab276863be83d82938d155f26c1adfef57d31461aca9810ca101220cb4ab67ef8e5ed79a64442b8ec85529ca8ff9eabc7979345fd3bf8e5a1e0e7b8,0000f6d56f390b98a0c4e5b86455fca5914ceb3a45d3aaafce5dd7151168784d75ca1012205446af71a78b775610095db555b00ba07a3ed965c4147d9780c2d2e4baf3b7a3,00022df9d8dd50c299f604bbd7258d67f9ffcf170a6f68918c4f2191763c5b58a4ca1012209f63b63506144433d84ae8c7e3d86381d64b664d7a8dd4b20406fec2892150b3,00027b56386bebe098d7e8734659d7d4bc549522c34813ffe54968e741be85cbe3ca10122046070e6fa582f25e0e900f2405463059b662caf955d78911b1febb772675cf43,00030f06c27c4b111fb3b4c68967072df38a82f3dba7a97d277be5c26eee721371ca101220b4a4bbb13066a6b36dffa0fef0328f982ee8b10c9e633dbb6075a57f74e31be2,000392bb1e6d6b44465757cbda101197d7056f76e0f23474d2eeb5b7657886e40eca101220a47642dc986b5c3a7981719f925e524242d516fb6059f34fde15da76891ee1bf,00042d89a60615938a6e57155ad4d679ae7c53d109ad66d8f83f5e0c4b2c0fa8c6ca101220a7633d59ed6afe1bf5a0d7af63d016760eef92363715ce70993853abf53809f2,00044e87675100b1862119e1deb64b1e22e69202e0d730bd54ffd4237cfb6914c0ca101220d3e84de2d76996dac6ad12938194c99a4998812e680efff788f6b03cba299223,0004811b10301b3f44dd16122fa7ebbbaa57b7d37bcc6ed343df3ff1c91d43bd3bca101220f4d738cb4b4fb75c23a624fc3bc999d7fec709cbaa4fb6fbac56408161f4a150,00049a0f285daa9474a9fc39708248b1b23577facde3fb1d660c37fa00ae75ac5cca101220145a4c236d6b1232bf6c1b70d3b3655d9bc6ba31b58bcef0860daf6bd1322940,0004ac58b417bd5dd8d9f3c9aa06ee4e3f24c23d9a4e6bc569c3037b4d63107057ca1012202c6097d0af02a241c18f5982427ef5cc142a20e2bcb45fd0d1536a1b3634b174,0004eff59249bf64728f08f037516cd3aa262bd876912ce5efee67616aa921622bca1012200bfb35555e68533d6b22c74bf6be6cc9a90348c3223d0163a75c6a2a73baa6eb,0005063459ff263492a436ad55119f085eabf5dbf8fe56c46968e6c338de29f7e9ca101220ec7d34f3819ea072b16dda92c8201ca3a15e81f37d1f236783a5d5a3c2da5807,000521f71d89df8d25df809b24084c725ecd1de29686dd761da8fa424d02d28496ca101220af7f5a73262287c7dc5a478b1aad97e47c3ab9939f9fefcfa62d4ebd5142f6e1,0005424c620e75312ec2939cccaa130637d4352cbdc7533bbcabd5a3384d052398ca101220daca7cfd5e28258e60f25962149cea8a494a24e63f60b793f6c1726d06d14efe,000548fd30ad8058cd1712ba7716260e6329461dd68c8585f1d8a03c1659abee37ca101220451e1a6527ad3de46ef2ca8abd6f93ffb03e663c4843b710ebcf42fbf2957f82,00058a1f7b8f5df6fef107ced38c0c2526680632b9144b444a555bfec905b5e8b2ca101220534e9c3cb9377e27996e9f861ef7e66a9a3386e4de28073229107e785510f546,0005a30647ebf899d23011ed8525c88147ffe85f5c126f73d2d43d502563bf265bca1012204557624e2f7aef09c91ee879872151ed7ea9cd14d45fd9b44ca444b69a40bf29}'::text[])))"
  },
  {
    "QUERY PLAN": "        Heap Fetches: 0"
  },
  {
    "QUERY PLAN": "Planning Time: 4.440 ms"
  },
  {
    "QUERY PLAN": "Execution Time: 0.530 ms"
  }
]
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
nicu-da: In the use I've seen we were limiting it to 20 now. But we can add a safety check that if we have really high batches then we split them.

ray-roestenburg-da: I've just added https://github.com/hyperledger-labs/splice/issues/2122 in case we later on want to do that as tech debt. For small batches you likely incur some overhead with that approach

PR #2112, opened 2025-09-03
add BQ monthly burn stat

isegall-da:
Can't have an integration test, as we use simulated time, and deducting 30 days from that goes negative and confuses sql.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/canton-network/src/bigQuery_functions.ts
+++ b/cluster/pulumi/canton-network/src/bigQuery_functions.ts
@@ -151,6 +151,22 @@ const up_to_time = new BQScalarFunction(
   `
 );
 
+const migration_id_at_time = new BQScalarFunction(
+  'migration_id_at_time',
+  [new BQFunctionArgument('as_of_record_time', TIMESTAMP)],
+  INT64,
+  `
+    -- Given a record time, find the latest migration ID that was active at that time. Takes the lowest ID that has updates
+    -- after the given time, therefore if the timestamp is during a migration, it will return the older migration ID.
+    (SELECT
+      MIN(migration_id)
+    FROM
+      ((SELECT record_time, migration_id FROM \`$$SCAN_DATASET$$.scan_sv_1_update_history_creates\`) UNION DISTINCT
+      (SELECT record_time, migration_id FROM \`$$SCAN_DATASET$$.scan_sv_1_update_history_exercises\`))
+    WHERE record_time > UNIX_MICROS(as_of_record_time))
+  `
+);
+
 const sum_bignumeric_acs = new BQScalarFunction(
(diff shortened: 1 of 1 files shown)


cluster/pulumi/canton-network/src/bigQuery_functions.ts
```diff
@@ -528,6 +529,7 @@ const all_stats = new BQTableFunction(
       \`$$FUNCTIONS_DATASET$$.minted\`(as_of_record_time, migration_id) as minted,
       \`$$FUNCTIONS_DATASET$$.minted\`(as_of_record_time, migration_id) + \`$$FUNCTIONS_DATASET$$.unminted\`(as_of_record_time, migration_id) as allowed_mint,
       \`$$FUNCTIONS_DATASET$$.burned\`(as_of_record_time, migration_id) as burned,
+      \`$$FUNCTIONS_DATASET$$.burned\`(as_of_record_time, migration_id) - \`$$FUNCTIONS_DATASET$$.burned\`(TIMESTAMP_SUB(as_of_record_time, INTERVAL 30 DAY), migration_id) as monthly_burn,
```
stephencompall-DA: This will work but is probably costlier than adding a lower bound to the `burned` function itself. It also assumes that there is no hard migration in the 30 day window.

isegall-da: > This will work but is probably costlier than adding a lower bound to the burned function itself

True, should probably be easy to add a lower bound there.

> assumes that there is no hard migration in the 30 day window.

I don't think so. the migration_id is always just an upper bound, a too-high migration ID should usually work, no?

stephencompall-DA: > the migration_id is always just an upper bound, a too-high migration ID should usually work, no?

If the migration ID is too high, then the record time will be effectively ignored. For example, if your migration happens 13 days into the window, then the subtractend here will not be 30 days prior, but 17 days prior.

isegall-da: Argh, you're right. Might need to guess the migration ID from the record time, which will be correct unless it falls exactly on the time of the migration where it might be ambiguous.

stephencompall-DA: I think you can always choose a consistent model that has funny behavior at the margins. I think this calls for a materialized view: the earliest and latest record time for each migration ID. Then if a requested record time falls in that window, you pick the ceiling, floor, or round halfway through, either way you get a "jump" of aggregated values that matches the length of the migration time (i.e. last record time on old mgid - first record time on new mgid). I think that's a fair compromise.

PR #2128, opened 2025-09-04
Prevent txlog backfill from including ACS imports

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/pull/2041#discussion_r2321926849

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
@@ -5,14 +5,14 @@ package org.lfdecentralizedtrust.splice.store.db
 
 import cats.data.{NonEmptyList, NonEmptyVector, OptionT}
 import org.apache.pekko.NotUsed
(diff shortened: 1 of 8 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -594,47 +594,58 @@ final class DbMultiDomainAcsStore[TXE](
...
-              .transactionally,
-            "destinationHistory.insert",
-          )
-        } yield DestinationHistory.InsertResult(
-          backfilledUpdates = trees.size.toLong,
-          backfilledEvents =
-            trees.foldLeft(0L)((sum, tree) => sum + tree.getEventsById.size().toLong),
-          lastBackfilledRecordTime = CantonTimestamp.assertFromInstant(nonEmpty.last.getRecordTime),
-        )
+            )
```
OriolMunoz-da, line 601: here be dragons, I don't understand why the previous was throwing an exception

rautenrieth-da (approved):
Makes sense, see my comment at https://github.com/hyperledger-labs/splice/pull/2041#issuecomment-3859182734

PR #2168, opened 2025-09-08
[static] always use computed migration IDs

isegall-da:
The BQ user typically does not care about migration IDs, they just want to provide a timestamp, and have all ambiguity of what that means handles behind the scenes in a consistent manner. This PR does that: it pushes all references to migration IDs to the bottom, i.e. to "in_time_window". As long as all time comparisons are done only there, then everything remains consistent and we don't need to worry about migration IDs all over the code.

(Ran the integration test locally)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/canton-network/src/bigQuery_functions.ts
+++ b/cluster/pulumi/canton-network/src/bigQuery_functions.ts
@@ -118,13 +118,13 @@ const migration_id_at_time = new BQScalarFunction(
       (SELECT
         MIN(migration_id)
       FROM
-        ((SELECT record_time, migration_id FROM \`$$SCAN_DATASET$$.scan_sv_1_update_history_creates\`) UNION DISTINCT
+        ((SELECT record_time, migration_id FROM \`$$SCAN_DATASET$$.scan_sv_1_update_history_creates\`) UNION
          (SELECT record_time, migration_id FROM \`$$SCAN_DATASET$$.scan_sv_1_update_history_exercises\`))
       WHERE record_time > UNIX_MICROS(as_of_record_time)),
       -- If none exists, return the migration ID of the last update.
       (SELECT migration_id FROM
         (
-          (SELECT record_time, migration_id FROM \`$$SCAN_DATASET$$.scan_sv_1_update_history_creates\`) UNION DISTINCT
+          (SELECT record_time, migration_id FROM \`$$SCAN_DATASET$$.scan_sv_1_update_history_creates\`) UNION
           (SELECT record_time, migration_id FROM \`$$SCAN_DATASET$$.scan_sv_1_update_history_exercises\`)
         ) ORDER BY record_time DESC LIMIT 1)
     )


cluster/pulumi/canton-network/src/bigQuery_functions.ts
```diff
@@ -107,33 +104,53 @@ const daml_record_numeric = new BQScalarFunction(
...
+    -- If no updates exist after the given time, returns the migration id of the last update.
+    IFNULL
+    ((SELECT
+      MIN(migration_id)
+    FROM
+      ((SELECT record_time, migration_id FROM \`$$SCAN_DATASET$$.scan_sv_1_update_history_creates\`) UNION DISTINCT
+      (SELECT record_time, migration_id FROM \`$$SCAN_DATASET$$.scan_sv_1_update_history_exercises\`))
+    WHERE record_time > UNIX_MICROS(as_of_record_time)),
+      (SELECT migration_id FROM
+        ((SELECT record_time, migration_id FROM \`$$SCAN_DATASET$$.scan_sv_1_update_history_creates\`) UNION DISTINCT
```
stephencompall-DA: Does `DISTINCT` help here? I think it can only hurt.

cluster/pulumi/canton-network/src/bigQuery_functions.ts
```diff
@@ -402,9 +390,8 @@ const burned = new BQScalarFunction(
                   AND e.template_id_module_name = 'Splice.Wallet.Subscriptions'
                   AND c.template_id_module_name = 'Splice.Amulet'
                   AND c.template_id_entity_name = 'Amulet'
-                  AND \`$$FUNCTIONS_DATASET$$.up_to_time\`(as_of_record_time, migration_id_arg,
-                        e.record_time, e.migration_id)
-                  AND c.record_time != -62135596800000000)))
```
stephencompall-DA, line 407: This check is not redundant; it checks `c` whereas the function call checks `e`. Since this is a cross-join leaving it out could multiply the value.

cluster/pulumi/canton-network/src/bigQuery_functions.ts
isegall-da: Hmmm... Seemed sensible not to have duplicates, no? 
But doesn't seem to matter too much either way, if you think it's safer without DISTINCT, I'm happy removing it too. 
Also the one four lines above?

cluster/pulumi/canton-network/src/bigQuery_functions.ts
isegall-da: Good catch, thanks

ray-roestenburg-da (approved):
Nice, thanks!

cluster/pulumi/canton-network/src/bigQuery_functions.ts
ray-roestenburg-da: nitpick, Maybe the formatting could be improved a little

ray-roestenburg-da: I don't understand @stephencompall-DA's issue with `union distinct`?

isegall-da: Tried addressing the readability. @stephencompall-DA, care to elaborate what problem you see with the `DISTINCT` ?

stephencompall-DA: It just adds extra processing as far as I can see, and doesn't make the query downstream of it any cheaper.

ray-roestenburg-da: There should not be a huge number of migrations so I would be surprised if that adds substantial processing, but fair point that it's not relevant for `select min(id)`, thanks

stephencompall-DA: This distincts (record_time, migration_id) tuples, not migration IDs alone.

PR #2172, opened 2025-09-09
[ci] Add ignore for non failures error categories for circuit breakers

nicu-da:
fixes https://github.com/hyperledger-labs/splice/issues/2019

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/project/ignore-patterns/canton_network_test_log.ignore.txt
+++ b/project/ignore-patterns/canton_network_test_log.ignore.txt
@@ -124,5 +124,7 @@ GENERIC_CONFIG_ERROR.*SpliceConfigTest
 # ExecutorServiceMetrics closes the gauges twice through shutdown and shutownNow in cases where shutdown times out
 .*InstrumentDescriptor.*name=daml.executor.* has called close\(\) multiple times
 
+Circuit breaker .* tripped after .* failures.*SpliceCircuitBreakerTest
+
 
 # Make sure to have a trailing newline


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
```diff
@@ -3,18 +3,65 @@
...
+    InvalidGivenCurrentSystemStateResourceExists,
+    InvalidGivenCurrentSystemStateResourceMissing,
+    InvalidGivenCurrentSystemStateSeekAfterEnd,
+  )
+
   def withCircuitBreaker[T](body: => Future[T]): Future[T] =
-    underlying.withCircuitBreaker(body)
+    underlying.withCircuitBreaker[T](
+      body,
+      isResultIgnored,
```
moritzkiefer-da: Does this have the right behavior? Afaict from reading the pekko code, this does not ignore those responses but considers them a success. So if you tripped the circuit breaker and now you get a ContractNotFound, you will close it again. It seems like the behavior we want is to ignore those errors no?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
```diff
@@ -3,18 +3,65 @@
...
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
 
 class SpliceCircuitBreaker(underlying: CircuitBreaker) {
 
+  private val errorCategoriesToIgnore: Set[ErrorCategory] = Set(
+    InvalidIndependentOfSystemState,
+    InvalidGivenCurrentSystemStateOther,
+    InvalidGivenCurrentSystemStateResourceExists,
```
moritzkiefer-da, line 32: Why is it save to ignore those? It's not obvious to me for example that a participant confirmation timeout really does not hit that.

nicu-da: Based on the description in the canton error category. 
Participant timeouts should be part of the following category:

```
  @Description(
    """The request could not be processed due to shared processing resources
                 |(e.g. locks or rate limits that replenish quickly) being occupied.
                 |If the resource is known (i.e. locked contract), it will be included as a resource info. (Not known
                 |resource contentions are e.g. overloaded networks where we just observe timeouts, but can’t pin-point the cause)."""
  )
  @RetryStrategy("Retry quickly (indefinitely or limited), but do not retry in load balancer.")
  @Resolution("""Expectation: this is processing-flow level contention that should be handled by
                |retrying the request with appropriate backoff.""")
  case object ContentionOnSharedResources
```


I would avoid adding those details into our code as they will get out of sync with time.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
nicu-da: I this this behavior does make sense though. 
A NOT_FOUND is a successful response, it should not trigger the circuit breaker. If we try it during open times it should also close the circuit breaker, as from the point of view of a client it did get a corresponding response. We should try to keep the breaker dumb imo and not try to make it too smart and cover too many corner cases.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
moritzkiefer-da: fair enough

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
moritzkiefer-da: > A NOT_FOUND is a successful response

The whole goal of the circuit breaker is to guard against overloads no? A NOT_FOUND gives you no useful information in that regard, your transaction usually fails locally for this so you didn't even make it to the synchronizer. And in fact a very common case to get a NOT_FOUND is if your node is lagging behind and you're using outdated contract ids which is also not something you want to consider a success.

I am definitely in favor of keeping it as dumb as possible but "ignore those errors and treat them neither as success or failure" doesn't seem significantly smarter than "treat them as success" no?

nicu-da: > And in fact a very common case to get a NOT_FOUND is if your node is lagging behind and you're using outdated contract ids which is also not something you want to consider a success.

But don't we pause the triggers during that time? Thus protecting against exactly this type of scenario. 
I'm also arguing against because the design of the pekko triggers have no way to use it in such a scenario, there's always a clear cut failure or success, no inbetween. And it sort of makes sense. We're actually trying to push the logic to check overloads too high in the system, it should be the sequencer that protects against such cases. And in fact with the sequencer circuit breaker if the node is falling behind it will reject requests. 

We're basically trying to decide on the client side what impact a request has on a participant and synchronizer based only on the response code. Not saying it's not possible but I also honestly fail to really see the benefit, if a NOT_FOUND closes the breaker, and submissions start failing it will just open again.  If it takes too long to open it again we can tune the configs.

moritzkiefer-da: > But don't we pause the triggers during that time?

I wouldn't rely on that forever, I'd actually like us to rip out that logic in favor of the circuit breakers given all the issues we have around time proofs. We also only pause triggers and not http endpoints while circuit breakers work for both.

> I'm also arguing against because the design of the pekko triggers have no way to use it in such a scenario, there's always a clear cut failure or success, no inbetween.

The pekko circuit breakers have explicit succeed and fail methods. Can't we write a 2-line replacement for withCircuitBreaker that calls one of those or nothing if it is ignored?

> Not saying it's not possible but I also honestly fail to really see the benefit, if a NOT_FOUND closes the breaker, and submissions start failing it will just open again.

Let's say I have 20 triggers. 10 of those fail with actual failures and cause load on the synchronizer and traffic cost for you. 10 of those fail with random local errors. If I have a single circuit breaker I get to 5min backoff or whatever we configured as the max until the synchronizer is healthy again. With your option, you'll permanently ping pong between closed and open and submit still fairly quickly with very limited backoff. And you can't really tune the configs to guard against as the crucial point is that you keep closing it again.

nicu-da: > I wouldn't rely on that forever, I'd actually like us to rip out that logic in favor of the circuit breakers given all the issues we have around time proofs. We also only pause triggers and not http endpoints while circuit breakers work for both.

Fair point, but the sequencer circuit breaker should be the one responsible for that. It's nice that we have that in the clients as well but that's not where we should move the responsibility for preventing load.  It's a nice to have once we have the breakers in the sequencers (even the participant should be smarter and check for example the overall sequencer connection delay to decide to reject stuff).

> The pekko circuit breakers have explicit succeed and fail methods. Can't we write a 2-line replacement for withCircuitBreaker that calls one of those or nothing if it is ignored?

Not quite that simple. You can call those but then the calls don't go through the circuit breaker infrastructure, so we will not be able to use timeouts in the future, we can't use the circuit breaker telemetry implementation because without going through those calls the callbacks aren't called.

> Let's say I have 20 triggers. 10 of those fail with actual failures and cause load on the synchronizer and traffic cost for you. 10 of those fail with random local errors. If I have a single circuit breaker I get to 5min backoff or whatever we configured as the max until the synchronizer is healthy again. With your option, you'll permanently ping pong between closed and open and submit still fairly quickly with very limited backoff. And you can't really tune the configs to guard against as the crucial point is that you keep closing it again.

You can increase the time it waits for the open call to be checked, but at the same time we get back to the first point.  

I think I can implement the change fairly low cost though, so I will do that.

cocreature: > Fair point, but the sequencer circuit breaker should be the one responsible for that. It's nice that we have that in the clients as well but that's not where we should move the responsibility for preventing load. It's a nice to have once we have the breakers in the sequencers (even the participant should be smarter and check for example the overall sequencer connection delay to decide to reject stuff).

That works if you're trying to protect against overload. It works much less well if you're trying to protect against a node wasting a lot of traffic due to a more node local issue (e.g. lagging behind).

To be clear: I hate all of this and in an ideal world, we wouldn't have any circuit breakers and this would be purely in the participant and sequencer. But afaik the participant side is not even anywhere in the plan yet. So it doesn't seem like we can get rid of this logic anytime soon so a bit of investment into it seems

PR #2185, opened 2025-09-09
Fix `ignored-expired-rewards-party-ids` to ignore beneficiaries

martinflorian-da:
...on `SvRewardCoupon`s

Fixes #2181

Query plan for the change (I ran this on CILR):

```
sv_sv_3=> explain analyze select reward_round, array_agg(contract_id)
  from dso_acs_store
WHERE
  store_id = 7
  AND migration_id = 6
  AND template_id_qualified_name = 'Splice.Amulet:SvRewardCoupon'
  AND assigned_domain = 'global-domain::12209471e1a52edc2995ad347371597a5872f2704cb2cb4bb330a849e7309598259e'
  AND reward_party IS NOT NULL
  AND reward_round IS NOT NULL
  and reward_party not in ('test')
  and create_arguments ->> 'beneficiary' not in ('tost')
GROUP BY
  reward_round
ORDER BY
  reward_round ASC
LIMIT
  1000;
                                                                                                                 QUERY PLAN                                                                                          
                        
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
------------------------
 Limit  (cost=0.55..163.01 rows=50 width=40) (actual time=0.086..0.140 rows=3 loops=1)
   ->  GroupAggregate  (cost=0.55..163.01 rows=50 width=40) (actual time=0.085..0.138 rows=3 loops=1)
         Group Key: reward_round
         ->  Index Scan using dso_acs_store_coupons on dso_acs_store  (cost=0.55..162.13 rows=50 width=150) (actual time=0.047..0.122 rows=36 loops=1)
               Index Cond: ((store_id = 7) AND (migration_id = 6) AND (template_id_qualified_name = 'Splice.Amulet:SvRewardCoupon'::text))
               Filter: ((reward_party <> 'test'::text) AND (assigned_domain = 'global-domain::12209471e1a52edc2995ad347371597a5872f2704cb2cb4bb330a849e7309598259e'::text) AND ((create_arguments ->> 'beneficiary'::
text) <> 'tost'::text))
 Planning Time: 0.430 ms
 Execution Time: 0.180 ms
(8 rows)
```

So it seems like it hits the index for template ID, which should make the performance risks here quite low AFAIU as we don't expect to have a terrible amount SV reward coupons lying around.

### DIFF
--- a/apps/sv/src/test/scala/org/lfdecentralizedtrust/splice/store/db/SvDsoStoreTest.scala
+++ b/apps/sv/src/test/scala/org/lfdecentralizedtrust/splice/store/db/SvDsoStoreTest.scala
@@ -677,7 +677,7 @@ abstract class SvDsoStoreTest extends StoreTest with HasExecutionContext {
       val sv1Closed = (1 to 3).map(_ => svRewardCoupon(round = 2, userParty(4), userParty(1), 1000))
       val sv2Closed = (1 to 3).map(_ => svRewardCoupon(round = 2, userParty(4), userParty(2), 1000))
       val sv3Closed = (1 to 3).map(_ => svRewardCoupon(round = 2, userParty(1), userParty(4), 1000))
-      val sv4Closed = (1 to 3).map(_ => svRewardCoupon(round = 2, userParty(1), userParty(1), 1000))
+      val sv4Closed = (1 to 3).map(_ => svRewardCoupon(round = 2, userParty(1), userParty(2), 1000))
       for {
         store <- mkStore()
         _ <- dummyDomain.create(closedRound)(store.multiDomainAcsStore)


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
```diff
@@ -520,7 +520,16 @@ class DbSvDsoStore(
     val opName = s"list${templateId.getEntityName}GroupedByRound"
     val partyFilter =
       if (ignoredParties.nonEmpty)
-        (sql"and reward_party not in " ++ inClause(ignoredParties)).toActionBuilder
+        companion match {
+          case SvRewardCoupon.COMPANION =>
+            // only SV reward coupons have a beneficiary field
+            (sql"and reward_party not in " ++ inClause(ignoredParties) ++
```
nicu-da, line 526: should this be a and? wouldn't the reward party be the sv so we get back to having to ignore the sv? :thinking:

martinflorian-da: I think yes? Something something boolean logic and negation... (X notin Y) and (Z notin Y) == neither X nor Z should be in Y ... maybe it's too late today and I'm crazy?

I kept `reward_party` here so that we can also still ignore the whole SV if we want to.

nicu-da: RIght, not and not, makes sense

nicu-da (approved):

apps/sv/src/test/scala/org/lfdecentralizedtrust/splice/store/db/SvDsoStoreTest.scala
```diff
@@ -667,11 +667,14 @@ abstract class SvDsoStoreTest extends StoreTest with HasExecutionContext {
       val app2Closed = (1 to 3).map(_ => appRewardCoupon(round = 2, userParty(2)))
 
       val sv1NotClosed =
-        (1 to 3).map(_ => svRewardCoupon(round = 3, userParty(1), userParty(1), 1000))
+        (1 to 3).map(_ => svRewardCoupon(round = 3, userParty(4), userParty(1), 1000))
```
moritzkiefer-da, line 670: do we have a test where both beneficiary and SV match? It should work fine but seems like a case worth testing and it's not completely obvious to me that we are atm.

martinflorian-da: > both beneficiary and SV match

We had that test, I removed it here... Did seem like a super realistic scenario to test but happy to add that, cheap enough.

martinflorian-da: added one where they match and we ignore that party; lmk if you meant something else

moritzkiefer-da: > 

added one where they match and we ignore that party; lmk if you meant something else


I meant one where beneficiary and sv are different and we ignore both

martinflorian-da: > I meant one where beneficiary and sv are different and we ignore both

ah right, that sounds slightly more realistic to test; the result will be unchanged but let me do this

PR #2187, opened 2025-09-09
BQ: split mint by reward type

isegall-da:
(integration test confirmed locally)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


cluster/pulumi/canton-network/src/bigQuery_functions.ts
```diff
@@ -316,6 +336,24 @@ const minted = new BQScalarFunction(
...
+  `
+    (SELECT
+      SUM(mint_amount)
+    FROM
+      UNNEST(ARRAY[
+        \`$$FUNCTIONS_DATASET$$.minted\`(as_of_record_time, migration_id).appRewardAmount,
+        \`$$FUNCTIONS_DATASET$$.minted\`(as_of_record_time, migration_id).validatorRewardAmount,
+        \`$$FUNCTIONS_DATASET$$.minted\`(as_of_record_time, migration_id).svRewardAmount,
+        \`$$FUNCTIONS_DATASET$$.minted\`(as_of_record_time, migration_id).unclaimedActivityRecordAmount
+      ]) AS mint_amount
```
stephencompall-DA, line 352: `minted` isn't a cheap function to call; I wouldn't be surprised if creating a function and calling it with the struct wasn't an easy way to save lots of slot time.

cluster/pulumi/canton-network/src/bigQuery_functions.ts
```diff
@@ -544,8 +584,10 @@ const all_stats = new BQTableFunction(
       \`$$FUNCTIONS_DATASET$$.unlocked\`(as_of_record_time, migration_id) as unlocked,
       \`$$FUNCTIONS_DATASET$$.locked\`(as_of_record_time, migration_id) + \`$$FUNCTIONS_DATASET$$.unlocked\`(as_of_record_time, migration_id) as current_supply_total,
       \`$$FUNCTIONS_DATASET$$.unminted\`(as_of_record_time, migration_id) as unminted,
-      \`$$FUNCTIONS_DATASET$$.minted\`(as_of_record_time, migration_id) as minted,
-      \`$$FUNCTIONS_DATASET$$.minted\`(as_of_record_time, migration_id) + \`$$FUNCTIONS_DATASET$$.unminted\`(as_of_record_time, migration_id) as allowed_mint,
+      \`$$FUNCTIONS_DATASET$$.minted\`(as_of_record_time, migration_id).appRewardAmount as minted_app_rewards,
+      \`$$FUNCTIONS_DATASET$$.minted\`(as_of_record_time, migration_id).validatorRewardAmount as minted_validator_rewards,
+      \`$$FUNCTIONS_DATASET$$.minted\`(as_of_record_time, migration_id).svRewardAmount as minted_sv_rewards,
+      \`$$FUNCTIONS_DATASET$$.minted\`(as_of_record_time, migration_id).unclaimedActivityRecordAmount as minted_unclaimed_activity_records,
```
stephencompall-DA, line 590: Likewise regarding `minted` and using functions because we sadly don't have real `let`.

cluster/pulumi/canton-network/src/bigQuery_functions_types.ts
```diff
@@ -49,6 +49,30 @@ export class BQArray extends BQType {
   }
 }
 
+export class BQStruct extends BQType {
+  private readonly structTypes: { name: string; type: BQType }[];
```
stephencompall-DA, line 53: I'm shocked that this has such clean syntax and you don't have to do really weird things to convince the API that this is OK. Very nice.

cluster/pulumi/canton-network/src/bigQuery_functions.ts
isegall-da: I was kinda hoping that BQ is smart enough to call it only once here, and do the caching for me. Is that too much to hope for?
Also, note that in the followup pr, I always call `minted` with a start_time of 24 hours earlier, so it becomes much cheaper.

isegall-da: (And I'm also getting rid of total_minted which I concluded doesn't make sense to have. We can always sum the four if needed elsewhere)

cluster/pulumi/canton-network/src/bigQuery_functions.ts
isegall-da: Likewise re hoping for magic optimizations, and for only 24-hour windows (but not for deleting this ;)

isegall-da: I believe I did try to make `minted` return a table instead of a struct, then calling it in the `FROM` section, and here only breaking it apart, but that didn't work. Do you have another idea for "using functions" ?

cluster/pulumi/canton-network/src/bigQuery_functions.ts
stephencompall-DA: I think one possible explanation for the earlier slot time explosion was exceeding some complexity limit that broke optimization; the exact nature of what breaks the barrier doesn't matter so much as it becomes the final +1.

There's also Google's optimization [advice](https://cloud.google.com/bigquery/docs/best-practices-costs#materialize_query_results_in_stages), which vaguely suggests it generally isn't that smart:

> If you create a large, multi-stage query, each time you run it, BigQuery reads all the data that is required by the query. You are billed for all the data that is read each time the query is run.

cluster/pulumi/canton-network/src/bigQuery_functions.ts
stephencompall-DA: If a function can't return a relation, maybe we have to consider stored procedures instead?

cluster/pulumi/canton-network/src/bigQuery_functions.ts
isegall-da: Yeah, the question is which dimension of "complexity" risks that. Any alternative to this that I can come up with seems in some way more complex than just calling the function four times.

cluster/pulumi/canton-network/src/bigQuery_functions.ts
isegall-da: Perhaps, but I propose we just run this on a limited scale, e.g. last 30 days, once, and see where we're standing in terms of costs. I expect that the 24 hour window restriction would reduce costs enough for this not to be worth investing in, nor making harder to understand.

PR #1907, opened 2025-08-13
Sketch: token API proxy

meiersi-da:
Sketch for a solution of #1901 

[no ci]

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #2194, opened 2025-09-10
Properly test and package featured app proxy for exchange integrations

meiersi-da:
WIP: please do not review yet

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


daml/splice-util-featured-app-proxies/daml/Splice/Util/FeaturedApp/DelegateProxy.daml
```diff
@@ -0,0 +1,116 @@
...
+
+-- | A proxy for a delegate to create featured app markers jointly with using token standard workflows.
+-- The delegate is given full control over the creation of the markers.
+--
+-- This is for example useful when using one or more parties for app operation purposes.
+-- Using this proxy allows the operational parties to properly attribute their activity to the featured app,
+-- which is represented by the app provider party.
+template DelegateProxy with
+    provider : Party -- ^ The app provider whose featured app right should be used
+    delegate : Party -- ^ The delegate interacting with the token standard workflow
```
moritzkiefer-da, line 43: iirc for wallets we recommend to have a single delegate contract instead of one per delegate (which would be one per wallet user in those cases). Is there a good reason to do things differently here? I see why this is sufficient for exchanges but having to maintain both proxy variants feels unfortunate.

moritzkiefer-da: If we do make the rules variant of that then we likely want to also consider whether there should be a way to restrict beneficiaries, I don't expect that giving full control over that to users is always desired.

Maybe a cheap option to start with would be to just have a flag on the rules contract that defines whether you either use a set of beneficiaries specified on the contract or let users chose it.

moritzkiefer-da (approved):
thanks, the approach makes sense although I'm vary of needing a different model for exchanges and wallets. As long as we don't upload it by default, it seems ok at least as it doesn't proliferate quite as badly.

daml/splice-util-featured-app-proxies/daml/Splice/Util/FeaturedApp/DelegateProxy.daml
meiersi-da: I initially intended to go in that direction of making a single highly configurable version of this TokenStandardProxy here: https://github.com/hyperledger-labs/splice/pull/1907/files#diff-4cbc6e851f73f40db384d63aa97dbf4ffb93c505b0ed0c3c360e7914f94f6201R42

However, it felt uneccessarily complicated compared to the option where we just define two different templates in different modules of the same utility package. This way the user just chooses what template to use and goes with it.

Note that the difference is not only in the template arguments, but also the proxy argument. 

Furthermore, I wouldn't be surprised if we find that we need a version 2 of one of the proxies, and we can just add that to the side w/o impacting existing users.

moritzkiefer-da: makes sense, maybe make it clear somehow that this is not the recommended template for wallets and link to your PR for that for now. I suspect people will get confused otherwise.

meiersi-da: https://github.com/hyperledger-labs/splice/issues/2197

PR #2207, opened 2025-09-10
Document `ignored-expired-rewards-party-ids` process

martinflorian-da:
Fixes #2180

[static]

I gave a bit more detail on "how to build the config" because this can be useful also just as documentation for our own team (that will likely be that "one SV" most of the time).

### DIFF
--- a/docs/src/sv_operator/sv_operations.rst
+++ b/docs/src/sv_operator/sv_operations.rst
@@ -343,6 +343,8 @@ Based on that:
 
       select distinct(reward_party) from dso_acs_store where reward_round < UPGRADE_ROUND;
 
+  Note however that this query overapproximates the set of parties to ignore, so you might need to remove some parties again later.
+
 - If one of the parties you want to ignore is an SV, you most likely want to ignore the beneficiary of that SV instead
   (or even better - talk to the SV's operator to make sure their beneficiary upgrades to a more recent version).
 - Once you are confident in the set of parties to ignore, remove the ``delegatelessAutomationExpiredRewardCouponBatchSize`` override and confirm that the


docs/src/sv_operator/sv_operations.rst
```diff
@@ -263,3 +263,101 @@ Upper bound
...
+
+Based on that:
+
+- You can wait until the trigger fails, then investigate which parties are involved in the failing transaction and add these to the set of parties to ignore.
+- For a more quick feedback loop, you can attempt queries against the SV app (postgres) database.
+  For example, you can list receivers of rewards that were not expired since the Daml upgrade:
+
+  .. code-block:: sql
+
+      select distinct(reward_party) from dso_acs_store where reward_round < UPGRADE_ROUND;
```
moritzkiefer-da, line 344: Note that this overapproximates: It could also be that the trigger just hasn't gotten to those contracts yet. So I'm a bit wary of documenting that mess.

docs/src/sv_operator/sv_operations.rst
```diff
@@ -263,3 +263,101 @@ Upper bound
...
+  trigger has a 100% success rate and does not log any warnings or errors.
+- If outdated validators eventually follow up on upgrading, their parties should be removed again from the ignore list to allow the
+  archival of expired rewards (both old and new) as well as to avoid blocking the archival of closed mining rounds.
+  The following database query shows expired rewards that were generated *after* the Daml upgrade,
+  which is an indicator that the hosting validator has upgraded and so these parties should not be ignored anymore:
+
+  .. code-block:: sql
+
+     select m.template_id_qualified_name, m.reward_party, min(m.reward_round) from (
+       select c.reward_party, c.reward_round, c.template_id_qualified_name from dso_acs_store c
```
moritzkiefer-da, line 360: I don't understand this query. What's the point of the join instead of just doing
```
select * from dso_acs_store where reward_round > upgrade_round
```

moritzkiefer-da (approved):
thanks!

docs/src/sv_operator/sv_operations.rst
martinflorian-da: The complexity is for giving me only things for closed rounds... Do you see a way to significantly reduce the complexity?

docs/src/sv_operator/sv_operations.rst
martinflorian-da: fair point, added a disclaimer;

the query still seems to be a useful tool, or at least a good starting point for better queries, so I'd like to keep it also for reminding ourselves

docs/src/sv_operator/sv_operations.rst
moritzkiefer-da: hm fair enough I guess

martinflorian-da: otherwise you get a lot of parties that are not really useful

PR #2225, opened 2025-09-12
include dashboard functions in BQ integration test

isegall-da:
Doesn't actually invoke them, as they don't play well with simtime, but by installing them in BQ we get basic testing of syntax, type-safety, etc., which would have actually caught most of the things that I have broken there so far.

Part of https://github.com/DACH-NY/canton-network-internal/issues/1461

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTotalSupplyBigQueryIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTotalSupplyBigQueryIntegrationTest.scala
@@ -103,6 +103,10 @@ class ScanTotalSupplyBigQueryIntegrationTest
         .build()
     bigquery.create(functionsDatasetInfo)
 
+    // Note that the dashboard tables are never actually populated in this test,
+    // but we do test creating them from the codegen'ed schemas, and creating the
+    // functions and procedures for populating them, so we get some sanity check
+    // on the queries for syntax and type errors.
     val dashboardsDatasetInfo =
       bq.DatasetInfo
         .newBuilder(dashboardsDatasetName)
@@ -125,9 +129,9 @@ class ScanTotalSupplyBigQueryIntegrationTest
     logger.info(s"Cleaning up BigQuery dataset: $datasetName")
 
     // Delete the temporary BigQuery datasets after tests
-//    bigquery.delete(datasetName, bq.BigQuery.DatasetDeleteOption.deleteContents())
-//    bigquery.delete(functionsDatasetName, bq.BigQuery.DatasetDeleteOption.deleteContents())
-//    bigquery.delete(dashboardsDatasetName, bq.BigQuery.DatasetDeleteOption.deleteContents())
+    bigquery.delete(datasetName, bq.BigQuery.DatasetDeleteOption.deleteContents())
+    bigquery.delete(functionsDatasetName, bq.BigQuery.DatasetDeleteOption.deleteContents())
+    bigquery.delete(dashboardsDatasetName, bq.BigQuery.DatasetDeleteOption.deleteContents())
     super.afterAll()
   }
 


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTotalSupplyBigQueryIntegrationTest.scala
```diff
@@ -117,8 +125,9 @@ class ScanTotalSupplyBigQueryIntegrationTest
     logger.info(s"Cleaning up BigQuery dataset: $datasetName")
 
     // Delete the temporary BigQuery datasets after tests
-    bigquery.delete(datasetName, bq.BigQuery.DatasetDeleteOption.deleteContents())
-    bigquery.delete(functionsDatasetName, bq.BigQuery.DatasetDeleteOption.deleteContents())
+//    bigquery.delete(datasetName, bq.BigQuery.DatasetDeleteOption.deleteContents())
```
martinflorian-da: Why not delete those anymore?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTotalSupplyBigQueryIntegrationTest.scala
```diff
@@ -101,6 +102,13 @@ class ScanTotalSupplyBigQueryIntegrationTest
         .setDefaultTableLifetime(1.hour.toMillis)
         .build()
     bigquery.create(functionsDatasetInfo)
+
+    val dashboardsDatasetInfo =
```
martinflorian-da, line 110: OK so AFAIU your idea is to test that we can create the dashboards? But without actually testing anything about them?

Wouldn't it be cleaner then to not put this in `beforeAll` but in a dedicated test case? Maybe it's more consistent with this test to keep it here, just asking. If you do keep it here I suggest adding a comment that it's not really used.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTotalSupplyBigQueryIntegrationTest.scala
martinflorian-da, line 1: You are changing an integration test, why is `[static]` enough?

isegall-da: Because it's an integration test that runs only periodically, not as part of [ci]. I did run it locally of course to confirm that it passes.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTotalSupplyBigQueryIntegrationTest.scala
isegall-da: Nice idea, will do

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTotalSupplyBigQueryIntegrationTest.scala
isegall-da: Argh, unintentional!

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTotalSupplyBigQueryIntegrationTest.scala
isegall-da: Naa, that would require another argument to the codegen whether to include the dashboard ones or not (or splitting the two in some other way), doesn't seem worth that added complexity. I'll just add a comment here on what we do and don't do in this test.

PR #2237, opened 2025-09-12
use new users when retrying auth0

isegall-da:
I'd say: Fixes https://github.com/DACH-NY/cn-test-failures/issues/3469, but we've closed that issue so many times already that who knows...

(see my blabbering at the bottom of the issue for why I think the previous attempts failed)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/tools/src/main/scala/org/lfdecentralizedtrust/splice/Auth0TestUserCleaner.scala
+++ b/apps/tools/src/main/scala/org/lfdecentralizedtrust/splice/Auth0TestUserCleaner.scala
@@ -101,7 +101,7 @@ object Auth0TestUserCleaner {
       clientSecret,
       loggerFactory,
       new Auth0Util.Auth0Retry {
-        override def retryAuth0CallsForTests[T](f: => T)(recover: Recover[T])(implicit
+        override def retryAuth0CallsForTests[T](f: => T)(implicit
             tc: TraceContext
         ): T =
           retryAuth0Calls(f)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/Auth0Util.scala
```diff
@@ -93,8 +97,7 @@ class Auth0Util(
   }
 
   private[this] def executeManagementApiRequest[T](
-      req: com.auth0.net.Request[T],
-      recover: retry.Recover[T] = PartialFunction.empty,
```
martinflorian-da, line 97: It's OK to remove because we never used this code path?

martinflorian-da: ah nvm you're removing the one usage in this PR, got it now

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/Auth0Util.scala
```diff
@@ -39,23 +39,27 @@ class Auth0Util(
...
+      val email = s"$username@canton-network-test.com"
+      user.setPassword(password.toCharArray)
+      user.setEmail(email)
+      user.setVerifyEmail(false) // avoid auth0 trying to send mails
+      user.setConnection("Username-Password-Authentication")
+      // Auth0 management API calls are rate limited, with limits much lower than
+      // the rate limits for the auth API calls.
+      // Here we simply assume a generic rate limit of 2 calls per second and
+      // wait before each management API call
+      Threading.sleep(500)
```
martinflorian-da, line 59: Hm I would have expected we only do this if we need to retry?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/Auth0Util.scala
isegall-da: We used it until now for trying to recover, but not any more as we're no longer recovering.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/Auth0Util.scala
isegall-da: Yeah, I was thinking the same, but apparently we've always had this sleep before the call (it was just somewhere else), and it's simpler than start adding try-catches, so I'd say an extra few seconds per test are acceptable here.

martinflorian-da: fair enough

PR #2258, opened 2025-09-15
add support for retrieving the balance w/o holding fees

meiersi-da:
Fixes #2255 

[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


moritzkiefer-da (commented):
Why do we want a client chosen parameter here instead of tying it to the active amulet rules version? Having the client chose that feels off given that they do not get a choice in whether holding fees are deducted or not.

meiersi-da:
> Why do we want a client chosen parameter here instead of tying it to the active amulet rules version? Having the client chose that feels off given that they do not get a choice in whether holding fees are deducted or not.

The problem I see is that the switch-over is quite tricky to detect on the client side: they need to parse the package-id of the `AmuletRules` that was used to execute the choice. I thought we're better off having the client decide (via feature-support) on when they want to switch-over.

Open to entertain the server side version, but I fear the mismatch questions from scan consumers.

moritzkiefer-da:
> Open to entertain the server side version, but I fear the mismatch questions from scan consumers.

You're not changing scan though, you're changing the wallet no? The wallet does not expose any historic information, it only exposes the current effective balance which afaik is primarily used in the wallet UI. In that case, isn't automatically switching over to no longer deducting holding fees as soon as the new versions become effective exactly what you want?

meiersi-da:
> > Open to entertain the server side version, but I fear the mismatch questions from scan consumers.
> 
> You're not changing scan though, you're changing the wallet no? The wallet does not expose any historic information, it only exposes the current effective balance which afaik is primarily used in the wallet UI. In that case, isn't automatically switching over to no longer deducting holding fees as soon as the new versions become effective exactly what you want?

Ah good point. I put them in the same bucket. You are right for the wallet. Let me change it there.

meiersi-da:
@moritzkiefer-da : could I ask for another review of this one. thx ;)

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/WalletSweepTrigger.scala
```diff
@@ -220,11 +242,14 @@ abstract class WalletSweepTrigger(
...
   final case class Task(
-      trackingId: String
+      trackingId: String,
+      deductHoldingFees: Boolean,
   ) extends PrettyPrinting {
     override def pretty: Pretty[Task] =
       prettyOfClass[Task](
-        param("Tracking Id", _.trackingId.unquoted)
+        param("Tracking Id", _.trackingId.unquoted),
+        // TODO(#2257): switch the param once the holding fees change landed on MainNet
```
moritzkiefer-da: is the comment stale? Now that you infer it from package version support it's not clear to me what we need to change other than the usual eventual cleanup of old version guards.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/util/TopupUtil.scala
```diff
@@ -36,14 +36,22 @@ object TopupUtil {
...
       ec: ExecutionContext,
       mat: Materializer,
   ): Future[BigDecimal] = for {
     latestRound <- scanConnection.getLatestOpenMiningRound()
     roundNum = latestRound.payload.round.number
-    walletBalance <- store.getAmuletBalanceWithHoldingFees(roundNum).map(_._1)
+    walletBalance <- store
+      .getAmuletBalanceWithHoldingFees(
+        roundNum,
+        // Note we are taking the conservative approach here to always deduct holding fees to avoid
```
moritzkiefer-da, line 49: It's not clear to me what you gain from doing that over querying the current version? The latter seems accurate no?

meiersi-da: Yeah, this was a shortcut, as topup util is used widely in the codebase and I wanted to avoid funneling the `PackageVersion` service everywhere. I'm also slightly worried about the load caused by it. 

It thought the under-approximation is the better trade-off.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/WalletSweepTrigger.scala
meiersi-da: yeah, the comment is about the eventual cleanup. Would you recommend not even keeping it, as we'll discover this as part of the cleanup. seems cheaper indeed.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/util/TopupUtil.scala
meiersi-da: going with this cheaper option

PR #2262, opened 2025-09-15
Deploy DA-1 by default on scratch, cimain, ciperiodic

martinflorian-da:
...and set DSO_SIZE to 3. So that we deploy by default now: sv-1 (DA-2), sv-da-1 (DA-1), sv-2 (DA-Eng-2), sv-3 (DA-Eng-3)

Biggest part of https://github.com/DACH-NY/canton-network-internal/issues/1484

You probably want to review together with https://github.com/DACH-NY/canton-network-internal/pull/1682 . I did cluster tests from there. I'll time the merges of the two PRs closely.

Limitations:

- DevNet preflights will not work on this version until we have moved DA-1 to us on DevNet; I plan to just pause the health checks and keep that pause short. Specifically I'll time merging so that we can reasonably get CILR testing of DA-1 and some coordination prep with the current DA-1 team *before* this gets packaged in a release and deployed on DevNet

TODOs shortly after these PRs:

- Override base versions temporarily so that CI upgrade and HDM tests work.
- Deploy DA-1 on CILR... (https://github.com/DACH-NY/canton-network-internal/issues/1721)
- Internal PR with README updates (and probably a CWT...)


cluster/deployment/scratchneta/config.yaml
```diff
@@ -13,3 +13,17 @@ validator1:
     maxDuration: "5m"
     retention: "1h" # Keep 1h
   deduplicationDuration: 30m
+svs:
```
moritzkiefer-da, line 16: hm it feels kind of off to copypasta this into every single config. Is there a good reason why we don't make it the default and then overwrite it where we don't want it?

cluster/scripts/vote-for-migration.sh
```diff
@@ -81,19 +81,39 @@ vote_data='{
...
+  fi
+  # the most local config.yaml takes precedence, and if ../config.yaml is the same as the SPLICE_ROOT one, merging is a NOP
+  yq ". *= load(\"$cluster_dir/../config.yaml\") | . *= load(\"$cluster_dir/config.yaml\")" < "$SPLICE_ROOT/cluster/deployment/config.yaml"
+}
 
-DSO_SIZE=${DSO_SIZE:-4}
 other_svs=()
+
+# standard (eng) SVs
+DSO_SIZE=${DSO_SIZE:-4}
```
moritzkiefer-da: do we want this to default to 4 if above you change it to 3?

martinflorian-da: Hm I don't think it matters a terrible lot but it seems nicer to fix for consistency, will do.

cluster/deployment/scratchneta/config.yaml
martinflorian-da: OK so for context we merge configs like this: https://github.com/hyperledger-labs/splice/blob/bbdbfeb50568e8a2d928ccd1d818d9a77376779b/cluster/pulumi/common/src/config/configLoader.ts#L22

So 1. I'm not fully sure how to overwrite, i.e. if something like `sv-da-1: null` would actually work, and 2. even if something like `sv-da-1: null` does work it seems sketchy that people have to remember to add it explicitly (on prod clusters mainly but still). I don't like the duplication but I'm not sure I prefer the alternative. WDYT?

moritzkiefer-da: Hm fair I guess. Can we realistically remove it after all prod clusters have switched over and change the default?

martinflorian-da: Added it as a follow-up on https://github.com/DACH-NY/canton-network-internal/issues/1483 ... tbh I'm not fully convinced that this is a good idea either but happy to give it another set of thoughts (when we're there).

cluster/scripts/vote-for-migration.sh
martinflorian-da: fixed + found another script to fix :facepalm:

PR #2285, opened 2025-09-16
Remove participant cpu limits

nicu-da:
Under load the participant is really cpu heavy due to all the signing and validating that happens and involves key signatures. Thus it can get cpu throttled fairly quickly.  Better to sacrifice cpu and avoid throttling

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/helm/splice-participant/values-template.yaml
+++ b/cluster/helm/splice-participant/values-template.yaml
@@ -13,7 +13,6 @@ pod:
 defaultJvmOptions: -XX:+UseG1GC -XX:MaxRAMPercentage=75 -XX:InitialRAMPercentage=75 -Dscala.concurrent.context.minThreads=8 -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/persistent-data
 resources:
   limits:
-    cpu: "4"
     memory: 32Gi
   requests:
     cpu: "1"


cluster/helm/splice-participant/values-template.yaml
```diff
@@ -13,7 +13,6 @@ pod:
 defaultJvmOptions: -XX:+UseG1GC -XX:MaxRAMPercentage=75 -XX:InitialRAMPercentage=75 -Dscala.concurrent.context.minThreads=8 -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/persistent-data
 resources:
   limits:
-    cpu: "4"
```
martinflorian-da, line 16: So just for slow me (after reading some docs): This means that we can take all CPU there is on the node but only "1" is guaranteed, which is a strict improvement to the status quo where we can use at most "4" but only "1" is guaranteed?

nicu-da: Exactly, you can starve the entire node in theory, but considering that this is the participant, in most scenarios you would prefer it instead of throttling it.

moritzkiefer-da (approved):

cluster/helm/splice-participant/values-template.yaml
stas-sbi: > but only "1" is guaranteed

Strictly speaking "requests" don't guarantee anything and used **only for placement decisions** by the scheduler

PR #2294, opened 2025-09-17
Hide fees with value 0 in scan UI

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/2274


apps/scan/frontend/src/__tests__/scan.test.tsx
```diff
@@ -158,3 +158,17 @@ test('backfilling indicator does not shows when response is unclear', async () =
   const backfillingIndicator = screen.queryByTestId('backfilling-alert');
   expect(backfillingIndicator).toBeNull();
 });
+
+// the server handler returns no fees by default
```
meiersi-da, line 165: Where do we test that the fees are shown if they are non-zero?

meiersi-da: Would we usually also test that the unit rendering matches? your call on what to do... I have to little experience with the frontend code to be of help here. Semantically it looks right :)

PR #2297, opened 2025-09-17
[ci] Increase client http timeouts

nicu-da:
By default the client times out after 40 seconds so even if we bumped the server timeout we still need to increase the client one. 

This bumps the timeout for all the http clients but I don't really see a big issue there as the server side timeout should still kick in. If we do notice any type of issue we might need to add custom http clients timeouts per operation but that sounds overly complex and not useful right now. 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/resources/include/svs/_sv.conf
+++ b/apps/app/src/test/resources/include/svs/_sv.conf
@@ -61,6 +61,10 @@
         max-failures = 20
       }
     }
+    timeouts {
+      # required if we want to also increase on the client side the timeout
+      request-timeout = 5 minutes
+    }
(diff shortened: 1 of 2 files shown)


moritzkiefer-da:
> This bumps the timeout for all the http clients but I don't really see a big issue there as the server side timeout should still kick in. If we do notice any type of issue we might need to add custom http clients timeouts per operation but that sounds overly complex and not useful right now.

iirc we added client side timeouts because we had cases where the connection was stuck in ways where the server timeout did not kick in and we really wanted it to time out client side. This seems to make this worse again.

Issue #2254, opened 2025-09-15
scan backend: add param to not deduct holding fees for balance & total endpoints


meiersi-da:
- concretely:
  - `/v0/total-amulet-balance`
  - `/v0/holdings/summary`
- use it in tests with zero fees to assert on exact balance
- note: we add a new param to make the change backwards compatible

TODO:
- [ ] investigate which of the endpoints actually traverses over the ACS
- [ ] only adapt the endpoints where we can just sum up the `amount.initialAmount` field for not deducting holding fees
- [ ] ensure both UI and token metadata report the right number of available amulet

OriolMunoz-da:
@meiersi-da 

> and token metadata

what are you referring to there?

OriolMunoz-da:
reopening: this ticket still requires UI work

PR #2318, opened 2025-09-18
Add balances to ACS snapshots

OriolMunoz-da:
Part of https://github.com/hyperledger-labs/splice/issues/2254

### DIFF
--- a/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/AcsSnapshotStoreTest.scala
+++ b/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/AcsSnapshotStoreTest.scala
@@ -854,6 +854,9 @@ class AcsSnapshotStoreTest
             2000L * n
           )
         )
+        val illegalDsoUnlocked = amulet(providerParty(42), 42, 42L, 0.42, dso = providerParty(42))
+        val illegalDsoLocked =
+          lockedAmulet(providerParty(42), 42, 42L, 0.42, dso = providerParty(42))
         val snapshotTimestamp = CantonTimestamp.Epoch.plusSeconds(100_000L)
 
         for {
@@ -875,6 +878,18 @@ class AcsSnapshotStoreTest
               Seq(PartyId.tryFromProtoPrimitive(amulet.payload.amulet.owner), dsoParty),
             )
           }
+          _ <- ingestCreate(
+            updateHistory,
+            illegalDsoLocked,
+            snapshotTimestamp.minusSeconds(2L),
+            Seq(PartyId.tryFromProtoPrimitive(illegalDsoUnlocked.payload.dso)),
+          )
+          _ <- ingestCreate(
+            updateHistory,
+            illegalDsoLocked,
+            snapshotTimestamp.minusSeconds(1L),
+            Seq(PartyId.tryFromProtoPrimitive(illegalDsoLocked.payload.amulet.dso)),
+          )
           _ <- store.insertNewSnapshot(
             None,
             DefaultMigrationId,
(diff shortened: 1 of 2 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -124,28 +124,44 @@ class AcsSnapshotStore(
...
+          sum(case when inserted_rows.template_id = ${PackageQualifiedName
+            .getFromResources(
+              Amulet.TEMPLATE_ID_WITH_PACKAGE_ID
+            )
+            .toString} and stakeholder=$dsoParty then (create_arguments->'record'->'fields'->2->'value'->'record'->'fields'->0->'value'->>'numeric')::numeric else 0 end),
+          sum(case when inserted_rows.template_id = ${PackageQualifiedName
+            .getFromResources(
+              LockedAmulet.TEMPLATE_ID_WITH_PACKAGE_ID
+            )
+            .toString} and stakeholder=$dsoParty then (create_arguments->'record'->'fields'->0->'value'->'record'->'fields'->2->'value'->'record'->'fields'->0->'value'->>'numeric')::numeric else 0 end)
```
meiersi-da: Can you share the query plan for this?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -124,28 +124,44 @@ class AcsSnapshotStore(
...
         from inserted_rows
-        having min(row_id) is not null;
+        join creates_to_insert on inserted_rows.create_id = creates_to_insert.row_id
```
moritzkiefer-da, line 158: I guess filtering by templates here isn't actually faster than just having the filter in sum since we d't have an index on the materialized view?

moritzkiefer-da (commented):
nice! Query plans and some timings on CILR would be great

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
OriolMunoz-da: you just can't: because the min and the max row_id have to be over the whole snapshot, not just filtered by template

OriolMunoz-da:
Here's a regular explain while I fail to run the analyze on cloudsql: https://www.pgexplain.dev/plan/f4527fda-e660-4941-840a-9f1f167cac21

OriolMunoz-da:
@moritzkiefer-da implemented in the last commit in https://github.com/hyperledger-labs/splice/pull/2334

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/AcsSnapshotStoreTest.scala
```diff
@@ -843,6 +843,156 @@ class AcsSnapshotStoreTest
...
+                _ <- ingestArchive(
+                  updateHistory,
+                  archiveUnlocked._1,
+                  archiveUnlocked._2.plusSeconds(1L),
+                )
+                _ <- ingestCreate(
+                  updateHistory,
+                  createLocked._1,
+                  createLocked._2,
+                  Seq(PartyId.tryFromProtoPrimitive(createLocked._1.payload.amulet.owner), dsoParty),
```
meiersi-da, line 958: Consider adding some `Amulet` that has the wrong stakeholder to test that the filtering works.

PR #2322, opened 2025-09-18
Sv App: reconnect to domain during startup only if we already host the dso party or if we don't see a proposal to host 

julientinguely-da:
fixes https://github.com/DACH-NY/cn-test-failures/issues/5574
fixes https://github.com/DACH-NY/canton-network-internal/issues/1551

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/SvApp.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/SvApp.scala
@@ -9,6 +9,33 @@ import cats.instances.future.*
 import cats.syntax.either.*
 import com.daml.grpc.adapter.ExecutionSequencerFactory
 import com.daml.ledger.javaapi.data.User
+import com.digitalasset.canton.concurrent.FutureSupervisor
+import com.digitalasset.canton.config.CantonRequireTypes.InstanceName
+import com.digitalasset.canton.config.{
+  CryptoConfig,
+  CryptoProvider,
+  NonNegativeFiniteDuration,
+  ProcessingTimeout,
+}
+import com.digitalasset.canton.lifecycle.{AsyncOrSyncCloseable, FlagCloseableAsync, SyncCloseable}
+import com.digitalasset.canton.logging.{NamedLoggerFactory, TracedLogger}
+import com.digitalasset.canton.resource.Storage
+import com.digitalasset.canton.time.Clock
+import com.digitalasset.canton.topology.store.TopologyStoreId.AuthorizedStore
+import com.digitalasset.canton.topology.transaction.{PartyToParticipant, TopologyMapping}
+import com.digitalasset.canton.topology.{ParticipantId, PartyId, SynchronizerId}
+import com.digitalasset.canton.tracing.{TraceContext, TracerProvider}
+import com.digitalasset.canton.util.MonadUtil
+import com.digitalasset.canton.version.ProtocolVersion
+import io.circe.Json
+import io.grpc.Status
+import io.opentelemetry.api.trace.Tracer
+import org.apache.pekko.actor.ActorSystem
+import org.apache.pekko.http.cors.scaladsl.CorsDirectives.cors
+import org.apache.pekko.http.cors.scaladsl.settings.CorsSettings
+import org.apache.pekko.http.scaladsl.model.HttpMethods
+import org.apache.pekko.http.scaladsl.server.Directive
+import org.apache.pekko.http.scaladsl.server.Directives.*
 import org.lfdecentralizedtrust.splice.admin.api.TraceContextDirectives.withTraceContext
 import org.lfdecentralizedtrust.splice.admin.http.{AdminRoutes, HttpErrorHandler}
 import org.lfdecentralizedtrust.splice.auth.{
@@ -21,18 +48,20 @@ import org.lfdecentralizedtrust.splice.automation.{
   DomainParamsAutomationService,
   DomainTimeAutomationService,
 }
+import org.lfdecentralizedtrust.splice.codegen.java.da.time.types.RelTime
 import org.lfdecentralizedtrust.splice.codegen.java.splice
 import org.lfdecentralizedtrust.splice.codegen.java.splice.dsorules.*
-import org.lfdecentralizedtrust.splice.codegen.java.da.time.types.RelTime
 import org.lfdecentralizedtrust.splice.config.SharedSpliceAppParameters
 import org.lfdecentralizedtrust.splice.environment.*
-import org.lfdecentralizedtrust.splice.http.{HttpClient, HttpRateLimiter}
+import org.lfdecentralizedtrust.splice.environment.BaseLedgerConnection.INITIAL_ROUND_USER_METADATA_KEY
 import org.lfdecentralizedtrust.splice.http.v0.sv.SvResource
 import org.lfdecentralizedtrust.splice.http.v0.sv_admin.SvAdminResource
+import org.lfdecentralizedtrust.splice.http.{HttpClient, HttpRateLimiter}
 import org.lfdecentralizedtrust.splice.migration.AcsExporter
 import org.lfdecentralizedtrust.splice.setup.{NodeInitializer, ParticipantInitializer}
-import org.lfdecentralizedtrust.splice.store.{AppStoreWithIngestion, UpdateHistory}
+import org.lfdecentralizedtrust.splice.store.AppStoreWithIngestion.SpliceLedgerConnectionPriority
 import org.lfdecentralizedtrust.splice.store.MultiDomainAcsStore.QueryResult
+import org.lfdecentralizedtrust.splice.store.{AppStoreWithIngestion, UpdateHistory}
 import org.lfdecentralizedtrust.splice.sv.admin.http.{HttpSvAdminHandler, HttpSvHandler}
 import org.lfdecentralizedtrust.splice.sv.automation.{
   DsoDelegateBasedAutomationService,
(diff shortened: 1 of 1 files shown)


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
```diff
@@ -459,6 +465,49 @@ class JoiningNodeInitializer(
     }
   }
 
+  // We can only reconnect the domains if it already hosts the dsoParty or doesn't have an active proposal to host it
+  def canProceedWithDomainReconnect(
+      participantAdminConnection: ParticipantAdminConnection
+  )(implicit tc: TraceContext, ec: ExecutionContext): Future[Boolean] = {
+    val initConnection = ledgerClient.readOnlyConnection(
```
nicu-da: don't we have a connection already that we can use? if not just making sure we don't need to close this connection or something

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
```diff
@@ -459,6 +465,49 @@ class JoiningNodeInitializer(
...
+      )
+      hasActiveProposalToHostDsoParty <- retryProvider.getValueWithRetries(
+        RetryFor.WaitingOnInitDependency,
+        "has_active_proposal_to_host_dso_party",
+        "Checks if there is an active proposal to host the dso party",
+        for {
+          participantId <- participantAdminConnection.getParticipantId()
+          activePartyToParticipantProposals <- participantAdminConnection
+            .listAllTransactions(
+              store = AuthorizedStore,
```
nicu-da: We should use the synchronizer store to check

nicu-da (approved):
LGTM! Thanks!

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
julientinguely-da: ah you're right in joinDsoAndOnboardNodes. Declaring it globally.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
```diff
@@ -459,6 +465,48 @@ class JoiningNodeInitializer(
...
+          participantId <- participantAdminConnection.getParticipantId()
+          synchronizerId <- participantAdminConnection.getSynchronizerId(
+            config.domains.global.alias
+          )
+          activePartyToParticipantProposals <- participantAdminConnection
+            .listAllTransactions(
+              store = TopologyStoreId.SynchronizerStore(synchronizerId),
+              proposals = true,
+              includeMappings = Set(TopologyMapping.Code.PartyToParticipant),
+              filterNamespace = Some(participantId.namespace),
```
julientinguely-da: small sanity check @nicu-da. Is it the right namespace here?

nicu-da: I think the party to participant namespace is the party namespace, and in that scenario it is not the right choice. why can;t we use the existing list party to participant endpoints? we have support for proposals there

julientinguely-da: :eyes:

julientinguely-da: in `listPartyToParticipant` we don't?

nicu-da: can you check `ensurePartyToParticipantAdditionProposalWithSerial` and see how we do it there?

julientinguely-da: oh thanks

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
julientinguely-da: The thing if I use the sync store I'm getting:
```
2025-10-02T16:22:51.049 [⋮] INFO - o.l.s.s.o.j.JoiningNodeInitializer:TimeBasedTestNetPreviewIntegrationTest/config=d42a0b9b/SV=sv4 (ff9a5b43674fb5f768200797e73d543f-has_dso_party-141e694b72632397-1090a9c82d9378d9) - The operation 'Get Checks if the dso party exists on SV onboarding' failed with a retryable error (full stack trace omitted): NOT_FOUND: Domain with alias Synchronizer 'global' is not connected
statusCode=NOT_FOUND 
```

julientinguely-da: Is it a case we reconnect without checking that we host the dso party?

nicu-da: Hmm, not ideal. So not sure we can use the authorized store, as checking the way we handle the party onboarding we always use the sync store.
What we really want is if the sv app crashes after this call https://github.com/hyperledger-labs/splice/blob/main/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeDsoPartyHosting.scala#L81 but before reaching this place https://github.com/hyperledger-labs/splice/blob/main/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeDsoPartyHosting.scala#L149 to not connect to the sync during init. 

We use user metadata to save this sort of state in other places, if we can't read the topology state without being connected to the sync I don't see any easy other way other than abusing user metadata.

julientinguely-da: ah well I think we can move the check after https://github.com/hyperledger-labs/splice/blob/c738ac4e996f27520464600e56a83ade8a33baa3/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala#L172 and reconnect there if needed?

nicu-da: Which check do you refer to?

julientinguely-da: https://github.com/hyperledger-labs/splice/pull/2322/files#:~:text=def%20canProceedWithDomainReconnect(

nicu-da: I think the url is wrong

julientinguely-da: sorry: 
My proposition is to move this [check](https://github.com/hyperledger-labs/splice/blob/cc01d5376db50431a89520c83aa02f3641420b91/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala#L464) and the call to `reconnect all Domains` just before https://github.com/hyperledger-labs/splice/blob/c738ac4e996f27520464600e56a83ade8a33baa3/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala#L172

nicu-da: yes,we could merge that logic. 
in that call we seem to use the topology state without connecting though https://github.com/hyperledger-labs/splice/blob/7ea141b40757ee6e2beea66c0bae28f71c45e8a0/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala#L967 so it does seem like it works :confused: or was it working just because we were actually just reconnecting blindly during startup

julientinguely-da: Yes it works indeed! 
So I merged both method into one. Can you please have a look if it does make sense? 
I am not sure about `dsoPartyToParticipantMapping.nonEmpty ^ activeDsoPartyToParticipantProposals.isEmpty` if it should be or or xor?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
```diff
@@ -168,7 +162,10 @@ class JoiningNodeInitializer(
           )
         ),
       ).tupled
-      decentralizedSynchronizerId <- connectToDomainUnlessMigratingDsoParty(dsoPartyId)
+      // It is possible that the participant left disconnected to domains due to party migration failure in the last SV startup.
```
nicu-da: this comment is a bit weird, if party migration failed in the startup then in most scenarios reconnecting is exactly what you don't want if you haven't imported the acs yet

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
```diff
@@ -459,6 +456,49 @@ class JoiningNodeInitializer(
     }
   }
 
+  // We can only reconnect the domains if the participant:
+  // - already hosts the dsoParty or
+  // - is not in the process to host it
```
nicu-da, line 461:
```suggestion
  // - is not in the process to host it
  // if not we risk reconnecting while the party was authorized but the acs was not imported yet thus breaking the participant
```

nicu-da (approved):
thanks! Looks sane

PR #2334, opened 2025-09-18
Determine Amulet total supply by ACS snapshot when possible

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/2279
Fixes https://github.com/hyperledger-labs/splice/issues/2254

compat tests: https://github.com/hyperledger-labs/splice/actions/runs/17920993127

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/plugins/UpdateHistorySanityCheckPlugin.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/plugins/UpdateHistorySanityCheckPlugin.scala
@@ -148,7 +148,7 @@ class UpdateHistorySanityCheckPlugin(
   private def checkScanTxLogScript(scan: ScanAppBackendReference)(implicit tc: TraceContext) = {
     val snapshotRecordTime = scan.forceAcsSnapshotNow()
     val amuletRules = scan.getAmuletRules()
-    val subtractHoldingFees: Boolean =
+    val amuletIncludesFees: Boolean =
       amuletRules.contract.payload.configSchedule.initialValue.packageConfig.amulet
         .split("\\.")
         .toList match {
@@ -186,9 +186,11 @@ class UpdateHistorySanityCheckPlugin(
             "--compare-acs-with-snapshot",
             snapshotRecordTime.toInstant.toString,
           ) ++ Option
-            .when(subtractHoldingFees)("--subtract-holding-fees-per-round")
+            .when(amuletIncludesFees)("--subtract-holding-fees-per-round")
             .toList ++ Option
-            .when(compareBalancesWithTotalSupply)("--compare-balances-with-total-supply")
+            .when(compareBalancesWithTotalSupply && !amuletIncludesFees)(
+              "--compare-balances-with-total-supply"
+            )
             .toList ++ ignoredRootCreates.flatMap { templateId =>
             Seq("--ignore-root-create", QualifiedName(templateId).toString)
           } ++ ignoredRootExercises.flatMap { case (templateId, choice) =>


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardMetadataHandler.scala
```diff
@@ -60,12 +65,37 @@ class HttpTokenStandardMetadataHandler(
...
+        if (deductHoldingFees) {
+          lookupTotalSupplyByLatestRound().value
+        } else {
+          lookupTotalSupplyByLatestAcsSnapshot().orElse(lookupTotalSupplyByLatestRound()).value
+        }
+    } yield result
+  }
 
   private def getAmuletInstrument()(implicit ec: ExecutionContext, tc: TraceContext) =
     for {
```
OriolMunoz-da, line 109: called in the line just below this

there aren't any tests... I'm also not sure there's much value in one since there already are for the endpoints that use the 2 options directly, but lmk

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardMetadataHandler.scala
```diff
@@ -60,12 +65,37 @@ class HttpTokenStandardMetadataHandler(
...
-  ).value
+
+  private def lookupTotalSupplyByLatestAcsSnapshot()(implicit tc: TraceContext) = {
+    for {
+      latestSnapshot <- OptionT(
+        acsSnapshotStore.lookupSnapshotBefore(acsSnapshotStore.currentMigrationId, clock.now)
+      )
+      unlocked <- OptionT.fromOption[Future](latestSnapshot.unlockedAmuletBalance)
+      locked <- OptionT.fromOption[Future](latestSnapshot.lockedAmuletBalance)
+    } yield (locked + unlocked) -> latestSnapshot.snapshotRecordTime.toInstant
```
OriolMunoz-da: double-check that addition is right here

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardMetadataHandler.scala
OriolMunoz-da: actually, at least it can be checked in the sanity plugin: https://github.com/hyperledger-labs/splice/pull/2318#pullrequestreview-3241093762

scripts/scan-txlog/scan_txlog.py
```diff
@@ -4324,6 +4341,16 @@ async def main():
                 if len(missing_in_script) > 0:
                     missing = [found_in_snapshot[cid] for cid in missing_in_script]
                     LOG.error(f"Contracts missing in script ACS: {missing}")
+                if args.scan_balance_assertions:
+                    # this will only work if a snapshot was taken, which is guaranteed by compare_acs_with_snapshot=True
+                    token_metadata = await scan_client.get_amulet_token_metadata()
+                    latest_per_party_balances = app_state.state.balance_end_of_round().values()
```
moritzkiefer-da: can you remind me what balance_end_of_round does here? I think ideally we would just check the acs state that the script has, no point in aligning anything to rounds.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardMetadataHandler.scala
```diff
@@ -60,12 +65,37 @@ class HttpTokenStandardMetadataHandler(
...
+      noHoldingFeesOnTransfers <- packageVersionSupport.noHoldingFeesOnTransfers(
+        store.key.dsoParty,
+        clock.now,
+      )
+      deductHoldingFees = !noHoldingFeesOnTransfers.supported
+      result <-
+        if (deductHoldingFees) {
+          lookupTotalSupplyByLatestRound().value
+        } else {
+          lookupTotalSupplyByLatestAcsSnapshot().orElse(lookupTotalSupplyByLatestRound()).value
```
moritzkiefer-da, line 103: hm, do we have any way to report back the time of this value? Is there some metadata field that we can put it into?

moritzkiefer-da: We'll need it at least in the scan UI. I guess the token standard endpoints also didn't mention the round before so maybe not a hard blocker for those but seems kind of annoying to not have it.

moritzkiefer-da (approved):
nice thanks!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardMetadataHandler.scala
meiersi-da: fill both of these: https://github.com/hyperledger-labs/splice/blob/c062908d98ab604ec4ab155a5c0c85733a78ec12/token-standard/splice-api-token-metadata-v1/openapi/token-metadata-v1.yaml#L140-L146

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardMetadataHandler.scala
meiersi-da: yes

meiersi-da: the arrow reads weird... returning the time as-of which it was computed as the key would make more sense.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardMetadataHandler.scala
moritzkiefer-da: ah perfect, forgot we had that

scripts/scan-txlog/scan_txlog.py
```diff
@@ -4324,6 +4341,16 @@ async def main():
                 if len(missing_in_script) > 0:
                     missing = [found_in_snapshot[cid] for cid in missing_in_script]
                     LOG.error(f"Contracts missing in script ACS: {missing}")
+                if args.scan_balance_assertions:
+                    # this will only work if a snapshot was taken, which is guaranteed by compare_acs_with_snapshot=True
+                    token_metadata = await scan_client.get_amulet_token_metadata()
```
meiersi-da, line 4368: This will only work once amulet-0.1.14 has been voted in correct? Consider gating the check by exposing whether that's the case in https://github.com/hyperledger-labs/splice/blob/c062908d98ab604ec4ab155a5c0c85733a78ec12/apps/scan/src/main/openapi/scan.yaml#L1520

meiersi-da (approved):
Thanks a lot!

scripts/scan-txlog/scan_txlog.py
meiersi-da: Hmm... probably better to not do that to avoid accidentally not testing the balance computation as part of the sanity check plugin. If needed we could add a separate parameter to disable this check in case it causes problems.

scripts/scan-txlog/scan_txlog.py
OriolMunoz-da:
```
def balance_end_of_round(self):
        amulets = self.list_contracts(TemplateQualifiedNames.amulet)
        locked_amulets = self.list_contracts(TemplateQualifiedNames.locked_amulet)
        per_party_balances = {}
        for amulet in amulets.values():
            owner = amulet.payload.get_amulet_owner()
            per_party_balances.setdefault(owner, PerPartyBalance())
            per_party_balances[owner].amulets += [amulet]
        for locked_amulet in locked_amulets.values():
            amulet = locked_amulet.payload.get_locked_amulet_amulet()
            owner = amulet.get_amulet_owner()
            per_party_balances.setdefault(owner, PerPartyBalance())
            per_party_balances[owner].locked_amulets += [locked_amulet]
        return per_party_balances
```
        
doesn't seem to be related to rounds at all - will rename

PR #2352, opened 2025-09-22
Pulumi: make more SV-related GCP secrets customizable

martinflorian-da:
Part of https://github.com/DACH-NY/canton-network-internal/issues/1484

Needed if, for example, you are reusing a GCP project for multiple clusters but using the same keys in those clusters is a no-go.

Drive-by: make all of the mock `config.yaml` used by moving the options there to DA-1

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


cluster/pulumi/sv-runbook/src/installNode.ts
```diff
@@ -112,10 +112,10 @@ export async function installNode(
 
   const imagePullDeps = imagePullSecret(xns);
 
-  const svKey = svKeyFromSecret('sv');
+  const svKey = svKeyFromSecret(svNamespaceStr.replace('-', '') + '-id');
```
moritzkiefer-da: I don't understand why this is the right logic and why it's specific to the sv runbook

cluster/pulumi/sv-runbook/src/installNode.ts
```diff
@@ -112,10 +112,10 @@ export async function installNode(
...
 
   const cometBftGovernanceKey = svAppConfig.externalGovernanceKey
-    ? svCometBftGovernanceKeyFromSecret(svNamespaceStr.replace('-', ''))!
+    ? svCometBftGovernanceKeyFromSecret(svNamespaceStr.replace('-', '') + '-id')
```
moritzkiefer-da: why is the cometbft governance key in the id secret? isn't that for the identity key?

martinflorian-da: This is just wrong, good catch... The interesting question is why did the cluster deploy still work, ~but I guess no one cares if we use the same secret for multiple SVs :(~

cluster/pulumi/sv-runbook/src/installNode.ts
martinflorian-da: Not sure I follow... the only thing that is specific is that the runbook doesn't support overriding the secret name and always picks (effectively) `sv-id`. Maybe I can just fix that as well; don't think we'll ever need this here but maybe it's just less confusing.

moritzkiefer-da: > Not sure I follow... the only thing that is specific is that the runbook doesn't support overriding the secret nam

I think that's the part I was missing. Definitely seems nicer if it would support it but I'm ok merging without it.

cluster/pulumi/sv-runbook/src/installNode.ts
martinflorian-da: Nvm that made 0 sense. The reason this worked is that we only use that secret if we deploy with KMS. So we would have found out very late... I hate it.

PR #2356, opened 2025-09-22
Use TokenMetadata totalSupply in Scan's total amulet

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/2254

image

This excludes holding fees _when using the ACS snapshot version_, as per the BE implmenetation: https://github.com/hyperledger-labs/splice/blob/42d13e343f7584603c7b41b6621e0f94e9f2d7c2/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardMetadataHandler.scala#L100

Therefore, there is no need to use the feature support endpoint for this change, as the response already contains the right data.

### DIFF
--- a/apps/app/src/test/resources/frontend-config.jsonnet
+++ b/apps/app/src/test/resources/frontend-config.jsonnet
@@ -72,6 +72,7 @@ local validatorNodes(clusterProtocol, clusterAddress, port) = {
   },
   scan: {
     scan: { url: 'http://localhost:5012/api/scan' },
+    tokenMetadata: { url: 'http://localhost:5012' },
   },
   sv1: {
     sv: { url: 'http://localhost:5114/api/sv' },
--- a/apps/scan/frontend/public/config.js
+++ b/apps/scan/frontend/public/config.js
@@ -7,6 +7,10 @@ window.splice_config = {
       // Edit this to the cluster you're trying to connect on.
       url: 'https://scan.sv-2.TARGET_HOSTNAME/api/scan',
     },
+    tokenMetadata: {
+      // Same as scan, '/registry' is already included in the openapi definitions
+      url: 'https://scan.sv-2.TARGET_HOSTNAME',
+    },
   },
   spliceInstanceNames: {
     networkName: 'Splice',
--- a/apps/scan/frontend/src/__tests__/setup/config.ts
+++ b/apps/scan/frontend/src/__tests__/setup/config.ts
@@ -10,6 +10,10 @@ const config = {
       // Edit this to the cluster you're trying to connect on.
       url: 'https://scan.sv-2.TARGET_HOSTNAME/api/scan',
     },
+    tokenMetadata: {
+      // Same as scan, '/registry' is already included in the openapi definitions
+      url: 'https://scan.sv-2.TARGET_HOSTNAME',
+    },
   },
   spliceInstanceNames: {
     amuletName: 'Teluma',
--- a/apps/scan/frontend/src/components/TotalAmuletBalance.tsx
+++ b/apps/scan/frontend/src/components/TotalAmuletBalance.tsx
@@ -1,40 +1,45 @@
 // Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 import { ErrorDisplay, Loading } from '@lfdecentralizedtrust/splice-common-frontend';
-import {
-  useAmuletPrice,
-  useTotalAmuletBalance,
-} from '@lfdecentralizedtrust/splice-common-frontend/scan-api';
+import { useAmuletPrice } from '@lfdecentralizedtrust/splice-common-frontend/scan-api';
 import BigNumber from 'bignumber.js';
 
 import { useScanConfig } from '../utils/config';
 import AmountSummary from './AmountSummary';
+import { useAmuletMetadata } from '../hooks/useAmuletMetadata';
 
 export const TotalAmuletBalance: React.FC = () => {
   const config = useScanConfig();
-  const totalAmuletBalanceQuery = useTotalAmuletBalance();
+  const amuletMetadataQuery = useAmuletMetadata();
   const amuletPriceQuery = useAmuletPrice();
 
-  const isLoading = totalAmuletBalanceQuery.isLoading || amuletPriceQuery.isLoading;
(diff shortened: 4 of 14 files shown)


party-allocator/package-lock.json
```diff
@@ -40,7 +40,7 @@
       }
     },
     "../token-standard/splice-api-token-transfer-instruction-v1/openapi-ts-client": {
-      "name": "transfer-instruction-openapi",
+      "name": "@lfdecentralizedtrust/transfer-instruction-openapi",
```
OriolMunoz-da, line 43: I have no idea what's up with this, but it looks right I guess

apps/app/src/test/resources/frontend-config.jsonnet
```diff
@@ -72,6 +72,7 @@ local validatorNodes(clusterProtocol, clusterAddress, port) = {
   },
   scan: {
     scan: { url: 'http://localhost:5012/api/scan' },
+    tokenMetadata: { url: 'http://localhost:5012' },
```
OriolMunoz-da: to reviewers: please check that the config files make sense

apps/scan/frontend/public/config.js
```diff
@@ -7,6 +7,10 @@ window.splice_config = {
       // Edit this to the cluster you're trying to connect on.
       url: 'https://scan.sv-2.TARGET_HOSTNAME/api/scan',
     },
+    tokenMetadata: {
+      // Same as scan, '/registry' is already included in the openapi definitions
+      url: 'https://scan.sv-2.TARGET_HOSTNAME',
```
OriolMunoz-da, line 8: does this need to be mentioned in the README? afaict no, because it should already be handled by this https://github.com/hyperledger-labs/splice/pull/2356/files#diff-47311e17778dbc412321e59bff9ee01384a0846c4dcd19b702b6acf74f0ccb7aR9

OriolMunoz-da: (but then idk what the purpose of this file is)

apps/scan/frontend/src/components/TotalAmuletBalance.tsx
```diff
@@ -1,40 +1,45 @@
...
+  } else if (isError || isDataUndefined) {
+    return <ErrorDisplay message={'Could not retrieve total amulet balance or amulet price'} />;
+  } else {
+    const totalSupply = new BigNumber(amuletMetadataQuery.data.totalSupply || '0');
+    const asOf = amuletMetadataQuery.data?.totalSupplyAsOf;
+    return (
+      <AmountSummary
+        title={title}
+        amount={totalSupply}
+        asOf={asOf}
```
OriolMunoz-da, line 35: the asOf will always be shown (whether computed via round totals or via ACS snapshots). I'd say that's fine, but shout if you disagree

build.sbt
```diff
@@ -308,7 +308,22 @@ lazy val `splice-api-token-metadata-v1-daml` =
     .in(file("token-standard/splice-api-token-metadata-v1"))
     .enablePlugins(DamlPlugin)
     .settings(
-      BuildCommon.damlSettings
+      BuildCommon.damlSettings,
+      templateDirectory := (`openapi-typescript-template` / patchTemplate).value,
+      Compile / sourceGenerators +=
```
OriolMunoz-da, line 313: doing the same thing as here: https://github.com/hyperledger-labs/splice/blob/42d13e343f7584603c7b41b6621e0f94e9f2d7c2/build.sbt#L338
and seems to work locally after a `make clean-all`...

apps/scan/frontend/public/config.js
meiersi-da: neither do I. The nicer config format would be if only the scan prefix was specified and the `api/scan` suffix was added internally.

meiersi-da: Who in the team can check this?

apps/scan/frontend/src/__tests__/setup/config.ts
```diff
@@ -10,6 +10,10 @@ const config = {
       // Edit this to the cluster you're trying to connect on.
       url: 'https://scan.sv-2.TARGET_HOSTNAME/api/scan',
     },
+    tokenMetadata: {
+      // Same as scan, '/registry' is already included in the openapi definitions
+      url: 'https://scan.sv-2.TARGET_HOSTNAME',
+    },
```
meiersi-da, line 12: Yeah, I think this is asking for trouble. Lots of downstream consumers might have trouble configuring this and might wonder why their config broke.

What do you think of making the config value optional and default it to the scan by dropping the `api/scan` suffix? And fail if that suffix is not present?

build.sbt
meiersi-da: probably need a message to the team after the merge then, or not?

apps/scan/frontend/src/components/TotalAmuletBalance.tsx
meiersi-da: agreed

party-allocator/package-lock.json
meiersi-da: @stephencompall-DA (FYI: @moritzkiefer-da): looking at https://github.com/hyperledger-labs/splice/commit/4aa118775e4933629b6c45efaabb8f2abed733b7 it seems that somehow your change to force the namespace didn't fully propagate; or it requires an `sbt clean` on the dev machines... do you have an idea what could be going wrong?

moritzkiefer-da: may also have been a concurrent merge

apps/scan/frontend/src/__tests__/setup/config.ts
moritzkiefer-da: > Yeah, I think this is asking for trouble. Lots of downstream consumers might have trouble configuring this and might wonder why their config broke.

Is it? We support only use through docker images and those don't require any change.

That said: Given that the docker image only requires the scan prefix already, making a breaking change to require only the prefix instead of two parameters might be just fine as it's not breaking to docker users. If we want to be extra cautious have it strip a trailing /api/scan.

apps/app/src/test/resources/frontend-config.jsonnet
moritzkiefer-da: see my comment below, given that the docker image only requires a single parameter without a suffix I would probably just change `scan` to be that URL and maybe for extra caution strip the `/api/scan` at the end if it is there.

PR #2383, opened 2025-09-23
Update Scan Holding text to depend on feature-support

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/2277

image

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
@@ -1392,17 +1392,25 @@ class HttpScanHandler(
             snapshotStore.currentMigrationId,
             snapshotTime,
           )
-          // note that this will make it so that the next snapshot is taken N hours after THIS snapshot.
-          // this is, in principle, not a problem:
-          // - this will only be used in tests
-          // - wall clock tests must take manual snapshots anyway, because they can't wait
-          // - simtime tests will advanceTime(N.hours)
-          _ = logger.info(s"Forcing ACS snapshot at $snapshotTime. Last snapshot: $lastSnapshot")
-          _ <- snapshotStore.insertNewSnapshot(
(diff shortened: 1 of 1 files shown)


apps/scan/frontend/src/components/NetworkInfo.tsx
```diff
@@ -224,7 +233,7 @@ const FeesTable: React.FC<{ amuletConfig: AmuletConfig<'USD'> }> = ({ amuletConf
             name="Holding Fee"
             value={BigNumber(amuletConfig.transferConfig.holdingFee.rate)}
```
OriolMunoz-da, line 234: afaiu the value itself doesn't change, but correct me if I'm wrong.
Reason I'm asking is the wording of the ticket:

> select display text **and fee deduction** based on /v0/feature-support

but I think that might be talking about the balance?

apps/scan/frontend/src/components/NetworkInfo.tsx
```diff
@@ -194,9 +189,23 @@ const NextConfigUpdate: React.FC = () => {
...
   const amuletName = config.spliceInstanceNames.amuletName;
+
+  if (featureSupport.isLoading) {
+    return <Loading />;
+  }
+  if (featureSupport.isError) {
+    return <ErrorDisplay message="Failed to fetch feature support info" />;
+  }
+  const holdingFeesDescription = featureSupport.data!.noHoldingFeesOnTransfers
+    ? 'A fixed fee for maintaining each active Canton Coin record, computed per round, but only charged if it surpasses the Coin amount.'
```
meiersi-da:
```suggestion
    ? 'A fixed fee for maintaining each active ${AmuletName} record, computed per round, but only charged if it surpasses the ${AmuletName} amount.'
```

apps/scan/frontend/src/components/NetworkInfo.tsx
meiersi-da: yes, that was about the balance.

meiersi-da (approved):
Nice work! Thanks!

PR #2388, opened 2025-09-24
Fix broken token-standard sanity plugin

OriolMunoz-da:
...by losing my sanity

I have seen it run concurrently between both and I don't trust npm does a sensible thing anymore.
It could also just be that token-standard's npm install just never ran, so the logs will help me debug it on CI.

### DIFF
--- a/project/BuildCommon.scala
+++ b/project/BuildCommon.scala
@@ -21,6 +21,8 @@ import xsbti.compile.CompileAnalysis
 import de.heikoseeberger.sbtheader.HeaderPlugin.autoImport.{headerResources, headerSources}
 import CantonDependencies.daml_ledger_api_value_proto
 
+import java.util.concurrent.atomic.AtomicInteger
+
 object BuildCommon {
 
   object defs {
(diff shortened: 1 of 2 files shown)


project/BuildCommon.scala
```diff
@@ -1601,12 +1605,15 @@ object BuildCommon {
     val cacheDir = streams.value.cacheDirectory / "npmInstall"
     val cache =
       FileFunction.cached(cacheDir, FileInfo.hash) { _ =>
+        val i = id.getAndIncrement()
+        println(s"NPM INSTALL FOR ${npmRootDir.value} CALLED. ID $i")
```
OriolMunoz-da: I'm keeping these, because somehow the PR introducing the build changes originally didn't break on PR CI run, but it does consistently on main

moritzkiefer-da: maybe add a todo to remove them at some point? Or alternatively if you really want to keep them forever maybe make them less capslocky

julientinguely-da (approved):
:crossed_fingers:

PR #2398, opened 2025-09-24
Change SV UI to set empty list of transfer fee steps

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/2264

image

### DIFF
--- a/docs/src/release_notes.rst
+++ b/docs/src/release_notes.rst
@@ -24,9 +24,8 @@ Upcoming
     - UI
 
       - When updating the ``AmuletRules`` config, the UI will omit any transfer fee steps with value zero from the ``AmuletRules`` config stored on-ledger.
-      Thereby making the ``AmuletRules`` contract smaller and saving
-      traffic for transactions using it. 
-      This is motivated by `CIP-0078 CC Fee Removal <https://github.com/global-synchronizer-foundation/cips/blob/main/cip-0078/cip-0078.md>`__ .
+        Thereby making the ``AmuletRules`` contract smaller and saving traffic for transactions using it.
+        This is motivated by `CIP-0078 CC Fee Removal <https://github.com/global-synchronizer-foundation/cips/blob/main/cip-0078/cip-0078.md>`__ .
 
 
 0.4.18


OriolMunoz-da:
> A bit confused by the diff in the screenshot. Is this going to be default for every new Amulet proposal?

No, only the first one to set fees to 0 (and then they'll stay like that for the foreseeable future). 

What I'm not so sure now that you asked is whether the FE will complain because those values are not going to be defined (as the array will be empty) 

@meiersi-da heads up

meiersi-da:
> > A bit confused by the diff in the screenshot. Is this going to be default for every new Amulet proposal?
> 
> No, only the first one to set fees to 0 (and then they'll stay like that for the foreseeable future).
> 
> What I'm not so sure now that you asked is whether the FE will complain because those values are not going to be defined (as the array will be empty)
> 
> @meiersi-da heads up

@fayi-da : this change should ideally make into tomorrow's release. It improves the CC fee removal delivery. Could you do me a favor and check whether both the SV UI and the Scan UI will deal properly with an empty list of steps in the transfer fees?

meiersi-da:
@fayi-da : actually scrap that request. The gods of CI seem to not be kind with us today; and this is a stretch goal. Let's wait until @OriolMunoz-da is back.

OriolMunoz-da:
@meiersi-da after the fields are deleted from the array (like in the screenshot above), the result is they can never be edited again (notice they're gone):
image

That seems fine by me, but better to double-check.

meiersi-da:
> @meiersi-da after the fields are deleted from the array (like in the screenshot above), the result is they can never be edited again (notice they're gone): image
> 
> That seems fine by me, but better to double-check.

Yes I confirm this is OK. We are not planning to get these fees back.

docs/src/release_notes.rst
```diff
@@ -21,6 +21,11 @@ Upcoming
 
       - Remove CPU limits from the helm charts for ``scan``, ``mediator`` and ``sequencer`` apps. This should avoid issues with cpu scheduling that might lead to performance degradations.
 
+    - UI
+
+        - When updating the AmuletRules config, the UI will omit any transfer fee steps with value zero from the payload
+          sent to the server. They will not be changeable after that either. This follows CIP-0078.
```
meiersi-da:
```suggestion
      - When updating the ``AmuletRules`` config, the UI will omit any transfer fee steps with value zero from the ``AmuletRules`` config stored on-ledger.
      Thereby making the ``AmuletRules`` contract smaller and saving
      traffic for transactions using it. 
      This is motivated by `CIP-0078 CC Fee Removal <https://github.com/global-synchronizer-foundation/cips/blob/main/cip-0078/cip-0078.md>`__ .
```

PR #2407, opened 2025-09-24
[ci] pre-calculating balances and leaderboards in aggregation

ray-roestenburg-da:
to improve performance of v0/wallet-balance, v0/total-amulet-balance, /v0/top-providers-by-app-rewards

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
```diff
@@ -596,48 +596,27 @@ class DbScanStore(
...
-              """.as[Option[BigDecimal]].headOption,
-              "getTotalAmuletBalanceForEarlierRound",
-            )
-          }
+        result <- ensureAggregated(asOfEndOfRound) { _ =>
+          storage.query(
+            // TODO(#800) change to query from round_totals when amulet expiry works again
+            // the round_total_amulet_balance is sparse and might not have an entry for the requested round
+            // which is why the first entry found <= asOfEndOfRound is used
+            sql"""
```
ray-roestenburg-da, line 604: Maybe better to fail with not_found if a early balance is requested, which would signal demand for this (and from that we can decide if we want to re-start scan aggregation for all rounds, or start a separate re-computation process for `round_total_amulet_balance`).

ray-roestenburg-da: We can also document from which round this is available for instance.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
```diff
@@ -596,48 +596,27 @@ class DbScanStore(
...
+              select greatest(
+                   0,
+                   sum_cumulative_change_to_initial_amount_as_of_round_zero -
+                   sum_cumulative_change_to_holding_fees_rate * ($asOfEndOfRound + 1)
```
ray-roestenburg-da: Pre-computed sum, so `sum_change * ($asOfEndOfRound + 1)` instead of `sum(change_for_round * ($asOfEndOfRound + 1))`: `(a+b)*c = ac + bc`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -413,8 +413,15 @@ class HttpScanHandler(
             HttpErrorHandler.onGrpcNotFound(s"Data for round ${asOfEndOfRound} not yet computed")
           )
       } yield {
-        definitions.GetTotalAmuletBalanceResponse(
-          Codec.encode(total)
+        total.fold(
```
ray-roestenburg-da, line 416: Changed to respond with NotFound for rounds that were not computed. The change in this PR calculates these going forward. Rather 'error-out' with `NotFound` than fallback to a very slow query. If we get a signal on this from logs that there is a lot of requests for older rounds, we can put some effort in computing all rounds in the round_total_amulet_balance table, but hopefully that will not be needed.

ray-roestenburg-da: (the UI already uses getInstrument and this only uses getTotalAmuletBalance if holdingfees are deducted.)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
```diff
@@ -593,51 +593,33 @@ class DbScanStore(
...
-                  group by party
+        result <- ensureAggregated(asOfEndOfRound) { _ =>
+          storage.query(
+            // TODO(#800) change to query from round_totals when amulet expiry works again
+            // the round_total_amulet_balance is sparse and might not have an entry for the requested round
+            // which is why the first entry found <= asOfEndOfRound is used
+            sql"""
+              select greatest(
+                0,
+                coalesce(
```
OriolMunoz-da: what are you coalescing here?
both `sum_cumulative*` columns are not null, and you're not left-joining

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
```diff
@@ -593,51 +593,33 @@ class DbScanStore(
...
+                  sum_cumulative_change_to_initial_amount_as_of_round_zero -
+                  sum_cumulative_change_to_holding_fees_rate * ($asOfEndOfRound + 1),
```
OriolMunoz-da: @meiersi-da 
we _do_ have the fees separate from the amount though?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
```diff
@@ -913,6 +920,44 @@ final class ScanAggregator(
...
+               sum(rpt.cumulative_change_to_holding_fees_rate)
+      from     round_party_totals rpt
+      join     active_parties_for_aggr_rounds ap
+      on       rpt.closed_round = ap.active_round
+      and      rpt.party = ap.party
+      and      rpt.store_id = $roundTotalsStoreId
+      group by ap.aggr_round
+      on conflict (store_id, closed_round)
+      do update set sum_cumulative_change_to_initial_amount_as_of_round_zero = excluded.sum_cumulative_change_to_initial_amount_as_of_round_zero,
+        sum_cumulative_change_to_holding_fees_rate = excluded.sum_cumulative_change_to_holding_fees_rate;
```
OriolMunoz-da: I don't understand when a conflict would happen and why. Why is it safe to replace the amounts?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
```diff
@@ -913,6 +920,44 @@ final class ScanAggregator(
...
+      union
+      -- all the parties that were active in the newly aggregated rounds
+      select party, round as aggr_round, round as active_round from temp_cumulative_totals
+      union
+      -- adding the lastClosedRound as aggr_round, for parties that were not active but have been aggregated in that round
+      select party,
+             $lastClosedRound as aggr_round,
+             closed_round as active_round
+      from   active_parties
+      where  store_id = $roundTotalsStoreId;
```
OriolMunoz-da: open and close parentheses `on commit drop as (...)`, this got very confusing when reviewing

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
```diff
@@ -747,6 +747,13 @@ final class ScanAggregator(
       */
     def aggregate(): DBIOAction[Unit, NoStream, Effect.Write] = {
       sqlu"""
+      create temp table active_parties_before on commit drop as
+      select party,
+             (select max(closed_round) from active_parties where store_id = $roundTotalsStoreId) as aggr_round,
```
OriolMunoz-da: this doesn't depend on the rest of the query, have you checked that the query plan doesn't run this for every party?
it might still be better to extract it

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
ray-roestenburg-da: ah thanks, cumulative_* in round_party_totals are null, probably misremembered. (it can't really hurt but will remove)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
moritzkiefer-da: yes but it's fees relative to round zero. If you don't deduct that your amount is too high that's no good

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
ray-roestenburg-da: Unlikely a conflict occurs, but all of this has to be idempotent because of 'reasons outside of my preference ;-)', so just safer, don't want to fail. it's on conflict pk so it would have to mean a previous aggregation already completed.
Since the aggregation calculates this from round_party_totals and most recent active_parties that is created by this aggregation (and all of this is transactional), it's safe to replace the amounts (it would be the same result).

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
ray-roestenburg-da: Ok I'll add those

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
ray-roestenburg-da: why for every party? This happens per aggregation, which is not per party. It aggregates everything for all rounds it finds.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
ray-roestenburg-da: done.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
ray-roestenburg-da: done.

ray-roestenburg-da:
Have another look please @OriolMunoz-da

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
ray-roestenburg-da: oh sorry you mean the sub select.. let me check to be safe

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
OriolMunoz-da: > it would be the same result

then you could as well ignore it, right? I was mostly surprised because [some lines above](https://github.com/hyperledger-labs/splice/pull/2407/files/f8748be6b86def950b4c79feaa41f9b0ff6a4482#diff-e009623f4eab792ba7720ee6708ba32fb58d6a25e48b8a099cbb2fc16f2df4c3R922) it's ignored. No strong feelings one way or the other

ray-roestenburg-da: ah, right, yes could just do on conflict do nothing, let me do that instead for clarity

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala
ray-roestenburg-da: image

Postgres is clever enough luckily (1 loop for aggregate and 1 loop for all the parties). the seq scan is fine cause all parties are needed. (one row per party in this table (per store_id))

ray-roestenburg-da: Another alternative is:
```
        with max_closed as (select max(closed_round) as aggr_round from scan_sv_1.active_parties where store_id = 1)
        select party,
               max_closed.aggr_round,
               closed_round as active_round
        from   scan_sv_1.active_parties,
               max_closed 
        where  store_id = 1

```
but this has the same explain plan, rather keep it simple with scalar sub-query.

PR #2436, opened 2025-09-26
Lock acs_snapshot_data while writing

rautenrieth-da:
Fixes https://github.com/DACH-NY/canton-network-internal/issues/1782

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AdvisoryLockIds.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AdvisoryLockIds.scala
@@ -1,0 +1,13 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package org.lfdecentralizedtrust.splice.store.db
+
+/** Registry for advisory lock identifiers used in splice applications.
+  */
+object AdvisoryLockIds {
+  // 0x73706c equals ASCII encoded "spl". Modeled after Canton's HaConfig, which uses ASCII "dml".
+  private val base: Long = 0x73706c00
+
+  final val acsSnapshotDataInsert: Long = base + 1
+}
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
@@ -175,10 +175,9 @@ class AcsSnapshotStore(
     *  Regular locks (e.g. obtained via `LOCK TABLE ... IN EXCLUSIVE MODE`) would conflict with harmless
     *  background operations like autovacuum or create index concurrently.
     *
-    *  In case the application crashes while holding the lock, the server should close the connection
+    *  In case the application crashes while holding the lock, the server _should_ close the connection
     *  and abort the transaction as soon as it detects a disconnect.
-    *  See [[com.digitalasset.canton.platform.store.backend.postgresql.PostgresDataSourceConfig]] for our connection keepalive settings.
-    *  With default settings, the server should detect a dead connection within ~15sec.
+    *  TODO(#2488): Verify that the server indeed closes connections in a reasonable time.
     */
   private def withExclusiveSnapshotDataLock[T, E <: Effect](
       action: DBIOAction[T, NoStream, E]
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
@@ -14,7 +14,7 @@ import org.lfdecentralizedtrust.splice.scan.store.AcsSnapshotStore.{
 }
 import org.lfdecentralizedtrust.splice.store.UpdateHistory.SelectFromCreateEvents
 import org.lfdecentralizedtrust.splice.store.{HardLimit, Limit, LimitHelpers, UpdateHistory}
-import org.lfdecentralizedtrust.splice.store.db.{AcsJdbcTypes, AcsQueries}
+import org.lfdecentralizedtrust.splice.store.db.{AcsJdbcTypes, AcsQueries, AdvisoryLockIds}
 import org.lfdecentralizedtrust.splice.util.{Contract, HoldingsSummary, PackageQualifiedName}
 import com.digitalasset.canton.data.CantonTimestamp
 import com.digitalasset.canton.lifecycle.{CloseContext, FutureUnlessShutdown}
@@ -158,7 +158,7 @@ class AcsSnapshotStore(
         join creates_to_insert on inserted_rows.create_id = creates_to_insert.row_id
         having min(inserted_rows.row_id) is not null;
              """).toActionBuilder.asUpdate
-      storage.update(withExclusiveSnapshotDataLock(statement), "insertNewSnapshot")
+      storage.queryAndUpdate(withExclusiveSnapshotDataLock(statement), "insertNewSnapshot")
     }.andThen { _ =>
       AcsSnapshotStore.PreventConcurrentSnapshotsSemaphore.release()
     }
@@ -169,8 +169,12 @@ class AcsSnapshotStore(
     *  Note: The acs_snapshot_data table must not have interleaved rows from two different acs snapshots.
     *  In rare cases, it can happen that the application crashes while writing a snapshot, then
     *  restarts and starts writing a different snapshot while the previous statement is still running.
+    *
     *  The exclusive lock prevents this.
-    *  Once obtained, a lock is held for the remainder of the current transaction.
+    *  We use a transaction-scoped advisory lock, which is released when the transaction ends.
+    *  Regular locks (e.g. obtained via `LOCK TABLE ... IN EXCLUSIVE MODE`) would conflict with harmless
+    *  background operations like autovacuum or create index concurrently.
+    *
     *  In case the application crashes while holding the lock, the server should close the connection
     *  and abort the transaction as soon as it detects a disconnect.
     *  See [[com.digitalasset.canton.platform.store.backend.postgresql.PostgresDataSourceConfig]] for our connection keepalive settings.
@@ -178,10 +182,16 @@ class AcsSnapshotStore(
     */
   private def withExclusiveSnapshotDataLock[T, E <: Effect](
       action: DBIOAction[T, NoStream, E]
-  ): DBIOAction[T, NoStream, E & Effect.Write & Effect.Transactional] =
+  ): DBIOAction[T, NoStream, Effect.Read & Effect.Transactional & E] =
     (for {
-      _ <- sqlu"LOCK TABLE acs_snapshot_data IN EXCLUSIVE MODE"
-      result <- action
+      lockResult <- sql"SELECT pg_try_advisory_xact_lock(${AdvisoryLockIds.acsSnapshotDataInsert})"
+        .as[Boolean]
+        .head
+      result <- lockResult match {
+        case true => action
+        // Lock conflicts should almost never happen. If they do, we fail immediately and rely on the trigger infrastructure to retry and log errors.
+        case false => DBIOAction.failed(new Exception("Failed to acquire exclusive lock"))
+      }
     } yield result).transactionally
 
   def deleteSnapshot(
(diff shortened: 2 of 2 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -158,12 +158,32 @@ class AcsSnapshotStore(
...
+    *  In case the application crashes while holding the lock, the server should close the connection
+    *  and abort the transaction as soon as it detects a disconnect.
+    *  See [[com.digitalasset.canton.platform.store.backend.postgresql.PostgresDataSourceConfig]] for our connection keepalive settings.
+    *  With default settings, the server should detect a dead connection within ~15sec.
+    */
+  private def withExclusiveSnapshotDataLock[T, E <: Effect](
+      action: DBIOAction[T, NoStream, E]
+  ): DBIOAction[T, NoStream, E & Effect.Write & Effect.Transactional] =
+    (for {
+      _ <- sqlu"LOCK TABLE acs_snapshot_data IN EXCLUSIVE MODE"
```
meiersi-da: An application level exclusive advisory lock (https://www.postgresql.org/docs/current/explicit-locking.html#ADVISORY-LOCKS) would probably be better as it would not block these background processes (from ChatGPT):

> 🔄 Background processes that could be blocked
> 
> AutoVacuum / Manual VACUUM
> 
> Regular VACUUM and autovacuum workers acquire SHARE UPDATE EXCLUSIVE.
> 
> That conflicts with your EXCLUSIVE lock, so they’ll wait until you release it.
> 
> 👉 Meaning: if you hold the lock too long, autovacuum for that table won’t run.
> 
> AutoAnalyze / ANALYZE
> 
> Also requires SHARE UPDATE EXCLUSIVE, so same as above: will be blocked.
> 
> CREATE INDEX CONCURRENTLY
> 
> This requires SHARE UPDATE EXCLUSIVE at points, so it would be blocked too.
> 
> DDL initiated by background workers (e.g. partition detach/attach on declarative partitioning, if you’re using time-based partitions managed by extensions)
> 
> Those use ACCESS EXCLUSIVE, so they’ll also wait.

meiersi-da (approved):
Thanks. Looks good to me, but I'm not deep enough in the code to judge whether it is really safe.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
rautenrieth-da: Thanks, didn't think about background processes. I'll rewrite it using an advisory lock.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -158,12 +158,42 @@ class AcsSnapshotStore(
...
         having min(inserted_rows.row_id) is not null;
              """).toActionBuilder.asUpdate
-      storage.update(statement, "insertNewSnapshot")
+      storage.queryAndUpdate(withExclusiveSnapshotDataLock(statement), "insertNewSnapshot")
     }.andThen { _ =>
       AcsSnapshotStore.PreventConcurrentSnapshotsSemaphore.release()
     }
   }
 
+  /** Wraps the given action in a transaction that holds an exclusive lock on the acs_snapshot_data table.
```
ray-roestenburg-da, line 167: I really would like to avoid this, let's have a discussion first

ray-roestenburg-da (changes requested):
I don't like locks so you need to convince me there is no other option :-)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -158,12 +158,42 @@ class AcsSnapshotStore(
...
+    *
+    *  Note: The acs_snapshot_data table must not have interleaved rows from two different acs snapshots.
+    *  In rare cases, it can happen that the application crashes while writing a snapshot, then
+    *  restarts and starts writing a different snapshot while the previous statement is still running.
```
ray-roestenburg-da, line 171: So this is not done transactionally? would that not be the solution instead of locks? transaction commits or rollsback.. why is the statement still running if the app crashes, connection close / disconnect / OS closes socket should abort it?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -158,12 +158,42 @@ class AcsSnapshotStore(
...
+  private def withExclusiveSnapshotDataLock[T, E <: Effect](
+      action: DBIOAction[T, NoStream, E]
+  ): DBIOAction[T, NoStream, Effect.Read & Effect.Transactional & E] =
+    (for {
+      lockResult <- sql"SELECT pg_try_advisory_xact_lock(${AdvisoryLockIds.acsSnapshotDataInsert})"
+        .as[Boolean]
+        .head
+      result <- lockResult match {
+        case true => action
+        // Lock conflicts should almost never happen. If they do, we fail immediately and rely on the trigger infrastructure to retry and log errors.
```
ray-roestenburg-da, line 191: Famous last words :-)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
ray-roestenburg-da: Ah, advisory locks are something else.. maybe this is ok.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AdvisoryLockIds.scala
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package org.lfdecentralizedtrust.splice.store.db
+
+/** Registry for advisory lock identifiers used in splice applications.
+  */
+object AdvisoryLockIds {
+  // 0x73706c equals ASCII encoded "spl". Modeled after Canton's HaConfig, which uses ASCII "dml".
```
ray-roestenburg-da, line 9: why spl

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -158,12 +158,42 @@ class AcsSnapshotStore(
...
+  /** Wraps the given action in a transaction that holds an exclusive lock on the acs_snapshot_data table.
+    *
+    *  Note: The acs_snapshot_data table must not have interleaved rows from two different acs snapshots.
+    *  In rare cases, it can happen that the application crashes while writing a snapshot, then
+    *  restarts and starts writing a different snapshot while the previous statement is still running.
+    *
+    *  The exclusive lock prevents this.
+    *  We use a transaction-scoped advisory lock, which is released when the transaction ends.
+    *  Regular locks (e.g. obtained via `LOCK TABLE ... IN EXCLUSIVE MODE`) would conflict with harmless
+    *  background operations like autovacuum or create index concurrently.
```
ray-roestenburg-da, line 176: Want to be sure this does not lock rows in any way for select / insert / update / delete

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AdvisoryLockIds.scala
rautenrieth-da: Picking random 3 characters from "splice".

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
rautenrieth-da: >  why is the statement still running if the app crashes, connection close / disconnect / OS closes socket should abort it?

[Some people are saying](https://dba.stackexchange.com/a/81424) a transaction can keep running and autocommit before the server checks for broken connections. FWIW, after some persistent prodding, I could get Gemini to confirm this:

> The statement you quoted is correct, and my previous answer was an oversimplification. I apologize for the inaccuracy.
>
> Here is the corrected explanation:
>
> Whether the query commits or aborts depends on *when* the server-side process detects the connection is broken, relative to when the query finishes execution.
>
> 1.  **Query Finishes Before Connection Error:** If your long-running statement (e.g., an `UPDATE` or `INSERT`) completes *before* the server-side process attempts a network write that fails (or before TCP keepalives time out), the implicit transaction in autocommit mode **will be committed**. The server process only discovers the connection is broken afterward, when it tries to send the completion status to the client or read the next command. By then, the commit has already happened.
>
> 2.  **Connection Error During Query Execution:** If the query needs to send data back to the client (like a large `SELECT`) and fills the server's TCP send buffer, the process will block. If the connection is broken, this block will eventually end with a network error. This error will cause the backend process to terminate, which rolls back the transaction and aborts the query. This is also what happens if the TCP keepalive mechanism detects the dead connection while the query is still running.
>
> In summary, the other source is correct. There is a race condition. A statement in autocommit mode **can successfully commit** even if the client has disconnected, provided the statement finishes execution before the server's backend process detects the broken connection.


Since this is a huge write-only statement that explanation sounds plausible. That or our `tcp_keepalives_*` settings don't work as expected (they are all configured to 0 server side, but AFAICT our connection pool overrides them to a relatively short timeout).

> So this is not done transactionally? would that not be the solution instead of locks?

The statement wasn't using transactions before. Simply adding an explicit transaction would probably solve the issue in practice, but in theory there could be cases where the app or the trigger is running twice due to some bug or misconfiguration. The advisory lock would additionally protect against that.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
rautenrieth-da: This is an advisory lock. It doesn't conflict with anything except other transactions that use the same advisory lock.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -158,12 +158,42 @@ class AcsSnapshotStore(
...
+    *
+    *  In case the application crashes while holding the lock, the server should close the connection
+    *  and abort the transaction as soon as it detects a disconnect.
+    *  See [[com.digitalasset.canton.platform.store.backend.postgresql.PostgresDataSourceConfig]] for our connection keepalive settings.
+    *  With default settings, the server should detect a dead connection within ~15sec.
```
OriolMunoz-da: will it though? We've seen queries stuck for longer than 15s after a pod kill/restart

meiersi-da: @nmarton-da: I remember that there was some trickery around keep alives that you figured out as part of the HA config for the IndexDB. Something around clients and servers having to set the right params? Do you remember?

meiersi-da: Actually, this ChatGPT convo has quite a bit of interesting info: https://chatgpt.com/share/68dbb7bd-48c8-8007-85e9-ae43fa3a5b54

Seems like it's worth checking our keep-alive configs for server and client -- independently of this PR.

rautenrieth-da: Added https://github.com/hyperledger-labs/splice/issues/2488 for double checking disconnect behavior, and leaving the advisory lock. We'd rather have stray locks than a corrupted database.

rautenrieth-da:
Actual behavior on CILR:

```
2025-10-02 07:11:52.069Z: trigger runs (never finishing)
2025-10-02T07:13:18.317Z: scan shuts down
2025-10-02 07:13:51.000Z: scan restarts
2025-10-02 07:15:14.951Z: trigger runs again on the same snapshot record time
2025-10-02 07:15:14.973Z: trigger fails with java.lang.Exception: Failed to acquire exclusive lock (not retrying the error)
2025-10-02 07:15:46.134Z: trigger runs again on the same snapshot record time
2025-10-02 07:19:12.841Z: trigger succeeds
```

I think that's enough to conclude that it takes a few minutes to release the lock, potentially leading to log warnings, but the trigger does not deadlock itself.

PR #2454, opened 2025-09-29
Add validation and default value for nextScheduledSynchronizerUpgradeTime

fayi-da:
https://github.com/user-attachments/assets/cd28f96b-ea3e-43ec-af14-26a3414a9e72

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/frontend/src/__tests__/governance/forms/set-dso-rules-form.test.tsx
+++ b/apps/sv/frontend/src/__tests__/governance/forms/set-dso-rules-form.test.tsx
@@ -1,7 +1,10 @@
 // Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { dateTimeFormatISO } from '@lfdecentralizedtrust/splice-common-frontend-utils';
+import {
+  dateTimeFormatISO,
+  nextScheduledSynchronizerUpgradeFormat,
+} from '@lfdecentralizedtrust/splice-common-frontend-utils';
 import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import dayjs from 'dayjs';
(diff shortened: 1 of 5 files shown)


apps/sv/frontend/src/__tests__/governance/forms/set-dso-rules-form.test.tsx
```diff
@@ -385,3 +388,146 @@ describe('Set DSO Config Rules Form', () => {
     expect(screen.queryByTestId('config-diffs-display')).toBeDefined();
   });
 });
+
+describe('Next Scheduled Synchronizer Upgrade', () => {
```
fayi-da, line 392: These tests are a bit difficult to follow without context so I've added some comments to make it easier

apps/sv/frontend/src/components/form-components/ConfigField.tsx
```diff
@@ -106,3 +115,29 @@ export const PendingConfigDisplay: React.FC<PendingConfigDisplayProps> = ({ pend
...
+interface NextScheduledSynchronizerUpgradeDisplayProps {
+  effectiveDate: string | undefined;
+  configChange: ConfigChange;
+}
+
+export const NextScheduledSynchronizerUpgradeDisplay: React.FC<
+  NextScheduledSynchronizerUpgradeDisplayProps
+> = props => {
+  const { effectiveDate } = props;
+  const defaultMigrationTime = dayjs(effectiveDate)
```
fayi-da, line 128: The default time we show is the current effective date converted to utc + 1hr

apps/sv/frontend/src/__tests__/governance/forms/set-dso-rules-form.test.tsx
```diff
@@ -385,3 +388,146 @@ describe('Set DSO Config Rules Form', () => {
...
+      <Wrapper>
+        <SetDsoConfigRulesForm />
+      </Wrapper>
+    );
+
+    const errorMessage = 'Upgrade Time must be at least 1 hour after the Effective Date';
+
+    const effectiveDateInput = screen.getByTestId('set-dso-config-rules-effective-date-field');
+    const effectiveDate = dayjs().add(10, 'day').format(dateTimeFormatISO);
+    console.log('tesco effectiveDate', effectiveDate);
```
ray-roestenburg-da: agree with the copilot 😄

apps/sv/frontend/src/components/form-components/ConfigField.tsx
```diff
@@ -6,11 +6,13 @@ import dayjs from 'dayjs';
 import relativeTime from 'dayjs/plugin/relativeTime';
 import { useFieldContext } from '../../hooks/formContext';
 import type { ConfigChange, PendingConfigFieldInfo } from '../../utils/types';
+import { nextScheduledSynchronizerUpgradeFormat } from '@lfdecentralizedtrust/splice-common-frontend-utils';
 
 dayjs.extend(relativeTime);
 
 export interface ConfigFieldProps {
   configChange: ConfigChange;
+  effectiveDate?: string | undefined;
```
ray-roestenburg-da, line 15: Should this not be `effectiveDate?: string` or `effectiveDate: string | undefined`, this seems 'double'

apps/sv/frontend/src/components/forms/formValidators.ts
```diff
@@ -89,3 +89,30 @@ export const validateGrantRevokeFeaturedAppRight = (value: string): string | fal
   const result = grantRevokeFeaturedAppRightSchema.safeParse(value);
   return result.success ? false : result.error.issues[0].message;
 };
+
+export const validateNextScheduledSynchronizerUpgrade = (
+  upgradeTime: string,
+  migrationId: string,
+  effectiveDate: string | undefined
+): string | false => {
```
ray-roestenburg-da, line 97: `string | false` is a very weird type for me.. I think I would even prefer string | undefined then, and change the name to indicate it returns an error or no error. Now validate returns false when it is valid? that's confusing, please change

ray-roestenburg-da: Maybe even better to just have an enum for the cases Empty, Incomplete, and UpgradeTimeIsNotAfterEffectiveDate (better names) and Valid or something like that, then another method that gets a message for the error cases, and no error message for the valid case.

pawelperek-da: note: The `enum` keyword is discouraged in modern TypeScript. From my experience, TS developers seem to have settled on using something like this instead:
```
type SynchronizerUpdateStatus = {
    status: "Success"
} | {
    status: "Error"
    reason: string
}

validateNextScheduledSynchronizerUpgrade(): SynchronizerUpdateStatus
```
although I believe`string | undefined` or `string | null` is good enough, I'd just change the name of the function to reflect that we are checking for an error

pawelperek-da (approved):

apps/sv/frontend/src/components/forms/formValidators.ts
ray-roestenburg-da: yeah I meant a type like an enum in Scala (3) and assumes something relatable existed, your suggestion is close enough for me

apps/sv/frontend/src/__tests__/governance/forms/set-dso-rules-form.test.tsx
fayi-da: oops lol

apps/sv/frontend/src/components/form-components/ConfigField.tsx
fayi-da: Both are required.

Not every field requires this parameter, hence the `?`. When effectiveDate is provided but is "at threshold", the value will be `undefined` hence the `string | undefined`.

ray-roestenburg-da: ok.

apps/sv/frontend/src/components/forms/formValidators.ts
fayi-da: Unfortunately folks, I don't like this type either(I noted it few PRs ago) but I didn't choose it :( . This is how the form lib works. I just extracted these separately for re-use. I think of the function "as this field invalid" which is what the form validator does. "Is this field invalid? if yes tell me why". Not pretty but 🤷 

I have a note to potentially use a zod schema directly for these validators. I don't think it'll work for a not-so-straightforward validation like this one but worth a try. I'll create a tech debt for it.

ray-roestenburg-da: ok at least change the name in this PR and add a comment for this weird type

fayi-da: Do you mind if we still keep the name so it's consistent with the other validators in this file? I't'll make refactoring a little easier later.

https://github.com/hyperledger-labs/splice/issues/2463

ray-roestenburg-da: ok do a follow up PR

PR #2458, opened 2025-09-29
Draft: Token Standards V2

bame-da:
A design to evolve the token standards in a  backwards compatible way.

So far, changes include:
- Configurable controllers for allocations
- Multi-leg allocations
- account info on holdings/transfer destinations
- changes to allocation/transfer timings/expiries

to come:
- multi-leg transfers
- published draft CIP
- completing backwards compatibility code
- Cross version compatibility testing

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2.daml
```diff
@@ -0,0 +1,627 @@
...
+allocation_specification_v2_to_v1 : AllocationV2.AllocationSpecification -> AllocationV1.AllocationSpecification
+allocation_specification_v2_to_v1 AllocationV2.AllocationSpecification{..} = 
+  AllocationV1.AllocationSpecification with
+    settlement = settlement_info_v2_to_v1 settlement
+    transferLegId
+    transferLeg = transfer_leg_v2_to_v1 transferLeg
+  where
+    (transferLegId, transferLeg')::other = if TextMap.size transferLegs > 0
+      then TextMap.toList transferLegs
+      else error "Cannot downcast an allocation specification without any legs."
```
yangzai: Hi @bame-da, I understand that it might be too esoteric to refine the type of `transferLegs` to `NonEmpty`, but maybe the workaround could be to provide the first leg as a separate, non-optional field (essentially keeping the V1 field + V2 list) so that we don't have to throw a runtime error here?

token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2.daml
```diff
@@ -0,0 +1,627 @@
...
+  None -> error ("Cannot parse time " <> t)
+
+insertOptional : Text -> Optional a -> TextMap.TextMap a -> TextMap.TextMap a
+insertOptional k ov m = case ov of
+  None -> m
+  Some v -> TextMap.insert k v m
+-- Holding
+----------
+
+-- Upcast
```
yangzai, line 48: Base on the fact that:
- all V2 should implicitly require V1
- and an initial look at the converters below, where:
  * `XXX_v1_to_v2` is generally unsafe (usage of `fromSome`, coercion of V1 to V2 contracts when V2 might not exist, etc.)
  * and `XXX_v2_to_v1` is generally safe (except for the `error` from the unrefined `transferLegs` below)

`XXX_v1_to_v2`s should be downcasts (unsafe) and `XXX_v2_to_v1`s should be upcasts (safe).

Ref: https://pling.jondgoodwin.com/post/practical-subtyping/#coercion

bame-da: You are right that at the technical level / type theoretical level v2_to_v1 is safe. It cannot fail as it just puts all extra info in metadata.
A v1_to_v2 can fail IF the metadata was manipulated. e.g. metadata says there are 3 transferLegs, but fields for leg 2 are not set. However, both directions are technically safe if:
- The v1 was created without manually setting the specific "reserved" metadata fields
- The metadata was created by the v2_to_v1 function and not manipulated thereafter

However, looking at the _semantic_ side v1_to_v2 is safe. If a v2 implementation is created through v1 interfaces, it will behave exactly as the v1 client that created it would have exptected.
v2_to_v1 is unsafe. The allocation may behave in ways not expected looking at it purely through the lens of the v1 interface (and ignoring all the metadata).

Maybe it's best just to stay away from loaded terms like "Upcast" and "Downcast" here and simply label them V1->V2 and V2->V1 conversions.

token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2.daml
yangzai: My bad, they are now a `TextMap`, which would bring in an issue of the mandatory leg having a duplicate key in the map, which I guess is fine if the overriding rule is standardised.

yangzai: And unfortunately we have no `NonEmptyMap`.

bame-da: I think in practice you can mitigate this runtime error by putting an ensure clause on the V2 implementation and thus require it to be non-empty. This could be added to the standard at the CIP level:

Allocations MUST contain at least one transfer leg.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,292 @@
...
+  receiver : Account
+  admin : Party
+  -- ^ the settlement to which this authorization applies. 
+
+
+
+-- Allocation
+------------------------
+
+-- | Convenience function to refer to the union of sender, receiver, and
```
meiersi-da, line 110: If we can, then we should minimize code in interface definitions, as bugs in it can't be fixed. So here this hints at not providing any defaults at the definition level, but instead provide convenience functions at the implementation level.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,292 @@
...
+allocationAdmin : AllocationSpecification -> Party
+allocationAdmin AllocationSpecification{..} = admin
+  where
+    admins = dedupSort $ map (._2.instrumentId.admin) (toList transferLegs)
+    admin = case admins of
+      [a] -> a
+      _ -> error "Allocations must involve instruments from exactly one admin"
+
+-- | A contract representing an allocation of some amount of asset holdings to
+-- a specific leg of a settlement.
```
meiersi-da: Outdated comment.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,292 @@
...
+    transferLegs: TextMap TransferLeg
+     -- ^ the transfers that this allocation satisfies
+  deriving (Show, Eq)
+
+-- | View of a funded allocation of assets to a specific leg of a settlement.
+-- To be standards compliant, an allocation must have a non-zero number of
+-- transfer legs, all with an instrumentId with the same adminId.
+data AllocationView = AllocationView with
+    allocation : AllocationSpecification
+      -- ^ The settlement for whose execution the assets are being allocated.
```
meiersi-da, line 89: Comment is off

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,292 @@
...
+    -- registry can make it to the settlementInfo's settlementDeadline.
+  deriving (Show, Eq)
+
+-- | View of a delegated authority from the signer to the executor to settle given transfer legs.
+-- authorizers are always the signatories so not made available in the viewtype.
+data AllocationTransferAuthorizationView = AllocationTransferAuthorizationView with
+  allocation : AllocationSpecification
+  receiver : Account
+  admin : Party
+  -- ^ the settlement to which this authorization applies.
```
meiersi-da: comments are off

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,292 @@
...
+interface Allocation where
+  viewtype AllocationView
+
+  allocation_executeTransferImpl : ContractId Allocation -> Allocation_ExecuteTransfer -> Update Allocation_ExecuteTransferResult
+  allocation_cancelImpl : ContractId Allocation -> Allocation_Cancel -> Update Allocation_CancelResult
+  allocation_withdrawImpl : ContractId Allocation -> Allocation_Withdraw -> Update Allocation_WithdrawResult
+  allocation_executeAuthorizeIncomingImpl : ContractId Allocation -> Allocation_AuthorizeIncoming -> Update (ContractId AllocationTransferAuthorization)
+
+  nonconsuming choice Allocation_AuthorizeIncoming : ContractId AllocationTransferAuthorization
+  -- ^ Transfer the authority from this Allocation to a AllocationTransferAuthorization which
```
meiersi-da:
```suggestion
  -- ^ Transfer the authority from this Allocation to an AllocationTransferAuthorization, which
```

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,292 @@
...
+
+  -- NOTE: This used to be consuming. Implementations now must archive self.
+  nonconsuming choice Allocation_ExecuteTransfer : Allocation_ExecuteTransferResult
+    -- ^ Execute the transfer of the allocated assets. Intended to be used to execute the settlement.
+    -- This choice SHOULD succeed provided the `settlement.settleBefore` deadline has not yet passed.
+    with
+      extraArgs : ExtraArgs
+        -- ^ Additional context required in order to exercise the choice.
+      extraAuth : [ContractId AllocationTransferAuthorization]
+        -- ^ Contract encapsulating extra authority needed to exercise the choice
```
meiersi-da: Seems like we support chaining of these, correct? let's either fix type or comment

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,292 @@
...
+  -- ^ Transfer the authority from this Allocation to a AllocationTransferAuthorization which
+  -- can then be used in another allocation's Allocation_ExecuteTransfer to authorize incoming
+  -- transfers.
+    with
+      extraArgs : ExtraArgs
+        -- ^ Additional context required in order to exercise the choice.
+    controller allocationControllers (view this).allocation
+    do allocation_executeAuthorizeIncomingImpl this self arg
+
+  -- NOTE: This used to be consuming. Implementations now must archive self.
```
meiersi-da: TODO: add this comment to the doc comment below.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,292 @@
...
+  with
+    senderHoldingCids : [ContractId Holding]
+      -- ^ The holdings that were released back to the sender.
+    meta : Metadata
+      -- ^ Additional metadata specific to the allocation, used for extensibility.
+  deriving (Show, Eq)
+
+
+-- Utility for matching with metadata
+-------------------------------------
```
meiersi-da: TODO: move that code out of the interface definition

token-standard/splice-api-token-allocation-v2/openapi/allocation-v2.yaml
```diff
@@ -0,0 +1,202 @@
...
+openapi: 3.0.0
+info:
+  title: allocation off-ledger API
+  description: |
+    Implemented by token registries for the purpose of the use and management of
+    allocations by wallets and apps orchestrating the settlement of asset transfers.
+  version: 1.0.0
+paths:
+
+  /registry/allocations/v1/{allocationId}/choice-contexts/execute-transfer:
```
meiersi-da, line 13: TODO: convert to v2

Also consider whether we should/can share the OpenAPI definitions

token-standard/splice-api-token-holding-v2/daml/Splice/Api/Token/HoldingV2.daml
```diff
@@ -0,0 +1,80 @@
...
+      -- ^ The party representing the registry app that administers the instrument.
+    id : Text
+      -- ^ The identifier used for the instrument by the instrument admin.
+      --
+      -- This identifier MUST be unique and unambiguous per instrument admin.
+  deriving (Eq, Ord, Show)
+
+-- | A data type to represent an on-chain managed account,
+--   for example in a traditional accounting structure, or simply a delegation
+--   to a provider to perform some services.
```
meiersi-da, line 24:
```suggestion
-- for example in a traditional accounting structure, or simply a delegation
-- to a provider to perform some services.
```
nit

daml/splice-amulet/daml/Splice/AmuletTransferInstruction.daml
```diff
@@ -41,39 +43,52 @@ template AmuletTransferInstruction
...
+      receiverHoldingCids = (map coerceInterfaceContractId receiverHoldingCids)
     meta
 
 abortAmuletTransferInstruction : AmuletTransferInstruction -> ExtraArgs -> Update TransferInstructionResult
 abortAmuletTransferInstruction instr extraArgs = do
   let twoStepTransfer = standardTransferToTwoStepTransfer instr.transfer
   senderHoldingCids <- abortTwoStepTransfer twoStepTransfer instr.lockedAmulet extraArgs
   pure TransferInstructionResult with
-    senderChangeCids = senderHoldingCids
+    -- TODO! This will change once there's a transfer instruction v2
```
meiersi-da, line 91: left-over?

daml/splice-amulet/daml/Splice/ExternalPartyAmuletRules.daml
```diff
@@ -276,6 +294,12 @@ amulet_transferFactory_transferImpl this _self arg = do
   -- inputHoldingCids: note that their detailed validation is done in the transfer itself
   require "At least one holding must be provided" (not $ null transfer.inputHoldingCids)
 
+  -- accounts :
+  require "sender account provider must be None" (isNone transfer.sender.provider)
+  require "sender account id must be None" (isNone transfer.sender.id)
```
meiersi-da, line 304: TODO: factor out into utils `isBasicAccount`

daml/splice-wallet/daml/Splice/Wallet/Install.daml
```diff
@@ -585,69 +586,70 @@ template WalletAppInstall
     -- This is a dedicated choice rather than going through AmuletOperation as it behaves a bit differently than all the other operations:
     -- 1. We cannot use rewards as inputs.
     -- 2. We can use expired locked amulets as inputs.
-    nonconsuming choice WalletAppInstall_TransferFactory_Transfer : Splice.Api.Token.TransferInstructionV1.TransferInstructionResult
+    nonconsuming choice WalletAppInstall_TransferFactory_Transfer : Splice.Api.Token.TransferInstructionV2.TransferInstructionResult
```
meiersi-da, line 589: TODO: add new choices so we have upgrade compat

token-standard/splice-api-token-allocation-request-v2/daml/Splice/Api/Token/AllocationRequestV2.daml
```diff
@@ -0,0 +1,65 @@
...
+        -- ^ Additional context required in order to exercise the choice.
+    controller (view this).settlement.executor
+    do allocationRequest_WithdrawImpl this self arg
+
+
+-- | View of `AllocationRequest`.
+--
+-- Implementations SHOULD make sure that at least all senders of the transfer legs
+-- are observers of the implementing contract, so that their wallet can show
+-- the request to them.
```
meiersi-da, line 51: Comment needs updating given that receivers now need to authorize.

token-standard/splice-api-token-holding-v2/daml/Splice/Api/Token/HoldingV2.daml
```diff
@@ -0,0 +1,80 @@
...
+    provider : Optional Party
+    -- ^ The provider of the account.
+    id : Optional Text
+    -- ^ Account number or similar. Should be unique per provider.
+    owner: Party
+    -- ^ The party that owns the account.
+  deriving (Eq, Show, Ord)
+
+basicAccount : Party -> Account
+basicAccount p = Account None None p
```
meiersi-da, line 36: consider pushing to utils and just follow the simple rule to only have code in interface definitions that is absolutely required for their definition.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,292 @@
...
+      -- ^ The extra account authorization this allocation needs for settlement on top of
+      -- the controllers specified via allocation.settlement.controllerOverride
+    expiresAt : Optional Time
+    -- ^ A time at which this allocation expires. SHOULD be as close as the
+    -- registry can make it to the settlementInfo's settlementDeadline.
+  deriving (Show, Eq)
+
+-- | View of a delegated authority from the signer to the executor to settle given transfer legs.
+-- authorizers are always the signatories so not made available in the viewtype.
+data AllocationTransferAuthorizationView = AllocationTransferAuthorizationView with
```
Sengoku11: Hi, missing ‘deriving (Show, Eq)’ for this view. Results in failing queryInterface

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,438 @@
...
+-- | A contract representing a delegated authority from the signatories to the executor to settle 
+-- given transfer legs.
+interface AllocationTransferAuthorization where
+  viewtype AllocationTransferAuthorizationView
+
+  allocationTransferAuthorization_transferImpl : ContractId AllocationTransferAuthorization -> AllocationTransferAuthorization_Transfer -> Update Allocation_ExecuteTransferResult
+
+  choice AllocationTransferAuthorization_Transfer : Allocation_ExecuteTransferResult
+  -- ^ Execute the transfer of the allocated assets. Intended to be used to execute the settlement.
+  -- This choice SHOULD succeed provided the `settlement.settleBefore` deadline has not yet passed.
```
meiersi-da: This interface is missing the choices to abort the workflow in case the settlement did not succeed. Ideally, we could make this a transient contract, but that's not supported in Daml.

However, what I'm wondering about is why we even need the separate `AllocationTransferAuthorization` interface. What would it take to push the improved transfer choice to the `Allocation` itself?

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,438 @@
...
+              extraObservers = if isSome auth_r then tl.receiver.owner::(optionalToList tl.receiver.provider) else []
+            return (r, sls)
+        -- Update the holdings maps
+        let 
+          hs' =
+            Map.insert (tl.instrumentId, tl.sender) res.senderHoldingCids $
+            Map.insert (tl.instrumentId, tl.receiver) 
+              (res.receiverHoldingCids ++ fromOptional [] (Map.lookup (tl.instrumentId, tl.receiver) hs)) $
+            hs
+          stls' = dedupSort (settled_legs ++ stls)
```
meiersi-da: Let's use a `Set` for this.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,438 @@
...
+
+settleAllocations : 
+  [TransferLeg] ->
+  Map (InstrumentId, Account) [ContractId Holding] ->
+  Map (Party, Account) (ContractId Allocation, ExtraArgs) ->
+  Update ((Map (InstrumentId, Account) [ContractId Holding]), [Metadata])
+settleAllocations 
+  transferLegs 
+  -- ^ transferLegs specifies he execution order of transfer legs.
+  -- The boolean indicates whether receiver auth is needed.
```
meiersi-da: Outdated comment? there's no boolean anymore.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,438 @@
...
+  matchLegs expected.transferLegs received.transferLegs
+
+-- Utility for settling allocations using the recommended algorithm
+--------------------------------------------------------------------
+-- TODO: Move away from interfaces to make easier to upgrade?
+
+settleAllocations : 
+  [TransferLeg] ->
+  Map (InstrumentId, Account) [ContractId Holding] ->
+  Map (Party, Account) (ContractId Allocation, ExtraArgs) ->
```
meiersi-da:
```suggestion
  Map Account (ContractId Allocation, ExtraArgs) ->
```

seems like the party is not used

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,438 @@
...
+  transferLegs 
+  -- ^ transferLegs specifies he execution order of transfer legs.
+  -- The boolean indicates whether receiver auth is needed.
+  holdings
+  -- ^ holdings are all input holdings, sorted by instrumentId. Typically these
+  -- match the instrumentIds on the allocations.
+  allocationsWithContext 
+  -- ^ these are the allocations to be used, grouped by admin and account.
+  = do
+    -- Start by sorting all transferauthorizations into a map.
```
meiersi-da: Missing assertion that leg-ids are unique.

daml/splice-amulet/daml/Splice/AmuletAllocation.daml
```diff
@@ -47,33 +72,65 @@ template AmuletAllocation
...
-    dso = allocationInstrumentAdmin allocation
-    sender = allocation.transferLeg.sender
-    receiver = allocation.transferLeg.receiver
-    amount = allocation.transferLeg.amount
-    provider = allocation.settlement.executor
-    transferBefore = allocation.settlement.settleBefore
-    transferBeforeDeadline = "allocation.settlement.settleBefore"
+    dso = admin
+    sender
+    outputs = map (\tl -> (tl.receiver.owner, tl.amount)) filteredLegs
```
meiersi-da: AFAIS, all legs on the same `AmuletAllocationTransferAuthorization` have the same receiver. Why don't we just create a single output for the sum of these?

token-standard/splice-token-standard-test-v2/daml/Splice/Tests/TestAmuletTokenDvP.daml
```diff
@@ -0,0 +1,375 @@
...
+          context = context.choiceContext
+          meta = beneficiaryMetadata
+    pure (allocCid, extraArgs, context.disclosures, alloc.holdingCids)
+  let 
+    disclosures = foldMap ((._3) . snd) (Map.toList richAllocationsWithContext)
+    holdingCids = Map.fromList $ 
+      concatMap 
+        (\((_, authorizer), (_, _, _, holdings_map)) -> 
+          map (\(instrId, hs') -> ((instrId, authorizer), hs')) (Map.toList holdings_map)) 
+        (Map.toList richAllocationsWithContext)
```
meiersi-da: This feels rather complex. Why can't we push the reference to the holdings to use for a transfer into the `AllocationTransferAuthorization` itself? Then every transfer can use the holdings from the previous step plus the holdings it references itself to fund the transfer; and the settlement code can be run by default without passing in any extra holdings.

daml/splice-amulet/daml/Splice/AmuletAllocation.daml
```diff
@@ -47,33 +72,65 @@ template AmuletAllocation
             senderHoldingCids
             meta = emptyMetadata
 
-allocationInstrumentAdmin : AllocationSpecification -> Party
-allocationInstrumentAdmin AllocationSpecification{..} = transferLeg.instrumentId.admin
-
+      allocation_getTransferAuthorizationsImpl _self Allocation_GetTransferAuthorizations{..} = do
+        let
+          grouped_tls = groupOn (\tl -> (tl.sender, tl.receiver)) allocation.transferLegs
```
meiersi-da: I'm not sure this works, as it can lead to executing underfunded transfers. Example:

1. A -> B: 10 CC
2. C -> A: 20 CC
3. A -> B: 10 CC

AFAIU your code it will execute legs 1 and 3 in a single transfer. However, the available holdings will only cover 10 CC.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
meiersi-da: I'm even wondering whether we couldn't go farther and create one allocation per transfer leg for the sender, and when required, another one per transfer for the receiver.

We however change the `AllocationFactory_Allocate` choice such that it creates all allocations for the same admin in a single call. AFAIS, this way we can also skip over the introduction of `AllocationTransferAuthorization`s; and instead have the `Allocation`s take that role.

Settlement proceeds then one leg after the other always using the one or two allocations matching the leg. 

I'd put the smarts for the bulk creation of all allocations for a given trade into a modified `WalletUserProxy` contract, so that wallets can instruct them all in one go. The user's node will have the .dars from all asset registries vetted, and can thus create all of the allocations in a single transaction.

daml/splice-amulet/daml/Splice/AmuletAllocation.daml
```diff
@@ -3,35 +3,60 @@
...
   where
-    signatory allocationInstrumentAdmin allocation, allocation.transferLeg.sender
+    signatory admin, sender
     observer allocation.settlement.executor
 
+    ensure 
+      all 
+        (\tl -> (
+            -- Sender needs to appear as sender or receiver of each leg.
+            sender `elem` [tl.sender.owner, tl.receiver.owner])
```
meiersi-da: `authorizer` seems the better name, as they can be both sender or receiver.

token-standard/splice-api-token-allocation-request-v2/daml/Splice/Api/Token/AllocationRequestV2.daml
```diff
@@ -0,0 +1,63 @@
...
+data AllocationRequestView = AllocationRequestView with
+    settlement : SettlementInfo
```
meiersi-da, line 55: This means the trading apps specify the controller override. Is this intentional? How will they make that decision? 

Probably answered in the CIP text, but need to check.

meiersi-da: It will be important for wallets to highlight overrides that do not include the wallet user party, as that means they are delegating the checking of atomic execution of the whole trade to controllers.

Consider whether we can choose a more evocative name; e.g., `settlementValidators`

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -0,0 +1,438 @@
...
+    settleAt : Time
+      -- ^ The earliest settlement time. Allocations should be made before this time. Settlement
+      -- happens at any point after this time.
+    meta : Metadata
+      -- ^ Additional metadata about the settlement, used for extensibility.
+    controllerOverride : Optional [Party]
+      -- ^ Additional party signatures in addition to executor that need to authorize
+      -- choices on the Allocation. Defaults to senders and receivers across all 
+      -- transferLegs.
+    settlementDeadline : Optional Time
```
meiersi-da, line 51: For Amulet, allowing for not setting a `settlementDeadline` is a DoS vector. Can we allow the registries to set an upper limit wrt for how long they track the state of in-flight settlements?

meiersi-da: and reject the creation of allocations with a `settlementDeadline` too far in the future?

meiersi-da: Ah just saw that the registry can set their own expiry. That sounds like a fine option.

Wallets can implement an extension choice by withdrawing the allocation and creating a new one immediately with the `senderHoldingCids` received from withdrawing.

PR #2476, opened 2025-09-30
fix exposing secret to the non-root images

isegall-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/5810

I'll rework CCI to allow testing the reonbaording workflow on a splice branch against a scratchnet, but happy for a review in the meantime so I can merge once confirmed to work.

The problem with the previous PR is that it tries to set ownership of the files in the mountpoint, but when mounting a secret, that's not a real FS, and is therefore read-only.

The approach I take here is to copy its content into an actual FS, which can then be chown'ed.

The alternative is to set fsGroup on the pod, which apparently is supposed to also set the owner of the mount point, but some internet dudes claim that's not universally supported, so I'm unsure how safe it is. Gemini suggested this approach as the second "common practice". It should work AFAICT. Do you happen to know what the common practice is here?

Successful SV reonboard test: https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/34258/workflows/ea9bb3be-3741-4203-bb40-a576ad99a901

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/helm/splice-sv-node/templates/sv.yaml
+++ b/cluster/helm/splice-sv-node/templates/sv.yaml
@@ -424,7 +424,7 @@ spec:
         {{- if .Values.migration.attachPvc }}
         - name: chown-domain-upgrade-dump
           image: busybox:1.37.0
-          command: ["chown", "-R", "1001:1001", "/domain-upgrade-dump"]
+          command: ["sh", "-c", "chown -R 1001:1001 /domain-upgrade-dump"]
           volumeMounts:
             - name: domain-upgrade-dump-volume
               mountPath: /domain-upgrade-dump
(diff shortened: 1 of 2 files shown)


cluster/helm/splice-sv-node/templates/sv.yaml
```diff
@@ -414,8 +414,10 @@ spec:
         {{- if .Values.participantIdentitiesDumpImport }}
         - name: chown-participant-bootstrapping-dump
           image: busybox:1.37.0
-          command: ["sh", "-c", "chown -R 1001:1001 /participant-bootstrapping-dump"]
+          command: ["sh", "-c", "cp -R /participant-bootstrapping-dump-secret/* /participant-bootstrapping-dump && chown -R 1001:1001 /participant-bootstrapping-dump"]
```
martinflorian-da: Maybe add a comment about why we need that extra detour

moritzkiefer-da (approved):
worth a try thx

isegall-da:
> @isegall-da
> 
> > tries to set ownership
> 
> Do you understand why it needs that?

We're using a non-root user in the images now, so need the files that it reads to be readable by that user.

PR #2478, opened 2025-09-30
Fix scan_txlog.py handling of holding fees

OriolMunoz-da:
Fixes #2380

Testing locally on 3.4...

### DIFF
--- a/scripts/scan-txlog/scan_txlog.py
+++ b/scripts/scan-txlog/scan_txlog.py
@@ -417,6 +417,18 @@ class DamlDecimal:
 class KnownPackageIds:
     amulet_0_1_0 = "48cac5ba4b6bf78df6c3a952ce05409a1d2ef39c05351074679adc0cf9cd1351"
     amulet_0_1_1 = "67bc95402ad7b08fcdff0ed478308d39c70baca2238c35c5d425a435a8a9e7f7"
+    amulet_0_1_2 = "1446ffdf23326cef2de97923df96618eb615792bea36cf1431f03639448f1645"
+    amulet_0_1_3 = "0d89016d5a90eb8bced48bbac99e81c57781b3a36094b8d48b8e4389851e19af"
+    amulet_0_1_4 = "a36ef8888fb44caae13d96341ce1fabd84fc9e2e7b209bbc3caabb48b6be1668"
+    amulet_0_1_5 = "b4867a47abbfa2d15482f22c9c50516f0af14036287f299342f5d336391e4997"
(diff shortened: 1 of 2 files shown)


scripts/scan-txlog/scan_txlog.py
```diff
@@ -1513,6 +1544,7 @@ def summary(self, subtract_holding_fees_per_round):
             output += ["amulets:"]
             for amulet in self.amulets:
                 amount = amulet.payload.get_amulet_amount()
+                subtract_holding_fees_per_round = KnownPackageIds.deducts_holding_fees(amulet.template_id.package_id)
```
moritzkiefer-da: this doesn't look right to me. The logic does not depend on the template id of the amulet. It depends on the template id in the ExerciseNode of AmuletRules_Transfer

scripts/scan-txlog/scan_txlog.py
```diff
@@ -2389,7 +2421,7 @@ def handle_transfer(self, transaction, event, description="Transfer"):
             initial_amulet_cc_input,
             amulet_cc_input,
             all_inputs,
-        ) = transfer_inputs.summary(self.args.subtract_holding_fees_per_round)
+        ) = transfer_inputs.summary(event.template_id.package_id)
```
OriolMunoz-da: amuletrules_transfer

scripts/scan-txlog/scan_txlog.py
```diff
@@ -2599,7 +2631,7 @@ def handle_buy_member_traffic(self, transaction, event):
...
```
OriolMunoz-da: AmuletRules_BuyMemberTraffic

scripts/scan-txlog/scan_txlog.py
```diff
@@ -2811,7 +2843,7 @@ def handle_transfer_preapproval_purchase(
...
```
OriolMunoz-da: AmuletRules_CreateExternalPartySetupProposal or AmuletRules_CreateTransferPreapproval or TransferPreapproval_Renew
(all in splice-amulet)

scripts/scan-txlog/scan_txlog.py
moritzkiefer-da: have you considered making `summar` take an `ExercisedEvent` as the argument and then read the package id from there? That feels like it may be more robust against e.g. the issue where you used the template id of the amulet.

moritzkiefer-da (approved):
Nice thanks for the fix! I assume you tested it against the failing test on 3.4.

PR #2479, opened 2025-09-30
Fix git repository definition for extra validators

moritzkiefer-da:
Did not test beyond expected files but the validator runbook is unchanged and the extra validator is now there as well so that seems like progress.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/deployment/src/stacks/splice.ts
+++ b/cluster/pulumi/deployment/src/stacks/splice.ts
@@ -5,8 +5,8 @@ import {
   CLUSTER_BASENAME,
   config,
   DeploySvRunbook,
-  DeployValidatorRunbook,
 } from '@lfdecentralizedtrust/splice-pulumi-common';
+import { deployedValidators } from '@lfdecentralizedtrust/splice-pulumi-common-validator';
 import {
   mustInstallSplitwell,
   mustInstallValidator1,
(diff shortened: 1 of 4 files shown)


cluster/pulumi/deployment/src/stacks/splice.ts
```diff
@@ -28,9 +28,9 @@ export function getSpliceStacksFromMainReference(): StackFromRef[] {
   if (config.envFlag('SPLICE_DEPLOY_MULTI_VALIDATOR', false)) {
     ret.push({ project: 'multi-validator', stack: `multi-validator.${CLUSTER_BASENAME}` });
   }
-  if (DeployValidatorRunbook) {
-    ret.push({ project: 'validator-runbook', stack: `validator-runbook.${CLUSTER_BASENAME}` });
-  }
+  deployedValidators.forEach(validator => {
+    ret.push({ project: 'validator-runbook', stack: `${validator}.${CLUSTER_BASENAME}` });
```
nicu-da: stack should be the same as here https://github.com/hyperledger-labs/splice/blob/main/cluster/pulumi/deployment/src/stacks/validators.ts#L22 :cry:

isegall-da: Why did we do that and not make them consistent?

nicu-da: follows the same structure as sv-canton and makes the stack name more evident when the project is not included, as sometimes pulumi uses it without the full name. 
the idea is to make it consistent by eventually just switching the validator-runbook to use the same naming but we hardcode that in a lot of parts in our ci setup

moritzkiefer-da: fixed, we should just change `deployeValidators` to be a list of objects with all fields we need instead of inlining special casing based on the name all over the place. But not for this pr

nicu-da: yeah, it started with a single case and then it was needed in another place and now it's the magic third time we need it

PR #2484, opened 2025-10-01
Tie validator onboarding secrets to a party hint

### LINKED ISSUE #1603: Tie validator onboarding secrets to a party hint and improve UI
Currently onboarding secrets work like this:

1. The SV generates a random secret. This is what is persisted on ledger in the onboarding contract as base64 encoding.
2. We base64 encode that together with the sv party id https://github.com/hyperledger-labs/splice/blob/4aa2cea6202f1d7ac6e773d89933489438d07f0b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvHandler.scala#L85. Note that this is not currently persisted. The sv party id never changes so we can generate it on the fly as needed.
3. This is then shared with the validator operator.
4. When the validator onboards we validate the SV party against the SV app and the secret against the on-ledger state.

The party hint is slightly different in that we need to persist it contrary to the sv party.

Concretely I would expect we can accomplish this by:

1. Extend the SV UI to specify a party hint when generating an onboarding secret.
2. When creating the on-ledger ``ValidatorOnboarding` contract don't just include the random secret but include the full JSON object as a secret `{"sv": …, "validator_party_hint": "…", "secret": ""}` in the `candidateSecret` field of the contract (no Daml changes required that way).
3. When decoding, check if the first character is a `{` which must be the case for the JSON encoding and can never be the case for the old base64 encoding so we can be backwards compatible with existing secrets. If it is, decode it as json, otherwise decode it as the old format.
4. When decoding as the new format, validate the secret against the party hint in `/v0/onboard/validator`. If it mismatch, fail, otherwise continue.

In addition to that:
In the SV UI, provide a way to copy not just a secret but a text consisting of:

```
your secret: base64 encoded json secret
network: dev/test/mainnet
sv sponsor url: sv sponsor url they need to use in their onboarding
```

This can be used by the SV sponsor to copy directly into a slack message to send to the  validator they are trying to onboard.

pawelperek-da:
Fixes #1603 

https://github.com/user-attachments/assets/ae28dcaf-8c80-4540-be3b-e82325b72421

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvOnboardingIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvOnboardingIntegrationTest.scala
@@ -165,7 +165,7 @@ class SvOnboardingIntegrationTest extends SvIntegrationTestBase {
       )
     }
 
-    clue("try to onboarding with a secret with wrong party hint") {
+    clue("try to onboard with a secret that has an incorrect party hint") {
       val partyId = PartyId
         .fromProtoPrimitive("invalid-name-1::dummy", "partyId")
         .getOrElse(sys.error("Could not parse PartyId"))
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvOnboardingIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvOnboardingIntegrationTest.scala
@@ -107,9 +107,12 @@ class SvOnboardingIntegrationTest extends SvIntegrationTestBase {
     val svParty = sv.getDsoInfo().svParty
     val sv1Party = sv1Backend.getDsoInfo().svParty
     sv.listOngoingValidatorOnboardings() should have length 0
+
+    val name = "dummy" + env.environment.config.name.getOrElse("")
+
     val (secret, onboardingContract) = actAndCheck(
       "the sv operator prepares the onboarding", {
-        sv.prepareValidatorOnboarding(1.hour)
+        sv.prepareValidatorOnboarding(1.hour, Some(name))
       },
     )(
       "a validator onboarding contract is created",
@@ -152,11 +155,29 @@ class SvOnboardingIntegrationTest extends SvIntegrationTestBase {
           ValidatorOnboardingSecret(
             sv1Party,
             onboardingContract.payload.candidateSecret,
+            None,
           ).toApiResponse,
           contactPoint,
         ),
         _.errorMessage should include(
           s"Secret is for SV $sv1Party but this SV is $svParty, validate your SV sponsor URL"
+        ),
+      )
+    }
+
+    clue("try to onboarding with a secret with wrong party hint") {
+      val partyId = PartyId
+        .fromProtoPrimitive("invalid-name-1::dummy", "partyId")
+        .getOrElse(sys.error("Could not parse PartyId"))
+
+      assertThrowsAndLogsCommandFailures(
+        sv.onboardValidator(
+          partyId,
+          onboardingContract.payload.candidateSecret,
+          contactPoint,
+        ),
+        _.errorMessage should include(
+          s"The onboarding secret entered does not match the secret issued for validatorPartyHint: $name"
         ),
       )
     }
--- a/apps/sv/frontend/src/__tests__/sv.test.tsx
+++ b/apps/sv/frontend/src/__tests__/sv.test.tsx
@@ -12,10 +12,14 @@ import { test, expect, describe } from 'vitest';
 
 import App from '../App';
 import { SvConfigProvider } from '../utils';
+import { onboardingInfo } from '../components/ValidatorOnboardingSecrets';
 import { svPartyId, voteRequests } from './mocks/constants';
 import { server, svUrl } from './setup/setup';
 import { changeAction } from './helpers';
-import { dateTimeFormatISO } from '@lfdecentralizedtrust/splice-common-frontend-utils';
+import {
+  dateTimeFormatISO,
+  getUTCWithOffset,
+} from '@lfdecentralizedtrust/splice-common-frontend-utils';
 import dayjs from 'dayjs';
 import { dsoInfo } from '@lfdecentralizedtrust/splice-common-test-handlers';
 
@@ -48,6 +52,64 @@ describe('SV user can', () => {
     render(<AppWithConfig />);
 
     await screen.findByText('You are on ScratchNet');
+  });
+
+  test('browse to the validator onboarding tab', async () => {
+    const user = userEvent.setup();
+    render(<AppWithConfig />);
+
+    expect(await screen.findByText('Validator Onboarding')).toBeDefined();
+    await user.click(screen.getByText('Validator Onboarding'));
+
+    expect(await screen.findByText('Validator Onboarding Secrets')).toBeDefined();
+  });
+
+  test('create a new validator secret with party hint', async () => {
+    const user = userEvent.setup();
+    render(<AppWithConfig />);
+
+    expect(await screen.findByText('Validator Onboarding')).toBeDefined();
+    await user.click(screen.getByText('Validator Onboarding'));
+
+    const partyHintInput = screen.getByTestId('create-party-hint');
+    await user.type(partyHintInput, 'wrong-input');
+
+    expect(screen.getByTestId('create-validator-onboarding-secret').hasAttribute('disabled')).toBe(
+      true
+    );
+
+    await user.clear(partyHintInput);
+    await user.type(partyHintInput, 'correct-input-123');
+
+    expect(screen.getByTestId('create-validator-onboarding-secret').hasAttribute('disabled')).toBe(
+      false
+    );
+  });
+
+  test('validator onboarding info has correct format', () => {
+    const validatorOnboardingInfo = onboardingInfo(
+      {
+        partyHint: 'splice-client-2',
+        secret: 'exampleSecret',
+        expiresAt: '2020-01-01 13:57',
+      },
+      'testnet'
+    );
+
+    expect(validatorOnboardingInfo).toBe(
+      `
+splice-client-2
+Network: testnet
+SPONSOR_SV_URL
+http://localhost:3000
+
+Secret
+exampleSecret
+
+Expiration
+2020-01-01 13:57 (${getUTCWithOffset()})
+    `.trim()
+    );
   });
 
   test('browse to the governance tab', async () => {
--- a/apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
+++ b/apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
@@ -57,12 +57,11 @@ const ValidatorOnboardingSecrets: React.FC = () => {
     return <p>Error, something went wrong while fetching onboarding secrets.</p>;
   }
 
-  const validatorOnboardings = validatorOnboardingsQuery.data.toSorted((a, b) => {
-    return (
+  const validatorOnboardings = validatorOnboardingsQuery.data.toSorted(
+    (a, b) =>
       new Date(b.contract.payload.expiresAt).valueOf() -
       new Date(a.contract.payload.expiresAt).valueOf()
-    );
-  });
+  );
 
   return (
     <Stack mt={4} spacing={4} direction="column" justifyContent="center">
@@ -107,8 +106,7 @@ const ValidatorOnboardingSecrets: React.FC = () => {
           fullWidth
           size="large"
           onClick={() => {
-            prepareOnboardingMutation.mutate(partyHint);
-            setPartyHint('');
+            prepareOnboardingMutation.mutateAsync(partyHint).then(() => setPartyHint(''));
           }}
         >
           Create a validator onboarding secret
--- a/apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
+++ b/apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
@@ -5,12 +5,21 @@ import {
   DisableConditionally,
   Loading,
   SvClientProvider,
-  CopyableTypography,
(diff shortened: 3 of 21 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvOnboardingIntegrationTest.scala
```diff
@@ -160,6 +164,23 @@ class SvOnboardingIntegrationTest extends SvIntegrationTestBase {
         ),
       )
     }
+
+    clue("try to onboarding with a secret with wrong party hint") {
```
ray-roestenburg-da:
```suggestion
    clue("try to onboard with a secret that has an incorrect party hint") {
```

apps/sv/frontend/src/__tests__/sv.test.tsx
```diff
@@ -50,6 +54,64 @@ describe('SV user can', () => {
...
+    const user = userEvent.setup();
+    render(<AppWithConfig />);
+
+    expect(await screen.findByText('Validator Onboarding')).toBeDefined();
+    await user.click(screen.getByText('Validator Onboarding'));
+
+    const partyHintInput = screen.getByTestId('create-party-hint');
+    await user.type(partyHintInput, 'wrong-input');
+
+    expect(screen.getByTestId('create-validator-onboarding-secret').hasAttribute('disabled')).toBe(
```
fayi-da, line 77: Surprised this works for you. I've had issues with hasAttribute so I stayed away from it.

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
```diff
@@ -5,28 +5,47 @@ import {
...
 import { useValidatorOnboardings } from '../hooks/useValidatorOnboardings';
 import { useSvConfig } from '../utils';
+import { useNetworkInstanceName } from '../hooks';
+import dayjs from 'dayjs';
+import {
+  dateTimeFormatISO,
+  getUTCWithOffset,
+} from '@lfdecentralizedtrust/splice-common-frontend-utils';
+
+const VALID_PARTY_ID_REGEX = /^[^-]+-[^-]+-\d+$/;
```
fayi-da, line 37: Let's put this in a util file in the common package. I have a feeling we'll be reusing this again.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/util/Secrets.scala
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+  private def decodeBase64Secret(str: String): Option[ValidatorOnboardingSecret] =
+    Try(Base64.getDecoder.decode(str)).toOption.flatMap { decodedBytes =>
+      val decodedStr = new String(decodedBytes, StandardCharsets.UTF_8)
+      decode[ValidatorOnboardingSecret](decodedStr).toOption
+    }
+
+  // There are two ways to create secrets:
+  // 1. Through the UI/API endpoints. The actual secret (excluding the wrapper that adds the SV party id) is a base64 encoded string of length 30.
```
ray-roestenburg-da: is it always length 30? do you verify this somewhere, or depend on it?

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
```diff
@@ -5,28 +5,47 @@ import {
...
 
 const ValidatorOnboardingSecrets: React.FC = () => {
   const ONBOARDING_SECRET_EXPIRY_IN_SECOND = 172800; // We allow validator to be onboarded in 48 hours
```
fayi-da, line 40: This should go in the same util file too.

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
```diff
@@ -38,7 +57,7 @@ const ValidatorOnboardingSecrets: React.FC = () => {
     return <p>Error, something went wrong while fetching onboarding secrets.</p>;
   }
 
-  const validatorOnboardings = validatorOnboardingsQuery.data.sort((a, b) => {
+  const validatorOnboardings = validatorOnboardingsQuery.data.toSorted((a, b) => {
     return (
```
fayi-da: Would something like this suffice here?

```ts
 return dayjs(b.contract.payload.expiresAt) - dayjs(a.contract.payload.expiresAt);
```

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
```diff
@@ -57,58 +76,152 @@ const ValidatorOnboardingSecrets: React.FC = () => {
...
       <DisableConditionally
-        conditions={[{ disabled: prepareOnboardingMutation.isPending, reason: 'Loading...' }]}
+        conditions={[
+          { disabled: prepareOnboardingMutation.isPending, reason: 'Loading...' },
+          { disabled: partyHint === '', reason: 'No Party Hint' },
+          {
+            disabled: !VALID_PARTY_ID_REGEX.test(partyHint),
+            reason: 'Party Hint is invalid',
+            severity: 'warning',
+          },
```
fayi-da, line 99: Do we need both conditions for partyHint? I think the second one is adequate. The empty string will fail the regex check

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
```diff
@@ -57,58 +76,152 @@ const ValidatorOnboardingSecrets: React.FC = () => {
...
           id="create-validator-onboarding-secret"
+          data-testid="create-validator-onboarding-secret"
           variant="pill"
           fullWidth
-          disabled={prepareOnboardingMutation.isPending}
           size="large"
-          onClick={() => prepareOnboardingMutation.mutate()}
+          onClick={() => {
+            prepareOnboardingMutation.mutate(partyHint);
+            setPartyHint('');
```
fayi-da: Probably not important but doesn't this clear the field if the request fails? I would have expected to mutateAsync and clear on success, show error on failure.

fayi-da: I think on failure, you  actually end up on the error boundary page. Not sure how important that is though

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/util/Secrets.scala
pawelperek-da: Actually, this is an incorrect figure, the length is 32 and not 30: https://github.com/hyperledger-labs/splice/blob/e7ba904872afba127fea4196e021edd105dd2828/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/util/SvUtil.scala#L59

I don't depend on it anywhere, it's just an implementation detail

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
```diff
@@ -57,58 +76,152 @@ const ValidatorOnboardingSecrets: React.FC = () => {
...
+  info += '\n\n';
+
+  info += 'Secret\n';
+  info += secret;
+  info += '\n\n';
+
+  info += 'Expiration\n';
+  info += `${dayjs(expiresAt).format(dateTimeFormatISO)} (${getUTCWithOffset()})`;
+
+  return info;
```
fayi-da, line 171: Can something like this work too?

```ts
 return [
    partyHint,
    `Network: ${instanceName ?? 'localnet'}`,
    `SPONSOR_SV_URL\n${window.location.origin}`,
    `Secret\n${secret}`,
    `Expiration\n${dayjs(expiresAt).format(dateTimeFormatISO)} (${getUTCWithOffset()})`
  ].join('\n\n');
```

apps/sv/src/main/openapi/sv-internal.yaml
```diff
@@ -633,6 +633,8 @@ components:
           description: |
             The contract storing the onboarding secret.
           $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/schemas/Contract"
+        party_hint:
```
fayi-da, line 636: Need documentation for this

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/util/Secrets.scala
ray-roestenburg-da: Ok maybe don't mention it in the comment since that will easily get outdated.

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
ray-roestenburg-da: This is specific to onboarding right? why put that in a util?

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
ray-roestenburg-da: Put it in there when you reuse it, not when you have no usage of it yet, would be my advice. (and even then I would imagine it makes more sense to reuse within sv app and not across apps)

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
fayi-da: Until we need to show something elsewhere like we always have. Not a big deal for me, we can always refactor later if needed.

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
fayi-da: In the past, i've seen regexes, numbers(e.g duration) used in sv and scan for instance but duplicated. Happy to leave for now

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
ray-roestenburg-da: Prefer to not put things in util that are specific to sv (this seems like one), can reuse within sv once it's needed. Or define some more specific lib than the util catch all thing (yeah it's still a pet peeve of mine)

apps/sv/frontend/src/__tests__/sv.test.tsx
pawelperek-da: Yeah, I think I know what you mean. The problem is that `<DisableConditionally>` component doesn't just modify `disabled` prop in button, but destroys the button and creates identical one but with `disabled={false}`, and that's why we cannot use 
```typescript
const button = screen.getByTestId("a-button")
expect(button.hasAttribute("disabled")).toBe(true)
// ... fill the form
expect(button.hasAttribute("disabled")).toBe(false)
```
but this should work: 
```typescript
expect(screen.getByTestId("a-button").hasAttribute("disabled")).toBe(true)
// ... fill the form
expect(screen.getByTestId("a-button").hasAttribute("disabled")).toBe(false)
```
...or maybe that's not what you are referring to 😄

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
pawelperek-da: Yeah, in theory regex is sufficient but I feel like this version has slightly better UX by not annoying users with an immediate warning. The user is going to be annoyed slightly later when they start typing instead

fayi-da: Ah i see what you did with `severity`.. Cool cool, works for me

apps/sv/frontend/src/__tests__/sv.test.tsx
fayi-da: Along the same lines but yh I think in general anything that recreates elements gives me headaches in tests.

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
pawelperek-da: This is very elegant but Wayne wanted 1 \n between the party hint, network instance and SPONSOR_SV_URL

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
ray-roestenburg-da: yeah that's not what code reuse is about IMHO unless you have a clear higher level plan for modules / libs where to put these

apps/sv/frontend/src/components/ValidatorOnboardingSecrets.tsx
pawelperek-da: `dayjs` by itself doesn't support subtracting, you could do something like this instead:
```typescript
dayjs(b.contract.payload.expiresAt).valueOf() - dayjs(a.contract.payload.expiresAt).valueOf()
```
but at this point I'm not convinced wether it's better than just a `new Date`

apps/sv/src/main/openapi/sv-internal.yaml
```diff
@@ -633,6 +633,11 @@ components:
...
+          description: |
+            Human-readable alias of the validator party, decoded from the stored secret and 
+            exposed for UI display in the ListOngoingValidatorOnboardingsResponse endpoint.
```
fayi-da: > and exposed for UI display in the ListOngoingValidatorOnboardingsResponse endpoint

Don't think we need this part

apps/sv/src/main/openapi/sv-internal.yaml
```diff
@@ -665,6 +675,11 @@ components:
           type: string
         contact_point:
           type: string
+        party_hint:
+          description: |
+            Alias of the validator party. Must match the persisted value in the 
+            onboarding secret created by PrepareValidatorOnboardingRequest endpoint.
```
fayi-da: > created by PrepareValidatorOnboardingRequest endpoint.

I'm against referring to the schema's object name  in these docs but confirm with @ray-roestenburg-da

PR #2501, opened 2025-10-02
Reset circuit breakers after a given amount of time

### LINKED ISSUE #2462: Circuit breaker failures should reset after a configurable timeout without failures
Currently they reset only if we got a successful submit.

Sometimes we might have low frequency situations with contention whenever there is work to do, causing failures that accumulate slowly over time.

So we want to also reset the number of failures if just nothing happened for a while.

moritzkiefer-da:
fixes #2462

[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
@@ -12,19 +12,30 @@ import com.digitalasset.base.error.ErrorCategory.{
   InvalidIndependentOfSystemState,
 }
 import com.digitalasset.base.error.utils.ErrorDetails
-import com.digitalasset.canton.logging.TracedLogger
+import com.digitalasset.canton.data.CantonTimestamp
+import com.digitalasset.canton.logging.{NamedLoggerFactory, NamedLogging}
+import com.digitalasset.canton.time.Clock
 import com.digitalasset.canton.tracing.TraceContext
 import io.grpc.StatusRuntimeException
 import org.apache.pekko.actor.Scheduler
 import org.apache.pekko.pattern.{CircuitBreaker, CircuitBreakerOpenException}
 import org.lfdecentralizedtrust.splice.config.CircuitBreakerConfig
 
+import java.util.concurrent.atomic.AtomicReference
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success}
 
-class SpliceCircuitBreaker(name: String, underlying: CircuitBreaker)(implicit
-    ec: ExecutionContext
-) {
+class SpliceCircuitBreaker(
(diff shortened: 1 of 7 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
```diff
@@ -47,11 +76,27 @@ class SpliceCircuitBreaker(name: String, underlying: CircuitBreaker)(implicit
     }
   }
 
-  private def callAndMark[T](body: => Future[T]) = {
+  private def callAndMark[T](body: => Future[T])(implicit tc: TraceContext) = {
+    lastFailure.updateAndGet(_.filter { lastFailureTime =>
+      val elapsed = clock.now - lastFailureTime
+      if (elapsed.compareTo(config.resetFailuresAfter.asJava) >= 0) {
```
nicu-da: won't this mean that we will always close the breaker when `resetTimeout` exceeds `resetFailuresAfter`?

moritzkiefer-da: Hm right, I guess we want to not run it in a half-open state? Then you still get the right behavior?

moritzkiefer-da: changed

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
```diff
@@ -47,11 +76,28 @@ class SpliceCircuitBreaker(name: String, underlying: CircuitBreaker)(implicit
...
+        } else {
+          true
+        }
+      })
+    }
+
     body.andThen {
       case Failure(exception) =>
         if (!isFailureIgnored(exception)) {
           underlying.fail()
```
nicu-da, line 99: should we move the check here actually? basically call fail only if the last failure if fresher than the timeout.

nicu-da: not a big deal just weird that we call succeed and fail in the same invocation basically

moritzkiefer-da: But isn't that what you want? Let's say I didn't have a failure for an hour. Now I have one. The number of failures now should be 1 not 0. So you do need the succeed + fail.

nicu-da: Yup, you're right. Ignore me :see_no_evil:

PR #2509, opened 2025-10-02
Improve authorization checks for CN app APIs

rautenrieth-da:
Fixes https://github.com/DACH-NY/canton-network-internal/issues/1639

### Summary of changes:
- `UserRightsProvider` is a new trait for fetching the rights for a user
- `AuthExtractor` has been refactored, now has several subclasses, each implementing on auth requirement:
  - `AuthenticationOnlyExtractor`: requires authenticated user, but no other special authorization. Useful for endpoints returning "public" data where we don't want to allow anonymous access for audit or rate limiting purposes.
  - `UserWalletAuthExtractor`: requires authenticated user, authorized to `actAs` the party associated with the user through a `WalletAppInstall` contract. Useful for endpoints that interact with some wallet.
  - `AdminAuthExtractor`: requires authenticated user, authorized to `actAs` the app operator party AND have participant admin rights
  - `ActAsKnownPartyAuthExtractor`: requires authenticated user, authorized to `actAs` a given party (usually the app service party)
- `Http*Handler`: all endpoints in the same handler require the same auth. Auth is enforced by using one of the above auth extractors when constructing the routes, and the auth extractors returning unrelated types. As a result, some endpoints have moved:
  - Two endpoints from `HttpWalletHandler` have moved to `HttpStatusWalletHandler` because they don't require the user to have a wallet installed.
  - Sv app now has 3 handlers: `HttpSvPublicHandler` (no auth), `HttpSvOperatorHandler` (actAs operator party), `HttpSvAdminHandler` (actAs operator party AND participant admin)

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PruningIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PruningIntegrationTest.scala
@@ -67,7 +67,7 @@ class PruningIntegrationTest
             svApps = config.svApps.updatedWith(InstanceName.tryCreate("sv1")) {
               _.map { config =>
                 config.copy(acsCommitmentReconciliationInterval =
-                  PositiveDurationSeconds.ofSeconds(60)
+                  PositiveDurationSeconds.ofSeconds(30)
                 )
               }
             },
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/runbook/RunbookSvPreflightIntegrationTestBase.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/runbook/RunbookSvPreflightIntegrationTestBase.scala
@@ -17,7 +17,7 @@ import java.net.http.{HttpClient, HttpRequest, HttpResponse}
 import scala.collection.immutable.ArraySeq
 import scala.concurrent.duration.*
 import scala.jdk.CollectionConverters.*
-import scala.util.{Random, Try}
+import scala.util.Random
 
 abstract class RunbookSvPreflightIntegrationTestBase
     extends FrontendIntegrationTestWithSharedEnvironment("sv")
@@ -200,42 +200,6 @@ abstract class RunbookSvPreflightIntegrationTestBase
         asOfRound should startWith("The content on this page is computed as of round: ")
         asOfRound should not be "The content on this page is computed as of round: --"
       }
-    }
-  }
-
-  "The Scan UI shows the same total balance as sv-1" in { implicit env =>
-    val svClient = sv_client("sv")
-    val sv1ScanClient = scancl("sv1Scan")
-    val initialRound = sv1ScanClient.getDsoInfo().initialRound.getOrElse("0").toLong
-
-    val svParty = svClient.getDsoInfo().svParty.toProtoPrimitive
-    val svInfo = svClient.getDsoInfo().dsoRules.payload.svs.asScala.get(svParty).value
-    val joinedAsOfRound = svInfo.joinedAsOfRound.number
-    val lastAggregatedRoundSv = Try(sv1ScanClient.getRoundOfLatestData()._1).getOrElse(initialRound)
-    logger.debug(
-      s"last aggregated round from sv1: $lastAggregatedRoundSv, sv runbook joined as of round: $joinedAsOfRound"
-    )
-    // We do +3 here instead of, say, +1 as this was racy once and we don't care that much about the first few rounds after SV onboarding.
-    if (lastAggregatedRoundSv >= joinedAsOfRound + 3) {
-      withFrontEnd("sv") { implicit webDriver =>
-        go to scanUrl
-        eventually(1.minutes) {
-          val asOfRound = find(id("as-of-round")).value.text
-          asOfRound should startWith("The content on this page is computed as of round: ")
-          asOfRound should not be "The content on this page is computed as of round: --"
-          asOfRound should not be "The content on this page is computed as of round: ??"
-          val round =
-            Try(asOfRound.split(" ").last.toLong)
-              .getOrElse(fail(s"Failed parsing round number from: $asOfRound"))
-          val totalAmuletBalanceSv = find(id("total-amulet-balance-amulet")).value.text
-          val totalAmuletBalanceSv1 = sv1ScanClient
-            .getTotalAmuletBalance(round)
-            .valueOrFail("Amulet balance not yet computed")
-          parseAmountText(totalAmuletBalanceSv, amuletNameAcronym) shouldBe totalAmuletBalanceSv1
-        }
-      }
-    } else {
-      logger.debug("Skipping total balance test, the gap between rounds in sv and sv1 is too large")
     }
   }
 
--- a/docs/src/validator_operator/validator_major_upgrades.rst
+++ b/docs/src/validator_operator/validator_major_upgrades.rst
@@ -119,6 +119,7 @@ Catching Up Before the Migration
 In order for the migration to the new synchronizer to be safe and successful, it is important that the validator node is fully caught up on the existing synchronizer before proceeding to :ref:`validator-upgrades-deploying`.
 
 * To ensure that the validator participant has caught up and the :ref:`migration dump <validator-upgrades-dumps>` has been created as expected, operators can check the logs of the ``validator-app`` pod for ``Wrote domain migration dump`` messages.
+  If you can't find it, you can alternatively go inside the container and check if it has the migration dump with the expected date: ``kubectl exec -it <validator-pod> -- bash;  ls -lha /domain-upgrade-dump``.
 * To ensure that the validator app has caught up, operators can check the logs of the ``validator-app`` pod for the message ``Ingested transaction``.
   If the latest such message is 10 or more minutes old, the validator app has very likely (with a large safety margin) caught up to the state on the participant, and hence to the state of the existing (paused) synchronizer.
 * Note that the sequencers of the existing (old) synchronizer will be kept available by SVs for a limited
--- a/nix/canton-sources.json
+++ b/nix/canton-sources.json
@@ -1,7 +1,7 @@
 {
-  "version": "3.4.9-snapshot.20251124.17412.0.v0564fd04",
+  "version": "3.4.9-snapshot.20251126.17420.0.v06d25393",
   "tooling_sdk_version": "3.3.0-snapshot.20250415.13756.0.vafc5c867",
   "daml_release": "v3.3.0-snapshot.20250417.0",
-  "enterprise_sha256": "sha256:1rkq7f3jkw4vfxr97qnndj8v2mi51ci294mh4kh261hpib2fbgs9",
-  "oss_sha256": "sha256:0y5zmb8ys1kfxh54mzyvb80xmsk9464zwsdy8an1ppcfnjk54mw4"
+  "enterprise_sha256": "sha256:0xs219bdnc3bpg4g7ndi749ssz9rp2rk26a1ncsf0zkv0z41k21p",
+  "oss_sha256": "sha256:167zj0cy11097agdh0zwsg9r3cln1pdhdy4kdi7h9cnsfdgkik9p"
 }


build-tools/splice-compose.sh
```diff
@@ -659,7 +659,7 @@ function subcmd_identities_dump {
   VALIDATOR_AUTH_AUDIENCE="$DEFAULT_AUDIENCE"
   export VALIDATOR_AUTH_AUDIENCE
 
-  token=$("${VALIDATOR_DIR}/get-token.py" administrator)
+  token=$("${VALIDATOR_DIR}/get-token.py" ledger-api-user)
```
rautenrieth-da, line 680: `administrator` is the wallet user, `ledger-api-user` is the app operator user.

apps/wallet/src/main/openapi/wallet-internal.yaml
```diff
@@ -10,7 +10,7 @@ paths:
   /v0/wallet/user-status:
     get:
       tags: [wallet]
-      x-jvm-package: wallet
+      x-jvm-package: status.wallet
```
rautenrieth-da, line 13: package `wallet`: requires `actAs` for the wallet party
package `status.wallet`: requires just an authenticated user, no rights checked

apps/sv/src/main/openapi/sv-internal.yaml
```diff
@@ -18,7 +18,7 @@ paths:
   /v0/admin/validator/onboarding/ongoing:
     get:
       tags: [sv]
-      x-jvm-package: sv_admin
+      x-jvm-package: sv_operator
```
rautenrieth-da, line 21: package `sv_public`: au auth whatsoever
package `sv_operator`: requires `actAs` on the sv operator party
package `sv_admin`: requires `actAs` on the sv operator party and participant admin

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/SvAppReference.scala
```diff
@@ -54,46 +55,51 @@ abstract class SvAppReference(
   def onboardValidator(validator: PartyId, secret: String, contactPoint: String): Unit =
     consoleEnvironment.run {
       httpCommand(
-        HttpSvAppClient.OnboardValidator(validator, secret, BuildInfo.compiledVersion, contactPoint)
+        HttpSvPublicAppClient.OnboardValidator(
```
ray-roestenburg-da, line 58: Maybe even call it `HttpSvNoAuthAppClient` (don't know which one is scarier)

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/SvAppReference.scala
```diff
@@ -54,46 +55,51 @@ abstract class SvAppReference(
...
     consoleEnvironment.run {
-      httpCommand(HttpSvAppClient.GetDsoInfo)
+      httpCommand(HttpSvPublicAppClient.GetDsoInfo)
     }
 
   @Help.Summary("Get the CometBFT node status")
   def cometBftNodeStatus(): definitions.CometBftNodeStatusResponse =
     consoleEnvironment.run {
-      httpCommand(HttpSvAppClient.GetCometBftNodeStatus())
+      httpCommand(HttpSvPublicAppClient.GetCometBftNodeStatus())
```
ray-roestenburg-da, line 96: why is the node status public? (just curious)

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/SvAppReference.scala
```diff
@@ -103,17 +109,17 @@ abstract class SvAppReference(
       params: Map[String, io.circe.Json] = Map.empty,
   ): definitions.CometBftJsonRpcResponse =
     consoleEnvironment.run {
-      httpCommand(HttpSvAppClient.CometBftJsonRpcRequest(id, method, params))
+      httpCommand(HttpSvPublicAppClient.CometBftJsonRpcRequest(id, method, params))
```
ray-roestenburg-da, line 112: This also seems strange to be public?

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/SvAppReference.scala
```diff
@@ -338,31 +344,31 @@ class SvAppBackendReference(
   def listOngoingValidatorOnboardings(): Seq[ValidatorOnboarding] =
     consoleEnvironment.run {
       httpCommand(
-        HttpSvAdminAppClient.ListOngoingValidatorOnboardings
+        HttpSvOperatorAppClient.ListOngoingValidatorOnboardings
       )
     }
 
   @Help.Summary("Prepare a validator onboarding and return an onboarding secret (via admin API)")
```
ray-roestenburg-da, line 354: is the comment still correct? (via admin API)

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorIntegrationTest.scala
```diff
@@ -264,77 +264,111 @@ class ValidatorIntegrationTest extends IntegrationTest with WalletTestUtil {
...
+      val userId = aliceValidatorBackend.participantClientWithAdminToken.ledger_api.users
+        .create(
+          s"testUser-${Random.nextInt()}",
+          actAs = Set.empty[PartyId],
+          primaryParty = Some(validatorParty),
+          participantAdmin = true,
+          isDeactivated = false,
+        )
+        .id
+      assertRejectedAsUnauthorized(userId)
```
ray-roestenburg-da, line 309: nice cleanup thanks

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/auth/ActAsKnownPartyAuthExtractor.scala
```diff
@@ -0,0 +1,96 @@
...
+  * that represent a known, active user on the participant.
+  *
+  * ==Authentication==
+  *
+  *  - request must have a valid JWT token authenticating the user
+  *
+  * ==Authorization==
+  *
+  *  - user must exist on the participant and be active
+  *  - primary party must be set for the user and equal to the app operator party
```
ray-roestenburg-da, line 24: you say app operator party, but it is actually the `party` parameter on this class right? slightly confusing cause you mix where this is used and what is defined here I think, rather refer to `party` with backticks

ray-roestenburg-da: (and use `@param` in the doc if needed)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/auth/AdminAuthExtractor.scala
```diff
@@ -3,89 +3,100 @@
...
+/** Auth extractor for APIs that perform administrative actions on the participant
+  *
+  * ==Authentication==
+  *
+  *  - request must have a valid JWT token authenticating the user
+  *
+  * ==Authorization==
+  *
+  *  - user must exist on the participant and be active
+  *  - primary party must be set for the user and equal to the app operator party
```
ray-roestenburg-da:
```suggestion
  *  - primary party must be set for the user and equal to the `adminParty`
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/auth/AdminAuthExtractor.scala
```diff
@@ -3,89 +3,100 @@
...
+  *  - user must have actAs rights for the app operator party
+  *  - user must have ParticipantAdmin rights
```
ray-roestenburg-da, line 25:
```suggestion
  *  - user must have ParticipantAdmin rights
  
  @param adminParty the app operator party
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/auth/ActAsKnownPartyAuthExtractor.scala
```diff
@@ -0,0 +1,96 @@
...
+    traceContext: TraceContext
+) extends AuthExtractor(verifier, loggerFactory, realm)(traceContext) {
+
+  def directiveForOperationId(
+      operationId: String
+  ): Directive1[ActAsKnownPartyAuthExtractor.ActAsKnownUserRequest] = {
+    authenticateLedgerApiUser(operationId)
+      .flatMap { authenticatedUser =>
+        onComplete(
+          rightsProvider.getUser(authenticatedUser) zip rightsProvider.listUserRights(
```
ray-roestenburg-da: if you always do this zip, you might as well change the rightsProvider and have a getUserAndRights or something like that

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/auth/UncachedUserRightsProvider.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import com.daml.ledger.javaapi.data.User
+import com.digitalasset.canton.tracing.TraceContext
+import io.grpc.StatusRuntimeException
+import org.lfdecentralizedtrust.splice.environment.SpliceLedgerConnection
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+class UncachedUserRightsProvider(
```
ray-roestenburg-da: nitpick but I would just call this something like LedgerUserRightsProvider

apps/sv/src/main/openapi/sv-internal.yaml
```diff
@@ -75,7 +75,7 @@ paths:
   /v0/admin/domain/cometbft/status:
     get:
       tags: [sv]
-      x-jvm-package: sv
+      x-jvm-package: sv_public
```
ray-roestenburg-da, line 78: So not sure about this one, why that's needed to be fully public

apps/sv/src/main/openapi/sv-internal.yaml
```diff
@@ -95,7 +95,7 @@ paths:
   /v0/admin/domain/cometbft/json-rpc:
     post:
       tags: [sv]
-      x-jvm-package: sv
+      x-jvm-package: sv_public
```
ray-roestenburg-da, line 98: So not sure about this one, why that's needed to be fully public

apps/sv/src/main/openapi/sv-internal.yaml
```diff
@@ -530,7 +530,7 @@ paths:
   /v0/devnet/onboard/validator/prepare:
```
ray-roestenburg-da, line 530: Can we prevent this API and other devnet-only ones to exist on non-devnet? or would we need to do many openapi spec files or something like that if we really wanted to? (or do we do, or can we do,  a proxy thing that 'removes this' and returns notfound or something on non-devnet)

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
```diff
@@ -76,12 +77,18 @@ class HttpSvHandler(
   override protected val votesStore: ActiveVotesStore = dsoStore
   override protected val workflowId: String = this.getClass.getSimpleName
 
-  def onboardValidator(
-      respond: v0.SvResource.OnboardValidatorResponse.type
+  /** Intended use: Other validators call this endpoint to onboard themselves,
+    * passing their onboarding secret in the request payload.
+    *
+    * Protection: The onboarding secret is unguessable and single-use.
+    * Note that we are still spending some resources on verifying the secret.
```
ray-roestenburg-da, line 84: So this is an easy target for an attack I guess.. but can prevent issues with rate limit etc.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
```diff
@@ -229,20 +240,24 @@ class HttpSvHandler(
...
           } yield res
       }
     }
   }
 
-  def getSvOnboardingStatus(
-      respond: v0.SvResource.GetSvOnboardingStatusResponse.type
+  /** Intended use: Used by other SV operators
+    *
+    * Protection: Endpoint is protected by IP allowlisting
```
ray-roestenburg-da, line 253: This comment will get out of date..

ray-roestenburg-da: why is this not sv operator?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
```diff
@@ -335,10 +356,15 @@ class HttpSvHandler(
     }
   }
 
-  def getDsoInfo(
-      respond: v0.SvResource.GetDsoInfoResponse.type
-  )()(extracted: TraceContext): Future[v0.SvResource.GetDsoInfoResponse] = {
-    implicit val tc = extracted
+  /** Intended use: The SV app UI.
+    *
+    * Protection: None for backwards compatibility reasons
```
ray-roestenburg-da, line 361: Maybe create an issue to deprecate and move

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
```diff
@@ -359,12 +385,16 @@ class HttpSvHandler(
     }
   }
 
-  def getCometBftNodeStatus(
-      respond: v0.SvResource.GetCometBftNodeStatusResponse.type
+  /** Intended use: Interacting with CometBFT node monitoring/debugging/testing purposes
+    *
+    * Protection: Endpoint is protected by IP allowlisting
```
ray-roestenburg-da, line 390: Can't this move to sv_operator as well? (or plan to make a move)

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
```diff
@@ -476,12 +516,16 @@ class HttpSvHandler(
       )
   }
 
-  def onboardSvSequencer(
-      respond: v0.SvResource.OnboardSvSequencerResponse.type
+  /** Intended use: Used by other SV operators
+    *
+    * Protection: Endpoint is protected by IP allowlisting
```
ray-roestenburg-da, line 521: I think I would remove these comments

ray-roestenburg-da: (maybe change it to a TODO to move it out of public)

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
```diff
@@ -1102,38 +1153,72 @@ class ValidatorApp(
...
                       metrics.httpServerMetrics
                         .withMetrics("walletInternal")(operation)
                         .tflatMap { _ =>
-                          AuthExtractor(verifier, loggerFactory, "splice wallet realm")(
+                          UserWalletAuthExtractor(
+                            verifier,
+                            walletManager,
+                            userRightsProvider,
+                            loggerFactory,
+                            "splice wallet realm",
```
ray-roestenburg-da: nitpick extract to constant / val

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/admin/http/HttpScanProxyHandler.scala
```diff
@@ -23,16 +23,16 @@ class HttpScanProxyHandler(
     ec: ExecutionContext,
     mat: Materializer,
     tracer: Tracer,
-) extends v0.ScanproxyHandler[TracedUser]
+) extends v0.ScanproxyHandler[AuthenticatedRequest]
```
ray-roestenburg-da, line 26: nice

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/admin/http/HttpValidatorAdminHandler.scala
```diff
@@ -84,7 +84,7 @@ class HttpValidatorAdminHandler(
     ec: ExecutionContext,
     mat: Materializer,
     tracer: Tracer,
-) extends v0.ValidatorAdminHandler[TracedUser]
+) extends v0.ValidatorAdminHandler[AdminUserRequest]
```
ray-roestenburg-da, line 86: nice, like it.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/admin/http/UserWalletAuthExtractor.scala
```diff
@@ -0,0 +1,128 @@
...
+) extends AuthExtractor(verifier, loggerFactory, realm)(traceContext) {
+
+  def directiveForOperationId(
+      operationId: String
+  ): Directive1[UserWalletAuthExtractor.WalletUserRequest] = {
+    authenticateLedgerApiUser(operationId)
+      .flatMap { authenticatedUser =>
+        onComplete(
+          // Double zip is a bit ugly...
+          rightsProvider.getUser(authenticatedUser) zip rightsProvider.listUserRights(
```
ray-roestenburg-da: if you change rightsProvide to have one method for this there is one less zip :-)

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/admin/http/UserWalletAuthExtractor.scala
```diff
@@ -0,0 +1,128 @@
...
+      )
+      .asRuntimeException()
+  }
+}
+
+object UserWalletAuthExtractor {
+
+  /** @param user         The authenticated user name.
+    * @param userWallet   The wallet associated with the authenticated user.
+    *                     We need to look up the wallet in the extractor to get the end user party,
```
ray-roestenburg-da: nitpick: here you are describing what code somewhere else is doing, I would prefer to document that differently, this should just describe the field

docs/src/release_notes.rst
```diff
@@ -11,6 +11,43 @@ Release Notes
...
+  - API security
+
+    - Tightened authorization checks for all non-public API endpoints.
+
+      All non-public endpoints now properly respect the current user rights
+      defined in the participant user management service.
+      Revoking user rights on the participant will revoke access to the corresponding API endpoints.
+
+      In general, endpoints that required authentication before will now check that the authenticated user
+      is not deactivated on the participant has has ``actAs`` rights for the relevant party
```
ray-roestenburg-da:
```suggestion
      is not deactivated on the participant has ``actAs`` rights for the relevant party
```

docs/src/release_notes.rst
```diff
@@ -11,6 +11,43 @@ Release Notes
...
+
+    - Administrative SV app endpoints now require participant admin rights.
+
+      The following SV app endpoints now require the user to have participant admin rights in
+      the participant user management service. This allows for finer grained access control
+      where users with ``actAs`` rights for the SV operator party but without participant admin
+      rights may use the SV or wallet UIs, but may not perform administrative actions like
+      hard synchronizer migrations.
+
+        - ``/v0/admin/domain/pause``
```
ray-roestenburg-da, line 27: I think this list should follow directly after 

> The following SV app endpoints now require the user to have participant admin rights in
      the participant user management service. 

And after the list the next paragraph

docs/src/release_notes.rst
```diff
@@ -11,6 +11,43 @@ Release Notes
...
+        - ``/v0/admin/domain/unpause``
+        - ``/v0/admin/domain/migration-dump``
+        - ``/v0/admin/domain/migration-dump``
+        - ``/v0/admin/domain/identities-dump``
+        - ``/v0/admin/domain/data-snapshot``
+
+      Note that only the service users of the SV and validator apps should automatically have participant admin rights.
+      If you are using other users to access the above endpoints, check their rights.
```
ray-roestenburg-da, line 40:
```suggestion
      If you are using other users to access the above endpoints, please check their rights.
```

docs/src/release_notes.rst
```diff
@@ -11,6 +11,43 @@ Release Notes
...
+
+    - Some endpoints will have changed authorization rules in an upcoming release.
+
+        - SV app ``/v0/dso`` is currently public, but will require authorization as SV operator,
+          similar to most other SV app endpoints.
+          Use the corresponding public endpoint in the scan app if you need to fetch DSO info.
```
ray-roestenburg-da: Maybe just state the actual endpoint in scan
```suggestion
          Use the public ``/v0/dso`` endpoint in the scan app if you need to fetch DSO info.
```

apps/sv/src/main/openapi/sv-internal.yaml
rautenrieth-da: This endpoint didn't have any auth before, here we're just renaming things to make it more clear.

Ideally this would have some kind of auth, but CometBFT will be replaced by a different consensus layer in the near future and we wanted to limit the number of breaking changes, so I left it as is.

apps/sv/src/main/openapi/sv-internal.yaml
rautenrieth-da: The implementation returns a `501 Not Implemented` on any non-devnet network where the `isDevNet` flag is false. Do you want to remove the endpoint because we don't trust this check, or because we want to avoid parsing the request in the first place?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
rautenrieth-da: > Intended use: Used by other SV operators

This is called by other SVs to onboard their node. Other SVs do not have any token for your sv app, so this can't have auth. We have to rely on IP whitelisting instead (like we do for many other requests between SV nodes).

apps/sv/src/main/openapi/sv-internal.yaml
ray-roestenburg-da: avoid handling it, but I guess it's not really important, I just thought it would be nicer if it responded like any other API that is not there

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
rautenrieth-da: There's one on the line below 🙂

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
rautenrieth-da: Same as above, endpoints that are called by other svs can't have auth.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
rautenrieth-da: Same as above: cometbft will be replaced, changing auth here is not worth the effort.

apps/sv/src/main/openapi/sv-internal.yaml
ray-roestenburg-da: ok

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/SvAppReference.scala
rautenrieth-da: "Admin API" is a bit overloaded, but the endpoints are called `/v0/admin/...`, so I'll leave.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/auth/UncachedUserRightsProvider.scala
rautenrieth-da: Renamed to `ParticipantUserRightsProvider`

apps/sv/src/main/openapi/sv-internal.yaml
rautenrieth-da: Same as for other cometbft endpoints: cometbft is getting replaced, we don't want to make changes in to API that could break other operator deployments.

docs/src/release_notes.rst
rautenrieth-da: Most style guides recommend not using "please" in technical documentation. Our release notes are inconsistent in this regard.

PR #2538, opened 2025-10-03
Use sbt --client in make files

OriolMunoz-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/5808, by preventing different sbts from stepping on each other's work 
see also: https://daholdings.slack.com/archives/C03J6NCPSE4/p1759505448124789

https://github.com/hyperledger-labs/splice/actions/runs/18228357489/job/51905429825?pr=2538
The `npm install` order makes sense


isegall-da (approved):
Nice catch, thank you!

nicu-da:
note: we had issues with this in the past as the client was kind of unstable and kept getting some network errors back and failing the build, just something to keep an eye on

PR #2563, opened 2025-10-06
Bump nix binary to 2.31.2

OriolMunoz-da:
Reverse of https://github.com/hyperledger-labs/splice/pull/1727/files:
- All the fixes there are still in code
- Newer nix version
- Bumping cache versions again (before merging, to 6, 6666 is for testing in case it ends up very corrupt)


apps/sv/frontend/src/__tests__/governance/forms/set-amulet-rules-form.test.tsx
```diff
@@ -235,45 +235,49 @@ describe('Set Amulet Config Rules Form', { timeout: 5000 }, () => {
...
 
-    expect(screen.getByTestId('proposal-submission-error')).toBeDefined();
-    expect(screen.getByText(/Submission failed/)).toBeDefined();
-    expect(screen.getByText(/Service Unavailable/)).toBeDefined();
-  });
+      expect(screen.getByTestId('proposal-submission-error')).toBeDefined();
+      expect(screen.getByText(/Submission failed/)).toBeDefined();
+      expect(screen.getByText(/Service Unavailable/)).toBeDefined();
+    },
+    { timeout: 10000 }
```
OriolMunoz-da: bumping a timeout because it flaked

.github/actions/nix/setup_nix/action.yml
```diff
@@ -168,6 +171,9 @@ runs:
         chown -R $(whoami):$(whoami) ~
         cat /tmp/nix-cache-key
         if [ ! -f "/cache/nix/$cache_key/cached" ]; then
+          echo "Garbage collecting to reduce cache size"
+          nix-store --gc
```
OriolMunoz-da: I don't understand why it was done above, as opposed to before saving
Maybe because it would clean the mounted /nix/store? But I don't see much of a point in that, as it shouldn't have changed, and if it did, the cache-key would change... I think

.github/actions/nix/run_bash_command_in_nix/action.yml
OriolMunoz-da, line 1: don't see the harm in keeping these

.github/actions/nix/setup_nix/action.yml
```diff
@@ -147,8 +152,6 @@ runs:
             target="default"
           fi
           nix develop path:nix#${target} -v --profile "$HOME/.nix-shell" --command echo "Done loading packages"
-          echo "Garbage collecting to reduce cache size"
-          nix-store --gc
```
OriolMunoz-da, line 151: I don't understand why it was done above, as opposed to before saving
Maybe because it would clean the mounted /nix/store? But I don't see much of a point in that, as it shouldn't have changed, and if it did, the cache-key would change... I think

.github/workflows/build.ts_cli_tests.yml
```diff
@@ -22,7 +22,6 @@ jobs:
 
       - name: Check out repository code
         uses: actions/checkout@v4
-        if: inputs.commit_sha != ''
```
OriolMunoz-da, line 25: this and below are the only ones that had this... what for?

.github/workflows/build.ui_tests.yml
```diff
@@ -28,7 +28,6 @@ jobs:
 
       - name: Check out repository code
         uses: actions/checkout@v4
-        if: inputs.commit_sha != ''
```
OriolMunoz-da, line 31: also this

.github/workflows/build.ts_cli_tests.yml
moritzkiefer-da: @isegall-da is that leftover? I could see it only being set when you run from a fork but then it should be used for every checkout and it's definitely not.

.github/actions/nix/setup_nix/action.yml
```diff
@@ -86,8 +91,8 @@ runs:
           # we use rsync here because it's simply faster to install
           rsync -avi /cache/nix/$cache_key/.nix-* $HOME/
           rsync -avi "/cache/nix/$cache_key/nix" $HOME/.config/
-          rsync -avi "/cache/nix/$cache_key/nix_store/var/" /nix/var
-          sudo mount --bind /cache/nix/$cache_key/nix_store/store /nix/store
+          # rsync -avi "/cache/nix/$cache_key/nix_store/var/" /nix/var
```
moritzkiefer-da, line 95: If we do uncomment that let's create an issue for it documenting how it fails and what you tried.

@martinflorian-da @nicu-da wdyt? I'm leaning towards saying getting an upgrade in with the slowdown and then moving the fix for this to the top of our prio list is better than to keep pushing out that upgrade for another few months (the pressure from slow CI also felt more so chances are higher we'll actually tackle it).

martinflorian-da: (With my limited context:) Agreed to merging this without this caching part, a bit worried about "just moving the fix to the top of our prio list". Can we give assign it to a concrete person already that will realistically get to it soon?

nicu-da: Agree with @martinflorian-da those 2 minutes were really annoying in the past, though I guess with the current CI timing it might be less so.

OriolMunoz-da: https://github.com/hyperledger-labs/splice/issues/2663 adding as TODOs

.github/workflows/build.ts_cli_tests.yml
OriolMunoz-da: ping @isegall-da , don't want to merge without knowing this is 100% fine

isegall-da: Discussed with @OriolMunoz-da now, just copying the conclusion here as well:
In these tests we do:
Checkout the PR head commit to get the commit message first
check the static label
Checkout commit_sha

In tests that run split_tests first, we check the static label in split_tests which is why we don't have these separate steps there, and we just checkout commit_sha there.

(in ts_cli we flipped the order between (2) and (3) which is wrong, we'll fix that)

PR #2573, opened 2025-10-07
Add subtitle to proposal summary field

fayi-da:
Fixes #2378

Decided to do some refactoring here to make this cleaner

Screenshot 2025-10-07 at 12 43 09

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/sv/frontend/src/utils/constants.ts
pawelperek-da, line 1: Does it make sense to create separate file just for those 2 consts?

fayi-da: It does feel lonely in the file but I guess we have to start somewhere. 

Got any ideas where they can live instead?

pawelperek-da: Is there a reason why they have to be constants? Afaik all different UI strings are just living in the components themselves, right?

fayi-da: In this case, it was mainly because we were using them in quite a few places

pawelperek-da: well, I'd go with non-const strings but if your conscience is telling you otherwise then LGTM 😛

PR #2598, opened 2025-10-08
Prevent flakes caused by TransferPreapproval not yet ingested by Scans

OriolMunoz-da:
Which then fails BFT fetch of it when calling .transferPreapprovalSend

Fixes https://github.com/DACH-NY/cn-test-failures/issues/5832

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/WalletTestUtil.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/WalletTestUtil.scala
@@ -1310,7 +1310,7 @@ trait WalletTestUtil extends TestCommon with AnsTestUtil {
   }
 
   /** @param partyWalletClient the wallet in which to create the transfer preapproval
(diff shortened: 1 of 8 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/WalletTestUtil.scala
```diff
@@ -1309,17 +1309,29 @@ trait WalletTestUtil extends TestCommon with AnsTestUtil {
       .supported
   }
 
-  def createTransferPreapprovalIfNotExists(
-      partyWalletClient: WalletAppClientReference
+  /** @param partyWalletClient the wallet in which to create the transfer preapproval
+    * @param checkValidator a validator to use to check that the preapproval was created and ingested by all Scans
```
isegall-da:
```suggestion
    * @param checkValidator a validator to use to check that the preapproval was created and ingested by enough Scans
```

Issue #2017, opened 2025-08-26
Change docker compose setup to only expose to localhost by default


moritzkiefer-da:
Too easy to shoot yourself in the foot otherwise.

PR #2601, opened 2025-10-08
Fix issue 2017: setting localhost as default in docker-compose based setups

pasindutennage-da:
Fixes  https://github.com/hyperledger-labs/splice/issues/2017

Set localhost as the default in `SV`, `localhost`, and `validator` docker-compose in `cluster/compose`

For `SV` and `Validator`, there is an option to expose externally using `-E` flag, in the respective `start.sh` scripts.

For `localnet`, it is hardcoded, because no extra traffic is expected.

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeFullNetworkFrontendIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeFullNetworkFrontendIntegrationTest.scala
@@ -95,6 +95,14 @@ class DockerComposeFullNetworkFrontendIntegrationTest
 
           }
         }
+        clue("full network (SV) stack should bind to localhost by default") {
+          withComposeFullNetwork() { () =>
+            var binding = getPortBinding("splice-sv-nginx-1", "80")
+            binding should startWith("127.0.0.1")
+            binding = getPortBinding("splice-validator-nginx-1", "80")
+            binding should startWith("127.0.0.1")
+          }
+        }
       }
     } finally {
       Seq("build-tools/splice-compose.sh", "stop_network", "-D", "-f").!
@@ -134,14 +142,6 @@ class DockerComposeFullNetworkFrontendIntegrationTest
   "docker-compose networking bindings work" in { implicit env =>
     registerHttpConnectionPoolsCleanup(env)
 
-    clue("full network (SV) stack should bind to localhost by default") {
-      withComposeFullNetwork() { () =>
-        var binding = getPortBinding("splice-sv-nginx-1", "80")
-        binding should startWith("127.0.0.1")
-        binding = getPortBinding("splice-validator-nginx-1", "80")
-        binding should startWith("127.0.0.1")
-      }
-    }
     clue("full network (SV) stack should bind to 0.0.0.0 with -E flag") {
       withComposeFullNetwork(startFlags = Seq("-E")) { () =>
         var binding = getPortBinding("splice-sv-nginx-1", "80")


cluster/compose/localnet/compose.yaml
```diff
@@ -70,9 +70,9 @@ services:
       - ${LOCALNET_DIR}/conf/nginx/sv.conf:/etc/nginx/templates/sv.c${SV_PROFILE}f.template
       - ${LOCALNET_DIR}/conf/nginx/swagger-ui:/etc/nginx/includes
     ports:
-      - "${SV_UI_PORT}:${SV_UI_PORT}"
-      - "${APP_PROVIDER_UI_PORT}:${APP_PROVIDER_UI_PORT}"
-      - "${APP_USER_UI_PORT}:${APP_USER_UI_PORT}"
+      - "127.0.0.1:${SV_UI_PORT}:${SV_UI_PORT}"
```
martinflorian-da: Ah your idea being that these should never be public? Hm not 100% sure about that change - perhaps someone do wants to make them public after all, as part of their testing?

Probably nicest to make them optionally public like for the other compose file?

cc @isegall-da WDYT?

pasindutennage-da: I have some mixed feeling about this, and the following is my reasoning.

1) In any realistic setting, `localnet` will never host a real deployment, and hence nothing bad can happen if no port is opened for the outside (the open ports accept traffic only from 127.0.0.1).

2) Now `validator` is a bit tricky -- `Validator` initiates the connection with the `SV` (and not the other way), and therefor it is OK if the `validator` only accepts new connection requests (or just requests) from `localhost`. But, I am curious about the case where any other component (a front end) connects with the `validator`, in which case `validator` does have to listen to new connections from outside it's local machine. If the later is a realistic scenario, then the user has to put `-E` in the `start.sh`.

3) The `SV` case: `SV` often gets external requests, so I believe it's default behavior should listen to 0.0.0.0 (all interfaces), instead of `localhost`. But a counter argument to that is, `SV` in production use helm charts, and in any realistic setting this docker-compose will be used for local testing. Hence for all practical purposes, I think it is ok to let `SV` accept connections from `localhost` only.

@moritzkiefer-da Please let me know what you think.

Thanks

isegall-da: I would actually drop SV and localnet from this PR altogether, and focus only on validators. Everything else is dev tooling, so I'd keep those simple (and vulnerable).

For validators, I think that your all-or-nothing approach of the -E flag makes sense.

moritzkiefer-da: I'd look at it this way:

The whole reason we're adding this is that it got flagged as a security issue. Imho the best course of action on that is to change the default for everything including localnet (so yes I strongly disagree with @isegall-da here) Yes it is only localnet but that doesn't really matter. If someone gets an RCE on your machine because you exposed localnet publicly that's bad.

Which then leaves the question of where we allow changing it. There I'd also keep it simple: Allow overwriting it everywhere. If nothing else this is useful for backwards compatibility and there is no real downside. Trying to justify why you really don't need it for e.g. localnet only to then have a user complain that we broke their setup with no option to fix it seems more trouble than just adding the flag.

docs/src/release_notes.rst
```diff
@@ -17,6 +17,8 @@ Upcoming
 
     - Replace ``-Dscala.concurrent.context.minThreads=8`` with ``-Dscala.concurrent.context.numThreads=8`` and set ``-XX:ActiveProcessorCount=8``  in the ``defaultJvmOptions`` for all the helm charts that deploy scala apps. This should ensure that the internal execution contexts spawn 8 threads to handle processing and that the JVM is configured for 8 CPUs as well. The previous behavior would spawn up to number of available processors, which can be up to the number of CPUs on the actual node if no CPU limit is set. This should avoid overloading the nodes during heavy processing.
 
+    - Docker-compose based deployment of LocalNet, Validator, and SuperValidator expose only to 127.0.0.1 by default. If you want to expose externally, use ``-E`` in Validator and SuperValidator ``start.sh``. For LocalNet, ``export HOST_BIND_IP=0.0.0.0`` manually.
```
martinflorian-da:
```suggestion
    - Docker-compose based deployments of LocalNet, validator, and SV now expose only to 127.0.0.1 by default. If you want to expose externally, use ``-E`` in validator and SV ``start.sh``. For LocalNet, set ``export HOST_BIND_IP=0.0.0.0`` manually.
```

wording nit

cluster/compose/localnet/compose.yaml
pasindutennage-da: Following the conversation, i made all default to localhost, with option for 0.0.0.0.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeFullNetworkFrontendIntegrationTest.scala
```diff
@@ -97,4 +101,55 @@ class DockerComposeFullNetworkFrontendIntegrationTest
...
+    val output = Try(reader.readLine()).toOption.getOrElse("")
+    if (process.waitFor() != 0 || output.isEmpty) {
+      fail(
+        s"Could not get port binding for $containerName:$internalPort. Is the container running?"
+      )
+    }
+    output
+  }
+
+  def withComposeFullNetwork[A](startFlags: Seq[String] = Seq.empty)(
```
martinflorian-da, line 125: Nice! Why not use this also for the  existing test case, to reduce duplication?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeFullNetworkFrontendIntegrationTest.scala
```diff
@@ -97,4 +101,55 @@ class DockerComposeFullNetworkFrontendIntegrationTest
...
+      test()
+    } finally {
+      Seq("build-tools/splice-compose.sh", "stop_network", "-D", "-f").!
+    }
+  }
+
+  "docker-compose networking bindings work" in { implicit env =>
+    registerHttpConnectionPoolsCleanup(env)
+
+    clue("full network (SV) stack should bind to localhost by default") {
```
martinflorian-da: I strongly suggest moving this check into the existing test case; starting a network just for this test seems to be overkill in terms of increasing test duration (=> CI time) and resource use (starting a network takes time...).

martinflorian-da (approved):
Nice work, thank you!

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeFullNetworkFrontendIntegrationTest.scala
pasindutennage-da: This tests two cases: 

1) deploy a network without -E, and check if validator and SV binds to only localhost

2) deploy a network with -E, and check if validator and SV binds to 0.0.0.0

you are right, that the "1" can indeed be moved to existing network, and I did it in the new commit.

for the second case (which i think is equally important), I kept the new network, but with only one check.

PR #2647, opened 2025-10-09
Validator app: fix delete user in offboard.

pasindutennage-da:
Fixes https://github.com/DACH-NY/canton-network-internal/issues/402

When the validator offboards a user, the corresponding user in the participant is also deleted.

[ci]

### DIFF
--- a/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/util/ValidatorUtil.scala
+++ b/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/util/ValidatorUtil.scala
@@ -323,7 +323,7 @@ private[validator] object ValidatorUtil {
             }
             .recover { case NonFatal(ex) =>
               logger
-                .warn(s"Skipping user deletion for '$endUserName' due to a non-fatal error.", ex)
+                .debug(s"Skipping user deletion for '$endUserName' due to a non-fatal error.", ex)
               ()
             }
         } yield {


apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/util/ValidatorUtil.scala
```diff
@@ -312,6 +313,7 @@ private[validator] object ValidatorUtil {
             // these commands could fail with PERMISSION_DENIED errors (#4425).
             Seq(Status.Code.PERMISSION_DENIED),
           )
+          _ <- connection.deleteUser(endUserName)
         } yield {
           logger.debug(s"User $endUserParty offboarded")
```
moritzkiefer-da: Not your fault but the log is wrong, it should log the user but it logs the party

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/WalletTestUtil.scala
```diff
@@ -174,18 +175,6 @@ trait WalletTestUtil extends TestCommon with AnsTestUtil {
     status.userOnboarded shouldBe false
     status.userWalletInstalled shouldBe false
 
-    val endUserParty = validatorAppBackend.participantClientWithAdminToken.ledger_api.users
```
moritzkiefer-da, line 177: don't you want to check that the user has been deleted here?

docs/src/release_notes.rst
```diff
@@ -25,6 +25,11 @@ Release Notes
 
       - The UI now provides a formatted message for easily sharing onboarding details with validator operators.
 
+  - Validator
+
+      - ``/v0/admin/users/offboard``:
+        Off boarding a user deletes the ledgerAPI user in the participant node.
```
```suggestion
        Offboarding a user deletes the ledgerAPI user in the participant node.
```

```suggestion
        Off boarding a user deletes the Ledger API user in the participant node.
```

docs/src/release_notes.rst
```diff
@@ -25,6 +25,11 @@ Release Notes
 
       - The UI now provides a formatted message for easily sharing onboarding details with validator operators.
 
+  - Validator
+
+      - ``/v0/admin/users/offboard``:
+        Offboarding a user deletes the ledger API user in the participant node.
```
martinflorian-da:
```suggestion
        Offboarding a user now also deletes the ledger API user in the participant node.
```

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/util/ValidatorUtil.scala
```diff
@@ -312,8 +314,9 @@ private[validator] object ValidatorUtil {
             // these commands could fail with PERMISSION_DENIED errors (#4425).
             Seq(Status.Code.PERMISSION_DENIED),
           )
+          _ <- connection.deleteUser(endUserName).recover { case NonFatal(_) => () }
```
martinflorian-da:
```suggestion
          _ <- connection.deleteUser(endUserName)
```

as discussed, then see how we can expect this in the test

martinflorian-da: or actually, this whole call is supposed to be idempotent; so probably what you want is to do a list users first, and only try to delete if the user is there

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/WalletTestUtil.scala
martinflorian-da: @pasindutennage-da I suggest still adding this check here to this method, if only for semantics; `assertUserFullyOffboarded` should definitely include "is the user gone" too now IMO.

apps/validator/src/main/openapi/validator-internal.yaml
```diff
@@ -106,7 +106,8 @@ paths:
       description: |
         As the validator operator, offboard the user specified in the request.
         Offboarding archives the daml contracts required for the user to use a wallet on this validator.
-        Offboarding does not delete the ledger API user, and does not archive any other daml contracts owned by the user.
+        Offboarding deletes the ledger API user
```
martinflorian-da:
```suggestion
        Offboarding deletes the ledger API user.
```

nit

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/util/ValidatorUtil.scala
```diff
@@ -312,8 +314,18 @@ private[validator] object ValidatorUtil {
             // these commands could fail with PERMISSION_DENIED errors (#4425).
             Seq(Status.Code.PERMISSION_DENIED),
           )
+          _ <- connection
+            .getUser(endUserName)
```
martinflorian-da, line 320: Nice! Consider adding a comment here that we're only deleting the user if it exists; seems easy to miss on a quick read. (nit)

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/util/ValidatorUtil.scala
```diff
@@ -312,8 +314,18 @@ private[validator] object ValidatorUtil {
...
+            .flatMap { _ =>
+              connection.deleteUser(endUserName)
+            }
+            .recover { case NonFatal(ex) =>
```
martinflorian-da, line 324: Honestly I think I would just fail here. But no super strong opinion.

If you do keep it let's make the log message a `warning` though.

martinflorian-da (approved):
Nice work, thank you!

Issue #2143, opened 2025-09-05
tune automation jitter to reduce tps spikes during rounds


nicu-da:
We must be careful with the config to not push triggers too much into the future and risk missing rewards

nicu-da:
after the new scheduling landed in cilr:

<img width="2253" height="1262" alt="Image" src="https://github.com/user-attachments/assets/7af38bca-398f-4714-bb4a-bd22796787af" />

<img width="2279" height="471" alt="Image" src="https://github.com/user-attachments/assets/f19ed746-3d58-4ddc-8f3d-39a8db4be8e6" />

<img width="2279" height="471" alt="Image" src="https://github.com/user-attachments/assets/002ae546-ac7f-43ce-999a-49b823f64fb8" />


before 

<img width="2265" height="977" alt="Image" src="https://github.com/user-attachments/assets/abbbed13-3d55-4ede-b793-be344c6125cb" />

after

<img width="2265" height="977" alt="Image" src="https://github.com/user-attachments/assets/fb38d4cf-5f11-4a7a-a7fd-d8d6665917f5" />

so we got about 10-15% off the peak extra (compared to running with 300s intervals)

PR #2653, opened 2025-10-10
[ci] Schedule reward operations over an interval of one tick

nicu-da:
We compute the start of the interval as the smallest future data from the opens/closing time of all the open rounds.

This ensures that you run the automation at least twice during the 20m for which a round is open.

The config is default enabled but can be disabled.

Also increase the reward interval default (used when the new scheduler is disabled) to 300s based on the values we tested in cilr. 

Part of https://github.com/hyperledger-labs/splice/issues/2143

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/EnvironmentDefinition.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/EnvironmentDefinition.scala
@@ -427,7 +427,7 @@ case class EnvironmentDefinition(
       )
       .addConfigTransform((_, conf) =>
         ConfigTransforms.updateAllAutomationConfigs(
-          _.focus(_.rewardOperationEnableRoundBasedInterval).replace(true)
+          _.focus(_.rewardOperationEnableRoundBasedInterval).replace(false)
         )(conf)
       )
       .withSequencerConnectionsFromScanDisabled(10_000)


apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/CollectRewardsAndMergeAmuletsTrigger.scala
```diff
@@ -38,7 +40,24 @@ class CollectRewardsAndMergeAmuletsTrigger(
...
   override protected def extraMetricLabels = Seq("party" -> store.key.endUserParty.toString)
 
-  override def isRewardOperationTrigger: Boolean = true
+  override protected val pollingScheduler: PollingTrigger.DelayedFutureScheduler =
+    if (context.config.rewardOperationEnableRoundBasedInterval)
+      new RoundBasedUniformFutureScheduler(
+        (ec: ExecutionContext, tc: TraceContext) =>
+          scanConnection
+            .getOpenAndIssuingMiningRounds()(ec, mat, tc)
+            .map(_._1.map(_.contract.payload))(ec),
```
moritzkiefer-da: that doesn't look right to me. Creating activity records as done in ReceiveFaucetCouponTrigger depends on the open round. But minting activity records as done here is done in the issuing round.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/RoundBasedUniformFutureScheduler.scala
```diff
@@ -0,0 +1,110 @@
...
+  )(implicit ec: ExecutionContext, tc: TraceContext): FutureUnlessShutdown[Unit] = {
+    FutureUnlessShutdown
+      .outcomeF(
+        openRoundReader.getOpenRounds
+      )
+      .flatMap { openRounds =>
+        val now = clock.now.toInstant
+        @SuppressWarnings(Array("org.wartremover.warts.IterableOps"))
+        val firstOpenRound = openRounds.minBy(round => {
+          if (round.opensAt.isAfter(now)) {
```
moritzkiefer-da: The logic here doesn't make sense to me. Why does the closing date matter for rounds that will open in the future?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/RoundBasedUniformFutureScheduler.scala
```diff
@@ -0,0 +1,110 @@
...
+      action: => Unit
+  )(implicit ec: ExecutionContext, tc: TraceContext): FutureUnlessShutdown[Unit] = {
+    FutureUnlessShutdown
+      .outcomeF(
+        openRoundReader.getOpenRounds
+      )
+      .flatMap { openRounds =>
+        val now = clock.now.toInstant
+        @SuppressWarnings(Array("org.wartremover.warts.IterableOps"))
+        val firstOpenRound = openRounds.minBy(round => {
```
moritzkiefer-da: nit and maybe irrelevant given the resolution to my next comment: it sounds like you're not looking for the first round but maybe "nextActionableRound"?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/RoundBasedUniformFutureScheduler.scala
```diff
@@ -0,0 +1,110 @@
...
+          } else round.targetClosesAt
+        })
+        val minSchedulingTimeForOpenRound = if (firstOpenRound.opensAt.isAfter(now)) {
+          firstOpenRound.opensAt
+        } else {
+          firstOpenRound.targetClosesAt
+        }
+        // the triggers always do a run during startup so we can ignore that scenario as it would cover any rounds missed previosly
+        // therefore we always schedule just for the next open round
+        scheduleBetween(
```
moritzkiefer-da: I'm not following the logic here. Let's say we have this:
1. The smallest round opens at T0, first tick after T1, closes at T2.
2. Let's say I run just after T0, I create my activity record.
3. Now you schedule again. In your logic firstOpenRound will still be the same. So minSchedulingTimeForOpenRound is opensAt. So now you schedule again at a random time between T0 and T1 but there is nothing I can do until T1.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/ReceiveSvRewardCouponTrigger.scala
```diff
@@ -49,7 +49,24 @@ class ReceiveSvRewardCouponTrigger(
     override val tracer: Tracer,
 ) extends PollingParallelTaskExecutionTrigger[ReceiveSvRewardCouponTrigger.Task] {
 
-  override def isRewardOperationTrigger: Boolean = true
+  override protected val pollingScheduler: PollingTrigger.DelayedFutureScheduler = {
+    if (context.config.rewardOperationEnableRoundBasedInterval)
+      new RoundBasedUniformFutureScheduler(
```
moritzkiefer-da: Wiring it through the polling interval sounds nice on the one hand since it's generic. But it also seems quite tricky to get right, see my comments on the scheduling logic. Have you considered to turn this around:

Run the trigger with a normal polling interval 30s. But then in the trigger just add a sleep, similar to what we do e.g. for the vetting triggers to randomize the load. The big advantage of that is that the trigger can see "I already created activity records up to round N so I need to wait for round N+1" whereas right now doing this through the polling scheduling logic seems to make it kinda confusing.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/CollectRewardsAndMergeAmuletsTrigger.scala
nicu-da: Ugh, yeah, I switched to open rounds for the faucet and sv rewards trigger and re-used for this one, switching tthis one to issuing rounds.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/RoundBasedUniformFutureScheduler.scala
nicu-da: True, for some reason in my minds i was mapping this to another round opening but the opening time should be the only one we care about.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/RoundBasedUniformFutureScheduler.scala
nicu-da: The idea was that when you run at T0 the scheduling should be right after T1 should be fixed in the next commit.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/ReceiveSvRewardCouponTrigger.scala
nicu-da: I think this has more downsides than upsides. You would still need to replicate the same logic in the sleep, but now you have some Frankenstein triggers that are even harder to follow.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/OpenRoundBasedUniformFutureScheduler.scala
```diff
@@ -0,0 +1,261 @@
...
+
+  override def scheduleNextPoll(
+      action: => Unit
+  )(implicit ec: ExecutionContext, tc: TraceContext): FutureUnlessShutdown[Unit] = {
+    retryProvider
+      .waitUnlessShutdown(
+        openRoundReader.getOpenRounds
+      )
+      .flatMap { openRounds =>
+        val firstRoundThatWillOpenInTheFuture =
```
moritzkiefer-da, line 152: I don't think your logic works for CollectRewardsTrigger. You are assuming that you need only a single run per round. That's true for ReceiveFaucetCouponTrigger and ReceiveSvRewardCouponTrigger but it is not true for CollectRewards as each run can only mint 100 rewards and there can be 1000s.

You can argue that the logic for workdone is enough to combat that but that seems quite risky to me to rely on around restarts and so on.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/OpenRoundBasedUniformFutureScheduler.scala
```diff
@@ -0,0 +1,261 @@
...
+
+    val delayFromMinSchedulingTime = java.time.Duration.ofNanos(
+      Random.nextLong(
+        nanosBetweenSchedulingInterval
+      )
+    )
+    val scheduleAt = minSchedulingTime.plus(delayFromMinSchedulingTime)
+
+    logger.info(
+      s"Schedling next poll at $scheduleAt for round $selectedRound interval $minSchedulingTime -> $maxSchedulingTime"
```
moritzkiefer-da:
```suggestion
      s"Scheduling next poll at $scheduleAt for round $selectedRound interval $minSchedulingTime -> $maxSchedulingTime"
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/OpenRoundBasedUniformFutureScheduler.scala
```diff
@@ -0,0 +1,261 @@
...
+              ),
+              maxSchedulingTimeForRound(round, issuingRounds),
+            )
+          case None =>
+            logger.info("No future issuing round, scheduling next poll with fixed interval")
+            scheduleBasedOnRegularInterval(action)
+        }
+      }
+  }
+  private def maxSchedulingTimeForRound(
```
moritzkiefer-da, line 231: I don't understand what you're trying to do here.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/ReceiveSvRewardCouponTrigger.scala
moritzkiefer-da: fair point but I find it very hard to convince myself that any of the logic is correct atm. For an individual sleep this is trivial to do for at least the ReceiveTriggers and pretty sure we can also find something easy for the CollectReward trigger

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/OpenRoundBasedUniformFutureScheduler.scala
nicu-da: I'm probably missing something but how is this different compared to the previous version? We have the comment that this is handled in the trigger itself https://github.com/hyperledger-labs/splice/blob/main/apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/CollectRewardsAndMergeAmuletsTrigger.scala#L43 and during restarts the triggers run once before the delay is applied

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/OpenRoundBasedUniformFutureScheduler.scala
nicu-da: We don't have the tick duration in the issuing rounds, we calculate the interval based on the round we're runnign for and the start time of the next round. If we can't find the next interval we just calculate based on the target close of the current round.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/ReceiveSvRewardCouponTrigger.scala
nicu-da: > that any of the logic is correct atm.

is that with the current state or previous state? 

> For an individual sleep this is trivial to do for at least the ReceiveTriggers and pretty sure we can also find something easy for the CollectReward trigger

I find this would be really confusing tbh, we already have the scheduling for the triggers, which are already pretty difficult to follow, and now we're just dropping random sleeps in the middle of triggers, it all just sounds like we're asking for trouble. 
I don't really see how is adapting the scheduling logic more complex than adding the sleep, we still have the same issues either way, we still need to calculate the interval for which to randomize for.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/OpenRoundBasedUniformFutureScheduler.scala
moritzkiefer-da: fair point

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/OpenRoundBasedUniformFutureScheduler.scala
moritzkiefer-da: I don't understand that, an issuing mining round is open for exactly two ticks https://github.com/hyperledger-labs/splice/blob/main/daml/splice-amulet/daml/Splice/AmuletRules.daml#L457

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/ReceiveSvRewardCouponTrigger.scala
moritzkiefer-da: I think to me the difference is context: If I have logic that is specific to one trigger it is very easy to reason about its effect compared to generic logic for 3 triggers with different requirements, e.g., the initial issue we had where it used open round times instead of issuing times or similar stuff.

Let me take another fresh look on Monday, you might be right that this is still the better tradeoff.

moritzkiefer-da: maybe a better way to phrase my concerns is to look at an example with ReceiveFaucetCouponTrigger:

With the sleep in the trigger, I can see that the on-ledger state says I minted round X so I can add a sleep until round X+1 and very easily convince myself that sleeping until then does not risk missing anything.

With the changed polling, the polling logic tracks some "last round processed". But that is not coupled to the on-ledger state of what you actually minted. So constructing an argument that this logic actually does the right thing becomes a lot harder at least for my stupid brain.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/OpenRoundBasedUniformFutureScheduler.scala
nicu-da: yes, I was simply trying to mimic the way the open one works. in the open round we have the tick duration available, here we don't. we could simply read the amulet rules and read the tick duration from there I guess so that we have the same logic everywhere.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/ReceiveSvRewardCouponTrigger.scala
nicu-da: Fair, though to point that the last round processed is just to cover a possible edge case where you run the trigger at the very end of a round and by the time you schedule the next round then the next round is already open so you will possibly miss a run, but the next run should cover the interval anyway. 

lets come back on it on monday, maybe I'm over estimating the complexity of the sleeps in the trigger as well. 

le: actually it makes no sense why we don't use the state from the ledger in the scheduler, will update to remove the state we keep there and use the ledger state to get the last round we processed

PR #2669, opened 2025-10-13
[ci] Run sv rewards trigger during a random time between the round open and one tick

nicu-da:
alternative to https://github.com/hyperledger-labs/splice/pull/2653 for https://github.com/hyperledger-labs/splice/issues/2143

Will extend to the other rewards triggers in another PR

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/ConfigTransforms.scala
+++ b/apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/ConfigTransforms.scala
@@ -26,7 +26,7 @@ import org.lfdecentralizedtrust.splice.validator.config.{
 import org.lfdecentralizedtrust.splice.wallet.config.WalletAppClientConfig
 import com.digitalasset.canton.SynchronizerAlias
 import com.digitalasset.canton.config.*
-import com.digitalasset.canton.config.RequireTypes.{NonNegativeInt, Port}
+import com.digitalasset.canton.config.RequireTypes.Port
 import monocle.macros.syntax.lens.*
 import org.apache.pekko.http.scaladsl.model.Uri
 import org.lfdecentralizedtrust.splice.sv.automation.singlesv.SvBftSequencerPeerOffboardingTrigger


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/RoundBasedTrigger.scala
```diff
@@ -0,0 +1,94 @@
...
+    ec: ExecutionContext,
+    mat: Materializer,
+    tracer: Tracer,
+) extends PollingParallelTaskExecutionTrigger[T] {
+  private val nextRunTime = new AtomicReference[Option[(Long, Instant)]](None)
+
+  override protected def retrieveTasks()(implicit tc: TraceContext): Future[Seq[T]] = {
+    if (shouldRun) {
+      retrieveAvailableTasksForRound().map(tasks => {
+        tasks.minByOption(_.roundDetails._1) match {
```
moritzkiefer-da: sorry for being so annoying but tracking rounds here independently from the on-ledger state still makes me nervous. To give a concrete example where it goes wrong: If you exceed the retries on a submission because you're disconnected, lagging, … your on-ledger state obviously does not advance. But you still advance here and then might miss one more round than needed. Yes the existing code with just an increased polling interval also didn't fully guard against that but the interval is still lower than what we end up with now (which obviously also came with higher spikes) so you have a higher chance to try again on the next polling interval.

Given that we do have the on-ledger state, why not use it? E.g., make `retrieAvailableTasksForRound` return the round for which the task is and then set the delay based on that. That doesn't seem meaningfully more complex but would avoid needing to reason about them being not being an sync.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/ReceiveSvRewardCouponTrigger.scala
```diff
@@ -47,14 +50,12 @@ class ReceiveSvRewardCouponTrigger(
...
-) extends PollingParallelTaskExecutionTrigger[ReceiveSvRewardCouponTrigger.Task] {
-
-  override def isRewardOperationTrigger: Boolean = true
+) extends RoundBasedTrigger[ReceiveSvRewardCouponTrigger.Task] {
 
   private val svParty = store.key.svParty
   private val dsoParty = store.key.dsoParty
 
-  override protected def retrieveTasks()(implicit
+  override protected def retrieveAvailableTasksForRound()(implicit
```
moritzkiefer-da, line 58: imho we still want a feature flag for this stuff

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/RoundBasedTrigger.scala
moritzkiefer-da: as discussed this is actually what's implemented, I just misread the code.

nicu-da: Discussed f2f, all good with the current approach

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/ReceiveSvRewardCouponTrigger.scala
nicu-da: added

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/EnvironmentDefinition.scala
```diff
@@ -425,6 +425,11 @@ case class EnvironmentDefinition(
               .replace(NonNegativeFiniteDuration.Zero)
           )(conf)
       )
+      .addConfigTransform((_, conf) =>
+        ConfigTransforms.updateAllAutomationConfigs(
+          _.focus(_.enableNewRewardTriggerScheduling).replace(false)
```
moritzkiefer-da, line 431: why do we need this? Shouldn't it also work in simtime?

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/ConfigTransforms.scala
```diff
@@ -333,6 +333,19 @@ object ConfigTransforms {
         }
     )
 
+  def updateInitialTickDuration(tick: NonNegativeFiniteDuration): ConfigTransform = {
+    ConfigTransforms.updateAllSvAppFoundDsoConfigs_(
+      _.copy(initialTickDuration = tick)
+    ) compose ConfigTransforms.updateAllAutomationConfigs(config =>
```
moritzkiefer-da: do you have an example where we need this? This seems more like a test is maybe missing a pause for some trigger. If you rely on timing that precise it feels like you're asking for flakes.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/EnvironmentDefinition.scala
nicu-da: The test I checked had rounds for 10m, and it was working because automation ran every second. with this change now we would need to update all the tests that had that behavior

nicu-da: works in wall clock time because we usually lower tick duration to 500ms

moritzkiefer-da: But you still poll every 10min to check if the simtime time has advanced then no? So if you advance simtime every 10s shouldn't you rerun the retrieveTasks function which will then include the task?

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/ConfigTransforms.scala
nicu-da: leftover from another issue that i fixed, though decided to keep it because in theory it could be that if you advance rounds fast enough (eg where tick time is reduced), some triggers might just not run for a round and actually hide issues. will remove it though, not sure what can of warms of flakes would this introduce.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/EnvironmentDefinition.scala
nicu-da: You need to advance time with a tick for the trigger to run. I can disable this and update the failing tests, though I have doubts about the value it adds.

moritzkiefer-da: alright let's keep it but add a comment

PR #2672, opened 2025-10-13
add a psql subcommand to cncluster script for convenient DB inspection

mblaze-da:
[static]

I've tested this by running the new subcommand against different databases in the cluster on a scratchnet.

> sequencer-0-pg
```
scratchnetc % cncluster psql sv-1 global-domain-0-sequencer                                                          
INFO: Retrieving DB connection info from pod description...
INFO: Retrieving DB password from secret [sequencer-0-pg-secrets]...
INFO: Deploying debug pod on version 0.4.19
pod/splice-debug created
INFO: Waiting for debug pod to become ready
pod/splice-debug condition met
INFO: Opening terminal on debug pod
psql (16.10 (Ubuntu 16.10-0ubuntu0.24.04.1), server 14.19 (Debian 14.19-1.pgdg13+1))
Type "help" for help.

cantonnet=# 
```

The optional search path parameter can be specified and it is set as shown in the following example:

> sv-app postgres
```
scratchnetc % cncluster psql sv-1 sv-app test
INFO: Retrieving DB connection info from pod description...
INFO: Retrieving DB password from secret [postgres-secrets]...
INFO: Deploying debug pod on version 0.4.19
pod/splice-debug created
INFO: Waiting for debug pod to become ready
pod/splice-debug condition met
INFO: Opening terminal on debug pod
psql (16.10 (Ubuntu 16.10-0ubuntu0.24.04.1), server 14.19 (Debian 14.19-1.pgdg13+1))
Type "help" for help.

cantonnet=# show search_path;
 search_path 
-------------
 test
(1 row)
```

### DIFF
--- a/build-tools/cncluster
+++ b/build-tools/cncluster
@@ -1458,7 +1458,7 @@ function subcmd_debug_shell() {
   fi
   repo="ghcr.io/digital-asset/decentralized-canton-sync-dev/docker"
 
-  local SHELL_COMMAND="${1:-/bin/bash}"
+  local SHELL_COMMAND="${*:-/bin/bash}"
 
   _info "Deploying debug pod on version $VERSION_NUMBER"
   kubectl apply --wait -f - <<EOF
@@ -2181,14 +2181,13 @@ subcommand_whitelist[psql]=""
 function subcmd_psql() {
   _cluster_must_exist
 
-  if [ $# -lt 2 ] || [ $# -gt 3 ]; then
-    _error "Usage: $SCRIPTNAME psql <namespace> <application> [search_path]"
+  if [ $# -ne 2 ]; then
+    _error "Usage: $SCRIPTNAME psql <namespace> <application>"
     exit 1
   fi
 
   local NAMESPACE="$1"
   local APPLICATION="$2"
-  local SEARCH_PATH=${3:-}
 
   _info "Retrieving DB connection info from pod description..."
   local DB_INIT_COMMAND
@@ -2297,16 +2296,29 @@ function subcmd_psql() {
     exit 1
   fi
 
-  subcmd_debug_shell " \
-    /bin/env
-      PGPASSWORD=$DB_PASSWORD \
-      ${SEARCH_PATH:+PGOPTIONS=--search_path=$SEARCH_PATH} \
-      psql \
-        --host=$DB_HOST \
-        --port=$DB_PORT \
-        --username=$DB_USERNAME \
-        --dbname=$DB_NAME \
-  "
+  case "${APPLICATION}" in
+    participant-*)
+      local SEARCH_PATH="participant"
+      ;;
+    sequencer-*)
+      local SEARCH_PATH="sequencer"
+      ;;
+    mediator-*)
+      local SEARCH_PATH="mediator"
+      ;;
+    *)
+      local SEARCH_PATH="$DB_NAME"
+      ;;
+  esac
+
+  subcmd_debug_shell /bin/env \
+    PGPASSWORD="$DB_PASSWORD" \
+    PGOPTIONS="--search_path=${SEARCH_PATH},public" \
+    psql \
+      --host="$DB_HOST" \
+      --port="$DB_PORT" \
+      --username="$DB_USERNAME" \
+      --dbname="$DB_NAME"
 }
 
 ################################


build-tools/cncluster
```diff
@@ -1479,10 +1481,11 @@ spec:
     effect: "NoSchedule"
 EOF
 
+  # shellcheck disable=SC2086
```
julientinguely-da, line 1484: let's try to remove the shellcheck disable=SC2086 by adding the parentheses (like "$SHELL_COMMAND")

build-tools/cncluster
```diff
@@ -2172,6 +2175,135 @@ EOF
     _update_cluster_config "$deployment_config" "operatorDeployment"
 }
 
+subcommand_whitelist[psql]=""
```
julientinguely-da, line 2180: nit: could we add the `###` on top to match with the other methods

build-tools/cncluster
```diff
@@ -2172,6 +2175,135 @@ EOF
...
+
+  if [ -z "${DB_NAME:-}" ]; then
+    _error "Failed to retrieve DB name from pod description."
+    exit 1
+  fi
+
+  local PASSWORD_REF
+  PASSWORD_REF=$(
+    kubectl get deployment \
+      --namespace sv-1 \
```
julientinguely-da: I think you forgot to parametrize the namespace here

build-tools/cncluster
```diff
@@ -2172,6 +2175,135 @@ EOF
...
+  fi
+
+  local NAMESPACE="$1"
+  local APPLICATION="$2"
+  local SEARCH_PATH=${3:-}
+
+  _info "Retrieving DB connection info from pod description..."
+  local DB_INIT_COMMAND
+  DB_INIT_COMMAND=$(
+    kubectl describe pod \
```
julientinguely-da, line 2195: suggestion: couldn't we use `kubectl get pod` instead of describe here, would seem less brittle?

build-tools/cncluster
mblaze-da: Then it would not work. We need this to expand.

build-tools/cncluster
mblaze-da: good catch

build-tools/cncluster
mblaze-da: I don't think it would make it much more reliable. We'd still need to `sed` the psql command out of a larger script in the `"command"` field. We'd just exchange a `grep -i psql` for a more complicated JSON path expression.

build-tools/cncluster
```diff
@@ -2172,6 +2175,137 @@ EOF
     _update_cluster_config "$deployment_config" "operatorDeployment"
 }
 
+###
+
+subcommand_whitelist[psql]=""
+function subcmd_psql() {
+  _cluster_must_exist
+
+  if [ $# -lt 2 ] || [ $# -gt 3 ]; then
```
martinflorian-da: The `-gt 3` seems to prevent me from adding flags?

martinflorian-da: A nvm the flags are not for this :) ignore me

build-tools/cncluster
```diff
@@ -2172,6 +2175,137 @@ EOF
...
+      -U=*|--username=*)
+        local DB_USERNAME="${1#*=}"
+        shift
+        ;;
+      -d|--dbname)
+        local DB_NAME="$2"
+        shift 2
+        ;;
+      -d=*|--dbname=*)
+        local DB_NAME="${1#*=}"
```
martinflorian-da: OK so at least on the deployment I tested on (`sv-app`) this is wrong: we don't want the default db (`cantonnet`), but the one that the init container is creating `sv_sv_1`:

```until errmsg=$(psql -h 10.60.65.197 -p 5432 --username=cnadmin --dbname=cantonnet -c 'create database sv_sv_1' 2\u003e\u00261); do\n  if [[ $errmsg == *\"already exists\"* ]]; then\n    echo \"Database sv_sv_1 already exists. Done.\"\n    break\n  fi\n\n  echo \"trying to create postgres database sv_sv_1, last error: $errmsg\";\n  sleep 2;\ndone\n"```

Comment (martinflorian-da) [reply build-tools/cncluster]: (sorry for the horrible paste, but I hope the point comes across; `create database sv_sv_1` is the interesting one)

File: build-tools/cncluster [hunk 7]
```diff
@@ -2172,6 +2175,137 @@ EOF
...
+
+  if [ -z "${DB_PASSWORD}" ]; then
+    _error "Failed to retrieve DB password."
+    exit 1
+  fi
+
+  subcmd_debug_shell " \
+    /bin/env
+      PGPASSWORD=$DB_PASSWORD \
+      ${SEARCH_PATH:+PGOPTIONS=--search_path=$SEARCH_PATH} \
```
martinflorian-da: I wonder if there is a nice way to infer the search path...

I often just guess/know it (is it perhaps always the same as database name for our deployments?).

But maybe there is some nice postgres way to just add all existing schemas to the search path or something...

build-tools/cncluster
mblaze-da: fixed

build-tools/cncluster
mblaze-da: Well, I tried to set the search path to include all schemas but I couldn't get it to work. Listing all schemas is simple:
```
PGPASSWORD=$DB_PASSWORD psql \
  --host=$DB_HOST \
  --port=$DB_PORT \
  --username=$DB_USERNAME \
  --dbname=$DB_NAME \
  --tuples-only \
  --pset format=unaligned \
  --command "SELECT string_agg(schema_name, ',') FROM information_schema.schemata"
```
as well as setting the search path:
```
PGOPTIONS=--search_path=$(...)
```
What I couldn't figure out was how to delay the evaluation of the schema listing till pod runtime while at the same time keeping the session interactive. I could get either one or the other to work. Not both at the same time. I gave up and implemented a simple mapping that uses `DB_NAME` with exceptions for participant, sequencer and mediator, as we've discussed on Slack. TBH I don't want to invest much more time into improving this as I don't think it is that important.

Perhaps this could be achieved by modifying the pod configuration but I did not want to duplicate any code from the debug_shell subcommand.

I tested it with participant and sv-app and it seems to be working.

mblaze-da:
@isegall-da there is a section about connecting to PG DBs with the help of `cncluster` in the internal repo that I could update. But this would be a separate PR, right?

build-tools/cncluster
martinflorian-da: Sounds like a perfectly reasonable decision to me!

martinflorian-da:
> But this would be a separate PR, right?

@mblaze-da Yep. Agree that it would be nice to do this as a follow-up.

martinflorian-da (approved):
I love it and will use it today; thanks a lot!

martinflorian-da:
(tested the use-cases that I need, they work! :) )

Issue #1572, opened 2025-07-18
Remove UpdateHistory from DbDsoStore


rautenrieth-da:
`DbDsoStore` and `DbScanStore` ingest data for the same party, therefore their UpdateHistory should contain exactly the same data. Except only the UpdateHistory of the `DbScanStore` is backfilled.

Duplicating the data and the backfilling process is a waste of storage and network resources.

- Document that the sv app will never be able to answer historical queries, and such queries will have to go through the scan API
- Stop ingesting UpdateHistory data in `DbDsoStore`
- Truncate the existing data in the database

PR #2675, opened 2025-10-13
Disable UpdateHistory for SV and Splitwell apps

OriolMunoz-da:
[preflight](https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/41199/workflows/056db189-a96e-4591-8d52-81201dbe2aca)

Fixes https://github.com/hyperledger-labs/splice/issues/1572

This PR removes the UpdateHistory from:
- sv-app SvSvStore
- sv-app SvDsoStore
- SplitwellStore

Includes also:
- Separate UpdateHistory from stores. The alternative seemed to involve cake pattern / visitor pattern / instance matching when registering the ingestion service. This seems way cleaner.
- No more passing around Storage and throwing on every spot that can only work with DbStorage
- No more Option[UpdateHistory] metrics, now that UpdateHistory is separate from stores
- See also https://github.com/hyperledger-labs/splice/pull/2686 for more annoyance


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/SequencerPruningTrigger.scala
```diff
@@ -24,34 +24,76 @@ import io.opentelemetry.api.trace.Tracer
...
+
   override def performWorkIfAvailable()(implicit traceContext: TraceContext): Future[Boolean] =
     for {
       synchronizerId <- sequencerAdminConnection.getStatus.map(_.trySuccess.synchronizerId)
-      recordTimeRangeO <- store.updateHistory
-        .getRecordTimeRange(migrationId)
-        .map(_.get(synchronizerId))
+      scanConnection <- createScanConnection()
+      recordTimeRangeO <- scanConnection
+        .getMigrationInfo(migrationId)
```
OriolMunoz-da, line 83: pay attention to this change

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V047__truncate_sv_app_history.sql
```diff
@@ -0,0 +1,26 @@
...
+DECLARE
+    descriptors TEXT[];
+BEGIN
+
+-- array equality (ordered) ensures that exactly these two, no more, no less, are there <=> it's the SV app
+select array_agg(store_name order by store_name) into descriptors
+from update_history_descriptors;
+
+IF descriptors  = '{"DbSvDsoStore", "DbSvSvStore"}' THEN
+    RAISE NOTICE 'Truncating update history tables as only SV app descriptors are present. Descriptors: %', descriptors::text;
```
OriolMunoz-da: fun fact, flyway logs these, very useful!

Image

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V047__truncate_sv_app_history.sql
```diff
@@ -0,0 +1,26 @@
...
+DO $$
+DECLARE
+    descriptors TEXT[];
+BEGIN
+
+-- array equality (ordered) ensures that exactly these two, no more, no less, are there <=> it's the SV app
+select array_agg(store_name order by store_name) into descriptors
+from update_history_descriptors;
+
+IF descriptors  = '{"DbSvDsoStore", "DbSvSvStore"}' THEN
```
OriolMunoz-da: Image

rautenrieth-da: This is more strict than it needs to be - some of the descriptors in the `update_history_descriptors` might not be used by the tables that we want to truncate.

Also, in case someone has a very stupid production setup where multiple apps write into the same database, this will fail to delete the unused data.

I still think it's fine as is, in real setups it should work and worst case it just leaves around unused data.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V051__truncate_sv_splitwell_history.sql
```diff
@@ -0,0 +1,26 @@
...
+-- array equality (ordered) ensures that exactly these two, no more, no less, are there <=> it's the SV app
+select array_agg(store_name order by store_name) into descriptors
+from update_history_descriptors;
+
+IF descriptors  = '{"DbSvDsoStore", "DbSvSvStore"}' THEN
+    RAISE NOTICE 'Truncating update history tables as only SV app descriptors are present. Descriptors: %', descriptors::text;
+    EXECUTE 'TRUNCATE TABLE update_history_assignments CASCADE';
+    EXECUTE 'TRUNCATE TABLE update_history_unassignments CASCADE';
+    EXECUTE 'TRUNCATE TABLE update_history_backfilling CASCADE';
+    EXECUTE 'TRUNCATE TABLE update_history_creates CASCADE';
```
rautenrieth-da, line 17: Note to self: The `acs_snapshot_data` table has a foreign key constraint to `update_history_creates`. We don't need to truncate `acs_snapshot_data` because the SV app doesn't write any data into it (and we'd rather fail if there is any unexpected data in there after all).

apps/splitwell/src/main/scala/org/lfdecentralizedtrust/splice/splitwell/SplitwellApp.scala
```diff
@@ -129,6 +129,7 @@ class SplitwellApp(
       migrationInfo,
       participantId,
     )
+    // splitwell does not need to have UpdateHistory
```
rautenrieth-da, line 133: Do we need to truncate any splitwell tables as well?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/SequencerPruningTrigger.scala
```diff
@@ -24,34 +24,76 @@ import io.opentelemetry.api.trace.Tracer
...
+    httpClient: HttpClient,
+    templateDecoder: TemplateJsonDecoder,
 ) extends PollingTrigger {
   val pruningMetrics = new SequencerPruningMetrics(
     context.metricsFactory
   )
 
+  // Similar to PublishScanConfigTrigger, this class creates its own scan connection
+  // on demand, because scan might not be available at application startup.
+  private def createScanConnection()(implicit
```
rautenrieth-da, line 65: The code is duplicated in `HttpSvAdminHandler`. Consider moving this to some helper/trait.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/ExternalPartyWalletService.scala
```diff
@@ -60,8 +63,22 @@ class ExternalPartyWalletService(
       participantId,
     )
 
+  val updateHistory = new UpdateHistory(
```
rautenrieth-da, line 67: Do we have any tests that cover the update history of external parties?

PR #2678, opened 2025-10-13
[ci] Update faucet triggers to use new scheduling

nicu-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/automation/ReceiveFaucetCouponTrigger.scala
+++ b/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/automation/ReceiveFaucetCouponTrigger.scala
@@ -4,8 +4,14 @@
 package org.lfdecentralizedtrust.splice.validator.automation
 
 import cats.data.OptionT
+import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
+import com.digitalasset.canton.time.Clock
+import com.digitalasset.canton.tracing.TraceContext
+import io.opentelemetry.api.trace.Tracer
+import org.apache.pekko.stream.Materializer
+import org.lfdecentralizedtrust.splice.automation.RoundBasedRewardTrigger.RoundBasedTask
 import org.lfdecentralizedtrust.splice.automation.{
-  PollingParallelTaskExecutionTrigger,
+  RoundBasedRewardTrigger,
   TaskOutcome,
   TaskSuccess,
   TriggerContext,
@@ -19,15 +25,12 @@ import org.lfdecentralizedtrust.splice.validator.store.ValidatorStore
 import org.lfdecentralizedtrust.splice.validator.util.ValidatorUtil
 import org.lfdecentralizedtrust.splice.wallet.UserWalletManager
 import org.lfdecentralizedtrust.splice.wallet.util.{TopupUtil, ValidatorTopupConfig}
-import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
-import com.digitalasset.canton.time.Clock
-import com.digitalasset.canton.tracing.TraceContext
-import io.opentelemetry.api.trace.Tracer
-import org.apache.pekko.stream.Materializer
 
+import java.time.temporal.ChronoUnit
+import java.time.{Duration, Instant}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.OptionConverters.*
-import math.Ordering.Implicits.*
+import scala.math.Ordering.Implicits.*
 
 class ReceiveFaucetCouponTrigger(
     override protected val context: TriggerContext,
@@ -41,17 +44,14 @@ class ReceiveFaucetCouponTrigger(
     override val ec: ExecutionContext,
     override val tracer: Tracer,
     materializer: Materializer,
-) extends PollingParallelTaskExecutionTrigger[ReceiveFaucetCouponTrigger.Task] {
-
-  override def isRewardOperationTrigger = true
+) extends RoundBasedRewardTrigger[ReceiveFaucetCouponTrigger.Task] {
 
   private val validatorParty = validatorStore.key.validatorParty
 
-  override protected def retrieveTasks()(implicit
+  override protected def retrieveAvailableTasksForRound()(implicit
       tc: TraceContext
-  ): Future[Seq[ReceiveFaucetCouponTrigger.Task]] = {
+  ): Future[Seq[ReceiveFaucetCouponTrigger.Task]] =
     retrieveNextRoundToClaim().value.map(_.toList)
-  }
 
   private def retrieveNextRoundToClaim()(implicit
       tc: TraceContext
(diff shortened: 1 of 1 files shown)


apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/CollectRewardsAndMergeAmuletsTrigger.scala
```diff
@@ -34,27 +40,45 @@ class CollectRewardsAndMergeAmuletsTrigger(
...
+  )
 
-  override def performWorkIfAvailable()(implicit traceContext: TraceContext): Future[Boolean] =
-    // Retry because we want to avoid missing rewards.
-    // We add the retry here instead of using a PeriodicTaskTrigger because
-    // PeriodicTaskTrigger does not allow defining whether work was performed
-    // that skips the polling interval. So if we have more rewards than we can
-    // get in one polling interval,
-    // it would wait until the next polling interval before trying to collect
-    // them which risks missing some.
```
nicu-da, line 50: moved this to the PollingParallelTaskExecutionTrigger that the reward triggered is implemetned on, which has the same behavior. it runs the tasks until it returns noop

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/CollectRewardsAndMergeAmuletsTrigger.scala
```diff
@@ -34,27 +40,45 @@ class CollectRewardsAndMergeAmuletsTrigger(
...
+    } yield {
+      val currentAmuletRules = AmuletConfigSchedule(amuletRules).getConfigAsOf(context.clock.now)
+      issuingRounds
+        .minByOption(_.payload.opensAt)
+        .map(issuingRound =>
+          CollectRewardsAndMergeAmuletsTask(
+            roundDetails =
+              Long.unbox(issuingRound.payload.round.number) -> issuingRound.payload.opensAt,
+            tickDuration =
+              Duration.of(currentAmuletRules.tickDuration.microseconds, ChronoUnit.MICROS),
```
moritzkiefer-da: any reason not to take targetClosesAt - opensAt / 2? afaict that's more accurate.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/CollectRewardsAndMergeAmuletsTrigger.scala
```diff
@@ -34,27 +40,45 @@ class CollectRewardsAndMergeAmuletsTrigger(
...
-  override def isRewardOperationTrigger: Boolean = true
+  override protected def retrieveAvailableTasksForRound()(implicit
+      tc: TraceContext
+  ): Future[Seq[CollectRewardsAndMergeAmuletsTask]] = {
+    for {
+      amuletRules <- scanConnection.getAmuletRules()
+      (_, issuingRounds) <- scanConnection.getOpenAndIssuingMiningRounds()
+    } yield {
+      val currentAmuletRules = AmuletConfigSchedule(amuletRules).getConfigAsOf(context.clock.now)
+      issuingRounds
```
moritzkiefer-da: hm why does this work? You take the minimum issuing round

Let's say I have issuing round N open at t0 and close at t0+2*tick and issuing round N+1 open at t0+tick. Won't you wait until t0+2*tick before you switch to issuing round N+1 at which point the first tick for N+1 is already past? I guess the argument is that it doesn't matter as long as you run every 10min but this seems kind of risky as issuing rounds are not fully aligned on tick boundaries.

The right option seems to be to look at the rewards that you still have to mint and determine the round from those no?

nicu-da: Hmm, isn't it that we actually run really aggressively? Of course we make the assumption that rounds are more or less distributed evently and they are overlapping but you can't have round N with closing T0 and round N+1 with closing T1 for which T1 < T0.

But in your example it will run between to and t0+1tick, after this runs assuming we have n + 1 from t0 + tick, the trigger eventually runs within the normal schedule, see that N+1 is open, and just schedule between t0+tick and t0 + 2tick. 

There's no scenario where you wouldn't run already for a round when time it after open + tick. 

> I guess the argument is that it doesn't matter as long as you run every 10min but this seems kind of risky as issuing rounds are not fully aligned on tick boundaries.

Well the trigger runs more or less every 30s so all it matters is the rounds it sees open then. it can run with an interval of more than 10m between actual work, if for one round it gets scheduled at the opening of the round and for the next one it gets scheduled close to the end of the first tick, but even so it's guaranteed to run for the first tick after opening.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/CollectRewardsAndMergeAmuletsTrigger.scala
nicu-da: no real reason, just to keep it consistent. will change

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/CollectRewardsAndMergeAmuletsTrigger.scala
moritzkiefer-da: You're right but then does this just do nothing in actually delaying anything? Let's say you finished minting for round N. While round N is still the lowest you'll just continue triggering every 30s. Once round N is archived, round N+1 is already in the middle (assuming they align) so then you schedule something in N+1 opensAt, opensAt + tick which is already the case so you trigger again every 30s? That seems worse than the current state because you might start very aggressively merging coin which both causes a fair bit of traffic and also has high traffic costs for the submitting node.

nicu-da: I'm probably missing something, sorry for being so slow.

We have 3 rounds:

R1 - opens at T0, closes T2
R2 - opens at T1 closes T3
R3 - opens T2 closes T4

We try to run as such:

For R1: between T0 and T1
For R2: Between T1 and T2 (at which R1 is still open)
For R3 between T2 and T3 (at which R2 is still open and R1 is archived)

At which point would it be problematic to run? (having issues understanding how this is worse compared to running more often, as in how we do it now)

nicu-da: Discussed f2f, issue is the use of the earliest issuing round, which is easily fixable, and also checking the rewards we have to ensure past rounds were collected in scenarios where the synchronizer had issue and the rounds are not really  overlapping over a long period.

PR #2686, opened 2025-10-14
Make scanConfig mandatory in SV app

OriolMunoz-da:
[scratch run for good measure](https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/36353/workflows/6b0c090f-49ae-4861-82dc-74dc38f45762)


docs/src/release_notes.rst
```diff
@@ -29,6 +29,12 @@ Upcoming
     - The SV app will no longer store the update history and such, will not be able to answer historical queries.
       All updates involving the DSO party will still be stored and returned by Scan.
 
+    - Deployment
+
+      - The helm values under ``scan``, that is ``publicUrl`` and ``internalUrl`` are now mandatory.
+        In practice we expect that all SVs had already set these values.
```
```suggestion
        All SVs already deploy scan on DevNet, TestNet and MainNet so this should have no impact.
```

moritzkiefer-da: do we make it mandatory in the json schema? I think not from a quick look, probably wanna change that then.

OriolMunoz-da: argh, I keep forgetting that's a thing, mb

OriolMunoz-da: re the suggestion:

> All SVs already deploy scan on DevNet, TestNet and MainNet so this should have no impact.

It's not just that, but also that they should have already configured the URLs in the SV app

moritzkiefer-da: Yeah but I don't think anyone even understands that difference tbh. But I'm fine with rephrasing. "we expect" just sounds a bit too weak given that we know they set it as scan urls are published in daml no?

PR #2720, opened 2025-10-15
Fix race condition in ScanEventStore.getEvents

OriolMunoz-da:
Situation:
* verdictsF = [1,2,3]
* updates = [1, 2] did not YET ingest 3
* updates ingests [3]
* cap=3
* result is [1, 2, 3], where 3 does not have an update

whereas the correct flow:
* cap=2
* verdictsF = [1,2] 3 is ingested but not included cause it's outside the cap
* updates = [1, 2] did not YET ingest 3, all good
* updates ingests [3], but is obviouslky not in the result above
* result is [1, 2], and all is good in this world

See also comment in code for explanation

Fixes https://github.com/hyperledger-labs/splice/issues/2513
Fixes https://github.com/DACH-NY/cn-test-failures/issues/5818

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanEventStore.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanEventStore.scala
@@ -61,27 +61,29 @@ class ScanEventStore(
       currentMigrationId: Long,
       limit: PageLimit,
   )(implicit tc: TraceContext): Future[Seq[Event]] = {
-    val verdictsF = verdictStore.listVerdicts(
-      afterO = afterO,
-      includeImportUpdates = false,
-      limit = limit.limit,
-    )
-    val updatesF = updateHistory.getUpdatesWithoutImportUpdates(afterO, limit)
-
     for {
+      // the cap must be computed before, to not include events that shouldn't be included in the response
       currentCap <- resolveCurrentMigrationCap(
         verdictStore.lastIngestedRecordTime,
         updateHistory.lastIngestedRecordTime,
         currentMigrationId,
       )
+      (updatesAll, verdictsAll) <- {
+        // launch them in parallel
+        val verdictsF = verdictStore.listVerdicts(
+          afterO = afterO,
+          includeImportUpdates = false,
+          limit = limit.limit,
+        )
+        val updatesF = updateHistory.getUpdatesWithoutImportUpdates(afterO, limit)
+        verdictsF.flatMap(verdicts => updatesF.map(_ -> verdicts))
+      }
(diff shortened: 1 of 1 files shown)


dfordivam:
Thank you! I had somewhat figured out the what is going on from the logs but couldn't get to the root cause. additional logging is very useful too!

dfordivam:
@OriolMunoz-da On a closer look I noticed that the `currentCap` was already being fetched before doing the DB read of verdicts and updates, so I did not understand how could this situation happen. I expect that the cap should be 2 even in the old code. 


> Situation:

    verdictsF = [1,2,3]
    updates = [1, 2] did not YET ingest 3
    updates ingests [3]
    cap=3
    result is [1, 2, 3], where 3 does not have an update

From the CI failure logs you shared earlier, I noticed that the mismatch event had an update without a verdict. This should have never happened and I could not get to the root cause of it. (I confirmed that the verdict for this event was sent by the mediator to the scan app). I did notice multiple DB contention errors during the time the verdict was supposed to be ingested. So even if the verdict ingestion had failed, I couldn't reason out why the event would be returned with only the update. This is still an open question, as the existing code should not let this situation happen.

dfordivam:
> @dfordivam where did you see:
> 
> > On a closer look I noticed that the currentCap was already being fetched before doing the DB read of verdicts and updates,

@ray-roestenburg-da I meant in the original code, the currentCap was being fetched before.  (https://github.com/hyperledger-labs/splice/pull/2039/files#diff-99092ad7438395e843f923bef0916d5c230af6ab0b7e5d25e805dbda8ee55792R72-R84) 

```
      currentCap <- resolveCurrentMigrationCap(
        verdictStore.lastIngestedRecordTime,
        updateHistory.lastIngestedRecordTime,
        currentMigrationId,
      )
      isAllowed = ScanEventStore.allowF(afterO, currentMigrationId, currentCap)
      verdicts <- verdictsF
      cappedVerdicts = verdicts.filter(v => isAllowed(v.migrationId, v.recordTime))
      // Fetch views only for filtered verdicts
      verdictsWithViews <- Future.traverse(cappedVerdicts)(v =>
        verdictStore.listTransactionViews(v.rowId).map(views => v -> views)
      )
      updatesAll <- updatesF
```

OriolMunoz-da:
>  I meant in the original code, the currentCap was being fetched before.

Unfortunately no, by doing [this](https://github.com/hyperledger-labs/splice/pull/2720/files#diff-99092ad7438395e843f923bef0916d5c230af6ab0b7e5d25e805dbda8ee55792L64-L70) before the currentCap, you're already querying both stores.
In Scala Futures (default ones, at least) are eagerly evaluated, not lazily.

PR #2726, opened 2025-10-16
Make alreadyExists check in verdict store more robust

OriolMunoz-da:
Strictly not necessary given that the trigger should already ensure this doesn't happen, but since we've seen test failures (due to the missing history_id) where the alreadyExists check was not quite right, this should make sure it doesn't happen again and instead gives us a unique key violation

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/plugins/EventHistorySanityCheckPlugin.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/plugins/EventHistorySanityCheckPlugin.scala
@@ -1,20 +1,25 @@
 package org.lfdecentralizedtrust.splice.integration.plugins
 
 import cats.data.Chain
-import org.lfdecentralizedtrust.splice.console.ScanAppBackendReference
-import org.lfdecentralizedtrust.splice.environment.SpliceEnvironment
-import org.lfdecentralizedtrust.splice.http.v0.definitions.DamlValueEncoding.members.CompactJson
-import org.lfdecentralizedtrust.splice.http.v0.definitions.EventHistoryItem
-import org.lfdecentralizedtrust.splice.http.v0.definitions.UpdateHistoryItemV2.members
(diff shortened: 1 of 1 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanVerdictStore.scala
```diff
@@ -256,7 +253,7 @@ class DbScanVerdictStore(
             "scanVerdict.insertVerdictAndTransactionViews.batch",
           )
       ).map { _ =>
-        val maxRt = items.map(_._1.recordTime).reduceOption((a, b) => if (a >= b) a else b)
+        val maxRt = items.map(_._1.recordTime).maxOption
```
ray-roestenburg-da, line 256: nice cleanup

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanVerdictStore.scala
```diff
@@ -213,24 +213,22 @@ class DbScanVerdictStore(
...
-             select exists(
-               select 1
+      val checkExist = (sql"""
+               select update_id
                from #${Tables.verdicts}
                where history_id = $historyId
-                 and update_id = ${headVerdict.updateId}
-             )
-           """.as[Boolean].head
+                 and update_id IN """ ++ inClause(items.map(_._1.updateId))).as[String]
```
ray-roestenburg-da, line 220: 👍 ~should~Will  use index just fine cause Oriol checked it 😄 (and this is like max 50 update_id's at a time (because of batch), so should be good)

OriolMunoz-da: checked, does

ray-roestenburg-da: Thanks

PR #2754, opened 2025-10-20
Merge main into canton-3.4

martinflorian-da:
Mainly I want the last commmit there to contain `[ci]`...


nicu-da (approved):

moritzkiefer-da (approved):

moritzkiefer-da:
> @moritzkiefer-da if you have a better idea don't let me slow you down; probably we could also just remove the branche protection for canton-3.4 briefly

not sure what you're trying to achieve here?

PR #2765, opened 2025-10-21
move grafana pg instance to infra nodes

isegall-da:
Because:
1. It makes more sense
2. On e.g. ci cluster, the node types for infra and app are different, and the storage type that we use (standard-rwo) is not supported on the c4 nodes that we use for apps.

When applied, this will probably kill the data but I believe that's fine for the grafana DB itself.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/infra/src/observability.ts
+++ b/cluster/pulumi/infra/src/observability.ts
@@ -865,6 +865,9 @@ function installPostgres(namespace: ExactNamespace): SplicePostgres {
     'grafana-pg',
     'grafana-pg-secret',
     { db: { volumeSize: '20Gi' } }, // A tiny pvc should be enough for grafana
-    true // overrideDbSizeFromValues
+    true, // overrideDbSizeFromValues
+    false, // disableProtection
+    undefined, // chart version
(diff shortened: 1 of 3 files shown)


cluster/pulumi/infra/src/observability.ts
```diff
@@ -865,6 +865,9 @@ function installPostgres(namespace: ExactNamespace): SplicePostgres {
     'grafana-pg',
     'grafana-pg-secret',
     { db: { volumeSize: '20Gi' } }, // A tiny pvc should be enough for grafana
-    true // overrideDbSizeFromValues
+    true, // overrideDbSizeFromValues
+    false, // disableProtection
```
isegall-da, line 869: Does this actually need to be true for this move to succeed? Or since from pulumi perspective it's just changing some helm values, it bypasses protection?

nicu-da: that's just pulumi protection, which prevents the pulumi resource from being deleted. as the changes are only an update on the resource it does not care

nicu-da (approved):

PR #2781, opened 2025-10-21
Fix ACS snapshots on short-lived migrations

OriolMunoz-da:


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTrigger.scala
```diff
@@ -167,6 +167,25 @@ class AcsSnapshotTrigger(
...
+        )
+      minTime = migrationRecordTimeRange
+        .map(_._2.min)
+        .minOption
+        .getOrElse(
+          throw new IllegalStateException(
+            s"SynchronizerId with no data in $migrationRecordTimeRange"
+          )
+        )
+      migrationLastedLongEnough = minTime
```
moritzkiefer-da: Hm why is this right? I thought we take a snapshot every 3h full hours not 3h after the start of the migration?

OriolMunoz-da: fixed

PR #2785, opened 2025-10-21
Improve throughput of featured app activity trigger on 0.4.x

### LINKED ISSUE #2757: Improve throughput of featured app activity marker trigger
### Motivation

lots of markers being created, and SVs having trouble keeping up

### Approach

More batching (100 instead of 5 markers by default) while avoiding contention as follows:

- assign an int32 hash to all markers by taking the first 4 bytes of their md5 hash
  - create a partial expression index on this hash for all `Splice.Amulet:FeaturedAppActivityMarker` contracts  
- if there are less than 10k markers:
  - assign a position `p` to every SV based on their index in the sorted list of SVs
  - have every SV pick markers whose hash is in range `[p * 2^32 / numSvs, (p + 1) * 2^32 / numSvs]` 
     - (with the obvious corrections for rounding errors and dealing with the fact that Java and PG only support signed ints)
  - create up to the configured parallelism of batches
- if there are more than 10k markers (because one of the SVs was not processing their load):
  - have every SV pick batches by selecting a random int32, and taking the next `parallelism * batchSize` nexts tasks after that number
  - resample if we were too close to the end of the range and fewer than `parallelism * batchSize` tasks could be retrieved

### Validation to be done

- [ ] time of index creation on MainNet with its large number > 600k of activity markers
- [ ] query plans

meiersi-da:
Fixes #2757

[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/FeaturedAppActivityMarkerIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/FeaturedAppActivityMarkerIntegrationTest.scala
@@ -10,12 +10,13 @@ import org.lfdecentralizedtrust.splice.integration.tests.SpliceTests.Integration
 import org.lfdecentralizedtrust.splice.sv.automation.delegatebased.{
   AdvanceOpenMiningRoundTrigger,
   ExpireIssuingMiningRoundTrigger,
+  FeaturedAppActivityMarkerTrigger,
 }
 import org.lfdecentralizedtrust.splice.util.*
 import com.digitalasset.canton.config.NonNegativeFiniteDuration
 import org.lfdecentralizedtrust.splice.codegen.java.splice.amulet
 import org.lfdecentralizedtrust.splice.codegen.java.splice.api.featuredapprightv1
-import org.lfdecentralizedtrust.splice.wallet.store.TransferTxLogEntry
+
 import scala.jdk.CollectionConverters.*
 
 @org.lfdecentralizedtrust.splice.util.scalatesttags.SpliceAmulet_0_1_9
@@ -28,7 +29,8 @@ class FeaturedAppActivityMarkerIntegrationTest
 
   override def environmentDefinition: SpliceEnvironmentDefinition =
     EnvironmentDefinition
-      .simpleTopology1Sv(this.getClass.getSimpleName)
+      // Using 4Svs so that we see whether they manage to jointly complete all work
+      .simpleTopology4Svs(this.getClass.getSimpleName)
       .addConfigTransforms((_, config) =>
         ConfigTransforms.updateAllSvAppFoundDsoConfigs_(
           _.copy(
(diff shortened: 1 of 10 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/FeaturedAppActivityMarkerIntegrationTest.scala
```diff
@@ -58,92 +69,79 @@ class FeaturedAppActivityMarkerIntegrationTest
...
-        aliceValidatorBackend.participantClientWithAdminToken.ledger_api_extensions.acs.filterJava(
-          amulet.AppRewardCoupon.COMPANION
-        )(alice, c => c.data.provider == alice.toProtoPrimitive) should have size 2
-        bobValidatorBackend.participantClientWithAdminToken.ledger_api_extensions.acs.filterJava(
-          amulet.AppRewardCoupon.COMPANION
-        )(bob, c => c.data.provider == bob.toProtoPrimitive) should have size 1
-      },
-    )
-
-    // Advance three times so the round the coupons are assigned to is issuing
```
meiersi-da, line 113: Note that I dropped the whole minting assertion, as we test the `AppRewardCoupon` issuene in other tests already.

moritzkiefer-da (approved):
excellent, thanks a lot for pushing this through!

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -325,6 +325,14 @@ case class SvAppBackendConfig(
       NonNegativeFiniteDuration.ofMillis(250),
     delegatelessAutomationExpectedTaskDuration: Long = 5000, // milliseconds
     delegatelessAutomationExpiredRewardCouponBatchSize: Int = 20,
+    // What batch size to target for converting app activity markers
+    delegatelessAutomationFeaturedAppActivityMarkerBatchSize: Int = 100,
```
meiersi-da, line 329: @moritzkiefer-da : with this config, we'll convert around 40 to 100 tasks per second, depending on the commit latency. 

Assuming 40 markers/s, and 500k outstanding markers, we'll fit them all into `500e3 / (40 * 600) = 20 rounds`, i.e. the rounds for ~4 hours. Any concerns with this?

moritzkiefer-da: > we'll convert around 40 to 100 tasks per second, depending on the commit latency.

Tbc, you are counting a task as a marker here not as a batch which is what a task would be in our trigger naming right? I think the relevant metric is more TPS. I think something like 5 TPS across SVs is probably fine, more would be problematic but in that case we can probably adjust parallelism to keep it under control or ask some SVs to disable the trigger.

meiersi-da: Yeah sorry. Should have said makers per second.

I was actually more concerned about how crowded the few rounds get. That said: it's OK. There are more rewards to be gained. Let's optimize for network stability first; and for that this config seems totally fine. Sounds like SV-2 will clean-up the ACS before the other ones even get to deploy 0.4.22.

moritzkiefer-da: this is now deployed on our node, we archive about 20-30k contracts per hour. If we assume 30k, it will take about 33h to archive 1 million markers. Haven't double checked your math but I suspect your assumptions on latency are quite a bit off. We have reasonably frequent timeouts and then record order publishing nonsense kicks in and latency goes kaputt

PR #2801, opened 2025-10-21
[ci] Add scan cache for open mining roudns

nicu-da:
This is the most expensive query that we currently run and has a performance impact on mainnet

https://github.com/DACH-NY/canton-network-internal/issues/2337

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/resources/include/scans/_scan.conf
+++ b/apps/app/src/test/resources/include/scans/_scan.conf
@@ -108,6 +108,10 @@
        ttl = 1s
        max-size = 1
       }
+      open-mining-rounds {
+       ttl = 1s
+       max-size = 1
+      }
   }
 
 }
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/CachingScanStore.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/CachingScanStore.scala
@@ -34,6 +34,7 @@ import org.lfdecentralizedtrust.splice.scan.config.{CacheConfig, ScanCacheConfig
 import org.lfdecentralizedtrust.splice.scan.store.db.{DbScanStoreMetrics, ScanAggregator}
 import org.lfdecentralizedtrust.splice.store.{
   Limit,
+  MiningRoundsStore,
   MultiDomainAcsStore,
   PageLimit,
   SortOrder,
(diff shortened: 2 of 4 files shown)


apps/app/src/test/resources/include/scans/_scan.conf
```diff
@@ -108,6 +108,10 @@
        ttl = 1s
        max-size = 1
       }
+      open-mining-rounds {
```
moritzkiefer-da, line 111: I wonder if we should change the config to warn on undefined endpoints unless they're explicitly specified to have no cache. Seem seasier to not miss something that way. Definitely not for this PR.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/CachingScanStore.scala
```diff
@@ -366,6 +367,16 @@ class CachingScanStore(
       store.lookupSvNodeState,
     ).get(svPartyId)
 
+  override def lookupOpenMiningRoundTriple()(implicit
```
moritzkiefer-da, line 370: do we have one for issuing rounds? if not we probably want ot add that as well.

moritzkiefer-da (approved):
nice thx

apps/app/src/test/resources/include/scans/_scan.conf
nicu-da: Issue is that we're explicitly adding caches to the store implementations actually, as we also want to re-use that internally. I do wonder if it actually makes sense to move the caches to the http layer, which would also allow us to implement this suggestion.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/CachingScanStore.scala
nicu-da: ugh, so this is mostly hit through the token standard apis. but indeed we have the open and issuing mining round endpoints in the scan handler which call the store directly with listContracts, so it will bypass any cache.

apps/app/src/test/resources/include/scans/_scan.conf
moritzkiefer-da: ah right. I don't think we can move the caches to the http layer, e.g amulet rules is hit from about 5 million different endpoints so caching that in the endpoints feels very inefficient. maybe some combination of store and endpoint caches

Issue #2409, opened 2025-09-25
Support trusting multiple SVs for Scan and Sequencer connections


meiersi-da:
### Status Quo

We only support "Trust Single" and "Full BFT".

### Problem

Some entities want more control, and only connect to a subset of SVs.

nicu-da:
> [@pasindutennage-da](https://github.com/pasindutennage-da) [@nicu-da](https://github.com/nicu-da) since [#3422](https://github.com/hyperledger-labs/splice/pull/3422) is moving close to closing all of this, I'm trying to zoom out a bit. What do we think about:
> 
>     * Exposing finer customization (such as the threshold to use) for sequencer connections as well? At this point it seems a bit confusing that we have it for scan but not for sequencer connections... (So I guess my question is: shall we also just do [Support configurable threshold for trusted sequencers in ValidatorDecentralizedSynchronizerConfig #3059](https://github.com/hyperledger-labs/splice/issues/3059) right away?)
> 
>     * Harmonizing the config... Shall we align the naming on the Helm/compose side; `trust-single`, `bft`, and `bft-custom` (the current names for scan) seem like a good fit for sequencer things as well? Having the same names seems nicer for clarity.

I would need a refresher a bit about the naming we use. I think it would make sense to use the same naming everywhere. Would be more obvious when we write the docs and we see the diff.

PR #2825, opened 2025-10-22
Add option to allow specific scans in the validator

pasindutennage-da:
Added new config option to enable a validator to specify the set of trusted scan connections. 

Solves scan related part of  https://github.com/hyperledger-labs/splice/issues/2409

scan-client{
  seedUrls = []
  trustedSvs = []
  threshold = []
  type = "trust-specific"
}

If selected the "trust-specific" option, then the validator will connect to the set of svs as specified in the trustedSvs list. At least threshold of the specified svs should be live.

If threshold is not set, then the default bft threshold will be used.

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
@@ -1105,11 +1105,13 @@ object BftScanConnection {
       val missingSvs = trustedSvsSet -- foundSvs
 
       logger.trace(s"Discovered the following trusted scans from the network: ${targetScans
-        .map(s => s"Name=${s.svName}, URL=${s.publicUrl}")
-        .mkString("; ")}")
+          .map(s => s"Name=${s.svName}, URL=${s.publicUrl}")
+          .mkString("; ")}")
 
       if (missingSvs.nonEmpty) {
-        logger.debug(s"Configured trusted SVs not found in the DSO rules: ${missingSvs.mkString(", ")}")
+        logger.debug(
+          s"Configured trusted SVs not found in the DSO rules: ${missingSvs.mkString(", ")}"
+        )
       }
 
       targetScans
@@ -1262,10 +1264,9 @@ object BftScanConnection {
         )
 
       case ts @ BftScanClientConfig.BftCustom(_, _, _, _, _) =>
-
         // We bootstrap with the set of provided seed-urls.
         // Since not all trusted SV seeds are provided (most likely), they will not be used in the initial scan connection checking.
-        // TODO: In the future, add a new threshold for how many trusted seed-urls should be there.
+        // In the future, add a new threshold for how many trusted seed-urls should be there.
 
         for {
           tempBftConnection <- bootstrapWithSeedNodes(


apps/app/src/pack/examples/recovery/manual-identities-dump.sc
```diff
@@ -5,11 +5,32 @@ import java.util.Base64
 val id = participant.id.toProtoPrimitive
 
 // This line needs to be adapted if your participant stores keys in an external KMS
-val keys = "[" + participant.keys.secret.list().filter(k => k.name.get.unwrap != "cometbft-governance-keys").map(key => s"{\"keyPair\": \"${Base64.getEncoder.encodeToString(participant.keys.secret.download(key.publicKey.fingerprint).toByteArray)}\", \"name\": \"${key.name.get.unwrap}\"}") .mkString(",") + "]"
+val keys = "[" + participant.keys.secret
```
nicu-da: Why did we need to change this?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrustSpecificScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,222 @@
...
+    with WalletTestUtil
+    with SvTestUtil
+    with HasExecutionContext
+    with HasActorSystem {
+
+  override protected def runEventHistorySanityCheck: Boolean = false
+
+  override def environmentDefinition: SpliceEnvironmentDefinition =
+    EnvironmentDefinition
+      .fromResources(Seq("simple-topology.conf"), this.getClass.getSimpleName)
```
nicu-da: We have `simpleTopology4Svs` that should take care of most things, you probably want to keep your custom transform and the manualStart one

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrustSpecificScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,222 @@
...
+      sv1Backend,
+      sv2ScanBackend,
+      sv2Backend,
+      sv3ScanBackend,
+      sv3Backend,
+      sv4ScanBackend,
+      sv4Backend,
+    )
+
+    sv2ScanBackend.stop()
```
nicu-da: any reason to start it and then stop it? we could just skip starting it

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrustSpecificScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,222 @@
...
+
+    loggerFactory.assertEventuallyLogsSeq(SuppressionRule.LevelAndAbove(Level.INFO))(
+      {
+        aliceValidatorBackend.startSync()
+      },
+      logs => {
+        val messages = logs.map(_.message)
+
+        messages.exists(_.contains(s"Connection to trusted sv ${getSvName(1)} made")) should be(
```
nicu-da: can we extract this in a nice helper method that receives the sv index and then if the log should exist or not? 

then we can easily generalize into something like 

```
connectionToSvWasStarted(1)
connectionToSvWasNotStarted(2)
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrustSpecificScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,222 @@
...
+//        case null => None
+//        case _ => findStatusRuntimeException(t.getCause)
+//      }
+//
+//    val statusRuntimeException =
+//      findStatusRuntimeException(exception).valueOrFail("Could not find StatusRuntimeException in cause chain")
+//
+//    val description = statusRuntimeException.getStatus.getDescription
+//    description should include("Failed to connect to required number of trusted scans.")
+//  }
```
nicu-da: lets just remove this

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -125,6 +125,27 @@ class BftScanConnection(
...
+            .fromConfig(ts.scansRefreshInterval),
+          loggerFactory,
+          retryProvider.timeouts,
+          "refresh_scan_list_trust_specific",
+        )({ tc =>
+          FutureUnlessShutdown.outcomeF(
+            retryProvider.retry(
+              RetryFor.LongRunningAutomation,
+              "refresh_trust_specific_scan_list",
+              "refresh_trust_specific_scan_list",
```
nicu-da: One of these two  is the description, i suggest we include the `ts` there for debugging purposes.

docs/src/release_notes.rst
```diff
@@ -25,6 +25,10 @@ Release Notes
 
       - The UI now provides a formatted message for easily sharing onboarding details with validator operators.
 
+  - Validator
+
+       - Validator now supports a new scan-config, that allows to connect to a pre-specified set of SCAN apps. A threshold can be specified, which indicates how many of the SCAN apps need to agree on the result.
```
nicu-da: I suggest we link to the docs related to this, which should contain info on how to set it up and more explanation about it.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -832,13 +870,301 @@ object BftScanConnection {
...
+    override def scanConnections: ScanConnections =
+      ScanConnections(Seq(scanConnection), 0)
 
     override protected def closeAsync(): Seq[AsyncOrSyncCloseable] = Seq(
       SyncCloseable("scan_connection", scanConnection.close())
     )
   }
 
+  class TrustSpecificScanList(
+      val sv_names: NonEmptyList[String],
```
```suggestion
      val svNames: NonEmptyList[String],
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -832,13 +870,301 @@ object BftScanConnection {
...
+      val seedUrls: NonEmptyList[Uri],
+      val threshold: Option[Int],
+      val retryProvider: RetryProvider,
+      val loggerFactory: NamedLoggerFactory,
```
nicu-da: lets make sure `val` is used only for trait implementations, no need to have it for other members

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -832,13 +870,301 @@ object BftScanConnection {
...
+      private val builder: (Uri, NonNegativeFiniteDuration) => Future[SingleScanConnection],
```
```suggestion
      private val connectionBuilder: (Uri, NonNegativeFiniteDuration) => Future[SingleScanConnection],
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -832,13 +870,301 @@ object BftScanConnection {
...
+
+    def initialize(
+        spliceLedgerClient: SpliceLedgerClient,
+        clock: Clock,
+    )(implicit
+        ec: ExecutionContextExecutor,
+        tc: TraceContext,
+        mat: Materializer,
+    ): Future[Unit] = {
+      val logger = loggerFactory.getTracedLogger(classOf[BftScanConnection])
```
nicu-da: the class should just extend `NamedLogging`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -832,13 +870,301 @@ object BftScanConnection {
...
-    override def scanConnections: ScanConnections = ScanConnections(Seq(scanConnection), 0)
+    override def scanConnections: ScanConnections =
+      ScanConnections(Seq(scanConnection), 0)
 
     override protected def closeAsync(): Seq[AsyncOrSyncCloseable] = Seq(
       SyncCloseable("scan_connection", scanConnection.close())
     )
   }
 
+  class TrustSpecificScanList(
```
nicu-da: I think we've taken this a step to far. Why can't we fully reuse the current `Bft` implementation? All we would need is to filter the resuling `Seq[DsoScan]` to contain only the SVs that are configured

docs/src/release_notes.rst
```diff
@@ -25,6 +25,10 @@ Release Notes
 
       - The UI now provides a formatted message for easily sharing onboarding details with validator operators.
 
+  - Validator
```
martinflorian-da: Let's only add a release notes entry once we've added the new feature to Helm as well + have some documentation for it. Until then I wouldn't count this as truly "released".

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1256,6 +1407,20 @@ object BftScanConnection {
       def setAmuletRulesCacheTimeToLive(ttl: NonNegativeFiniteDuration): TrustSingle =
         copy(amuletRulesCacheTimeToLive = ttl)
     }
+
+    case class TrustSpecific(
```
martinflorian-da: So the full/any BFT option is called just `Bft`.

WDYT about naming this simply `Custom`?

Wording-wise, `TrustSpecific` seems confusingly close to `TrustSingle` IMO.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -785,13 +785,25 @@ object BftScanConnection {
...
-      val f = connections.f
-      BftCallConfig(
-        connections = connections.open,
-        requestsToDo = 2 * f + 1,
-        targetSuccess = f + 1,
-      )
+      connections.totalInSubset match {
+        case Some(n) =>
+          // Use the static total number of trusted SVs (n) for the threshold
+          val threshold = connections.threshold.getOrElse((n / 3) + 1)
```
martinflorian-da: Any reason not to use a formula from `apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/Thresholds.scala`? Seems to be a more general issue for scan; IMO it's nice to have a single files with all thresholds we use, including scan.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1256,6 +1407,20 @@ object BftScanConnection {
       def setAmuletRulesCacheTimeToLive(ttl: NonNegativeFiniteDuration): TrustSingle =
         copy(amuletRulesCacheTimeToLive = ttl)
     }
+
+    case class TrustSpecific(
+        seedUrls: NonEmptyList[Uri], // by default only one seed_url is provided
+        threshold: Option[Int] = None, // default to len(seedUrls)/3+1
```
martinflorian-da, line 1502: Hm shouldn't this default to `formula(len(trustedSvs))` instead?

Not sure if it makes a lot of sense to have any threshold for the seedUrls honestly. In practice there is usually just one atm :clown_face:

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftCustomScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,207 @@
...
+        val messages = logs.map(_.message)
+        messages.exists(
+          _.contains(
+            s"Successfully established initial connection to trusted scan: ${getSvName(1)}"
+          )
+        ) should be(true)
+        messages.exists(
+          _.contains(
+            s"Successfully established initial connection to trusted scan: ${getSvName(2)}"
+          )
```
nicu-da: lets extract this check into a helper method

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -979,10 +992,13 @@ object BftScanConnection {
...
           connectionBuilder(scan.publicUrl)
-            .transformWith { result =>
-              // logging
-              result.failed.foreach { err =>
-                // TODO(#815): abstract this pattern into the RetryProvider
+            .transformWith {
+              case Success(conn) =>
+                logger.info(
+                  s"Successfully connected to scan of ${scan.svName} (${scan.publicUrl})."
+                )
```
nicu-da, line 1022: Is this logged during every refresh? If that's the case i would actually lower this to debug. We should also lower the log above to debug in that scenario. And add a info log that logs all the scans that we're trying to connect to at the same time. And we will have the info logs for failures.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1015,16 +1026,100 @@ object BftScanConnection {
...
+
+    def scanConnections: ScanConnections
 
     override protected def closeAsync(): Seq[AsyncOrSyncCloseable] =
       initialScanConnections.zipWithIndex.map { case (connection, i) =>
         SyncCloseable(s"scan_connection_$i", connection.close())
       }
   }
 
+  class BftAny(
```
```suggestion
  class AllDsoScansBft(
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1015,16 +1026,100 @@ object BftScanConnection {
...
+      }
+    }
+
+    override def scanConnections: ScanConnections = {
+      val connections = currentConnectionsState
+      connections.copy(totalInSubset = None, threshold = None)
+    }
+  }
+
+  class BftCustom(
```
```suggestion
  class ConfigurationProvidedScansBft(
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1015,16 +1026,100 @@ object BftScanConnection {
...
+      override val loggerFactory: NamedLoggerFactory,
+  )(implicit override val ec: ExecutionContext, tc: TraceContext)
+      extends Bft {
+
+    private val trustedSvsSet = trustedSvs.toList.toSet
+
+    override protected def filterScans(allScans: Seq[DsoScan]): Seq[DsoScan] = {
+      val targetScans = allScans.filter(scan => trustedSvsSet.contains(scan.svName))
+
+      logger.info(s"Discovered the following trusted scans from the network: ${targetScans
```
nicu-da: I would avoid this as a info log if it's logged during every refresh, we could lower it to debug. 
We should log at info level only during the init the SVs that we're going to use.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1015,16 +1026,100 @@ object BftScanConnection {
...
+          .map(s => s"Name=${s.svName}, URL=${s.publicUrl}")
+          .mkString("; ")}")
+      targetScans
+    }
+
+    override protected def validateState(state: BftState): Unit = {
+      val successfulCount = state.openConnections.size
+      val requiredCount = threshold.getOrElse((trustedSvs.size / 3) + 1)
```
nicu-da: same point as @martinflorian-da 's comment, lets make sure we have one place where we calculate thresholds (I think we have a Thresholds helper object)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1015,16 +1026,100 @@ object BftScanConnection {
...
+
+        throw Status.UNAVAILABLE
+          .withDescription(
+            s"Failed to connect to required number of trusted scans. " +
+              s"Required: $requiredCount, Connected: $successfulNames. Scan apps failed or missing from network: $failedNames."
+          )
+          .asRuntimeException()
+      } else {
+        logger.info(s"created threshold number of scan connections.")
+      }
```
nicu-da: this seems to be pretty similar so i would generalize it and keep in the specific subclasses just the threshold calculation

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1080,57 +1175,160 @@ object BftScanConnection {
...
+          retryProvider,
+          loggerFactory,
+        )
+
+      case ts @ BftScanClientConfig.BftCustom(_, _, _, _, _) =>
+        for {
+          bootstrapConn <- ts.seedUrls
+            .traverse(uri => builder(uri, ts.amuletRulesCacheTimeToLive).transform(Success(_)))
+            .map(_.collectFirst { case Success(conn) => conn })
+            .flatMap {
```
nicu-da: this still has the issue that you trust a single connection from the seeds, we should do minimum do a bft read from all the seeds.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1080,57 +1175,160 @@ object BftScanConnection {
...
+              discoveredScans
+            }
+          }
+
+          trustedScans = allScans.filter(scan => ts.trustedSvs.toList.contains(scan.svName))
+
+          trustedScanDetails = trustedScans
+            .map(s => s"  - Name: ${s.svName}, URL: ${s.publicUrl}")
+            .mkString("\n")
+          _ = logger.info(s"all available trusted scans on booststrap:\n$trustedScanDetails")
```
nicu-da, line 1291: we seem to duplicate here what we're doing in the BftCustom implementation, I would try to generalize a bit this init code and reduce the duplication

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1080,57 +1175,160 @@ object BftScanConnection {
...
-                failed.toMap,
-                uri => builder(uri, amuletRulesCacheTimeToLive),
-                Bft.getScansInDsoRules,
-                scansRefreshInterval,
-                retryProvider,
-                loggerFactory,
-              )
+              .map(_ => ()),
+            loggerFactory.getTracedLogger(classOf[BftScanConnection]),
+          )
```
nicu-da, line 1352: same as the point above, a lot of this code seems to be shared with the BftAny init so I would try to generalize it

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrustSpecificScanConnectionIntegrationTest.scala
pasindutennage-da: Yes, we can

but then when the test fails (if in future), it does not pin point to the exact failing assertion, but to this new method. and it does not point to call stack showing which assertion failed.

given that there is little number of assertions, i suggest to just keep it verbose (and repetitive)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: added a new requiredNumScanThreshold

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftCustomScanConnectionIntegrationTest.scala
pasindutennage-da: Done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrustSpecificScanConnectionIntegrationTest.scala
pasindutennage-da: Done

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Done

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: with the new bft read in intialization, they are indeed very different in the new version.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1080,57 +1177,210 @@ object BftScanConnection {
...
+
+      case ts @ BftScanClientConfig.BftCustom(_, _, _, _, _) =>
+        /*
+         * The initialization for `BftCustom` is more complex than for `AllDsoScansBft`.
+         *
+         * `AllDsoScansBft` can bootstrap with a single seed URL because its BFT threshold is
+         * dynamic (f+1). On the first `refresh`, if it only knows about one node, the threshold becomes 1,
+         * allowing it to make a non-BFT discovery call to find the rest of the network.
+         *
+         * This does not work for `BftCustom` because its threshold is static and determined by the
```
nicu-da: Sorry, should've made this a bit clearer, but for the initial discovery the previous bft was creating a bft connection from the seeds and had a threshold based on the number of seeds. What's stopping us from doing the same thing here?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/Thresholds.scala
```diff
@@ -59,4 +59,8 @@ object Thresholds {
     val f = floor((svNum - 1) / 3.0).toInt
     PositiveInt.tryCreate(ceil((svNum + f + 1) / 2.0).toInt)
   }
+
+  def requiredNumScanThreshold(svNum: Int) = {
+    (svNum / 3) + 1
```
martinflorian-da: Why is this not `FPlus1Threshold`?

martinflorian-da: Hm this furthermore deviates from the formula used by scan? https://github.com/hyperledger-labs/splice/blob/0fe2c8dbaf769b74835b3b978774d58858d60166/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala#L821

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
martinflorian-da: Extra boy scout points if you also move the existing threshold calculations there :) Just seems nice to me to have a single place to see all relevant thresholds across all layers.

cc @OriolMunoz-da

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -816,7 +829,12 @@ object BftScanConnection {
     }
   }
 
-  case class ScanConnections(open: Seq[SingleScanConnection], failed: Int) {
+  case class ScanConnections(
+      open: Seq[SingleScanConnection],
+      failed: Int,
+      threshold: Option[Int] = Some(0),
+      totalInSubset: Option[Int] = None,
```
martinflorian-da: This name is not self-explanatory; either pick a better name or add a comment.

martinflorian-da: To me, `targetNumber` or `targetTotalNumber` sounds more intuitive.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1080,57 +1177,212 @@ object BftScanConnection {
...
+            if (successfulResponses.isEmpty) {
+              Future.failed(
+                new IllegalStateException(
+                  "Failed to get scan list from any of the connected seed nodes."
+                )
               )
+            } else {
+              val groupedResponses = successfulResponses.groupBy(identity)
+
+              val consensusEntryOpt = groupedResponses.maxByOption { case (_, group) => group.size }
```
martinflorian-da: This is not "majority agreement" under most definitions of the term; this is "the option with the most votes".

Specifically:

- you would not fail if there is no reasonable majority for anything
- you'd get a random response if two versions have the same number of votes

To be clear the goal here is IMO to go for a simple implementation and an easy to understand (and document) logic.

At the end maybe it will be the easiest to just add an extra config parameter to allow customizing the threshold for the seed connections (defaulting to what we do for full DSO connections).

project/ignore-patterns/canton_network_test_log.ignore.txt
```diff
@@ -114,6 +114,7 @@ Failed to connect to scan of Digital-Asset-Eng-3.*AppUpgradeIntegrationTest
 # Ignore scan issues in disaster recovery tests as we don't start scan there
 Failed to connect to scan of .*DisasterRecoveryIntegrationTest
 Failed to connect to scan of .*AppUpgradeIntegrationTest
+Failed to connect to scan of .*ConfigurationProvidedBftScanConnectionIntegrationTest
```
martinflorian-da, line 118: I see you asserting logs very diligently in your test; (why) do you still need this?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ConfigurationProvidedBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,148 @@
...
+  def refreshConnectionEstablished(svName: Int, message: Seq[String]) = {
+    message.exists(
+      _.contains(
+        s"Successfully connected to scan of ${getSvName(svName)}"
+      )
+    ) should be(true)
+  }
+
+  "simple threshold normal case validator onboarding succeeds" in { implicit env =>
+    sv1Backend.startSync()
```
martinflorian-da: Let's start all the things we need in parallel, to save testing time. See for example: https://github.com/hyperledger-labs/splice/blob/ab87bfe18ebde80062f80a59b48e895a2808b0f7/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SpliceTests.scala#L502

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ConfigurationProvidedBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,148 @@
...
+      logs => {
+        val messages = logs.map(_.message)
+        connectionEstablished(1, messages)
+        connectionEstablished(2, messages)
+        connectionEstablished(3, messages)
+        connectionNotEstablished(4, messages)
+      },
+    )
+
+    eventuallySucceeds() {
```
martinflorian-da: consider adding `clue()`s to your test for both documentation and nicer logging; I'm guessing the onboard user below is for testing that the validator is operational, but having it clearly spelled out would help readability

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ConfigurationProvidedBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,148 @@
...
+  override protected def runEventHistorySanityCheck: Boolean = false
+
+  override def environmentDefinition: SpliceEnvironmentDefinition =
+    EnvironmentDefinition
+      .simpleTopology4Svs(this.getClass.getSimpleName)
+      .addConfigTransforms((_, config) =>
+        ConfigTransforms.updateAllValidatorConfigs {
+          case (name, c) if name == "aliceValidator" =>
+            val bftCustomConfig = BftScanClientConfig.BftCustom(
+              seedUrls = NonEmptyList.one(Uri("http://127.0.0.1:5012")),
```
martinflorian-da: The current iteration of this PR seems to be circling a lot on the question of what to do with the seed URLs, however the test here is testing only the simplest scenario possible...

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ConfigurationProvidedBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,148 @@
...
+              trustedSvs =
+                NonEmptyList.of(s"${getSvName(1)}", s"${getSvName(2)}", s"${getSvName(3)}"),
+              threshold = Some(2),
```
martinflorian-da, line 40: Is it feasible to add a test for "it stops working if two of the trusted SVs shut down"?

martinflorian-da: I also wonder if there is some easy way to ensure we never try to talk to sv-4's scan at all. Perhaps we can use toxiproxy to just disallow connections to it? Will we get log warnings then if we do try?

martinflorian-da (approved):
Nice work, I think we're getting there!

My main concern is currently the logic around seed connections...

Close second: the size and complexity that this PR already has reached.

How about:

- hard code the seed connection threshold to 1 in this PR; adding a TODO comment that this is just a temp limitation
- follow up with a second PR that focuses only that

We want smaller PRs whenever we can have them, and this sounds like a reasonable cut to me?

Also recall that this is not really "released" until we document it, so it's OK that it lands on main in unfinished form.

project/ignore-patterns/canton_network_test_log.ignore.txt
pasindutennage-da: Yes, there is a race condition between 2 action-assertions -- between 2 calls a refresh takes places, not captured in test log, hence causing an error. hence we need to manually ignore it here

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ConfigurationProvidedBftScanConnectionIntegrationTest.scala
pasindutennage-da: > Is it feasible to add a test for "it stops working if two of the trusted SVs shut down"?

I had this before, but when validator fails to start (due to not enough SCANs), it is not possible to catch that in the test, instead the entire test fails. I tried wrapping the error, and assert, but did not work

@nicu-da can you please verify this?

pasindutennage-da: > I also wonder if there is some easy way to ensure we never try to talk to sv-4's scan at all.

I have assertions checking if a connection is made. For SV4, i check the absense of the assertion. Is it fine?

nicu-da: >     Is it feasible to add a test for "it stops working if two of the trusted SVs shut down"?

You could add something that validates that we don't reach consensus after the validator has already started. But you can't add a test that the validators fails to start as our current test infra assumes it's fine to just fail the test for that.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: with unified logic, this logic no longer exists

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ConfigurationProvidedBftScanConnectionIntegrationTest.scala
martinflorian-da: > something that validates that we don't reach consensus after the validator has already started

Yeah I had something like that in mind. Note that you can, for example, `.stop()` nodes that you have started.

Let's add a test like this (either here or in a follow-up); seems valuable.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -785,13 +785,26 @@ object BftScanConnection {
...
     def default(connections: ScanConnections): BftCallConfig = {
-      val f = connections.f
-      BftCallConfig(
-        connections = connections.open,
-        requestsToDo = 2 * f + 1,
-        targetSuccess = f + 1,
-      )
+      connections.targetTotalNumber match {
+        case Some(n) =>
+          // Use the static total number of trusted SVs (n) for the threshold
```
martinflorian-da:
```suggestion
```

Comment is wrong/incomplete and the code line below is anyway clear enough IMO.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ConfigurationProvidedBftScanConnectionIntegrationTest.scala
pasindutennage-da: thank you. I will add it now

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ConfigurationProvidedBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,239 @@
...
+          connectionEstablished(1, messages)
+        }
+        withClue("Validator should connect to sv2:") {
+          connectionEstablished(2, messages)
+        }
+        withClue("Validator should connect to sv3:") {
+          connectionEstablished(3, messages)
+        }
+        withClue("Validator should NOT connect to sv4:") {
+          connectionNotEstablished(4, messages)
```
martinflorian-da, line 110: Nice!

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ConfigurationProvidedBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,239 @@
...
+            messages.exists(
+              _.contains(
+                "Consensus not reached"
+              )
+            ) should be(true)
+          }
+        },
+      )
+
+      aliceValidatorBackend.stop()
```
martinflorian-da, line 236: Any specific reason for the explicit `stop()`? The test harness will do that for you usually.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ConfigurationProvidedBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,239 @@
...
+          sv2ScanBackend.stop()
+        },
+        logs => {
+          val messages = logs.map(_.message)
+          withClue(
+            "Validator should fail to reach consensus when trusted scans drop below the threshold:"
+          ) {
+            messages.exists(
+              _.contains(
+                "Consensus not reached"
```
martinflorian-da, line 229: Nice!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1015,16 +1050,78 @@ object BftScanConnection {
...
+      override val loggerFactory: NamedLoggerFactory,
+  )(implicit override val ec: ExecutionContext, tc: TraceContext)
+      extends Bft {
+
+    private val trustedSvsSet = trustedSvs.toList.toSet
+
+    override protected def filterScans(allScans: Seq[DsoScan]): Seq[DsoScan] = {
+      val targetScans = allScans.filter(scan => trustedSvsSet.contains(scan.svName))
+
+      logger.debug(s"Discovered the following trusted scans from the network: ${targetScans
```
martinflorian-da: I slightly wonder about log spam here as we do call `refresh` quite often. Maybe we can change this to `trace` and instead add a `debug` (or even `info`?) log about SVs that we configured but are missing from the DSO rules?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1080,57 +1237,128 @@ object BftScanConnection {
...
+          ttl,
+          scanList,
+          clock,
+          retryProvider,
+          loggerFactory,
+        )
+
+      case ts @ BftScanClientConfig.BftCustom(_, _, _, _, _) =>
+        for {
+          tempBftConnection <- bootstrapWithSeedNodes(
```
martinflorian-da, line 1272: Maybe add a TODO comment describing the used threshold logic for the bootstrap phrase and the plans for improving that.

Unless I misunderstood / forgot: your intention was to make that threshold configurable as well, wasn't it?

martinflorian-da (approved):
Really nice work, thanks a lot for pushing this through!

PR #2831, opened 2025-10-23
Fix ledger audience in auth0 M2M apps on MainNet

martinflorian-da:
Quick fix to unbreak our SV...

Of course we have a different config on MainNet than on other networks, see https://github.com/hyperledger-labs/splice/blob/264cb6428f47f7bd183d12bd17098d5f2a04e6f9/cluster/pulumi/infra/src/auth0.ts#L384

[static]

### DIFF
--- a/cluster/pulumi/infra/src/auth0.ts
+++ b/cluster/pulumi/infra/src/auth0.ts
@@ -137,17 +137,18 @@ function newM2MApp(
       }
     );
 
-    if (ledgerApiAudValue !== 'https://canton.network.global') {
+    // TODO(DACH-NY/canton-network-internal#2206): Of course on MainNet we use a different default audience...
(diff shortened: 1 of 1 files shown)


cluster/pulumi/infra/src/auth0.ts
```diff
@@ -151,6 +151,19 @@ function newM2MApp(
           provider: auth0DomainProvider,
         }
       );
+    } else if (isMainNet) {
```
moritzkiefer-da: Any reason not to just combine the two ifs? They do the exact same thing except for the audience field. So if you only put the if around that this becomes simpler.

martinflorian-da: potato potato; happy to change so this has fewer lines of code

PR #2832, opened 2025-10-23
add the info endpoint to the sv-runbook Pulumi project

mblaze-da:
fixes https://github.com/DACH-NY/canton-network-internal/issues/1857

Please see the two TODO comments. I want to resolve them before merging this PR.

### DIFF
--- a/apps/app/src/pack/examples/sv-helm/info-values.yaml
+++ b/apps/app/src/pack/examples/sv-helm/info-values.yaml
@@ -5,8 +5,6 @@ runtimeDetails:
 
 deploymentDetails:  # optional, exposes static configuration under path /
   network: TARGET_CLUSTER  # One of "main", "test" or "dev"
-  sv:
-    version: "SV_VERSION"
   configDigest:
     allowedIpRanges:
       type: "md5"


apps/app/src/pack/examples/sv-helm/info-values.yaml
```diff
@@ -5,16 +5,18 @@ runtimeDetails:
 
 deploymentDetails:  # optional, exposes static configuration under path /
   network: TARGET_CLUSTER  # One of "main", "test" or "dev"
+  sv:
+    version: "SV_VERSION"
```
martinflorian-da: So if you add this then you break this for users afaict.

The comments [here](https://github.com/hyperledger-labs/splice/blob/8ec48fe6a4e22cf99a251a77254b3e3001d3c8fe/cluster/helm/splice-info/values-template.yaml#L20) tell me it is optional; is it not?

apps/app/src/pack/examples/sv-helm/info-values.yaml
```diff
@@ -5,16 +5,18 @@ runtimeDetails:
...
   configDigest:
     allowedIpRanges:
       type: "md5"
-      value: MD5_HASH_OF_ALLOWED_IP_RANGES
+      value: "MD5_HASH_OF_ALLOWED_IP_RANGES"
```
martinflorian-da: Any particular reason for adding the quotes? It's a placeholder, so meant to be replaced anyway... (So if quotes are need one could add them while replacing?)

cluster/pulumi/canton-network/src/info.ts
```diff
@@ -73,7 +71,7 @@ export function installInfo(
           ? {
               chainIdSuffix: getChainIdSuffix(),
               migrationId: decentralizedSynchronizerMigrationConfig.upgrade.id,
-              version: cnChartVerstionToString(
+              version: CnChartVersion.stringify(
```
martinflorian-da, line 76:
```suggestion
              version: CnChartVersion.toString(
```

some bikeshedding; we definitely use the `toString` wording in other places, whereas the `stringify` one sounded alien to me (like from some external library)

cluster/pulumi/common/src/artifacts.ts
```diff
@@ -1,18 +1,28 @@
...
     };
 
-export function parsedVersion(version: string | undefined): CnChartVersion {
-  return version && version.length > 0 && version !== 'local'
-    ? {
-        type: 'remote',
-        version: version,
-      }
-    : { type: 'local' };
+// TODO: I'd propose to disable this rule globally.
```
martinflorian-da: Hmm I'm not convinced: https://typescript-eslint.io/rules/no-namespace/

> Namespaces are an outdated way to organize TypeScript code. ES2015 module syntax is now preferred

So if you strongly prefer this style (I don't have a strong opinion either way), can't you go for a module instead of a namespace?

cluster/pulumi/sv-runbook/src/installNode.ts
```diff
@@ -532,3 +541,38 @@ async function installSvAndValidator(
...
+  scan: pulumi.Resource
+): void {
+  function readFileOrEmptyString(path: string | undefined): string {
+    return path !== undefined ? fs.readFileSync(path, 'utf-8') : '';
+  }
+
+  const infoValues = {
+    ...loadYamlFromFile(`${SPLICE_ROOT}/apps/app/src/pack/examples/sv-helm/info-values.yaml`, {
+      TARGET_CLUSTER: clusterNetwork,
+      SV_VERSION: CnChartVersion.stringify(decentralizedSynchronizerMigrationConfig.active.version),
```
martinflorian-da: In case its hard to make this get picked up otherwise: you can still add your own extra keys here, beyond overrides... I'm happier with that route than with adding new things to the example file.

cluster/pulumi/canton-network/src/info.ts
mblaze-da: I'd agree if we were talking about a method. `version.toString()` would definitely fit the JS/TS ecosystem. Since we're using a simple structural type here we cannot add methods. In JS there is for example `JSON.stringify` so that's where I took inspiration. Since you're more familiar with the codebase I'm willing to trust you that `toString` might be a better fit but I wanted to present my reasoning.

cluster/pulumi/sv-runbook/src/installNode.ts
```diff
@@ -532,3 +541,38 @@ async function installSvAndValidator(
...
+      MIGRATION_ID: decentralizedSynchronizerMigrationConfig.active.id.toString(),
+      MD5_HASH_OF_ALLOWED_IP_RANGES: createHash('md5')
+        .update(readFileOrEmptyString(externalIpRangesFile()))
+        .digest('hex'),
+      MD5_HASH_OF_APPROVED_SV_IDENTITIES: createHash('md5')
+        .update(readFileOrEmptyString(approvedSvIdentitiesFile()))
+        .digest('hex'),
+      CHAIN_ID_SUFFIX: getChainIdSuffix(),
+    }),
+    // TODO: I'm not sure which properties should be templated in the example info-values.yaml
```
martinflorian-da: IMO for this issue the goal state is to not change anything that is customer-facing, unless we actually found a bug in the example and/or documentation. So I think your solution here is perfect.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/runbook/RunbookSvPreflightIntegrationTestBase.scala
```diff
@@ -331,6 +335,29 @@ abstract class RunbookSvPreflightIntegrationTestBase
       }
     }
   }
+
+  "Info service is reachable and returns a JSON object" in { _ =>
```
martinflorian-da, line 339: For the record: This could test more but I actually agree that adding more things here is probably not a good investment.

apps/app/src/pack/examples/sv-helm/info-values.yaml
mblaze-da: That kind of answers my question [here](https://github.com/hyperledger-labs/splice/blob/aea89bdc1a1be41aabce6b2278301a92c7a372f4/cluster/pulumi/sv-runbook/src/installNode.ts#L567). Yeah, I guess it is optional.

apps/app/src/pack/examples/sv-helm/info-values.yaml
mblaze-da: I started adding them when I templated `CHAIN_ID_SUFFIX` with a number and it complained that it wants a string. Perhaps that was a mistake. I was not sure if it is better to include quoting in values or in the template.

cluster/pulumi/canton-network/src/info.ts
martinflorian-da: Your explanation makes more sense, I think I confused myself there about what this really is.

apps/app/src/pack/examples/sv-helm/info-values.yaml
martinflorian-da: Yeah all the schema definitions I can find tell me it should be a `string`... So you can probably substitute with something like `"\"1\""` in your Pulumi code? Still preferable than changing the example IMO.

cluster/pulumi/common/src/artifacts.ts
mblaze-da: Then I think we'd have to define the type and its associated functions in separate files. If there is another way I don't yet know how to achieve it. Maybe some more sophisticated export statements. My rationale here is that having functions associated with a type defined in a namespace that has the same name as the type enchances readability and is similar to what we'd do in Scala.

martinflorian-da: Hm yeah maybe you're right...

> enchances readability

It seems like with our current ideas that comes at the cost of either disabling a lint (moving us away from more "modern" Typescript) or complexity (to workaround the linter).

I think my preference then is to just stick to what we've done so far, and establish the connection between function and type via function naming (in case it seems to generic by default / the argument list is not enough).

mblaze-da: I managed to keep the syntax and remove the `namespace` usage. If you prefer the previous way to define and call those functions please tell me and I'll refactor that.

apps/app/src/pack/examples/sv-helm/info-values.yaml
martinflorian-da: So let's avoid the changes to this file if we easily can

cluster/pulumi/common/src/artifacts.ts
martinflorian-da: No I don't mind your approach; thanks!

PR #2834, opened 2025-10-23
Bump nix binary to 2.32.1 on GHA

OriolMunoz-da:


.github/actions/nix/setup_nix/action.yml
OriolMunoz-da, line 48: cache key will be different

PR #2839, opened 2025-10-23
Stop spamming splitwell groups

rautenrieth-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/5610


isegall-da (approved):
Nice catch, thanks!

PR #2849, opened 2025-10-23
Replace gsutil => gcloud storage

OriolMunoz-da:
Will test tomorrow since this affects cloudsql

see also https://github.com/DACH-NY/canton-network-internal/pull/2370

PR #2851, opened 2025-10-23
Add mediator scan to BigQuery

isegall-da:
Tests:
- [x] deploy from main, then from here -> mediator scan gets added
- [x] deploy directly from here, for sanity

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/canton-network/bigquery-cloudsql.sh
+++ b/cluster/pulumi/canton-network/bigquery-cloudsql.sh
@@ -18,7 +18,6 @@ REQUIRED_ARGS=(
   "compute-region"
   "service-account-email"
   "tables-to-replicate-joined"
-  "db-name"
   "schema-name"
   "postgres-user-name"
   "publication-name"
@@ -26,13 +25,11 @@ REQUIRED_ARGS=(
   "replicator-user-name"
   "postgres-instance-name"
   "scan-app-database-name"
-  "tables-to-wait-for-record-time-list"
-  "tables-to-wait-for-record-time-length"
+  "flyway-migration-to-wait-for"
 )
 PRIVATE_NETWORK_PROJECT=""
 COMPUTE_REGION=""
 SERVICE_ACCOUNT_EMAIL=""
-DB_NAME=""
 SCHEMA_NAME=""
 POSTGRES_USER_NAME=""
 PUBLICATION_NAME=""
@@ -41,8 +38,7 @@ REPLICATOR_USER_NAME=""
 POSTGRES_INSTANCE_NAME=""
 SCAN_APP_DATABASE_NAME=""
 TABLES_TO_REPLICATE_JOINED=""
-TABLES_TO_WAIT_FOR_RECORD_TIME_LIST=""
-TABLES_TO_WAIT_FOR_RECORD_TIME_LENGTH=""
+FLYWAY_MIGRATION_TO_WAIT_FOR=""
 
 # Track which arguments have been provided
 declare -A PROVIDED_ARGS
@@ -77,9 +73,6 @@ while [ "$#" -gt 0 ]; do
     service-account-email)
       SERVICE_ACCOUNT_EMAIL="$param_value"
       ;;
-    db-name)
-      DB_NAME="$param_value"
-      ;;
     schema-name)
       SCHEMA_NAME="$param_value"
       ;;
@@ -104,11 +97,8 @@ while [ "$#" -gt 0 ]; do
     scan-app-database-name)
       SCAN_APP_DATABASE_NAME="$param_value"
       ;;
-    tables-to-wait-for-record-time-list)
-      TABLES_TO_WAIT_FOR_RECORD_TIME_LIST="$param_value"
-      ;;
-    tables-to-wait-for-record-time-length)
-      TABLES_TO_WAIT_FOR_RECORD_TIME_LENGTH="$param_value"
+    flyway-migration-to-wait-for)
+      FLYWAY_MIGRATION_TO_WAIT_FOR="$param_value"
       ;;
     *)
       echo "Unknown parameter: --$param_name" >&2
@@ -170,6 +160,7 @@ gcloud storage buckets add-iam-policy-binding "gs://$TMP_BUCKET" \
 case "$SUBCOMMAND" in
   create-pub-rep-slot)
     cat > "$TMP_SQL_FILE" <<EOT
+      SET search_path TO $SCHEMA_NAME;
       DO \$\$
       DECLARE
         migration_complete BOOLEAN := FALSE;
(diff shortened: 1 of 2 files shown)


cluster/pulumi/canton-network/bigquery-cloudsql.sh
```diff
@@ -150,21 +150,22 @@ GCS_URI="gs://$TMP_BUCKET/$(basename "$TMP_SQL_FILE")"
 # things to clean up
 cleanup() {
   echo 'Cleaning up temporary GCS object and bucket'
-  gsutil rm "$GCS_URI" || true
-  gsutil rb "gs://$TMP_BUCKET" || true
+  gcloud storage  rm "$GCS_URI" || true
```
isegall-da, line 143: These are @OriolMunoz-da 's changes from https://github.com/hyperledger-labs/splice/pull/2849, just to test them as part of my testing of this PR.

cluster/pulumi/canton-network/bigquery-cloudsql.sh
```diff
@@ -202,7 +203,10 @@ case "$SUBCOMMAND" in
       DO \$\$
       BEGIN
         -- TODO (#453) drop slot, pub if table list doesn't match
-        IF NOT EXISTS (SELECT 1 FROM pg_publication WHERE pubname = '$PUBLICATION_NAME') THEN
+        IF EXISTS (SELECT 1 FROM pg_publication WHERE pubname = '$PUBLICATION_NAME') THEN
+          ALTER PUBLICATION $PUBLICATION_NAME
```
isegall-da, line 201: Unfortunately, the way we write the Command in pulumi, it actually never updates the resource, but deletes it first on modification, but I still figured I'd include this change here, in case we change pulumi to update things in-place.

cluster/pulumi/canton-network/src/bigQuery.ts
```diff
@@ -161,7 +167,7 @@ function installDatastream(
           },
           // editing dataFreshness does not alter existing BQ tables, see its
           // docstring or https://github.com/hyperledger-labs/splice/issues/2011
-          dataFreshness: '14400s',
+          dataFreshness: clusterProdLike ? '14400s' : '0s',
```
isegall-da, line 170: Just for some convenience on a scratchnet to get the data quickly for testing

cluster/pulumi/canton-network/src/bigQuery.ts
```diff
@@ -473,6 +480,7 @@ function createPublicationAndReplicationSlots(
     {
       deletedWith: postgres.databaseInstance,
       dependsOn: [scan, postgres.databaseInstance, replicatorUser],
+      deleteBeforeReplace: true,
```
isegall-da, line 481: As mentioned above, since we include all inputs as part of the CLI call, pulumi considers any change a change in the resource itself, and always deletes&creates the resource instead of updating it in place, so we want it to delete it before creating the new one.

cluster/pulumi/canton-network/src/bigQuery.ts
```diff
@@ -40,7 +40,13 @@ const replicatorUserName = 'bqdatastream';
 const replicationSlotName = 'update_history_datastream_r_slot';
 const publicationName = 'update_history_datastream_pub';
 // what tables from Scan to replicate to BigQuery
-const tablesToReplicate = ['update_history_creates', 'update_history_exercises'];
+const tablesToReplicate = [
+  'update_history_creates',
+  'update_history_exercises',
+  'scan_verdict_store',
+  'scan_verdict_transaction_view_store',
```
OriolMunoz-da, line 47: I'm OOTL as to why we want these, so just in case, these are the columns:
```
verdict_row_id
view_id
informees
confirming_parties
sub_views
``` 
what do we want these for?

OriolMunoz-da (approved):
thanks! I think a review from Stephen might still be of value

cluster/pulumi/canton-network/src/bigQuery.ts
isegall-da: These are the two tables that we expose via the mediator scan. This PR basically clones them into BQ.

The columns thing was specific to the update history, to make sure we don't start consuming before a specific flyway migration that made some breaking changes that we wanted to make sure don't break BQ.

cluster/pulumi/canton-network/bigquery-cloudsql.sh
stephencompall-DA: `Command` has an `update` argument alongside `create` and `delete`. Of course, [its docs](https://github.com/pulumi/pulumi-command/blob/83a7bf7305a661127c7640b8d930445652c785c5/sdk/nodejs/local/command.ts#L271-L281) also claim that `create` is used as the default for `update`, but might be worth a try specifying it explicitly.

isegall-da: Tried that, didn't help.

cluster/pulumi/canton-network/src/bigQuery.ts
stephencompall-DA: `update` might also solve this.

cluster/pulumi/canton-network/bigquery-cloudsql.sh
isegall-da: (for the same reason. Since we pass the args as part of the command, pulumi thinks that the command itself changed, so does delete&create, and not update)

cluster/pulumi/canton-network/src/bigQuery.ts
```diff
@@ -40,7 +40,13 @@ const replicatorUserName = 'bqdatastream';
...
+];
+const tablesToWaitForRecordTime = ['update_history_creates', 'update_history_exercises'];
```
stephencompall-DA:
```suggestion
const tablesToWaitForRecordTime = ['update_history_creates', 'update_history_exercises'];
const tablesToReplicate = [
  ...tablesToWaitForRecordTime,
  'scan_verdict_store',
  'scan_verdict_transaction_view_store',
];
```

cluster/pulumi/canton-network/bigquery-cloudsql.sh
```diff
@@ -179,11 +179,11 @@ case "$SUBCOMMAND" in
...
           -- Check if all tables exist AND have the record_time column
           -- this is added by V037__denormalize_update_history.sql
-          SELECT COUNT(*) = $TABLES_TO_REPLICATE_LENGTH INTO migration_complete
+          SELECT COUNT(*) = $TABLES_TO_WAIT_FOR_RECORD_TIME_LENGTH INTO migration_complete
             FROM information_schema.columns
             WHERE table_catalog = '$DB_NAME'
               AND table_schema = '$SCHEMA_NAME'
-              AND table_name IN ($TABLES_TO_REPLICATE_LIST)
+              AND table_name IN ($TABLES_TO_WAIT_FOR_RECORD_TIME_LIST)
               AND column_name = 'record_time';
```
stephencompall-DA: The record time migration happens before the mediator scan tables migration. Therefore, when the mediator scan tables exist, by definition the record time column is there. So a simpler solution (and one that doesn't race) might be to go back to `TABLES_TO_REPLICATE_LENGTH` and `TABLES_TO_REPLICATE_LIST`, use `information_schema.tables`, and drop the `column_name` check.

Then we could do away with `tablesToWaitForRecordTime` as well.

isegall-da: > The record time migration happens before the mediator scan tables migration

Why's that?

stephencompall-DA: `scan_verdict_store` is in V043, record_time is in V037.

isegall-da: Ahh of course :facepalm:

isegall-da: But then I'm kinda tempted to replace these two args by a simple "migration_to_wait_for", and just wait for that migration. Seems easier to understand later what exactly we're waiting for and why, and how to bump it in the next change.

PR #2865, opened 2025-10-24
[ci] Specify encoding for the dump write

nicu-da:
fixes https://github.com/hyperledger-labs/splice/issues/2864

reproduced locally with stacktrace

```


An exception or error caused a run to abort: Requested array size exceeds VM limit 
java.lang.OutOfMemoryError: Requested array size exceeds VM limit
	at java.base/java.lang.String.encodeWithEncoder(String.java:902)
	at java.base/java.lang.String.encode(String.java:875)
	at java.base/java.lang.String.getBytes(String.java:1818)
	at better.files.File.write(File.scala:373)
```

the stacktrace goes though `encodeWithEncoder` which was surprising because the utf-8 encoding should go through `encodeUTF8`. the file write api uses the `Charset.defaultCharset()` which in theory should be utf-8 for java 17+ but it does not seem to work. 

so the root issue is here https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/String.java#L871 the `Charset.defaultCharset()` result is UTF_8 but it fails the check `==  UTF_8.INSTANCE` (equals does return true)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/BackupDump.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/BackupDump.scala
@@ -4,12 +4,13 @@
 package org.lfdecentralizedtrust.splice.util
 
 import better.files.File
-import org.lfdecentralizedtrust.splice.config.BackupDumpConfig
 import com.digitalasset.canton.logging.NamedLoggerFactory
 import com.digitalasset.canton.tracing.TraceContext
 import io.circe.Decoder
+import org.lfdecentralizedtrust.splice.config.BackupDumpConfig
 
 import java.io.FileNotFoundException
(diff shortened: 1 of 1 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/BackupDump.scala
```diff
@@ -40,7 +41,7 @@ object BackupDump {
     import better.files.File
     val file = File(path)
     file.parent.createDirectories()
-    file.write(content)
+    file.write(content)(File.OpenOptions.default, StandardCharsets.UTF_8)
```
nicu-da, line 44: will also add a big warn here not to modify it when i port this to main to avoid another build here

PR #2900, opened 2025-10-28
pulumi-auth0 overhaul

isegall-da:
This is actually still a refactoring, in the sense that it should not actually change any actual deployments, only how we organize and use the data, so that in followup PRs I can cleanup our deployments in ways that are easy to understand and review.
Apologies for the bombdrop of a PR, couldn't think of any sensible way to break it into reviewable steps.

Note that this PR is backward-compatible in the sense that the deployment stacks can still read the output of "old" infra stacks, to make it less noisy on prod networks, previews, and e.g. HDM tests which go back and forth on infra versions.

[Cluster test](https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/38601/workflows/f0ab963b-1b56-40fb-85ad-6f034d159c27)
Successful HDM test: https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/38580/workflows/dda868e5-60b3-4425-83d7-292f5b243244

(requires the changes in internal, from here: https://github.com/DACH-NY/canton-network-internal/pull/2445)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


cluster/pulumi/common/src/auth0/auth0.ts
```diff
@@ -59,6 +61,20 @@ export class Auth0Fetch implements Auth0Client {
     return this.cfg;
   }
 
+  // Any namespace that we deploy using validator-runbook stack currently reuses the
```
isegall-da, line 64: @nicu-da this should fix the party-allocator issue on CILR

cluster/pulumi/common/src/auth0/auth0types.ts
isegall-da, line 1: I'd say the changes in this file are the gist of the change in how we manage the data, and is probably the main part to understand when reviewing this.

build-tools/get-auth0-token.py
```diff
@@ -57,125 +60,156 @@ def get_pulumi_infra_outputs(self) -> dict:
       self.pulumi_infra_outputs = json.loads(infra_outputs_str)
     return self.pulumi_infra_outputs
 
+  def init_pulumi_stack(self, args: argparse.Namespace) -> str:
+    if args.namespace == "sv":
+      self.pulumi_stack = "svRunbook"
+    elif args.namespace == "validator":
```
nicu-da: how does this handle other validators deployed using the validator runbook?

cluster/expected/sv-runbook/expected.json
isegall-da, line 1: The mock config actually has `cloudsql: enabled`, so while I'm not 100% sure why that was ignored before, this does seem correct to me. We have similar values in canton-network/expected.json already, so not too worried about this.

nicu-da: the default sv runbook cloudsql is enabled false so this looks like we broke something

isegall-da: But doesn't the default:enabled:true in config.yaml override that?

nicu-da: that overrides the default for all the projects, but the config.yaml in the deployment directory also defines a sv runbook default which is more specific so it will still override the mock config.yaml default

isegall-da: Found it, fixed. It's the usual imports nightmare...

PR #2927, opened 2025-10-28
fix fixed-tokens cache

isegall-da:
Apparently, if the secret is completely missing, we also don't bother saving the cache to a new one, and it is missing on 3 out of the 5 scratchnets :shrug: 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/common/src/auth0/auth0.ts
+++ b/cluster/pulumi/common/src/auth0/auth0.ts
@@ -8,7 +8,7 @@ import { Output } from '@pulumi/pulumi';
 import { AuthenticationClient, ManagementClient, TokenSet } from 'auth0';
 
 import { config, isMainNet } from '../config';
-import { CLUSTER_BASENAME } from '../utils';
+import { CLUSTER_BASENAME, fixedTokens } from '../utils';
 import type {
   Auth0Client,
   Auth0SecretMap,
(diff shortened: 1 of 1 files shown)


cluster/pulumi/common/src/auth0/auth0.ts
```diff
@@ -111,22 +111,25 @@ export class Auth0Fetch implements Auth0Client {
       this.auth0Cache = cacheMap;
       await pulumi.log.debug('Auth0 cache loaded...');
     } catch (e) {
+      await this.k8sApi.createNamespacedSecret('default', {
```
martinflorian-da: 2 things seems sketchy here:

- we have two code paths that create that secret; at least extract it to a function but I guess we could also just do it only once?
- creating a secret when what we're really doing is a "read"...

Would it not be sufficient to fix the save method so it always does an "ensure Secret exists"? AFAICT by removing the `return` there we already get that?

cluster/pulumi/common/src/auth0/auth0.ts
```diff
@@ -111,22 +111,25 @@ export class Auth0Fetch implements Auth0Client {
...
-      await pulumi.log.debug('No auth0 cache loaded in Auth0Fetch');
-      return;
-    }
-
-    if (!this.hasDiffsToSave) {
-      await pulumi.log.debug('No auth0 cache diffs to save');
+    if (this.auth0Cache && !this.hasDiffsToSave) {
+      await pulumi.log.debug('Auth0 cache was loaded, and no cache diffs to save');
       return;
     }
```
martinflorian-da, line 138: (on line below)

Does that work if `this.auth0Cache` is `undefined`?

cluster/pulumi/common/src/auth0/auth0.ts
isegall-da: That's what I thought too, but that failed because it somehow tried to create it twice, so I went for the easiest of just sticking to the assumption we have on the write that the secret already exists.

PR #2935, opened 2025-10-29
[ci] Increase number of retries for app init dso rules step to account for ingestion

nicu-da:
fixes https://github.com/hyperledger-labs/splice/issues/2939

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/RetryFor.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/RetryFor.scala
@@ -39,6 +39,16 @@ object RetryFor {
     resetRetriesAfter = None,
   )
 
+  /** A retry intended for app initialization steps that might take a long time to execute.
+    * Eg: after a migration waiting for store ingestion
+    */
+  val WaitingOnInitDependencyLong: RetryFor = RetryFor(
+    maxRetries = 500,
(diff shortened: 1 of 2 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/RetryFor.scala
```diff
@@ -39,6 +39,16 @@ object RetryFor {
     resetRetriesAfter = None,
   )
 
+  /** A retry intended for app initialization steps that might take a long time to execute.
+    * Eg: after a migration waiting for store ingestion
+    */
+  val WaitingOnInitDependencyLong: RetryFor = RetryFor(
+    maxRetries = 500,
```
moritzkiefer-da, line 46: Wondering if we gain much by not making it infinite. Your SV app doesn't go ready if you don't hit it so it seems like you should get alerts for this anyway so then it seems like the downside is pretty limited.

nicu-da: This should in theory give it somewhere around 30m before a restart, I tried to avoid making it infinite just in case there some weird issue in the sv app itself, and a restart never hurt anyone :see_no_evil:

PR #2946, opened 2025-10-29
Add more ingestion metrics

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/2649

image
image


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceMetrics.scala
```diff
@@ -3,10 +3,48 @@
...
+
+import scala.concurrent.ExecutionContext
 
 object SpliceMetrics {
 
   val MetricsPrefix: MetricName = MetricName("splice")
 
+  private type CantonTimestampMicros = Long
+
+  def cantonTimestampGauge(
```
OriolMunoz-da, line 16: annoyance of mine

nicu-da: Can we just upstream it to canton?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/metrics/ScanMediatorVerdictIngestionMetrics.scala
```diff
@@ -6,21 +6,21 @@ package org.lfdecentralizedtrust.splice.scan.metrics
...
-  val lastIngestedRecordTime: Gauge[CantonTimestampMicros] = metricsFactory.gauge(
+  val lastIngestedRecordTime: Gauge[CantonTimestamp] = SpliceMetrics.cantonTimestampGauge(
+    metricsFactory,
     MetricInfo(
       name = prefix :+ "last_record_time_us",
       summary = "Latest ingested mediator verdict record_time",
       qualification = Traffic,
     ),
-    initial = 0L,
+    initial = CantonTimestamp.MinValue,
```
rautenrieth-da, line 23: Changing the initial value might mess up some existing alerts/dashboards, but all other timestamp gauges have `MinValue` as the initial value and this metric is new, so changing it makes sense.

PR #2955, opened 2025-10-29
[ci] Support http proxy

ray-roestenburg-da:
Fixes #2931 

Added unit tests using `tinyproxy` that prove that the NodeBase httpClient can make requests through an HTTP forward proxy.

- [x] add an integration test with `http.proxy...` set to show validator can tap, and that many services are proxied to. 
 
Does not support Proxy authentication (which is also not currently supported in canton).

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorProxyIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorProxyIntegrationTest.scala
@@ -1,7 +1,7 @@
 package org.lfdecentralizedtrust.splice.integration.tests
 
 import better.files.File
-
+import org.apache.pekko.http.scaladsl.model.Uri
 import org.lfdecentralizedtrust.splice.config.ConfigTransforms
 import org.lfdecentralizedtrust.splice.integration.EnvironmentDefinition
 import org.lfdecentralizedtrust.splice.integration.tests.SpliceTests.IntegrationTest
@@ -11,6 +11,7 @@ import org.lfdecentralizedtrust.splice.unit.http.{
   SystemPropertiesSupport,
   TinyProxySupport,
 }
+
 import scala.concurrent.duration.*
 
 class ValidatorProxyIntegrationTest
@@ -41,7 +42,6 @@ class ValidatorProxyIntegrationTest
           "validator-proxy-test",
           "JAVA_TOOL_OPTIONS" -> props.toJvmOptionString,
         ) {
-          println("jvm options:" + props.toJvmOptionString)
           aliceValidatorBackend.start()
           clue("Wait for validator initialization") {
             // Need to wait for the participant node to startup for the user allocation to go through
@@ -83,12 +83,9 @@ class ValidatorProxyIntegrationTest
             aliceValidatorBackend.participantClient.config.ledgerApi.port.unwrap,
           ) shouldBe true
 
-          val urlPattern = "^http://.*:(\\d+)$".r
-          val sequencerPort =
-            sv1Backend.config.localSynchronizerNode.value.sequencer.externalPublicApiUrl match {
-              case urlPattern(portStr) => portStr.toInt
-              case _ => fail("Could not extract port from sequencer external URL")
-            }
+          val sequencerPort = Uri(
+            sv1Backend.config.localSynchronizerNode.value.sequencer.externalPublicApiUrl
+          ).effectivePort
 
           eventuallySucceeds() {
             proxy.proxiedConnectRequest(
--- a/docs/src/release_notes.rst
+++ b/docs/src/release_notes.rst
@@ -19,8 +19,8 @@ Upcoming
 
       - ``/v0/admin/users/offboard``:
         Offboarding a user now also deletes the ledger API user in the participant node.
-      - If you need to use an HTTP proxy in your environment, you can now add `https.proxyHost` and `https.proxyPort`
-        to the `additionalJvmOptions` in the validator helm chart to use the HTTP proxy for outgoing connections.
+      - If you need to use an HTTP proxy in your environment, you can now use `https.proxyHost` and `https.proxyPort` Java system properties.
+        Please see :ref:`HTTP Proxy configuration <validator-http-proxy-helm>` for Kubernetes-Based deployment and :ref:`HTTP Proxy configuration <validator-http-proxy-compose>` for Docker Compose-Based deployment.
 
   - Scan
 
--- a/docs/src/validator_operator/validator_compose.rst
+++ b/docs/src/validator_operator/validator_compose.rst
@@ -82,6 +82,8 @@ the example here. As long as you have docker-compose 2.26.0 or newer you should 
 
 Additional parameters describing your own setup as opposed to the connection to the network are described below.
 
+.. _validator-http-proxy-compose:
+
 HTTP Proxy configuration
 ------------------------
 
--- a/docs/src/validator_operator/validator_helm.rst
+++ b/docs/src/validator_operator/validator_helm.rst
@@ -69,7 +69,7 @@ Create the application namespace within Kubernetes.
     The validator deployment assumes one validator per namespace.
     If you wish to run multiple validators in the same cluster, please create a separate namespace for each.
 
-.. _validator-http-proxy:
+.. _validator-http-proxy-helm:
 
 HTTP Proxy configuration
 ------------------------


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
```diff
@@ -472,4 +361,139 @@ object NodeBase {
       )
 
   }
+
+  def buildHttpClient(
```
nicu-da: note: this change makes it harder to see what actually chagned so i will just review the full method

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
```diff
@@ -472,4 +361,139 @@ object NodeBase {
...
+          case "https" => ConnectionContext.httpsClient(SSLContext.getDefault)
+          case _ => ConnectionContext.noEncryption()
+        }
+
+        val settings = createClientConnectionSettings()
+        val connectionFlow: Flow[HttpRequest, HttpResponse, Future[Http.OutgoingConnection]] =
+          Http()
+            .outgoingConnectionUsingContext(host, port, connectionContext, settings)
+
+        // A new stream is materialized, creating a new connection for every request. The connection is closed on stream completion (success or failure)
```
nicu-da: nit: I don't think this is correct, from what i remember the pekko http client has an internal pool cache

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
```diff
@@ -472,4 +361,139 @@ object NodeBase {
...
+              Future.failed(e)
+            }
+        val start = System.currentTimeMillis()
+        dispatchRequest(request).map { response =>
+          val responseTraceCtx = traceContextFromHeaders(response.headers)
+          val end = System.currentTimeMillis()
+          logger.trace(msg(s"HTTP request took ${end - start} ms to complete"))(responseTraceCtx)
+          logger.debug(msg(s"Received response with status code: ${response.status}"))(
+            responseTraceCtx
+          )
```
nicu-da: Can we merge these logs? Have the time it took for the request in the debug log would be really useful and i don't understand why have it at trace if we log anyway

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
```diff
@@ -472,4 +361,139 @@ object NodeBase {
...
+    def host(scheme: String) = s"$scheme.proxyHost"
+    def port(scheme: String) = s"$scheme.proxyPort"
+    def user(scheme: String) = s"$scheme.proxyUser"
+    def password(scheme: String) = s"$scheme.proxyPassword"
+    def prop(property: String): Option[String] =
+      Option(System.getProperty(property)).map(_.trim)
+    def props(scheme: String) =
+      (prop(host(scheme)), prop(port(scheme)), prop(user(scheme)), prop(password(scheme)))
+    def proxySettings(scheme: String) = props(scheme) match {
+      case (Some(host), Some(port), Some(user), Some(password)) =>
```
nicu-da: this basically reuses the java http ones right? What if the pekko config for the proxy are set? Should we add support for that as well?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
```diff
@@ -472,4 +361,139 @@ object NodeBase {
...
+        }
+      }
+    }
+  private def traceContextFromHeaders(headers: immutable.Seq[HttpHeader]) = {
+    W3CTraceContext
+      .fromHeaders(headers.map(h => h.name() -> h.value()).toMap)
+      .map(_.toTraceContext)
+      .getOrElse(TraceContext.empty)
+  }
+  private def createClientConnectionSettings()(implicit ac: ActorSystem) = {
```
nicu-da: We should add docs for this

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/unit/http/TinyProxySupport.scala
```diff
@@ -0,0 +1,124 @@
...
+import java.nio.file.Files
+import scala.annotation.tailrec
+import scala.collection.mutable
+import scala.concurrent.blocking
+import scala.concurrent.duration.{Duration, FiniteDuration}
+import scala.sys.process.{Process, ProcessLogger}
+import scala.concurrent.duration.*
+
+trait TinyProxySupport {
+  object HttpProxy {
```
nicu-da, line 18: can we add a bit of scala docs on what we're doing here

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/unit/http/HttpProxyTest.scala
```diff
@@ -0,0 +1,212 @@
...
+import org.apache.pekko.http.scaladsl.server.Directives.{complete, *}
+import org.apache.pekko.http.scaladsl.server.Route
+import org.lfdecentralizedtrust.splice.environment.NodeBase
+import org.lfdecentralizedtrust.splice.http.HttpClient
+import org.scalatest.wordspec.AsyncWordSpec
+
+import scala.concurrent.duration.*
+import scala.concurrent.{ExecutionContext, Future}
+
+class HttpProxyTest
```
nicu-da: I think we want a full integration test where we configure the validator and participant with the proxy settings and validate that they can connect to scan and the sequencers using the proxy, we need to guarantee a validator node can work with a proxy config and not just the http client

nicu-da (commented):
thank you! looks good just a few things mostly around docs and testing

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
ray-roestenburg-da: you mean the comment? I think you are thinking of the host level client side API, which has a pool, the connection level API does not offer this. Hmm searching for it I found the comment twice, seems like we have duplicate code in ConsoleHttpCommandRunner so I should probably fix it there too to support this http proxy stuff, just to be safe, for it tests.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
ray-roestenburg-da: Sure can do that

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
ray-roestenburg-da: I think we can just instruct users to use the 'java http ones', since this is also what canton supports for gRPC, so you set everything all at once. But having said that, setting the pekko config will probably just work out of the box.
I'll add a test and see what happens

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/unit/http/HttpProxyTest.scala
ray-roestenburg-da: Ok will add.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
ray-roestenburg-da: I'll actually move it to http package since ConsoleHttpCommandRunner duplicates this code

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
ray-roestenburg-da: (and yes, this reuses the java http ones)

ray-roestenburg-da: [This](https://github.com/grpc/grpc-java/blob/v1.67.1/core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java) is what we use for gRPC FYI (it adapts to the `http.proxyXYZ` settings)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
ray-roestenburg-da: Interesting: https://github.com/hyperledger-labs/splice/blob/56dfa765184ea156f7bcca7766c305db9cf17507/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala#L134-L136 This was broken, but not in ConsoleHttpCommandRunner

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
ray-roestenburg-da: done

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/NodeBase.scala
ray-roestenburg-da: Moved to HttpClient which is a far more logical place for this

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/unit/http/TinyProxySupport.scala
ray-roestenburg-da: done

ray-roestenburg-da: done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/unit/http/HttpProxyTest.scala
ray-roestenburg-da: I've added an integration test which shows that validator, wallet, sv1, and scan are proxied.

ray-roestenburg-da:
Please have another look, I'll add some docs now

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorProxyIntegrationTest.scala
```diff
@@ -0,0 +1,68 @@
...
+            .set("http.nonProxyHosts", "")
+            .set("https.nonProxyHosts", "")
+        withProperties(props) {
+          initDsoWithSv1Only()
+          aliceValidatorBackend.startSync()
+          aliceValidatorBackend.onboardUser(aliceWalletClient.config.ledgerApiUser).discard
+          aliceWalletClient.tap(10)
+          val aliceTxs = aliceWalletClient.listTransactions(None, 10)
+          aliceTxs should not be empty
+          // the proxy should have seen connections to scan, sv1, alice's wallet and validator, and the participant (at least once each)
```
nicu-da: we also should see for the sequencers right? and I assume we cannot easily test for the auth stuff here

nicu-da (approved):
Nice, thank you! LGTM!

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorProxyIntegrationTest.scala
ray-roestenburg-da: Added verification of that as well, thanks!

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorProxyIntegrationTest.scala
```diff
@@ -0,0 +1,73 @@
...
+        val props =
+          SystemProperties()
+            .set("https.proxyHost", "localhost")
+            .set("https.proxyPort", proxy.port.toString)
+            // localhost and tcp loopback addresses are not proxied by default in gRPC and Java URL connections
+            // so we need to override the nonProxyHosts to ensure our proxy is used for all connections
+            .set("http.nonProxyHosts", "")
+            .set("https.nonProxyHosts", "")
+        withProperties(props) {
+          initDsoWithSv1Only()
```
nicu-da: Hmm, shouldn't we start this without the proxy? And wrap with the proxy just the alice validator start. That way we ensure that the assertions we do in the proxy were done by the validator and not the SV

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorProxyIntegrationTest.scala
```diff
@@ -0,0 +1,73 @@
...
+            host,
+            aliceValidatorBackend.config.clientAdminApi.port.unwrap,
+          ) shouldBe true
+          proxy.proxiedConnectRequest(
+            host,
+            sv1ScanBackend.config.participantClient.ledgerApi.clientConfig.port.unwrap,
+          ) shouldBe true
+          proxy.proxiedConnectRequest(
+            host,
+            sv1Backend.sequencerClient.config.clientAdminApi.port.unwrap,
```
nicu-da: we should extract the port from this url right?

```
    sv1Backend.config.localSynchronizerNode.value.sequencer.externalPublicApiUrl
```

ray-roestenburg-da: Would that be the same as `sv1Backend.sequencerClient.config.publicApi.port.unwrap`? I'll check what `sv1Backend.config.localSynchronizerNode.value.sequencer.externalPublicApiUrl` is

nicu-da: our setup declares:

```
   internal-api.port = 5108
      external-public-api-url = "http://localhost:5108"
```

While `sv1Backend.sequencerClient.config.publicApi.port.unwrap` would be the same, it goes through even one more indirection as that's a client we create in the test code, it just happens that we use the localsynchronizer ports. So I would choose to at least go through ` sv1Backend.config.localSynchronizerNode.value.sequencer` to read it from the config directly. Though ideally we would read it from the external public api url because that's the url used to configure the participant sequencer connections.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorProxyIntegrationTest.scala
```diff
@@ -0,0 +1,137 @@
...
+          .set("http.nonProxyHosts", "")
+          .set("https.nonProxyHosts", "")
+      withProperties(props) {
+        withCanton(
+          cantonArgs,
+          cantonExtraConfig,
+          "validator-proxy-test",
+          "JAVA_TOOL_OPTIONS" -> props.toJvmOptionString,
+        ) {
+          println("jvm options:" + props.toJvmOptionString)
```
nicu-da: we probs want to remove this

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorProxyIntegrationTest.scala
```diff
@@ -0,0 +1,137 @@
...
+          aliceValidatorBackend.start()
+          clue("Wait for validator initialization") {
+            // Need to wait for the participant node to startup for the user allocation to go through
+            eventuallySucceeds(timeUntilSuccess = 120.seconds) {
+              EnvironmentDefinition.withAllocatedValidatorUser(aliceValidatorBackend)
+            }
+            aliceValidatorBackend.waitForInitialization()
+          }
```
nicu-da, line 52: would `startSync` not work?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorProxyIntegrationTest.scala
```diff
@@ -0,0 +1,137 @@
...
+            aliceValidatorBackend.participantClient.config.ledgerApi.port.unwrap,
+          ) shouldBe true
+
+          val urlPattern = "^http://.*:(\\d+)$".r
+          val sequencerPort =
+            sv1Backend.config.localSynchronizerNode.value.sequencer.externalPublicApiUrl match {
+              case urlPattern(portStr) => portStr.toInt
+              case _ => fail("Could not extract port from sequencer external URL")
+            }
+
```
nicu-da: nit: We can probably parse it in a pekko url and then access the port directly there

ray-roestenburg-da: yeah ok

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorProxyIntegrationTest.scala
ray-roestenburg-da: (this was also fixed in the end)

PR #2958, opened 2025-10-29
Enabling selected sequencer connections in validator

pasindutennage-da:
Added new option in the existing domain config to enable a validator to specify the set of trusted sequencers.

Solves sequencer related part of https://github.com/hyperledger-labs/splice/issues/2409

If both `url` and `sequencerNames` are set in `ValidatorDecentralizedSynchronizerConfig`, then the validator will fail the bootstrap

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
@@ -50,7 +50,7 @@ class ValidatorSequencerConnectionIntegrationTest
       )
       .withManualStart
 
-  "validator with 'sequencerNames' config connects to specified sequencers and tracks URL changes" in {
+  "validator with 'sequencerNames' set in config connects to specified sequencers and tracks URL changes of sequencers" in {
     implicit env =>
       startAllSync(
         sv1Backend,


apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/config/ValidatorAppConfig.scala
```diff
@@ -97,6 +97,12 @@ case class ValidatorDecentralizedSynchronizerConfig(
       */
     trafficBalanceCacheTimeToLive: NonNegativeFiniteDuration =
       NonNegativeFiniteDuration.ofSeconds(1),
+
+    /** An optional statically specified sequencerNames for a sequencer to use to connect.
```
martinflorian-da: Not sure I understand this line...

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceConfig.scala
```diff
@@ -686,6 +686,13 @@ object SpliceConfig {
         val participantIdentifier =
           ValidatorCantonIdentifierConfig.resolvedNodeIdentifierConfig(conf).participant
         for {
+          _ <- Either.cond(
+            !(conf.domains.global.url.isDefined && conf.domains.global.sequencerNames.isDefined),
+            (),
+            ConfigValidationFailed(
+              "Configuration error: `url` and `sequencerNames` cannot both be specified for the global domain."
```
martinflorian-da, line 693: Did you consider using something like a sealed trait instead of this extra rule? Make the code much easier to reason about, in principle.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SequencerNameFilteringIntegrationTest.scala
```diff
@@ -0,0 +1,138 @@
...
+      }
+
+      withClue("Alice's validator should be functional and able to onboard a user") {
+        eventuallySucceeds() {
+          aliceValidatorBackend.onboardUser(aliceWalletClient.config.ledgerApiUser)
+        }
+      }
+  }
+
+  "validator connects to available sequencers when a configured one is down" in { implicit env =>
```
martinflorian-da: This is slightly misleading.

What you're testing here is that it works if the whole *SV* is not *onboarded*. Perfectly good and useful test IMO, it's just different from "sequencer is down" (the detection and handling of that happens inside the participant).

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceConfig.scala
nicu-da: That would make it harder to keep backwards compatibility right?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SequencerNameFilteringIntegrationTest.scala
nicu-da: I woudl remove this test honestly. we don't care about testing the participant here

note: each test is fairly expensive time wise and our CI is a bit overwhelmed already

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SequencerNameFilteringIntegrationTest.scala
```diff
@@ -0,0 +1,138 @@
...
+import org.lfdecentralizedtrust.splice.integration.tests.SpliceTests.IntegrationTest
+import org.lfdecentralizedtrust.splice.util.{SvTestUtil, WalletTestUtil}
+import com.digitalasset.canton.SynchronizerAlias
+import com.digitalasset.canton.config.NonNegativeFiniteDuration
+import com.digitalasset.canton.sequencing.GrpcSequencerConnection
+import org.apache.pekko.http.scaladsl.model.Uri
+
+import scala.concurrent.duration.*
+
+class SequencerNameFilteringIntegrationTest
```
nicu-da: Do we not have any other test scenarios that validate the participant uses all the sequencers from scan? 
I would ideally try to colocate those tests

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SequencerNameFilteringIntegrationTest.scala
```diff
@@ -0,0 +1,138 @@
...
+              ),
+              automation =
+                c.automation.copy(pollingInterval = NonNegativeFiniteDuration.ofSeconds(1)),
+            )
+          case (_, c) => c
+        }(config)
+      )
+      .withManualStart
+
+  "validator with 'sequencerNames' config connects only to specified sequencers" in {
```
nicu-da: what we don't test is that changing the sequencer url of one of the configured svs is picked up by the validator app and eventually by the participant

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/config/ValidatorAppConfig.scala
```diff
@@ -97,6 +97,12 @@ case class ValidatorDecentralizedSynchronizerConfig(
...
+      * specifies the trusted set of sequencers the participant should connect to.
+      * If set, the url parameter should not be set.
+      */
+    sequencerNames: Option[Seq[String]] = None,
```
nicu-da: if we go for the option we should make the seq a nonemptyseq

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
```diff
@@ -192,8 +192,20 @@ class DomainConnector(
           // so this is just an extra safeguard.
           sequencers.synchronizerId == decentralizedSynchronizerId
         )
+      val sv_filteredSequencers = config.domains.global.sequencerNames match {
```
nicu-da: lets make sure we follow scala naming conventions

```suggestion
      val svFilteredSequencers = config.domains.global.sequencerNames match {
```

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
```diff
@@ -192,8 +192,20 @@ class DomainConnector(
...
+        case Some(allowedNames) =>
+          val allowedNamesSet = allowedNames.toSet
```
nicu-da: we might want to have this is a set in the config directly

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceConfig.scala
pasindutennage-da: In this iteration we will have a simple condition check

in a next PR, a new sealed trait will be introduced, together with threshold

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SequencerNameFilteringIntegrationTest.scala
pasindutennage-da: Test removed

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
pasindutennage-da: you mean have this in the config class parsing?

but whenever we refresh, we still need to check for the names filter right?

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceConfig.scala
martinflorian-da: > That would make it harder to keep backwards compatibility

Hm I guess that's a fair point :thinking: More generally though I wonder if we should care about backwards compatibility of splice configs at all.

In this particular case it seems rare to me that someone will interact with this part of the config manually (as opposed to just doing what our Helm charts and docker compose logic does). So it seems OK for me to break it (with an appropriate note in the release notes of course). (Not a high-confidence opinion.)

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/config/ValidatorAppConfig.scala
pasindutennage-da: Changed the comment.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SequencerNameFilteringIntegrationTest.scala
pasindutennage-da: As discussed, the existing tests focus on the SV participant-sequencer connections, together with mediator.

In this PR, we will keep the 2 tests seperate. But in the future, if required, lets merge them.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SequencerNameFilteringIntegrationTest.scala
pasindutennage-da: Done

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/config/ValidatorAppConfig.scala
pasindutennage-da: Done

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
pasindutennage-da: Done

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/config/ValidatorAppConfig.scala
pasindutennage-da: The custom generic type NonEmpty[Seq] is not recognized by the ConfigReader  because it requires an explicit ConfigReader for its underlying structure. 

NonEmptyList works well because cats provides pre-existing, imported implicit type classes.

martinflorian-da (approved):
Nice work, thank you!

PR #2961, opened 2025-10-29
Smallish fixes to public reset docs + updating a `foundDso` default

martinflorian-da:
Based on experience from the latest TestNet reset.

See for example https://github.com/DACH-NY/canton-network-internal/issues/2060


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -94,7 +94,7 @@ object SvOnboardingConfig {
       bootstrappingDump: Option[SvBootstrapDumpConfig] = None,
       initialPackageConfig: InitialPackageConfig = InitialPackageConfig.defaultInitialPackageConfig,
       initialTransferPreapprovalFee: Option[BigDecimal] = None,
-      initialFeaturedAppActivityMarkerAmount: Option[BigDecimal] = None,
+      initialFeaturedAppActivityMarkerAmount: Option[BigDecimal] = Some(BigDecimal(1.0)),
```
martinflorian-da, line 97: ugh sorry forgot that I also changed this; triggered CI now;

the context is that this is what we currently have on MainNet + there is no convenient way to override this default during deployment

isegall-da (approved):
Thank you

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
rautenrieth-da: I assume we don't need to mention this in release notes because it's always us founding networks?

martinflorian-da: Yeah I think no one will really care about this change... we're basically aligning the default with what we do on MainNet.

PR #2985, opened 2025-10-30
Unbump resources on scratch networks

martinflorian-da:
Save some money.

Part of #2943

Notably this also fixes the logic for `SEQUENCER_LOW_RESOURCES` (yes it was broken) ... and removes that flag altogether in favor of explicit values in the config. Because it seems to me that those flags and `if`s in the pulumi add more complexity than is worth it.

Tested expected behaviour via a manual scratch deploy.

[static]


cluster/pulumi/sv-canton/src/decentralizedSynchronizerNode.ts
```diff
@@ -108,20 +107,7 @@ abstract class InStackDecentralizedSynchronizerNode
             driver: driver,
             tokenExpirationTime: sequencerTokenExpirationTime,
             additionalEnvVars: svConfig.sequencer?.additionalEnvVars,
-            resources: lowResourceSequencer
-              ? {
-                  resources: {
```
martinflorian-da, line 113: double nexted resources means it does nothing, without error

PR #2989, opened 2025-10-30
make operator's pulumi projects scope configurable and add the GHA project

mblaze-da:
In principle this change should not cause any change in Pulumi resources right now. It is needed though for [875](https://github.com/DACH-NY/canton-network-internal/issues/875) so that deploying an operator in the `splice` cluster that manages only the `gha` project is possible. I'm testing this functionality on a scratchnet manually.

### DIFF
--- a/cluster/pulumi/common/src/operator/config.ts
+++ b/cluster/pulumi/common/src/operator/config.ts
@@ -33,12 +33,29 @@ const ProjectFilterSchema = z.union([
 
 export type ProjectFilter = z.infer<typeof ProjectFilterSchema>;
 
-const defaultProjectFilters = [
-  'canton-network',
-  'infra',
-  'sv-canton',
-  'validator-runbook',
-] as const;
+function* iterateDefaultProjectFilters(): Generator<ProjectFilter> {
+  yield 'canton-network';
+  yield 'infra';
+  yield 'sv-canton';
+  yield 'validator-runbook';
+
+  // TODO(DACH-NY/canton-network-internal#875): Remove the following env. var. based additions when
+  //   it is certain that all cases of their usage is replaced with explicit configuration.
+  //   Note that their inclusion is purposefully limited to default project filters so that explicit
+  //   configuration always takes precedence to maximize clarity.
+  if (DeploySvRunbook) {
+    yield 'sv-runbook';
+  }
+  if (config.envFlag('SPLICE_DEPLOY_MULTI_VALIDATOR', false)) {
+    yield 'multi-validator';
+  }
+  if (mustInstallValidator1) {
+    yield 'validator1';
+  }
+  if (mustInstallSplitwell) {
+    yield 'splitwell';
+  }
+}
 
 export const OperatorDeploymentConfigSchema = z.object({
   operatorDeployment: z.object({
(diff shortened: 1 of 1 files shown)


cluster/pulumi/deployment/src/config.ts
```diff
@@ -0,0 +1,57 @@
...
+export type ProjectFilter = z.infer<typeof ProjectFilterSchema>;
+
+const defaultProjectFilters = [
+  'canton-network',
+  'infra',
+  'sv-canton',
+  'validator-runbook',
+] as const;
+
+const DeploymentConfigSchema = z.object({
```
nicu-da: we should probably merge within the same file as OperatorDeploymentConfigSchema

I agree it's a bit awkward the location because of reasons but better than having the config split

cluster/pulumi/deployment/src/config.ts
```diff
@@ -0,0 +1,57 @@
...
+  deployment: z
+    .object({
+      projectWhitelist: z.set(ProjectFilterSchema),
```
nicu-da:
```suggestion
      projectsToDeploy: z.set(ProjectFilterSchema),
```

cluster/pulumi/deployment/src/config.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const ProjectFilterSchema = z.union([
+  z.literal('canton-network'),
+  z.literal('gha'),
+  z.literal('infra'),
+  z.literal('multi-validator'),
+  z.literal('splitwell'),
+  z.literal('sv-canton'),
+  z.literal('sv-runbook'),
+  z.literal('validator-runbook'),
```
nicu-da: This is actually a good time to get rid of validator-runbook naming here. Can we set for the config the name of validators? 

Reason is that this actually controls mutliple validators but validator-runbook is just the legacy name when it was just one of them

cluster/pulumi/common/src/operator/config.ts
```diff
@@ -9,6 +17,29 @@ export const StackConfigSchema = z.object({
   parallelism: z.number().optional(),
 });
 
+const ProjectFilterSchema = z.union([
+  z.literal('canton-network'),
```
martinflorian-da, line 21: Nit: can this be refactored into something slightly less verbose, such as `['a','b','c'].map(z.literal)`?

martinflorian-da (approved):
Beautiful stuff, thank you!

In part because it's maybe a nice exercise, but mostly for my OCD, I suggest to run the following checks before merging:

- hdm_test on a scratch (that one uses the operator): https://github.com/DACH-NY/canton-network-internal/blob/90abd4cd073cfddb206e77046446752aa44ce6e5/README.md?plain=1#L131
- deployments preview on a prod cluster, triggered on a (draft) PR on internal that pins its `splice` submodule to this branch
- ~same as :point_up: but for the `splice` cluster (I think you did this already but my memory is bad... and also you didn't comment about this on your PR here?)~ nvm, I remembered/rediscovered https://github.com/DACH-NY/canton-network-internal/pull/2512 now

cluster/pulumi/common/src/operator/config.ts
mblaze-da: This breaks proper literal typing.

Issue #989, opened 2025-06-05
Batch inserts


canton-network-da:
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#5643). Original author: @OriolMunoz-da ]

See: https://github.com/DACH-NY/canton-network-node/pull/5624#discussion_r1227088411

PR #2998, opened 2025-10-30
Batch ACS ingestion (non-import)

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/3049

1 by 1 ingestion:
```
Ingested 2671 in 29289112 ns, average time: 19731946.05 ns over 2672 records, total time: 52723759853 ms

=> 19.73 ms per item
```

batches of 100 latest results after commits from Nov 4:
`Ingested batch 6624 (86 elements) in 31703632 ns, average per-item time: 303890.00 ns over 662387 records, total time: 201292788597 ns`

meaning that on my laptop, to ingest the cilr ACS dump:
- 0.3ms per item :warning: 
- 3.35m total

batches of 1000 were slower than 100 on my machine

See here for how the tests were run: https://github.com/hyperledger-labs/splice/pull/2998#issuecomment-3481504620

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
@@ -102,7 +102,7 @@ class UpdateIngestionService(
         logger.debug(s"Processing batch of ${batch.size} elements")
         ingestionSink.ingestUpdateBatch(batch.map(_.updateOrCheckpoint))
       case None =>
-        logger.warn("Received empty batch of updates to ingest. This is never supposed to happen.")
+        logger.error("Received empty batch of updates to ingest. This is never supposed to happen.")
         Future.unit
     }
   }
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
@@ -1636,7 +1636,8 @@ final class DbMultiDomainAcsStore[TXE](
     )
 
     case class AcsInsertEntry(
-        offset: Long, // not always the same as `createdEvent.getOffset`
+        // not always the same as `createdEvent.getOffset`: when `ingestACS`ing, it's `participantBegin`
+        offset: Long,
         createdEvent: CreatedEvent,
         stateData: ContractStateRowData,
     )


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DecentralizedSynchronizerMigrationIntegrationTest.scala
```diff
@@ -435,9 +435,11 @@ class DecentralizedSynchronizerMigrationIntegrationTest
         .getExternalPartyBalance(externalParty)
         .totalUnlockedCoin shouldBe "0.0000000000"
       walletClient.transferPreapprovalSend(externalParty, 40.0, UUID.randomUUID.toString)
-      validatorBackend
-        .getExternalPartyBalance(externalParty)
-        .totalUnlockedCoin shouldBe "40.0000000000"
+      eventually() {
```
rautenrieth-da, line 438: I assume those are fixes for existing flakes that now occur more frequently with your changes?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/AutomationConfig.scala
```diff
@@ -108,3 +109,8 @@ case class AutomationConfig(
   def topupTriggerPollingInterval_ : NonNegativeFiniteDuration =
     topupTriggerPollingInterval.getOrElse(pollingInterval)
 }
+
+case class IngestionConfig(
+    maxBatchSize: Int = 100,
+    batchWaitTime: PositiveFiniteDuration = PositiveFiniteDuration.ofMillis(50L),
+)
```
rautenrieth-da, line 118: Default looks sensible

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -319,72 +320,76 @@ class UpdateHistory(
...
-        val offset: Long = updateOrCheckpoint.offset
-        val recordTime = updateOrCheckpoint match {
-          case TreeUpdateOrOffsetCheckpoint.Update(ReassignmentUpdate(reassignment), _) =>
-            Some(reassignment.recordTime)
-          case TreeUpdateOrOffsetCheckpoint.Update(TransactionTreeUpdate(tree), _) =>
-            Some(CantonTimestamp.assertFromInstant(tree.getRecordTime))
-          case TreeUpdateOrOffsetCheckpoint.Checkpoint(_) => None
-        }
+        MonadUtil
+          .sequentialTraverse(batch.toList) { updateOrCheckpoint =>
```
rautenrieth-da, line 327: This is not doing any actual batching, is it? Do we have an issue for also batching update history inserts, or is that currently not a bottleneck?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -2193,4 +2295,36 @@ object DbMultiDomainAcsStore {
...
   }
+
+  sealed trait BatchStep
+  case class UpdateCheckpoint(checkpoint: TreeUpdateOrOffsetCheckpoint.Checkpoint) extends BatchStep
+  case class IngestReassignment(update: ReassignmentUpdate, synchronizerId: SynchronizerId)
+      extends BatchStep
+  case class IngestTransactionTreesBatch(batch: NonEmptyVector[IngestTransactionTree])
+      extends BatchStep
+  case class IngestTransactionTree(tree: Transaction, synchronizerId: SynchronizerId)
+  def batchInsertionSteps(
```
rautenrieth-da: If I understand correctly, this rolls up consecutive transaction tree updates into a single "step". Can we do even more? For example, can we first remove all offset checkpoints that are not at the tail of the list? That would allow us to roll up even more transaction trees.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DecentralizedSynchronizerMigrationIntegrationTest.scala
OriolMunoz-da, line 438: (since this is the first file)
The `groupedWithin` makes some ingestion flakes more likely to appear, so you'll see some eventuallys added here and there...
That being said, I reduced the waitTime in one of the latter commits ,so this might not be necessary anymore but doesn't hurt either

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
```diff
@@ -85,16 +86,29 @@ class UpdateIngestionService(
           } yield offset
       }
     } yield new SpliceLedgerSubscription(
-      source = connection.updates(subscribeFrom, filter),
+      source = connection
+        .updates(subscribeFrom, filter)
+        .groupedWithin(
+          config.ingestion.maxBatchSize,
+          config.ingestion.batchWaitTime.asFiniteApproximation,
+        ),
```
OriolMunoz-da: lmk if you have better ideas
`batchWaitTime` is 50ms

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsQueries.scala
```diff
@@ -268,21 +268,19 @@ trait AcsQueries extends AcsJdbcTypes {
...
-      sp: SetParameter[V]
-  ): SQLActionBuilder = {
+  protected def sqlCommaSeparated(
+      seq: Iterable[SQLActionBuilder]
+  ): SQLActionBuilderChain = {
     seq
-      .map(v => sql"$v")
+      .map(SQLActionBuilderChain(_))
       .reduceOption { (acc, next) =>
-        (acc ++ sql"," ++ next).toActionBuilder
```
OriolMunoz-da, line 279: I removed a bunch of noisy `toActionBuilder`

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsRowData.scala
```diff
@@ -24,6 +25,10 @@ trait AcsRowData {
 }
 
 object AcsRowData {
+  trait HasIndexColumns[R] {
```
OriolMunoz-da, line 28: the problem (with Scan as example):
- ScanAcsRowData.indexColumns returns all columns, with either Some or None as value, which is fine
- The interface fallback returns NO columns, which means it will be missing columns when batch inserting

So I needed to associate which column names to insert per AcsRowData type, and this seemed the most straightforward way (otherwise I need to propagate type signatures to every AppStore and that was too noisy for no benefit). For the interface fallback, they are assigned to `null`.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1229,289 +1251,317 @@ final class DbMultiDomainAcsStore[TXE](
...
+                            ),
+                          summary,
+                        )
+                      )
+                    }
+                } yield ()
+              case Delete(exercisedEvent) =>
+                doDeleteContract(exercisedEvent, summary)
+            })*),
+            DBIO.seq(txLogEntries.map { case (recordTime, (txe, offset, synchronizerId)) =>
```
OriolMunoz-da: https://github.com/hyperledger-labs/splice/issues/3048

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
```diff
@@ -739,14 +751,9 @@ object MultiDomainAcsStore extends StoreErrors {
         incompleteIn: Seq[IncompleteReassignmentEvent.Assign],
     )(implicit traceContext: TraceContext): Future[Unit]
 
-    def ingestUpdate(update: TreeUpdateOrOffsetCheckpoint)(implicit
+    def ingestUpdateBatch(batch: NonEmptyList[TreeUpdateOrOffsetCheckpoint])(implicit
         traceContext: TraceContext
     ): Future[Unit]
-
-    final def ingestUpdate(synchronizerId: SynchronizerId, update: TreeUpdate)(implicit
```
OriolMunoz-da, line 746: moved to test-only code

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/commands/HttpScanAppClient.scala
```diff
@@ -1410,6 +1410,8 @@ object HttpScanAppClient {
     override def handleOk()(implicit decoder: TemplateJsonDecoder) = {
       case http.GetUpdateByIdV1Response.OK(response) =>
         Right(response)
+      case http.GetUpdateByIdV1Response.NotFound(_) =>
+        Left(s"Update with ID $updateId not found")
```
OriolMunoz-da, line 1414: somehow managed to get a MatchError flake, looks like thsi was missing

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
moritzkiefer-da: hm why do we need to batch with a wait time? I was expecting we have a buffer of maxBatchSize. We stream getUpdates into that. We then take the full buffer contents and insert them, concurrently we stream getUpdates into the buffer again. After you finished the insert you take your full buffer again.
That way on low throughput you don't batch anything and you get low latency. But as throughput increases you batch more.

OriolMunoz-da: hmmm... is that as simple as the `source.batch` from pekko streams?

moritzkiefer-da: I believe so yes

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DecentralizedSynchronizerMigrationIntegrationTest.scala
OriolMunoz-da: there's no `groupedWithin` anymore, see here: https://github.com/hyperledger-labs/splice/pull/2998#discussion_r2487183832

so I guess I just fixed very unlikely flakes :shrug:

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
```diff
@@ -85,16 +86,26 @@ class UpdateIngestionService(
           } yield offset
       }
     } yield new SpliceLedgerSubscription(
-      source = connection.updates(subscribeFrom, filter),
+      source = connection
+        .updates(subscribeFrom, filter)
+        .batch(config.ingestion.maxBatchSize.toLong, Vector(_))(_ :+ _),
```
moritzkiefer-da, line 91: Let's add a histogram metric for the batch size.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1562,57 +1612,121 @@ final class DbMultiDomainAcsStore[TXE](
...
+        contractFilter.getInterfaceViewsIndexColumnNames
+      )
+      for {
+        rawContractIdToEventNumber <- (sql"""
+                insert into #$acsTableName(store_id, migration_id, contract_id, template_id_package_id, template_id_qualified_name, package_name,
+                                           create_arguments, created_event_blob, created_at, contract_expires_at,
+                                           assigned_domain, reassignment_counter, reassignment_target_domain,
+                                           reassignment_source_domain, reassignment_submitter, reassignment_unassign_id
+                                           #$acsIndexColumnNames)
+                values """ ++ joinedAcsTableValues ++ sql" returning contract_id, event_number").toActionBuilder
```
moritzkiefer-da, line 1665: is this subject to the 65k query param limit for a statement? 100 txs with 1k events each doesn't seem that unrealistic based on some of the transactions I've seen on mainnet.

I think Canton uses a `          MonadUtil.batchedSequentialTraverse` to guard against this usually.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1562,57 +1612,121 @@ final class DbMultiDomainAcsStore[TXE](
...
+        .map(entry => entry.createdEvent.getContractId -> getInsertValues(entry))
+      val acsTableValues = insertValues.map(_._2.acsTable)
+      val joinedAcsTableValues = acsTableValues.reduceLeft(_ ++ sql"," ++ _)
+      // column names are hardcoded so they can be raw-interpolated
+      val acsIndexColumnNames = mkIndexColumnNames(contractFilter.getAcsIndexColumnNames)
+      val interfaceViewsIndexColumnNames = mkIndexColumnNames(
+        contractFilter.getInterfaceViewsIndexColumnNames
+      )
+      for {
+        rawContractIdToEventNumber <- (sql"""
```
moritzkiefer-da: this is probably a naive question, Canton has `DbStorage.bulkOperation`. Would it make sense to use this here? It seems a bit more clever. It doesn't support the returning stuff I guess, dunno if we need that.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
moritzkiefer-da: probably worth adding a test that tries to insert large batches to see that it doesn't blow up

moritzkiefer-da (commented):
didn't finish review but publishing my comments so far, overall looks great!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: unfortunately I need the returning (or a query afterwards, but that's worse than returning) to insert the corresponding interface views, as they're referenced by event_number (PK)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: added a test here: https://github.com/hyperledger-labs/splice/pull/2998/commits/e0f2a04edebc1a499ff776406e26dfc002203e36
it still works with 10k created events per batch, but it's too slow
at 10x more, it OOMs on my machine before hitting any other limit

I'll still check to add the `MonadUtil.batchedSequentialTraverse` for extra safety though

OriolMunoz-da: > I'll still check to add the MonadUtil.batchedSequentialTraverse for extra safety though

actually doesn't look like I can do that in the current implementation:
- DBIO doesn't have a Parallel instance
- Moving that to the Future means that this is not transactional with the txlog

lmk if I'm misisng something

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1562,57 +1613,121 @@ final class DbMultiDomainAcsStore[TXE](
...
 
-    private def doIngestAcsInsert(
-        offset: Long,
+    case class AcsInsertEntry(
+        offset: Long, // not always the same as `createdEvent.getOffset`
         createdEvent: CreatedEvent,
         stateData: ContractStateRowData,
+    )
+    private def doIngestAcsInserts(
+        entries: NonEmptyList[AcsInsertEntry],
```
rautenrieth-da, line 1645: This method is always called as
```scala
DBIO.seq(doIngestAcsInserts(NonEmptyList.of(singleElement)))
```

is that just preparation for some future refactoring or am I missing something? Don't we end up executing the exact same SQL statements, just more of them wrapped in a single transaction? What change in this PR is responsible for the 10x improvement that you measured?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1229,289 +1251,318 @@ final class DbMultiDomainAcsStore[TXE](
...
-              st
-            }
-          },
+      val workTodo = trees.batch
+        .map(tree =>
+          Trees
+            .foldTree(
+              tree.tree,
+              VectorMap.empty[String, OperationToDo],
+            )(
```
rautenrieth-da: Is there a reason why we can't calculate a single `workTodo`, processing the events from all trees together? AFAICT, that would have two benefits:

1. It would allow us to apply the `optimization: a delete on a contract cancels-out with the corresponding insert` for contracts that are created and archived in different transactions from the same batch. Looking at the [distribution of MainNet contract life times](https://docs.google.com/document/d/11BE-R2kVL3pnC20fDKWhzr-PN1iBf7Su-DtcmXUpUGk/edit?tab=t.0#heading=h.15ln8q7khx97), this is unlikely to optimize away anything

2. It would allow us to convert the `DBIO.seq(DBIO.seq(workTodo.map(... DBIO.seq(...))))` into batch statements

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DecentralizedSynchronizerMigrationIntegrationTest.scala
OriolMunoz-da: see here : https://github.com/hyperledger-labs/splice/pull/2998#discussion_r2487213101

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/AutomationConfig.scala
OriolMunoz-da: note batchWaitTime has been removed

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
OriolMunoz-da: we only care about the ACS import (that doesn't go through ingestAcs) for now, though ofc everything else needs to be worked on
I am tracking them in https://github.com/hyperledger-labs/splice/issues/989

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: I'm very mystified as to how I even saw a performance increase - you're right, that's definitely wrong

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: > It would allow us to apply the optimization: a delete on a contract cancels-out with the corresponding insert for contracts that are created and archived in different transactions from the same batch.

this is already the case

> It would allow us to convert the DBIO.seq(DBIO.seq(workTodo.map(... DBIO.seq(...)))) into batch statements

odds are this is just me being stupid like above

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: I was indeed missing something: https://github.com/hyperledger-labs/splice/pull/2998#discussion_r2490626920
my stupidity clouds my brain

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
OriolMunoz-da, line 91: another improvement I can think here is to use `batchWeighted`, computing the weight something like every event adds X to the cost

wdyt?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: fixed and made sure that it doesn't blow up, though by adding some `grouped` in the batching statements code...

OriolMunoz-da:
latest results after commits from today:
`Ingested batch 6624 (86 elements) in 31703632 ns, average per-item time: 303890.00 ns over 662387 records, total time: 201292788597 ns`

meaning that on my laptop, to ingest the cilr ACS dump:
- 0.3ms per item :warning: 
- 3.35m total

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
rautenrieth-da: Sounds like a good idea, would also allow you to set an upper bound on the number of events in the batch, to prevent `SELECT ... WHERE ... IN (a,b,c,d,...)` statements from blowing up because of too many parameters.

ray-roestenburg-da: Sounds like a good idea, you mean look into updates and see the size of them basically in terms of events and calculate some weight based on that right? Sounds like something to do to finetune though, would be nice to see how this performs first

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
```diff
@@ -85,16 +86,26 @@ class UpdateIngestionService(
...
-  )(implicit traceContext: TraceContext) =
-    ingestionSink.ingestUpdate(msg.updateOrCheckpoint)
+      msgs: Seq[GetTreeUpdatesResponse]
+  )(implicit traceContext: TraceContext): Future[Unit] = {
+    NonEmptyList.fromFoldable(msgs) match {
+      case Some(batch) =>
+        logger.debug(s"Processing batch of ${batch.size} elements")
+        ingestionSink.ingestUpdateBatch(batch.map(_.updateOrCheckpoint))
+      case None =>
+        logger.warn("Received empty batch of updates to ingest. This is never supposed to happen.")
```
ray-roestenburg-da: This is an error IMHO

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
ray-roestenburg-da: Maybe just the batching in reading already helped a lot? We definitely need to understand this.

ray-roestenburg-da: ah wait this is outdated, in ingestTransactionTrees you do some form of batching

ray-roestenburg-da: So I guess `DBIO.sequence(entries.grouped(ingestionConfig.maxEntriesPerInsert).map { entries => ...` is causing the performance increase?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1562,57 +1635,123 @@ final class DbMultiDomainAcsStore[TXE](
       reassignmentUnassignId = Some(String255.tryCreate(event.unassignId)),
     )
 
-    private def doIngestAcsInsert(
-        offset: Long,
+    case class AcsInsertEntry(
+        offset: Long, // not always the same as `createdEvent.getOffset`
```
ray-roestenburg-da: This comment needs a little more context, why is this?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1562,57 +1635,123 @@ final class DbMultiDomainAcsStore[TXE](
...
+        val acsTableValues = insertValues.map(_._2.acsTable)
+        val joinedAcsTableValues = acsTableValues.reduceLeft(_ ++ sql"," ++ _)
+        // column names are hardcoded so they can be raw-interpolated
+        val acsIndexColumnNames = mkIndexColumnNames(contractFilter.getAcsIndexColumnNames)
+        val interfaceViewsIndexColumnNames = mkIndexColumnNames(
+          contractFilter.getInterfaceViewsIndexColumnNames
+        )
+        for {
+          rawContractIdToEventNumber <- (sql"""
+                insert into #$acsTableName(store_id, migration_id, contract_id, template_id_package_id, template_id_qualified_name, package_name,
```
ray-roestenburg-da, line 1660: You can probably improve this by using addBatch in a prepared statement?

ray-roestenburg-da: You can do that in a separate PR, this is already great of course

ray-roestenburg-da: Please run a cluster test

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1631,39 +1770,35 @@ final class DbMultiDomainAcsStore[TXE](
...
-          and contract_id = ${lengthLimited(event.getContractId)}
-      """.map {
-        case 1 =>
-          summary.ingestedArchivedEvents.addOne(event)
-        case _ =>
-          // there was actually no contract with that id. This can happen because:
-          // `contractFilter.mightContain` in `getIngestionWork` can return true for a template,
-          // but that might still satisfy some other filter, so the contract was never inserted
-          summary.numFilteredArchivedEvents += 1
+    private def doDeleteContracts(events: Seq[ExercisedEvent], summary: MutableIngestionSummary) = {
```
ray-roestenburg-da, line 1782: Probably in a different PR, I would be interested in metrics on the number of these deletes executed (in the context of the ingestion), as well as the nr of inserts.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/StoreMetrics.scala
```diff
@@ -45,6 +46,14 @@ class StoreMetrics(metricsFactory: LabeledMetricsFactory)(metricsContext: Metric
     )
   )(metricsContext)
 
+  val batchSize: Histogram = metricsFactory.histogram(
```
ray-roestenburg-da, line 49: so yeah per batch the number of inserts and deletes would be interesting

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanTables.scala
```diff
@@ -67,26 +67,71 @@ object ScanTables extends AcsTables {
...
+      ScanAcsStoreRowData.IndexColumns.member_traffic_domain -> memberTrafficDomain,
+      ScanAcsStoreRowData.IndexColumns.total_traffic_purchased -> totalTrafficPurchased,
+      ScanAcsStoreRowData.IndexColumns.validator_license_rounds_collected -> validatorLicenseRoundsCollected,
+      ScanAcsStoreRowData.IndexColumns.sv_party -> svParty,
+      ScanAcsStoreRowData.IndexColumns.vote_action_requiring_confirmation -> voteActionRequiringConfirmation,
+      ScanAcsStoreRowData.IndexColumns.vote_requester_name -> voteRequesterName.map(lengthLimited),
+      ScanAcsStoreRowData.IndexColumns.vote_request_tracking_cid -> voteRequestTrackingCid,
+      ScanAcsStoreRowData.IndexColumns.transfer_preapproval_receiver -> transferPreapprovalReceiver,
+      ScanAcsStoreRowData.IndexColumns.transfer_preapproval_valid_from -> transferPreapprovalValidFrom,
+      ScanAcsStoreRowData.IndexColumns.wallet_party -> walletParty,
```
ray-roestenburg-da, line 86: Thanks for cleanup!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: Note that the query is using RETURNING, which doesn't seem to be compatible with preparedstatement batches judging by this comment from Canton:
https://github.com/hyperledger-labs/splice/blob/e03b0cc172bf1d83ac7a8ee76feed20c79a9157c/canton/community/base/src/main/scala/com/digitalasset/canton/resource/Storage.scala#L688

> The operation must not return a result set!

Furhtermore, this is the signature of `executeBatch`:
`int[] executeBatch() throws SQLException;`

This is unlike:
`ResultSet executeQuery() throws SQLException`
which returns a ResultSet

ray-roestenburg-da: yeah, reading up, the `... returning ...` query in postgres is not supported in addBatch in this way.

ray-roestenburg-da: This is probably already good enough and close to the performance you get from that anyway

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: already done :smile: : https://github.com/hyperledger-labs/splice/pull/2946/files#diff-5b8e9f3158653279c0c943bd4e5c5a2045ac0bbd84db623fbf2da5f7329bce38R2259

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/StoreMetrics.scala
OriolMunoz-da: same as above: https://github.com/hyperledger-labs/splice/pull/2946/files#diff-5b8e9f3158653279c0c943bd4e5c5a2045ac0bbd84db623fbf2da5f7329bce38R2259

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
ray-roestenburg-da: Oh nice! yeah archived == deleted, cool.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/StoreMetrics.scala
ray-roestenburg-da: Thanks yeah read over that one

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: > So I guess DBIO.sequence(entries.grouped(ingestionConfig.maxEntriesPerInsert).map { entries => ... is causing the performance increase?

no, that's just for safety, to avoid passing more than the 65k parameters to postgres and was added later

there was an initial performance increase caused by ¿¿¿some change??? 
then batching was properly finalized in https://github.com/hyperledger-labs/splice/pull/2998/commits/9a6f3edf0812abd31ae6543161ff1223fb6f48c3 and then gave an even further performance increase

ray-roestenburg-da: Thanks

ray-roestenburg-da: Yeah I meant doIngestAcsInserts, and of course then you have to actually call it with a batch of entries, not just one :-)

OriolMunoz-da:
Successful cluster test: https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/39546/workflows/58f344c7-bb98-42d6-99a1-fc118458a7aa (had to rerun because docker failed to pull postgres :rage4: )

PR #3034, opened 2025-10-31
introduce an automation for GHA runner version update

mblaze-da:
This introduces a new workflow that will fire roughly once every two weeks. It will update the version of the GHA runner base image that is used to build `splice-test-docker-runner` and `splice-test-runner-hook` images.

The shell script was tested locally. I'm not sure though if the whole solution is correct but I don't know how to test the actual workflow without creating it.

### DIFF
--- a/scripts/bump-gha-runner-version.sh
+++ b/scripts/bump-gha-runner-version.sh
@@ -13,9 +13,10 @@ runner_version=$(
   | sed 's/^v//' # remove the 'v' prefix
 )
 
+# gets the multiplatform image digest
 runner_digest=$(
-  docker manifest inspect "ghcr.io/actions/actions-runner:${runner_version}" \
-  | jq -r '.manifests[] | select(.platform.architecture == "amd64") | .digest'
+  docker buildx imagetools inspect "ghcr.io/actions/actions-runner:${runner_version}" \
+  | yq '.Digest'
 )
 
 docker_runner_file="${SPLICE_ROOT}/cluster/images/splice-test-docker-runner/Dockerfile"


.github/workflows/bump_gha_runner_version.yml
```diff
@@ -0,0 +1,24 @@
+name: Bump GHA runner version (WIP)
+on:
+  # schedule:
+  #   - cron: '30 6 1,15 * *'  # Monthly on the 1st and 15th at 06:30
+  workflow_dispatch:
+
+jobs:
+  bump_gha_runner_version:
```
martinflorian-da, line 8: Do we get notified if this fails? Not 100% sure but I think you need some more boilerplate for that...

S.a.: https://github.com/hyperledger-labs/splice/blob/92d74d52a21095ffb5e6779b608a6a3ad3e10885/.github/actions/tests/failure_notifications/action.yml#L1

martinflorian-da: (Unless it's super annoying to add, let's add notifications; we're very likely to not notice this getting hard broken otherwise.)

scripts/bump-gha-runner-version.sh
```diff
@@ -0,0 +1,46 @@
...
+
+if git diff --exit-code --quiet "${docker_runner_file}" "${runner_hook_file}"; then
+  echo "GHA runner version is up to date."
+  exit 0
+fi
+
+git add --all
+updated_branch="gha-runner-version-bump-$(date +%Y-%m-%d)"
+git switch -c "${updated_branch}"
+git commit -m "[static] bump GHA runner version to the latest (auto-generated)" -s
```
martinflorian-da, line 44: Is "build docker images" part of `[static]`?

If not we might want to either:

- add that (follow-up PR?)
- do a docker-build here on (only) `splice-test-docker-runner`?

I think I like the second option more (we change images ~rarely and when we do we typically also do full cluster tests which do build docker images), assuming that 1. we can do that without forcing a full splice build ( :crossed_fingers: ) and 2. we get notifications and a [failures board](https://github.com/orgs/DACH-NY/projects/48) entry in case this flow fails (see my other comment).

isegall-da:
> > > The shell script was tested locally. I'm not sure though if the whole solution is correct but I don't know how to test the actual workflow without creating it.
> > 
> > 
> > I think you could temporarily remove the schedule and manually trigger the action from the UI to see that it works as expected, no?
> 
> Can't you also trigger a GHA run from a branch? We do that for CircleCI runs all the time. But clicking around [with your branch](https://github.com/hyperledger-labs/splice/actions?query=branch%3Amblaze-da%2Fgha-runner-version-updater) the new workflow doesn't seem to show up, sad. @isegall-da do you know of a good way to test this without merging to main first? (Which I'm not against here, for the record, just seems nicer if we can have a feedback loop that doesn't involve merging a PR.)

I think you need to add `on: dispatch` or something like that to the worklow to make it runnable directly (probably imprecise wording, I'm commuting)

.github/workflows/bump_gha_runner_version.yml
mblaze-da: I've copied the job from other workflows. I hope it works.

scripts/bump-gha-runner-version.sh
mblaze-da: Good idea! I've added explicit `docker-build` for these two images.

isegall-da:
> > > > > > The shell script was tested locally. I'm not sure though if the whole solution is correct but I don't know how to test the actual workflow without creating it.
> > > > > 
> > > > > 
> > > > > I think you could temporarily remove the schedule and manually trigger the action from the UI to see that it works as expected, no?
> > > > 
> > > > 
> > > > Can't you also trigger a GHA run from a branch? We do that for CircleCI runs all the time. But clicking around [with your branch](https://github.com/hyperledger-labs/splice/actions?query=branch%3Amblaze-da%2Fgha-runner-version-updater) the new workflow doesn't seem to show up, sad. @isegall-da do you know of a good way to test this without merging to main first? (Which I'm not against here, for the record, just seems nicer if we can have a feedback loop that doesn't involve merging a PR.)
> > > 
> > > 
> > > I think you need to add `on: dispatch` or something like that to the worklow to make it runnable directly (probably imprecise wording, I'm commuting)
> > 
> > 
> > I was close, it's `on: workflow_dispatch` see [here](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch)
> 
> @isegall-da I do have this configured and I'm still not able to trigger it through `gh workflow run`. It is not listed as a workflow. I think the problem is something that they mentioned in a note under the doc you've linked:
> 
> > This event will only trigger a workflow run if the workflow file exists on the default branch.
> 
> So it seems that I need some form of dummy definition for a workflow on `main` for it to be runnable from other branches. So I guess testing changes to existing workflows can be done easily but for new workflows it requires some prior changes on `main`.

Argh, that's annoying. Oh, well, seems like "merge to main and test there" is the answer.

PR #3053, opened 2025-11-03
Make SequencerPruningIntegrationTest less flaky

martinflorian-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/6243

[ci]


martinflorian-da:
@nicu-da This could work but I have a better idea now: just reduce the sv status report reporting interval to something like 15 or 20 seconds. Still this fix could be better than nothing, so feel free to merge if you like it the flake is too annoying to wait for me (I plan to finish this very early in my day tomorrow.)

PR #3062, opened 2025-11-04
Make SequencerPruningIntegrationTest less flaky

martinflorian-da:
...by causing more on-ledger traffic so that the pruning trigger notices more quickly that it is safe to prune.

Fixes https://github.com/DACH-NY/cn-test-failures/issues/6243 (see deeper analysis there)

[ci]


nicu-da (approved):
thx

PR #3082, opened 2025-11-05
Validator local DB schema

pasindutennage-da:
Adding a new DB schema for validator local DB operations.

Scan connections of the validator are stored in the database.

### DIFF
--- a/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorConfigProvider.scala
+++ b/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorConfigProvider.scala
@@ -5,22 +5,22 @@ package org.lfdecentralizedtrust.splice.validator.store
 
 import cats.data.OptionT
 import com.digitalasset.canton.tracing.TraceContext
-import io.circe.{Decoder, Encoder}
+import io.circe.Codec
 import scala.concurrent.Future
+import io.circe.generic.semiauto.*
 
 final case class ScanUrlInternalConfig(
     svName: String,
     url: String,
 )
 
+object ScanUrlInternalConfig {
+  implicit val scanUrlCodec: Codec[ScanUrlInternalConfig] = deriveCodec[ScanUrlInternalConfig]
+}
+
 class ValidatorConfigProvider(config: ValidatorInternalStore) {
 
   private val scanInternalConfigKey = "validator_scan_internal_config_key"
-
-  implicit val scanUrlEncoder: Encoder[ScanUrlInternalConfig] =
-    Encoder.forProduct2("svName", "url")(s => (s.svName, s.url))
-  implicit val scanUrlDecoder: Decoder[ScanUrlInternalConfig] =
-    Decoder.forProduct2("svName", "url")(ScanUrlInternalConfig.apply)
 
   final def setScanUrlInternalConfig(
       value: Seq[ScanUrlInternalConfig]


apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V050__create_scan_config.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE validator_scan_config
+(
+    sv_name       text NOT NULL,
+    scan_url      text NOT NULL,
+    restart_count int  NOT NULL,
```
nicu-da: don't see a real reason for this

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,99 @@
...
+    extends ValidatorInternalStore {
+
+  import storage.api.*
+
+  private val setScanConfigRow: PositionedParameters => ScanConfigRow => Unit = pp =>
+    row => {
+      pp.setString(row.svName)
+      pp.setString(row.scanUrl)
+      pp.setInt(row.restartCount)
+    }
```
nicu-da: we should use slick for this

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  private implicit val getScanConfigRow: GetResult[ScanConfigRow] = GetResult { r =>
+    ScanConfigRow(
+      svName = r.nextString(),
+      scanUrl = r.nextString(),
+      restartCount = r.nextInt(),
+    )
+  }
+
+  private def toFuture[A](
```
nicu-da: this is in some utility already if i remember correctly

nicu-da (commented):
I think it's first worth to have the review from @rautenrieth-da or Oriol

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V050__create_scan_config.sql
pasindutennage-da: you mean the restart count? 

it is used as a versioning -- we need a way to denote the most recent versions.

else, every time we update, we have to delete entire database and then insert - not a good practice IMO

nicu-da: why can't we upsert the entries? we just care about the latest entry for an sv

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: PositionedParameters is indeed from slick.jdbc

This is the implicit used whenever a ScanConfigRow needs to be converted to a database row, a pattern I observed in existing implementations

apps/validator/src/test/scala/org/lfdecentralizedtrust/splice/store/ValidatorInternalStoreTest.scala
```diff
@@ -0,0 +1,142 @@
...
+    new DbValidatorInternalStore(
+      storage = storage,
+      loggingContext = elc,
+      closeContext = cc,
+    )
+  }
+
+  override protected def cleanDb(
+      storage: DbStorage
+  )(implicit traceContext: TraceContext): FutureUnlessShutdown[?] = resetValidatorInternalTables(
```
rautenrieth-da: Easier to add the new table to `SpliceDbTest.resetAllAppTables` and call that here - that's how all other tests are doing it.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
moritzkiefer-da: you shouldn't need that, you can just do something ilke
```
sql"insert into… values(${config.svName}, …)"
```
Note the use of the sql interpolation. Check for example DbMultiDomainAcsStore.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V050__create_scan_config.sql
rautenrieth-da: > else, every time we update, we have to delete entire database and then insert - not a good practice IMO

So your idea is to have an append-only table where you never delete rows, and `restart_count` is used to figure out what the current value is? That will result in a table with unbounded size, and querying the data will get progressively more expensive.

Can't you have something like

```
CREATE TABLE validator_scan_config
(
    sv_name      text primary key NOT NULL,
    scan_url      text NOT NULL,
}
```

And then every time you want to update the content, you call `insert into validator_scan_config ... on conflict do update` for each row that should exist, and `delete from validator_scan_config ...` for each row that should not exist (if a sv gets offboarded)? I don't have much context, but the table is expected to contain little data and is not expected to be updated often, right? You don't need to optimize it for massive throughput.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
```diff
@@ -0,0 +1,16 @@
...
+import com.digitalasset.canton.tracing.TraceContext
+import org.lfdecentralizedtrust.splice.validator.store.db.ValidatorInternalTables.ScanConfigRow
+
+import scala.concurrent.Future
+
+trait ValidatorInternalStore {
+
+  def setScanConfigs(rows: Seq[ScanConfigRow])(implicit tc: TraceContext): Future[Unit]
+
+  def getScanConfigs()(implicit tc: TraceContext): Future[Seq[ScanConfigRow]]
```
rautenrieth-da: This interface should hide implementation details. Consider returning a `Map[String, String]` (map of sv name to scan url) or removing the restartCount from `ScanConfigRow`.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,99 @@
...
+      insertSql,
+      rows,
+      storage.profile,
+    )(setScanConfigRow)
+
+    val transactionalAction: DBIOAction[Unit, slick.dbio.NoStream, slick.dbio.Effect.All] =
+      bulkAction.transactionally
+
+    val futureUnlessShutdown: FutureUnlessShutdown[Unit] = storage.profile match {
+      case _: Postgres =>
```
rautenrieth-da: Why do you need to match for Postgres? Other stores have this code inside the class to allow the use of `.transactionally`

```scala
  override val profile: JdbcProfile = storage.profile.jdbc
```

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
rautenrieth-da: `FutureUnlessShutdownUtil.futureUnlessShutdownToFuture`

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/ValidatorInternalTables.scala
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package org.lfdecentralizedtrust.splice.validator.store.db
+
+object ValidatorInternalTables {
+
+  val tableName: String = "validator_scan_config"
```
rautenrieth-da: I would inline this value everywhere. We usually only interpolate the table name if we have the same query running against different tables.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  override def setScanConfigs(rows: Seq[ScanConfigRow])(implicit tc: TraceContext): Future[Unit] = {
+
+    val insertSql =
+      s"""INSERT INTO $tableName(sv_name, scan_url, restart_count)
+         |VALUES (?, ?, ?)
+         |ON CONFLICT (scan_url, sv_name, restart_count) DO NOTHING
+         |""".stripMargin
+
+    val bulkAction: DbAction.All[Unit] = DbStorage.bulkOperation_(
```
rautenrieth-da: Was it a requirement to use bulk operations? It sounds to me like inserting rows individually using
```scala
DBIO.sequence(
  rows.map(r =>
    sqlu"insert into validator_scan_config ..."
  )
).transactionally
```

would result in significantly simpler code, without having a measurable effect on performance.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V050__create_scan_config.sql
nicu-da: (it will also be queries only during startup so keeping it as basic as possible should work). 
We could even go one step further and actually just store a json under a config key. so turn this into a meta config table, that way we don't even care about deleting as we update all the data  at once :shrug:

pasindutennage-da: Resolved -- I used the approach Nicu has proposed -- use a key and json. So now we only need write,update,read only a single row

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved, used FutureUnlessShutdownUtil.futureUnlessShutdownToFuture

apps/validator/src/test/scala/org/lfdecentralizedtrust/splice/store/ValidatorInternalStoreTest.scala
pasindutennage-da: Resolved, added to SpliceDbTest.resetAllAppTables

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
pasindutennage-da: Resolved, used Map[String, String]

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved, removed the match

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved. Not using bulkOperation_ anymore

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Used the approach @moritzkiefer-da proposed

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V050__create_scan_config.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE validator_internal_config
+(
+    config_key   text primary key not null,
+    config_value jsonb            not null,
+    CONSTRAINT uc_scan_config UNIQUE (config_key)
```
rautenrieth-da: A primary key is an automatic unique constraint. Either remove this constraint, or remove the `primary key` from the column definition above and change this constraint to
```suggestion
    CONSTRAINT uc_scan_config PRIMARY KEY (config_key)
```

Both have the same effect, defining the primary key as an explicit constraint lets you name the constraint so that it's easier to modify it later (as unlikely as changing the primary key is).

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,83 @@
...
+
+    val jsonString: String = configData.toJson.noSpaces
+
+    val action = sql"""INSERT INTO #$tableName (config_key, config_value)
+        VALUES ($key, $jsonString::jsonb)
+        ON CONFLICT (config_key) DO UPDATE
+        SET config_value = excluded.config_value""".asUpdate
+
+    val f = storage.update(action, "set-validator-config")
+    FutureUnlessShutdownUtil.futureUnlessShutdownToFuture(f).map(_ => ())
```
rautenrieth-da:
```suggestion
    storage.update(action, "set-validator-config").map(_ => ())
```

`FutureUnlessShutdownUtil.futureUnlessShutdownToFuture` is an implicit method, I _think_ you can just import it and it will auto-convert the `FutureUnlessShutdown` to a `Future`.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  override def getConfig(key: String)(implicit tc: TraceContext): Future[Map[String, String]] = {
+
+    val queryAction = sql"""SELECT config_key, config_value
```
rautenrieth-da: You don't need to select the `config_key`, you know the keys are unique and you are filtering for `config_key=$key`.

This could potentially simplify the reading of `ValidatorInternalConfig`.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,83 @@
...
+          ValidatorInternalConfig
+            .fromJson(json)
+            .fold(
+              err =>
+                throw new Exception(
+                  s"Failed to decode JSON to ValidatorInternalConfig: ${err.getMessage}"
+                ),
+              identity,
+            )
+      }
```
rautenrieth-da: Have a look at https://github.com/hyperledger-labs/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsQueries.scala#L57. That's a `GetResult[]` instance that contains much simpler code for reading JSON object.

The implicit for reading Json objects (required for the `<<[Json]` part) is defined at https://github.com/hyperledger-labs/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsJdbcTypes.scala#L276. I think it's fine to use `extends AcsJdbcTypes` for `DbValidatorInternalStore`, we use AcsJdbcTypes in many other places that have nothing to do with the ACS.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
```diff
@@ -0,0 +1,15 @@
...
+
+package org.lfdecentralizedtrust.splice.validator.store
+
+import com.digitalasset.canton.tracing.TraceContext
+
+import scala.concurrent.Future
+
+trait ValidatorInternalStore {
+
+  def setConfig(key: String, values: Map[String, String])(implicit tc: TraceContext): Future[Unit]
```
rautenrieth-da: This store is limited to storing simple key-value maps like
```json
{
  "key1": "value1",
  "key2": "value2"
}
```

That will be difficult to use if we want to store more complicated values (nested objects, arrays, ...). I think this store should either store and read arbitrary Json values:

```suggestion
  def setConfig(key: String, value: Json)(implicit tc: TraceContext): Future[Unit]
```

or (my preference) it should provide a specific method for storing the scan urls:

```suggestion
  def setScanUrls(value: Seq[ScanUrl])(implicit tc: TraceContext): Future[Unit]
  final case class ScanUrl(
    svName: String,
    url: String
  )
```

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V050__create_scan_config.sql
pasindutennage-da: Resolved: Added Primary Key to constraint

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved: used implicit futureUnlessShutdownToFuture

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved, not selecting the key anymore

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved: Mixed in AcsJdbcTypes

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
pasindutennage-da: I prefer your first suggestion "def setConfig(key: String, value: Json)", because that allows any future internal config to be stored, without worrying about the DB layer.

pasindutennage-da: Resolved: Kept key based json, and then added custom scan savings

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Decided to return only json, hence now can use the implicit without additional details

apps/validator/src/test/scala/org/lfdecentralizedtrust/splice/store/ValidatorInternalStoreTest.scala
```diff
@@ -0,0 +1,159 @@
...
+        retrievedValue shouldEqual initialValue
+      }
+    }
+
+    "return an empty JSON for a non-existent key" in {
+      for {
+        store <- mkStore()
+        retrievedValue <- store.getConfig("non-existent-key")
+      } yield {
+        retrievedValue shouldEqual Json.obj()
```
nicu-da: Any reason why return an empty json instead of returning an option with None in this case? (in certain scenarios we would probably want to distinguish between the two)

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  val profile: JdbcProfile = storage.profile.jdbc
+
+  def setConfig(key: String, value: Json)(implicit tc: TraceContext): Future[Unit] = {
+    val action = sql"""INSERT INTO #$tableName (config_key, config_value)
+        VALUES ($key, $value)
+        ON CONFLICT (config_key) DO UPDATE
+        SET config_value = excluded.config_value""".asUpdate
+
+    storage.update(action, "set-validator-internal-config").map(_ => ())
```
nicu-da: lets add at least a debug log here for setting the config

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  implicit val scanUrlEncoder: Encoder[ScanUrlInternalConfig] =
+    Encoder.forProduct2("svName", "url")(s => (s.svName, s.url))
+  implicit val scanUrlDecoder: Decoder[ScanUrlInternalConfig] =
+    Decoder.forProduct2("svName", "url")(ScanUrlInternalConfig.apply)
+
+  def setScanUrlInternalConfig(
+      value: Seq[ScanUrlInternalConfig]
+  )(implicit tc: TraceContext): Future[Unit] = {
+    setConfig(scanInternalConfigKey, value.asJson)
```
nicu-da: lets add an info log here

apps/validator/src/test/scala/org/lfdecentralizedtrust/splice/store/ValidatorInternalStoreTest.scala
pasindutennage-da: Resolved: Used Future[Option[Json]] as return type

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved: added a debug log

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
pasindutennage-da: Resolved: Added info log

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,57 @@
...
+    with AcsJdbcTypes {
+
+  val tableName = "validator_internal_config"
+
+  val profile: JdbcProfile = storage.profile.jdbc
+
+  override val logger: TracedLogger = loggerFactory.getTracedLogger(getClass)
+
+  def setConfig(key: String, value: Json)(implicit tc: TraceContext): Future[Unit] = {
+    val action = sql"""INSERT INTO #$tableName (config_key, config_value)
```
rautenrieth-da:
```suggestion
    val action = sql"""INSERT INTO validator_internal_config (config_key, config_value)
```

let's not use raw string interpolation unless we have to

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,57 @@
...
+        VALUES ($key, $value)
+        ON CONFLICT (config_key) DO UPDATE
+        SET config_value = excluded.config_value""".asUpdate
+    val updateAction = storage.update(action, "set-validator-internal-config")
+    logger.debug(s"saving validator config in database $value")
```
rautenrieth-da:
```suggestion
    logger.debug(s"saving validator config in database with key '$key' and value '$value'")
```

and possibly `${value.noSpaces}` instead of `$value`, not sure how Json objects are converted to strings.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
```diff
@@ -0,0 +1,54 @@
...
+  def getConfig(key: String)(implicit tc: TraceContext): Future[Option[Json]]
+
+  private val scanInternalConfigKey = "validator_scan_internal_config_key"
+
+  implicit val scanUrlEncoder: Encoder[ScanUrlInternalConfig] =
+    Encoder.forProduct2("svName", "url")(s => (s.svName, s.url))
+  implicit val scanUrlDecoder: Decoder[ScanUrlInternalConfig] =
+    Decoder.forProduct2("svName", "url")(ScanUrlInternalConfig.apply)
+
+  def setScanUrlInternalConfig(
```
rautenrieth-da:
```suggestion
  final def setScanUrlInternalConfig(
```

Not required, but protected/final help communicate how these methods are meant to be used.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
```diff
@@ -0,0 +1,54 @@
...
+import io.circe.Json
+import io.circe.syntax.*
+import io.circe.{Decoder, Encoder}
+import scala.concurrent.{ExecutionContext, Future}
+
+trait ValidatorInternalStore {
+
+  def logger: TracedLogger
+
+  def setConfig(key: String, value: Json)(implicit tc: TraceContext): Future[Unit]
```
rautenrieth-da:
```suggestion
  protected def setConfig(key: String, value: Json)(implicit tc: TraceContext): Future[Unit]
```

We don't expect users to ever call this directly, right? The key `scanInternalConfigKey` is not public.

If you want the method to be accessible from test code, you can also use `protected[store]`, then you can call it from any class in the `org.lfdecentralizedtrust.splice.validator.store` package.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
```diff
@@ -0,0 +1,54 @@
...
+    logger.info(s"saving scan internal config in the database $value")
+    setConfig(scanInternalConfigKey, value.asJson)
+  }
+
+  def getScanUrlInternalConfig(
+  )(implicit tc: TraceContext, ec: ExecutionContext): Future[Option[Seq[ScanUrlInternalConfig]]] = {
+
+    getConfig(scanInternalConfigKey).map {
+      case Some(json) =>
+        if (json.isNull || json.asObject.exists(_.isEmpty)) {
```
rautenrieth-da: I'd skip this (non-exhaustive) validation and rely on `json.as[Seq[ScanUrlInternalConfig]].toOption` throwing an exception if it can't parse the json. Only this class can write the values, so unless you have a bug, the parsing should always succeed.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
```diff
@@ -0,0 +1,54 @@
...
+  def setConfig(key: String, value: Json)(implicit tc: TraceContext): Future[Unit]
+
+  def getConfig(key: String)(implicit tc: TraceContext): Future[Option[Json]]
+
+  private val scanInternalConfigKey = "validator_scan_internal_config_key"
+
+  implicit val scanUrlEncoder: Encoder[ScanUrlInternalConfig] =
+    Encoder.forProduct2("svName", "url")(s => (s.svName, s.url))
+  implicit val scanUrlDecoder: Decoder[ScanUrlInternalConfig] =
+    Decoder.forProduct2("svName", "url")(ScanUrlInternalConfig.apply)
```
nicu-da: lets move this along with `ScanUrlInternalConfig` to it's own file

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
```diff
@@ -0,0 +1,54 @@
...
+import io.circe.{Decoder, Encoder}
+import scala.concurrent.{ExecutionContext, Future}
+
+trait ValidatorInternalStore {
+
+  def logger: TracedLogger
+
+  def setConfig(key: String, value: Json)(implicit tc: TraceContext): Future[Unit]
+
+  def getConfig(key: String)(implicit tc: TraceContext): Future[Option[Json]]
```
nicu-da: We could switch to OptionT[Future, Json] to have a nicer API

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
```diff
@@ -0,0 +1,54 @@
...
+  )(implicit tc: TraceContext): Future[Unit] = {
+    logger.info(s"saving scan internal config in the database $value")
+    setConfig(scanInternalConfigKey, value.asJson)
+  }
+
+  def getScanUrlInternalConfig(
+  )(implicit tc: TraceContext, ec: ExecutionContext): Future[Option[Seq[ScanUrlInternalConfig]]] = {
+
+    getConfig(scanInternalConfigKey).map {
+      case Some(json) =>
```
nicu-da: let's avoid pattern matching here and just use a .map

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
nicu-da: this just implies that the json is invalid. lets instead use just `json.as` and handle the left of the Either

nicu-da: I would actually move the full encoding/decoding directly in the Db store implementation

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
nicu-da: I would actually take one step further, keep only the get and set for the config in the store itself, and extract the specific to getting and setting the scan urls into a `ValidatorConfigProvider`

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,57 @@
...
+    extends ValidatorInternalStore
+    with AcsJdbcTypes {
+
+  val tableName = "validator_internal_config"
+
+  val profile: JdbcProfile = storage.profile.jdbc
+
+  override val logger: TracedLogger = loggerFactory.getTracedLogger(getClass)
+
+  def setConfig(key: String, value: Json)(implicit tc: TraceContext): Future[Unit] = {
```
rautenrieth-da:
```suggestion
  override def setConfig(key: String, value: Json)(implicit tc: TraceContext): Future[Unit] = {
```

`override` makes sure the compiler complains if the super method is removed or its signature changed.

apps/validator/src/test/scala/org/lfdecentralizedtrust/splice/store/ValidatorInternalStoreTest.scala
```diff
@@ -0,0 +1,160 @@
...
+    val otherKey = "other-config"
+    val otherValue = Json.fromString("payload2")
+
+    "set and get a json payload successfully" in {
+      for {
+        store <- mkStore()
+        _ <- store.setConfig(configKey, initialValue)
+        retrievedValue <- store.getConfig(configKey)
+      } yield {
+        retrievedValue.getOrElse(Json.Null) shouldEqual initialValue
```
rautenrieth-da:
```suggestion
        retrievedValue shouldBe Some(initialValue)
```

`shouldBe` is slightly better than `shouldEqual` if you don't need to customize equality.
Moving the `Some` to the right side makes the check closer to the interface of your store.

rautenrieth-da (commented):
Nice work, left a few minor comments

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/ValidatorInternalTables.scala
pasindutennage-da: Fixed: removed variable and used inline in queries

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved: Used table name inline

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
pasindutennage-da: Resolved, used the final keyword

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved

apps/validator/src/test/scala/org/lfdecentralizedtrust/splice/store/ValidatorInternalStoreTest.scala
pasindutennage-da: Resolved, used ShouldBe

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
pasindutennage-da: Fixed, used OptionT[Future, Json]

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,57 @@
...
+    protected val loggerFactory: NamedLoggerFactory,
+)(implicit val ec: ExecutionContext)
+    extends ValidatorInternalStore
+    with AcsJdbcTypes {
+
+  val tableName = "validator_internal_config"
+
+  val profile: JdbcProfile = storage.profile.jdbc
+
+  override val logger: TracedLogger = loggerFactory.getTracedLogger(getClass)
```
nicu-da: This should be provided by extending `NamedLogging`

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
nicu-da: discussed with Robert, we can skip this one

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
pasindutennage-da:
```
json.as[Seq[ScanUrlInternalConfig]].fold(
            failure => throw new RuntimeException(s"FATAL: Corrupt configuration data found for $scanInternalConfigKey. Cannot decode.", failure), 
            success => Some(success)
          )
```

This way, json passing is simple, but at the same time an Error is thrown (without hiding it in a Option None)

Note that in very rare cases, even a correctly written json object will fail decoding, because of DB corruption / packet corruption (rare cases, but better to throw the exception)

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved: Extended to named login

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
pasindutennage-da: Updated to

```
OptionT(
      getConfig(scanInternalConfigKey)
        .map { json =>
          json.as[Seq[ScanUrlInternalConfig]].fold(
            failure => throw new RuntimeException(s"FATAL: Corrupt configuration data found for $scanInternalConfigKey. Cannot decode.", failure),
            success => Some(success)
          )
        }
        .getOrElse(None)
    )
```

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
pasindutennage-da: Resolved: Moved to a different class.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
pasindutennage-da: Resolved: Used ValidatorConfigProvider new class

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
pasindutennage-da: Resolved: Moved to ValidatorConfigProvider

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
nicu-da: Not sure I fully follow, with NamedLogging you no longer need to declare the logger right?

rautenrieth-da: Yes, if you `extends NamedLogging` on the trait, you can remove this line.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
pasindutennage-da: Test classes are in package org.lfdecentralizedtrust.splice.store. Hence protected cannot be used?

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,55 @@
...
+    implicit val closeContext: CloseContext,
+    protected val loggerFactory: NamedLoggerFactory,
+)(implicit val ec: ExecutionContext)
+    extends ValidatorInternalStore
+    with AcsJdbcTypes
+    with NamedLogging {
+
+  val profile: JdbcProfile = storage.profile.jdbc
+
+  override def setConfig(key: String, value: Json)(implicit tc: TraceContext): Future[Unit] = {
```
nicu-da: Sorry to be a stickler for this, but we should strive to not have `Json` in our type definition, it's more or less the equivalent of having object. We should instead pass the generic type with a codec implicit directly 
```suggestion
  override def setConfig[T:Encoder](key: String, value: T)(implicit tc: TraceContext): Future[Unit]
```

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -0,0 +1,55 @@
...
+        ON CONFLICT (config_key) DO UPDATE
+        SET config_value = excluded.config_value""".asUpdate
+    val updateAction = storage.update(action, "set-validator-internal-config")
+    logger.debug(
+      s"saving validator config in database with key '$key' and value '${value.noSpaces}'"
+    )
+    updateAction.map(_ => ())
+  }
+
+  override def getConfig(key: String)(implicit tc: TraceContext): OptionT[Future, Json] = {
```
nicu-da:
```suggestion
  override def getConfig[T:Decoder](key: String)(implicit tc: TraceContext): OptionT[Future, T] = {
```

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorConfigProvider.scala
```diff
@@ -0,0 +1,56 @@
...
+)
+
+class ValidatorConfigProvider(config: ValidatorInternalStore) {
+
+  private val scanInternalConfigKey = "validator_scan_internal_config_key"
+
+  implicit val scanUrlEncoder: Encoder[ScanUrlInternalConfig] =
+    Encoder.forProduct2("svName", "url")(s => (s.svName, s.url))
+  implicit val scanUrlDecoder: Decoder[ScanUrlInternalConfig] =
+    Decoder.forProduct2("svName", "url")(ScanUrlInternalConfig.apply)
```
nicu-da: nit: Ideally we would have this in the companion object for `ValidatorConfigProvider`, and `ScanUrlInternalConfig` as well.
also we should use semi automatic derivation to make our life easier
```suggestion
  implicit val scanUrlCodec: Codec[ScanUrlInternalConfig] = deriveCodec[ScanUrlInternalConfig]
    
```

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved: Used generics

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved: Used generics

nicu-da (approved):
LGTM!

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorConfigProvider.scala
pasindutennage-da: Resolved, moved the codec to companion object

Issue #2472, opened 2025-09-30
[canton-3.4] Enable force ACS pruning of 30 days for SVS


moritzkiefer-da:
follow-up: 
- [x] #3109
- [ ] enable default 30 days pruning only for SVs (not validators)
- [ ] we have to wait for https://github.com/DACH-NY/canton/issues/29513

canton issue https://github.com/DACH-NY/canton/issues/28249

martinflorian-da:
...and add integration test to confirm it works (see https://github.com/hyperledger-labs/splice/issues/3066 which you probably want to complete first).

moritzkiefer-da:
integration test definitely makes sense but honestly I'd consider cluster alerts for the oldest event being older than the pruning window to be more important here.

cristina-da:
[Pruning with configurable commitment state](https://github.com/DACH-NY/canton/pull/29135/) is merged to 3.4

PR #3089, opened 2025-11-05
Add integration test for participant pruning

julientinguely-da:
fixes https://github.com/hyperledger-labs/splice/issues/3066

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #3090, opened 2025-11-06
Enable participant pruning with 30d retention

julientinguely-da:
fixes https://github.com/hyperledger-labs/splice/issues/2472

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


cluster/deployment/config.yaml
```diff
@@ -158,6 +158,10 @@ svs:
           memory: '2Gi'
         limits:
           memory: '4Gi'
+      participantPruningSchedule:
```
nicu-da: Aren't we missing canton support for force pruning? 🤔

martinflorian-da: I'd feel a lot better if we had some confirmation this works before enabling it on CILR... (Like an integration test?)

martinflorian-da: But I guess you're only adding the possibility here, and we can test by enabling only on a single SV?

Did you do a scratchnet test of some sort, maybe with a very short retention period + a validator that does some taps and transfers and then disappears?

nicu-da: we need https://github.com/DACH-NY/canton/issues/28249 before we enable pruning for SV participants

julientinguely-da: yes indeed we need that. 

The integration test is coming: https://github.com/hyperledger-labs/splice/pull/3089

julientinguely-da: I'm really confused the test works when reverting https://github.com/hyperledger-labs/splice/pull/3065 (removing the adminClaim)

nicu-da: Added a comment on the test, we need to use loggerfactory.assertlogs

cluster/deployment/config.yaml
```diff
@@ -158,6 +158,10 @@ svs:
...
+        cron: "0 5 * * * ?" # Runs daily at 5AM
```
moritzkiefer-da: We should set it in the validator config default not in our cluster config. I discussed it with Simon and we want to switch the default to 30 days pruning for everyone, not just SVs, because a significant majority of performance issues people have is due to lack of pruning. Ofc needs to be clearly pointed out in release notes.

That said: At this point I think trying to get this into 0.5.0 is too risky so I'd keep it for 0.5.1.

As for force pruning: I think it's fine to ship that later. Worst case your node just won't prune as expected but for most nodes it probably will be fine.

In terms of monitoring: We should make sure we have alerts for the max event age metric being older than 30 days.

julientinguely-da: yes, makes sense to have it for all SVs

> That said: At this point I think trying to get this into 0.5.0 is too risky so I'd keep it for 0.5.1.

makes sense.

> In terms of monitoring: We should make sure we have alerts for the max event age metric being older than 30 days.

following up on this: https://github.com/hyperledger-labs/splice/issues/3109

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/config/ValidatorAppConfig.scala
```diff
@@ -193,7 +194,14 @@ case class ValidatorAppBackendConfig(
       NonNegativeFiniteDuration.ofSeconds(5),
     // Identifier for all Canton nodes controlled by this application
     cantonIdentifierConfig: Option[ValidatorCantonIdentifierConfig] = None,
-    participantPruningSchedule: Option[PruningConfig] = None,
+    // Participant pruning is enabled by default for all validators and SVs
+    participantPruningSchedule: Option[PruningConfig] = Some(
+      PruningConfig(
+        cron = "0 5 * * * ?", // Runs daily at 5AM
```
moritzkiefer-da: any reason not to copy it from the mediator config where we run it every 10min with a max duration of 5min? More smaller pruning steps usually work better as they distribute load

julientinguely-da: no reason, thanks for the suggestion

moritzkiefer-da (commented):
Let's please add clear release notes on this and how to overwrite it

julientinguely-da:
@moritzkiefer-da I guess we wait https://github.com/DACH-NY/cn-test-failures/issues/6386#issuecomment-3511598600 is resolved before to merge this PR?

moritzkiefer-da:
don't think we necessarily need to wait for the change but we should wait until after 0.5.1

docs/src/release_notes.rst
```diff
@@ -5,8 +5,17 @@
 
 .. _release_notes:
 
+Upcoming
+--------
+
+- Canton Participant
+
+  - Participants now prune data to only retain the last 30 days matching the 30 day pruning interval of sequencers and mediators. Y
+    You can overwrite this default value by adding the following env variable in the `additionalEnvVars` helm values of the validator app: `ADDITIONAL_CONFIG_PARTICIPANT_PRUNING`
```
nicu-da, line 14: shouldn't we include an example of the value here as well?

julientinguely-da: added, let me know if it's good

nicu-da: lgtm thanks

PR #3092, opened 2025-11-06
Add advisory lock to SqlIndexInitializationTrigger

rautenrieth-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/6328

The new test proves that `SqlIndexInitializationTrigger` would detect if an index is being built and not do anything. This PR still adds an advisory lock to SqlIndexInitializationTrigger, as in test code we are running multiple instances of `SqlIndexInitializationTrigger` against the same database.

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AdvisoryLockIds.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AdvisoryLockIds.scala
@@ -3,11 +3,62 @@
 
 package org.lfdecentralizedtrust.splice.store.db
 
+import com.digitalasset.canton.resource.DbStorage
+import slick.dbio.{DBIOAction, Effect, NoStream}
+import slick.jdbc.JdbcProfile
+import slick.jdbc.canton.ActionBasedSQLInterpolation.Implicits.actionBasedSQLInterpolationCanton
+
+import scala.concurrent.ExecutionContext
+
 /** Registry for advisory lock identifiers used in splice applications.
   */
 object AdvisoryLockIds {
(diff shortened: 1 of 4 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AdvisoryLockIds.scala
```diff
@@ -3,11 +3,62 @@
...
   private val base: Long = 0x73706c00
 
-  final val acsSnapshotDataInsert: Long = base + 1
+  final case class AdvisoryLockId(value: Long) extends AnyVal
+  final case class FailedToAcquireAdvisoryLock(lockId: AdvisoryLockId)
+      extends Exception(s"Failed to acquire exclusive lock with id ${lockId.value}")
+
+  final val acsSnapshotDataInsert = AdvisoryLockId(base + 1)
+
+  final val sqlIndexInitializationTrigger = AdvisoryLockId(base + 2)
```
OriolMunoz-da, line 25: for the acs snapshots you're guaranteed that only scan is acquiring locks on the scan DB, but for the index init trigger, won't all apps try to acquire it (and slow themselves down)? Afaict the "value" will be the same for all apps

In production we have a single physical DB (cn-apps), but separate logic DBs (scan/sv/validator for SVs). Afaict advisory locks are per logical-DBs, not server-wide, so it should be fine, but please double-check...

rautenrieth-da:
```
org.postgresql.util.PSQLException: ERROR: CREATE INDEX CONCURRENTLY cannot run inside a transaction block
```

No locking for SqlIndexInitializationTrigger then.

PR #3097, opened 2025-11-06
Add log ignore for dropping invalid indexes

rautenrieth-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/6328

I don't want to demote the warning, as otherwise we get no feedback that the trigger is not progressing, and we really don't expect the trigger to fail in production where each app gets its own database.


OriolMunoz-da (approved):
sgtm

PR #3096, opened 2025-11-06
Canton default change: `max-token-lifetime=Inf`

martinflorian-da:
According to upstream, setting this to `Inf` should remove the requirement of adding an `exp` field to tokens (which might be a breaking change for some validators).

[ci]

Signed-off-by: Martin Florian <martin.florian@digitalasset.com>

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #3105, opened 2025-11-06
Don't set expiration on auth tokens

martinflorian-da:
To make sure that https://github.com/hyperledger-labs/splice/pull/3096 has the desired effect.

[ci]


moritzkiefer-da:
lgtm, this sets us back to the state before the canton change so this is quite well tested, I would just merge it even before 0.5.0

moritzkiefer-da (approved):

moritzkiefer-da:
I guess I don't understand why clicking "merge" now would delay the release but :shrug:

martinflorian-da:
> delay the release

Well if, contrary to our expectation, this does cause some of our flows to fail somewhere, we might need to fix it.

Also I guess I mean not just the release itself but also the upgrades to it.

PR #3120, opened 2025-11-07
Confirmation requests alerting fixes

martinflorian-da:
1. Fix per-member alerts to actually read the correct config
2. Align defaults with what makes sense on MainNet today

Context: https://github.com/DACH-NY/cn-test-failures/issues/6378

[static]


cluster/deployment/config.yaml
```diff
@@ -74,10 +74,10 @@ monitoring:
       # we alert if the rate is higher than expected to spot potential overload situations
       confirmationRequests:
         total:
-          rate: 10
+          rate: 15
```
martinflorian-da, line 77: We configure this on MainNet today

cluster/deployment/config.yaml
```diff
@@ -74,10 +74,10 @@ monitoring:
...
           overMinutes: 5
         perMember:
-          rate: 2
+          rate: 3
```
martinflorian-da, line 80: "2" would give false positives about once every other day, so seemed better to have this less noisy: https://grafana.global.canton.network.digitalasset.com/goto/7g-fzxkvg?orgId=1

martinflorian-da: image

PR #3176, opened 2025-11-11
Add log checker query for JWTs

martinflorian-da:
First half of https://github.com/DACH-NY/canton-network-internal/issues/2647

Tested manually with mock logs...

[ci]

### DIFF
--- a/.github/actions/scripts/check-logs.sh
+++ b/.github/actions/scripts/check-logs.sh
@@ -93,11 +93,14 @@ find_exceptions |
 sed -i 's/secret=test/secret=hidden/g' "$LOGFILE"
 
 find_secrets() {
-  set +o pipefail # rg returns 1 if there were not matches
+  set +o pipefail # rg returns 1 if there were no matches
+  # Common x=y format
   rg -o -e "(secret|token|private-key|password)=[^,[:space:]]*" "$LOGFILE" |
     # we mask secrets as "****" in our logs and testcontainers obfuscates secrets as "hidden non-blank value"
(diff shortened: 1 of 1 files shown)


.github/actions/scripts/check-logs.sh
```diff
@@ -93,11 +93,14 @@ find_exceptions |
...
 
 find_secrets() {
-  set +o pipefail # rg returns 1 if there were not matches
+  set +o pipefail # rg returns 1 if there were no matches
+  # Common x=y format
   rg -o -e "(secret|token|private-key|password)=[^,[:space:]]*" "$LOGFILE" |
     # we mask secrets as "****" in our logs and testcontainers obfuscates secrets as "hidden non-blank value"
     # (https://github.com/testcontainers/testcontainers-java/blob/bf5605a2031d7f29f86a85430e3509a198c6e125/core/src/main/java/org/testcontainers/utility/AuthConfigUtil.java#L33)
     rg -v -e "=\\\\\"\*\*\*\*\\\\\"" -e "=hidden" || true
+  # JWTs; `eyJhbGc` is a base64-endcoded JSON object that starts with `{"alg`
```
moritzkiefer-da, line 102: you get no guarantee that jwts are sorted alphabetically though …

martinflorian-da: Yup; seems common enough to still check IMO, in case there isn't a `Bearer` in front.

And also, importantly: I find it reasonably unlikely that we'll get false-positives with this.

PR #3177, opened 2025-11-11
Avoid using the same database in tests that start a local validator participant

martinflorian-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/6394

...but also does some refactoring to make it more ergonomic and (hopefully) a bit less error-prone to write tests that use a throwaway validator participant.

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorProxyIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorProxyIntegrationTest.scala
@@ -45,7 +45,7 @@ class ValidatorProxyIntegrationTest
       withProperties(props) {
         withCanton(
           Seq(
-            testResourcesPath / "standalone-participant-extra.conf",
+            testResourcesPath / "standalone-participant-extra.conf"
           ),
           Seq.empty,
           "validator-proxy-test",


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletSurviveCantonRestartIntegrationTest.scala
```diff
@@ -1,80 +1,64 @@
...
-      "canton.participants.validatorParticipant.admin-api.port=7502",
-      "canton.participants.validatorParticipant.sequencer-client.use-new-connection-pool=true",
-    )
-
-  override protected def extraPortsToWaitFor: Seq[(String, Int)] = Seq(
-    ("ParticipantLedgerApi", 7501),
-    ("ParticipantAdminApi", 7502),
-  )
+  // Can sometimes be unhappy when doing funky `withCanton` things; disabling them for simplicity
+  override protected def runTokenStandardCliSanityCheck: Boolean = false
```
moritzkiefer-da, line 24: I don't see that we disabled this so far, why do we need to disable it now?

martinflorian-da: The comment is my honest answer: It failed once in a local test run, and I timeboxed trying to understand why... It could be that the validator app became too unhappy because its participant was down at the end of the test.

The other test I'm fixing here also disabled both of these things; seemed to me to be a good reference for future "local participant" tests as these tests are hard enough to get right even without the extra hidden checks...

moritzkiefer-da: hm seems kind of surprising that it never failed before but now it does fail but I guess I don't care enough

martinflorian-da: > hm seems kind of surprising that it never failed before but now it does fail but I guess I don't care enough

I agree. My best hypotheses are:

- somehow the wiring with the `aliceValidatorLocal` config (instead of `aliceValidator` with rewrites) confused it
- I messed up something else in my local test run

...and I didn't/don't think that addressing either of these is a good use of time.

Issue #1133, opened 2025-06-16
Take a periodic topology snapshot from Dev/Test/Main for Canton tests to consume


isegall-da:
A couple of months ago we took topology snapshots from Dev/Test/Main, which Canton checked-in, and are now testing imports from in their CI. Would be good to periodically update that with a fresh snapshot.

moritzkiefer-da:
We've just moved this up in priority so hopefully soonish yes

meiersi-da:
Side note: @isegall-da with our work on cold storage for Scan being served from GCS, we can think about implementing serving topology transactions on scan, and then the corresponding cold storage would correspond quite directly to the snapshots. The difference being that they would just batch the topology transactions.

We could even just do a first option, where Scan just regularly creates these snapshots and serves URLs to them.

moritzkiefer-da:
Not sure I'll find the time to tackle it before my vacation so mainly just jumping the steps here.

PR #3188, opened 2025-11-12
Add trigger for periodic topology snapshots

julientinguely-da:
fixes https://github.com/hyperledger-labs/splice/issues/1133

test bucket: https://console.cloud.google.com/storage/browser/da-splice-topology-snapshots;tab=objects?forceOnBucketsSortingFiltering=true&project=da-cn-splice&prefix=&forceOnObjectsSortingFiltering=false 

The plan would be to activate it on sv1 in our production clusters (will be available from 0.5.3).

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


cluster/pulumi/common/src/topology-snapshot.ts
```diff
@@ -0,0 +1,16 @@
...
+import {
+  config,
+  BackupConfig,
+  bootstrapDataBucketSpec,
+} from '@lfdecentralizedtrust/splice-pulumi-common';
+
+export async function topologySnapshotConfig(prefix: string): Promise<BackupConfig> {
+  const bucketSpec = await bootstrapDataBucketSpec(
+    config.optionalEnv('TOPOLOGY_SNAPSHOT_PROJECT') || 'da-cn-devnet',
+    config.optionalEnv('TOPOLOGY_SNAPSHOT_BUCKET') || 'da-cn-topology-snapshots'
```
martinflorian-da: Let's add these to the config schema! We want to avoid adding new env vars whenever it's reasonable...

cluster/pulumi/common/src/topology-snapshot.ts
```diff
@@ -0,0 +1,16 @@
...
+  );
+
+  return { backupInterval: '2h', location: { bucket: bucketSpec, prefix: prefix } };
```
martinflorian-da: Let's also make the backupInterval configurable in the config.yaml schema. Seems cheap enough to do, and if we need it we'll already have it.

cluster/pulumi/common-sv/src/config.ts
```diff
@@ -84,6 +84,8 @@ export interface SvConfig extends StaticSvConfig, SingleSvConfiguration {
   onboardingPollingInterval?: string;
   cometBftGovernanceKey?: CnInput<SvCometBftGovernanceKey>;
   initialRound?: string;
+  periodicTopologySnapshot?: boolean;
```
martinflorian-da: Doesn't the existence of a `periodicTopologySnapshotConfig` already imply true/false here? Why the extra variable?

cluster/helm/splice-sv-node/templates/sv.yaml
martinflorian-da, line 1: It looks right to me but I would probably have added a small Helm test for this. (Don't mind it if you think it's not necessary.)

cluster/deployment/config.yaml
```diff
@@ -151,6 +151,8 @@ svs:
           memory: '12Gi'
         limits:
           memory: '18Gi'
+    svApp:
+      periodicTopologySnapshot: false
```
martinflorian-da: Suggestion: How about using this here for all defaults that you currently have in pulumi? 

And making something like `enabled` one of the parameters of (what would have to become) something like `periodicTopologySnapshotConfig`.

Actually I think we can make this a bit more future-proof by having a config path like `svApp.periodicSnapshots.topology`. This way it's clear how to add ACS later on, for example.

test-full-class-names.log
```diff
@@ -15,6 +15,7 @@ org.lfdecentralizedtrust.splice.integration.tests.ExternalPartySetupProposalInte
 org.lfdecentralizedtrust.splice.integration.tests.ExternallySignedPartyOnboardingTest
 org.lfdecentralizedtrust.splice.integration.tests.FeaturedAppActivityMarkerIntegrationTest
 org.lfdecentralizedtrust.splice.integration.tests.GcpBucketPeriodicBackupIntegrationTest
+org.lfdecentralizedtrust.splice.integration.tests.GcpBucketPeriodicTopologySnapshotIntegrationTest
```
martinflorian-da: So for the bucket where the integration tests write to: Let's make sure it gets garbage collected!

So I think you want [this page here](https://console.cloud.google.com/storage/browser/da-splice-topology-snapshots;tab=lifecycle?forceOnBucketsSortingFiltering=true&project=da-cn-splice&prefix=&forceOnObjectsSortingFiltering=false)

To look like for the identities dump bucket:

image

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
```diff
@@ -81,6 +83,25 @@ class SequencerAdminConnection(
     ).flatMap(_ => responseObserver.resultFuture.map(_.map(_.chunk)))
   }
 
+  def getSummaryOfTransactions(store: TopologyStoreId)(implicit
```
martinflorian-da:
```suggestion
  def getTopologyTransactionsSummary(store: TopologyStoreId)(implicit
```

Mostly "transactions" can mean many things.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PeriodicTopologySnapshotIntegrationTest.scala
```diff
@@ -0,0 +1,70 @@
...
+          within = {
+            initDsoWithSv1Only()
+          },
+          assertion = logEntries => {
+            forAtLeast(
+              1,
+              logEntries,
+            )(logEntry =>
+              logEntry.message should (include(s"Took a new topology snapshot for $utcDate") or
+                include("Today's topology snapshot already exists."))
```
martinflorian-da, line 57: Hm so for identity dumps we seem to be fine with taking multiple per day - why limit ourselves on topology snapshots?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
```diff
@@ -0,0 +1,111 @@
...
+            BackupDump.write(
+              config.location,
+              Paths.get(s"$folderName/authorized"),
+              authorizedStore.toStringUtf8,
+              loggerFactory,
+            ),
+            BackupDump.write(
+              config.location,
+              Paths.get(s"$folderName/transactions-summary"),
+              summary.toString,
```
martinflorian-da: I wonder: Shall we also add the sequencer ID here? Might be informative, and that file is only for human consumption anyway.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
```diff
@@ -0,0 +1,111 @@
...
+      folderName: String,
+      utcDate: String,
+      synchronizerId: SynchronizerId,
+  )(implicit traceContext: TraceContext): Future[TaskSuccess] =
+    for {
+      sequencerId <- sequencerAdminConnection.getSequencerId
+      onboardingState <- sequencerAdminConnection.getOnboardingState(sequencerId)
+      authorizedStore <- sequencerAdminConnection.exportAuthorizedStoreSnapshot(sequencerId.uid)
+      // list a summary of the transactions state at the time of the snapshot to validate further imports
+      summary <- sequencerAdminConnection.getSummaryOfTransactions(
```
martinflorian-da: Don't you have a race here? You want the summary of transactions at exactly the time where the snapshots were taken.

Probably also want to add an appropriate TimeQuery the authorized store snapshot if possible.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  private def takeTopologySnapshot(
+      sequencerAdminConnection: SequencerAdminConnection,
+      folderName: String,
+      utcDate: String,
+      synchronizerId: SynchronizerId,
+  )(implicit traceContext: TraceContext): Future[TaskSuccess] =
+    for {
+      sequencerId <- sequencerAdminConnection.getSequencerId
+      onboardingState <- sequencerAdminConnection.getOnboardingState(sequencerId)
```
martinflorian-da: Huh don't we want `getGenesisState` here?

This mistake makes me think that you might want to test a topology import from the export that you're doing here; one time manually should be more than enough.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
```diff
@@ -0,0 +1,111 @@
...
+      _ <- Future {
+        blocking {
+          val fileDesc =
+            s"dumping current topology state into gcp bucket"
+          logger.debug(s"Attempting to write $fileDesc")
+          val paths = Seq(
+            BackupDump.write(
+              config.location,
+              Paths.get(s"$folderName/genesis-state"),
+              onboardingState.toStringUtf8,
```
martinflorian-da: We can't just dump bytes to make it smaller and more easier to use/import?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
martinflorian-da: Copy pasting what we did last time on the console:

```
import com.digitalasset.canton.data.CantonTimestamp
val now = CantonTimestamp.now()
val genesisState = sequencer.topology.transactions.genesis_stateV2(None, Some(now))
sequencer.topology.transactions.list(store = sequencer.synchronizer_id, proposals = false, timeQuery = TimeQuery.Snapshot(now)).result.groupMapReduce(_.mapping.code)(_ => 1)(_ + _)
genesisState.writeTo(new java.io.FileOutputStream("/tmp/state-export/genesis-state"))
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PeriodicTopologySnapshotIntegrationTest.scala
martinflorian-da: OK I found the explanation for this now on the issue...

Let's add a comment about our reasoning here though (to the trigger code).

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
martinflorian-da: OK I read about it now on the issue...

So do you understand how to test an import from this? Would we also need keys from the sequencer (probably not?)?

And why do we actually need the authorized store? (I'd avoid dumping something we don't know anyone will use.)

martinflorian-da: Oh and aren't you missing this part of Moritz's design?

> stream the snapshot so we never hold the full one in memory.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
martinflorian-da: Maybe also the specific timestamp of the dump? The bucket has a granularity of "days", might be nice to quickly see what was the time more specifically.

cluster/pulumi/common/src/topology-snapshot.ts
martinflorian-da: Maybe also add a comment to that config option that it will produce a dump at most once per day, even if you set the interval to <24h

cluster/deployment/config.yaml
julientinguely-da: sounds good, included your suggestion and will follow up with https://github.com/hyperledger-labs/splice/issues/3224 later

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
julientinguely-da: > OK I read about it now on the issue...

Sorry for that, should better inline these comments in the code next time

> Would we also need keys from the sequencer (probably not?)? 

I've discussed it with Moritz last week, his exact words: 
`Authorized store makes sense as well I guess, I'd probably lean towards not exporting keys. Seems dodgy from a security pov.`

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
julientinguely-da: > Shall we also add the sequencer ID here? Might be informative, and that file is only for human consumption anyway.

Isn't it always the same, and we can get it from the sv ui info tab? 

> Maybe also the specific timestamp of the dump? 

I guess we could, but can also easily see when the file was written. A bit annoying because we have this mechanism to check if today's backup already exists before taking one, it's done by name and `gcpBucket.fileExists(fileName)` uses string and not regex..

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
julientinguely-da: > Don't you have a race here? 

I don't believe so, what makes you think that?

> Probably also want to add an appropriate TimeQuery the authorized store snapshot if possible. 

good catch!

test-full-class-names.log
julientinguely-da: good catch, done: 
Screenshot from 2025-11-19 11-04-33

cluster/helm/splice-sv-node/templates/sv.yaml
julientinguely-da: good point, adding one now

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
julientinguely-da: Actually good point, I think we can.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
martinflorian-da: > I don't believe so, what makes you think that?

I mean just that: you take snapshot at time X, then 3 seconds later you print your summary, but now your summary doesn't match your snapshot.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
martinflorian-da: But anyway I think we want to try to somehow do streaming here... holding the topology and acs in memory will add significant memory pressure on our live nodes...

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
martinflorian-da: > Isn't it always the same, and we can get it from the sv ui info tab?

Yeah, but it also seemed trivial to add, and might help the Canton team when they need to consume these dumps. Note that MainNet ID will hopefully never change, but TestNet and DevNet will change on each reset.

martinflorian-da: > easily see when the file was written

Not so easy in general; requires talking to GCP instead of just looking at a file that you might have gotten from some other process.

My point is just: it's easy to add, don't see a downside but potential upsides.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
martinflorian-da: (I'm talking about the "stream the snapshot so we never hold the full one in memory." part of the design.)

julientinguely-da: on it

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
julientinguely-da: I just added it along the transaction history and renamed the file to metadata. Added a comment.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
```diff
@@ -81,16 +103,139 @@ class SequencerAdminConnection(
...
     ).flatMap(_ => responseObserver.resultBytes)
   }
 
+  /** Streams onboarding state from the gRPC admin service directly to a bucket without writing to memory
+    */
+  def streamOnboardingState(
+      sequencerIdOrTimestamp: Either[SequencerId, CantonTimestamp],
+      backupDumConfig: BackupDumpConfig,
+      fileName: String,
+  )(implicit
```
julientinguely-da, line 140: @martinflorian-da, I'm done processing your review. Do you want to have a final look? Or ok for me to merge it?
The plan is to enable it on ci in the followup PR.

apps/app/src/main/resources/application.conf
```diff
@@ -1,4 +1,4 @@
 # This is deliberately lower than ConsoleCommandTimeout.defaultRequestTimeout
 pekko.http.server.request-timeout = 38 seconds
 pekko.http.server.parsing.error-handler="org.lfdecentralizedtrust.splice.http.PekkoHttpParsingErrorHandler$"
-pekko.http.server.parsing.ignore-illegal-header-for = ["user-agent"]
+pekko.http.server.parsing.ignore-illegal-header-for = ["user-agent", "etag"]
```
martinflorian-da, line 4: oh boy... glad you figured that out!

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PeriodicTopologySnapshotTest.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  protected val topologySnapshotInterval: NonNegativeFiniteDuration =
+    NonNegativeFiniteDuration.ofMinutes(10)
+
+  protected def topologySnapshotConfig: PeriodicBackupDumpConfig =
+    PeriodicBackupDumpConfig(topologySnapshotLocation, topologySnapshotInterval)
+
+  protected def topologySnapshotLocation: T
+
+  protected def readDump(filename: String): String
```
martinflorian-da: Where is that used? :man_with_probing_cane:

... Or could this test be missing a check for "the dump actually exists on Gcp now" ... :)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
```diff
@@ -81,16 +102,138 @@ class SequencerAdminConnection(
...
+          filterSigningKey = "",
+          protocolVersion = None,
+        ),
+        filterNamespace = "",
+        excludeMappings = Seq.empty,
+      )
+    ).map(_.result.groupMapReduce(_.mapping.code)(_ => 1)(_ + _))
+  }
+
+  def getOnboardingState(sequencerIdOrTimestamp: Either[SequencerId, CantonTimestamp])(implicit
```
martinflorian-da, line 124: I find that use of `Either` quite a bit weird, but since it's just what Canton uses ... I guess I'm not allowed to complain?

Except maybe I can complain that you touch this line at all; do you need that change? (nit-level)

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
```diff
@@ -0,0 +1,138 @@
...
+        config.location,
+        Paths.get(s"$folderName/genesis-state").toString,
+      )
+      authorizedStore <- sequencerAdminConnection.exportAuthorizedStoreSnapshot(sequencerId.uid)
+      // list a summary of the transactions state at the time of the snapshot to validate further imports
+      summary <- sequencerAdminConnection.getTopologyTransactionsSummary(
+        TopologyStoreId.Synchronizer(synchronizerId),
+        clock.now,
+      )
+      sequencerId <- sequencerAdminConnection.getSequencerId
```
martinflorian-da:
```suggestion
```

That shouldn't change that quickly :) (dup to line 94)

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/PeriodicTopologySnapshotTrigger.scala
```diff
@@ -0,0 +1,138 @@
...
+      synchronizerId: SynchronizerId,
+  )(implicit
+      traceContext: TraceContext,
+      esf: ExecutionSequencerFactory,
+      actorSystem: ActorSystem,
+  ): Future[TaskSuccess] =
+    for {
+      sequencerId <- sequencerAdminConnection.getSequencerId
+      // uses onboardingStateV2 so we don't lose information when exporting
+      storageObject <- sequencerAdminConnection.streamOnboardingState(
```
martinflorian-da: Can we somehow have a log line once we start streaming this and a log line once we're done? Only if it's cheap...

I also suggest making the logs here INFO; at once per day that seems non-verbose enough, and having these logs might be useful for understanding timing? (Only medium confidence in this comment.)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
julientinguely-da: Not sure here honestly; to my understanding we need to specify the timestamp to match with: 
```
import com.digitalasset.canton.data.CantonTimestamp
val now = CantonTimestamp.now()
val genesisState = sequencer.topology.transactions.genesis_stateV2(None, Some(now))
sequencer.topology.transactions.list(store = sequencer.synchronizer_id, proposals = false, timeQuery = TimeQuery.Snapshot(now)).result.groupMapReduce(_.mapping.code)(_ => 1)(_ + _)
genesisState.writeTo(new java.io.FileOutputStream("/tmp/state-export/genesis-state"))
```

julientinguely-da:
> Maybe I can have my last 20% once you add one more check to your test?

Thanks again for the reviews! I processed the last ones, extended the test to also include a topology import (not sure if it does anything though). Please go ahead with the rest. :pray:

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
martinflorian-da: I mean in the signature of the method defined on this line. Are we using this method anywhere else?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PeriodicTopologySnapshotTest.scala
```diff
@@ -0,0 +1,98 @@
...
+        dumps.exists(_.getName.endsWith("authorized")) shouldBe true
+        dumps
+          .find(_.getName.endsWith("onboarding-state"))
+          .getOrElse(throw new RuntimeException("Onboarding state dump not found."))
+          .getContent()
+      })
+
+      clue("the topology snapshot import works.")({
+        sv1Backend.appState.localSynchronizerNode.value.sequencerAdminConnection
+          .importTopologySnapshot(
```
martinflorian-da: Hmm importing into the already initialized sequencer seems like it should be a NOP; in any case something we should never want to do.

How about starting a fresh sequencer here with `withCanton`?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
julientinguely-da: yes, in one other place and are passing the sequencerId

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PeriodicTopologySnapshotTest.scala
```diff
@@ -0,0 +1,98 @@
...
+import org.lfdecentralizedtrust.splice.integration.EnvironmentDefinition
+import org.lfdecentralizedtrust.splice.integration.tests.SpliceTests.IntegrationTest
+import org.lfdecentralizedtrust.splice.util.GcpBucket
+import org.slf4j.event.Level
+
+import java.nio.charset.StandardCharsets
+import java.time.{ZoneOffset, ZonedDateTime}
+import scala.concurrent.duration.DurationInt
+
+abstract class PeriodicTopologySnapshotIntegrationTestBase[T <: BackupDumpConfig]
```
martinflorian-da: Actually: Why did you add an `abstract class` in the first place, as opposed to just the single `GcpBucketPeriodicTopologySnapshotIntegrationTest`? So we're prepared in case we want to support a different backend some day?

If this is the reason: I think I'm leaning toward making the code as simple as possible for today's requirements, and worrying about generalizing this only once we need to. (~nit)

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PeriodicTopologySnapshotTest.scala
julientinguely-da: Seems a bit too much extra work honestly, could not find easy way to export and import private keys.  I will go for the more straightforward approach, testing it manually once. Please let me know if you disagree :)

julientinguely-da: The manual test worked. Just needed to use `globalSequencerSv1.setup.assign_from_onboarding_stateV2` instead of `globalSequencerSv1.setup.assign_from_genesis_stateV2`.

martinflorian-da: > could not find easy way to export and import private keys

Do you need that? Can't you just initialize it with the right synchronizer ID, i.e., have it use fresh keys?

But happy to time box this.

julientinguely-da: > Do you need that? Can't you just initialize it with the right synchronizer ID, i.e., have it use fresh keys? 

Not sure here, I believe the onboarding state is encrypted with the keys it was generated from. Happy to sync on how you'd do it

julientinguely-da: Screenshot from 2025-11-27 16-09-56

PR #3195, opened 2025-11-13
Fix restore of users without rights

martinflorian-da:
Happened to validators during DevNet migration

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DecentralizedSynchronizerMigrationIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DecentralizedSynchronizerMigrationIntegrationTest.scala
@@ -1416,7 +1416,12 @@ class DecentralizedSynchronizerMigrationIntegrationTest
             false,
           )
           participant.ledger_api.users.create(
-            s"fake-user-4-${suffix}"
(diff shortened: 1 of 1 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DecentralizedSynchronizerMigrationIntegrationTest.scala
```diff
@@ -1415,6 +1415,9 @@ class DecentralizedSynchronizerMigrationIntegrationTest
             "",
             false,
           )
+          participant.ledger_api.users.create(
```
martinflorian-da, line 1418: I confirmed that the test failed on this as expected, before doing the actual fix.

moritzkiefer-da, line 1418: how about testing one user with no rights and a primary partyt and one with no rights and no primary party? Just seems like another thing we can mess up relatively easily

martinflorian-da: sure, adding users here is cheap

PR #3197, opened 2025-11-13
Adding Scan Connection persistence to validator

pasindutennage-da:
Adding support for validator to save the scan urls in the DB

Fixes https://github.com/issues/assigned?issue=DACH-NY%7Ccanton-network-internal%7C451

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
@@ -870,7 +870,7 @@ object BftScanConnection {
     protected val initialScanConnections: Seq[SingleScanConnection]
     protected val initialFailedConnections: Map[Uri, Throwable]
     protected val connectionBuilder: Uri => Future[SingleScanConnection]
-    protected val refreshScanUrlsCallback: Option[Seq[DsoScan] => Future[Unit]]
+    protected val refreshScanUrlsCallback: Option[Seq[(String, String)] => Future[Unit]]
     protected val getScans: BftScanConnection => Future[Seq[DsoScan]]
     val scansRefreshInterval: NonNegativeFiniteDuration
     val retryProvider: RetryProvider
@@ -945,7 +945,11 @@ object BftScanConnection {
 
         filteredScans = filterScans(scansInDsoRules)
 
-        _ = refreshScanUrlsCallback.map(f => f(filteredScans))
+        dsoScanSeq: Seq[(String, String)] = filteredScans.map(scan =>
+          (scan.svName, scan.publicUrl.toString)
+        )
+
+        _ = refreshScanUrlsCallback.map(f => f(dsoScanSeq))
 
         newState <- computeNewState(retriedCurrentState, filteredScans)
       } yield {
@@ -1066,7 +1070,7 @@ object BftScanConnection {
       override val initialScanConnections: Seq[SingleScanConnection],
       override val initialFailedConnections: Map[Uri, Throwable],
       override val connectionBuilder: Uri => Future[SingleScanConnection],
-      protected val refreshScanUrlsCallback: Option[Seq[DsoScan] => Future[Unit]],
+      protected val refreshScanUrlsCallback: Option[Seq[(String, String)] => Future[Unit]],
       override val getScans: BftScanConnection => Future[Seq[DsoScan]],
       override val scansRefreshInterval: NonNegativeFiniteDuration,
       override val retryProvider: RetryProvider,
@@ -1093,7 +1097,7 @@ object BftScanConnection {
       override val initialScanConnections: Seq[SingleScanConnection],
       override val initialFailedConnections: Map[Uri, Throwable],
       override val connectionBuilder: Uri => Future[SingleScanConnection],
-      protected val refreshScanUrlsCallback: Option[Seq[DsoScan] => Future[Unit]],
+      protected val refreshScanUrlsCallback: Option[Seq[(String, String)] => Future[Unit]],
       override val getScans: BftScanConnection => Future[Seq[DsoScan]],
       override val scansRefreshInterval: NonNegativeFiniteDuration,
       override val retryProvider: RetryProvider,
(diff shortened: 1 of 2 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1172,22 +1176,33 @@ object BftScanConnection {
...
+      lastPersistedScanUrlList: Option[Seq[DsoScan]],
+      persistScanUrlCallback: Option[Seq[DsoScan] => Future[Unit]],
   )(implicit
       ec: ExecutionContextExecutor,
       tc: TraceContext,
       mat: Materializer,
   ): Future[BftScanConnection] = {
     val logger = loggerFactory.getTracedLogger(getClass)
+
+    val persistedUris: NonEmptyList[Uri] = lastPersistedScanUrlList match {
```
martinflorian-da:
```suggestion
    val bootstrapUris: NonEmptyList[Uri] = lastPersistedScanUrlList match {
```

(turbo-)nit

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -870,6 +870,7 @@ object BftScanConnection {
     protected val initialScanConnections: Seq[SingleScanConnection]
     protected val initialFailedConnections: Map[Uri, Throwable]
     protected val connectionBuilder: Uri => Future[SingleScanConnection]
+    protected val persistScanUrlCallback: Option[Seq[DsoScan] => Future[Unit]]
```
```suggestion
    protected val refreshScanUrlsCallback: Option[Seq[DsoScan] => Future[Unit]]
```

the fact that we persist them as part of that callback doesn't matter here

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1172,22 +1176,33 @@ object BftScanConnection {
...
-      seedUrls: NonEmptyList[Uri],
+      seedUris: NonEmptyList[Uri],
       amuletRulesCacheTimeToLive: NonNegativeFiniteDuration,
       spliceLedgerClient: SpliceLedgerClient,
       scansRefreshInterval: NonNegativeFiniteDuration,
       clock: Clock,
       retryProvider: RetryProvider,
       loggerFactory: NamedLoggerFactory,
       builder: (Uri, NonNegativeFiniteDuration) => Future[SingleScanConnection],
+      lastPersistedScanUrlList: Option[Seq[DsoScan]],
```
nicu-da, line 1191: This shouldn't leak here, it should just be passed as the seed urls.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
```diff
@@ -199,6 +206,14 @@ class ValidatorApp(
       initialSynchronizerTime <-
         withParticipantAdminConnection { participantAdminConnection =>
           for {
+            _ <- Future.unit
```
nicu-da: this is a bit confusing, I would just move the internalStore definition outside the for in this case

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
```diff
@@ -683,6 +700,45 @@ class ValidatorApp(
...
+
+  private def getPersistedScanList(
+      validatorConfigProvider: ValidatorConfigProvider
+  )(implicit traceContext: TraceContext): Option[Seq[DsoScan]] = {
+    val futureOption: Future[Option[Seq[ScanUrlInternalConfig]]] =
+      validatorConfigProvider.getScanUrlInternalConfig().value
+
+    val maybeInternalConfigs: Option[Seq[ScanUrlInternalConfig]] =
+      try {
+        Await.result(futureOption, 10.seconds) // what is the max waiting time?
```
nicu-da: We should never wait for futures in non test code. We should always include them in a for comprehension as we already run everything more or less in an async context

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
```diff
@@ -683,6 +700,45 @@ class ValidatorApp(
...
+      } catch {
+        case e: Throwable =>
```
nicu-da: Let's avoid catch all scenarios. 
Note: in scala catching throwable would also catch scala flow control exceptions such as return. Worst case scenario we should use catch NonFatal. Also we should never silently ignore exceptions

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Resolved: Used refreshScanUrlsCallback

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Resolved: Used bootstrapUris

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Resolved: Passed as List[String, String] in apply, and then later converted to DSO

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
pasindutennage-da: Resolved, moved internal store outside the for

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
pasindutennage-da: Thanks

Used for comprehensions

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
pasindutennage-da: Thanks

Since I removed the Await, i no longer have this catch clause

Issue #2694, opened 2025-10-14
Improve detection of ACS import-related failure in reonboard from key


martinflorian-da:
Latest context: https://digitalasset.atlassian.net/browse/DA-2467

The problem with this flow is that it is a bit brittle. We move the party, then attempt to import the ACS. If we fail now and restart, on the next run we see that the party was migrated and don’t attempt to re-import the ACS again.

The best idea about this I have: Check the size of the ACS on the participant. If it's empty, at least log a warning, at least if it's the validator party. Perhaps also just try resume the reonboard from key in such cases (perhaps make this controllable via a config flag).

nicu-da:
> Out of curiosity. What is the context and problem that you have?

During the reonboarding we unhost the parties from the old node, and then import the acs into the new node. if the process fails at that step, we check what parties to migrate by listing them from the old node, but then we're already toast as no parties is left there so we won't migrate anything. but this is purely in our apps, not canton

PR #3201, opened 2025-11-13
recovery from key - store the parties being migrated in the local db for recovery in case of failure

nicu-da:
This ensures the process is resilient to a restart after we unhosted the parties from the old node

fixes https://github.com/hyperledger-labs/splice/issues/2694

Tried to add a unit test but mocking the participant admin connection is a full on terror 
testing in the integration tests that we fail at the right time it's also a pain 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
+++ b/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
@@ -94,6 +94,8 @@ import org.lfdecentralizedtrust.splice.wallet.util.ValidatorTopupConfig
 import org.lfdecentralizedtrust.splice.wallet.{ExternalPartyWalletManager, UserWalletManager}
 import org.lfdecentralizedtrust.tokenstandard.allocation.v1.Resource as TokenStandardAllocationResource
 import org.lfdecentralizedtrust.tokenstandard.allocationinstruction.v1.Resource as TokenStandardAllocationInstructionResource
+import org.lfdecentralizedtrust.tokenstandard.metadata.v1.Resource as TokenStandardMetadataResource
+import org.lfdecentralizedtrust.tokenstandard.transferinstruction.v1.Resource as TokenStandardTransferInstructionResource
 
 import scala.concurrent.{ExecutionContextExecutor, Future}
 import scala.util.{Failure, Success}


apps/validator/src/test/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStoreTest.scala
nicu-da, line 1: I have no idea why it shows the files as deleted and then added, as the commits contain the rename separately, but I just added two more tests and extended the BaseTest

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -4,70 +4,76 @@
...
-          None
-        }
-      }
-    }
+      """.as[Json].headOption,
+        "get-validator-internal-config",
+      )
+      .mapK(new ~>[FutureUnlessShutdown, Future] {
+        def apply[A](fa: FutureUnlessShutdown[A]): Future[A] =
+          FutureUnlessShutdownUtil.futureUnlessShutdownToFuture(fa)
```
pasindutennage-da: `futureUnlessShutdownToFuture` can be used implicitly right?

test-full-class-names-non-integration.log
```diff
@@ -43,4 +41,6 @@ org.lfdecentralizedtrust.splice.util.SpliceUtilTest
 org.lfdecentralizedtrust.splice.util.TransactionTreeExtensionsTest
 org.lfdecentralizedtrust.splice.util.ValueJsonCodecCodegenTest
 org.lfdecentralizedtrust.splice.util.ValueJsonCodecProtobufTest
+org.lfdecentralizedtrust.splice.validator.store.DbValidatorInternalStoreTest
```
pasindutennage-da, line 44: Do we put package names or the path in this file?

Note the package name was `package org.lfdecentralizedtrust.splice.store`

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
nicu-da: Ah good point, I missed that

test-full-class-names-non-integration.log
nicu-da: I've also updated the package, it should follow the same package as the class it's testing, which is also in splice.validator.store and it makes sense as it's the validator store.

btw: this file is auto generated, if you installed the pre commit hook it should always make sure this is up to date

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/FutureUnlessShutdownUtil.scala
```diff
@@ -22,6 +22,7 @@ object FutureUnlessShutdownUtil {
   )(implicit ec: ExecutionContext): OptionT[Future, A] = {
     OptionT(futureUnlessShutdownToFuture(f.value))
   }
+
```
martinflorian-da, line 25: Very important to have newlines! :stuck_out_tongue:

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/util/SpliceRateLimiterTest.scala
```diff
@@ -10,7 +10,6 @@ import org.apache.pekko.stream.Materializer
 import org.apache.pekko.stream.scaladsl.{Sink, Source}
 import org.apache.pekko.stream.testkit.StreamSpec
 import org.lfdecentralizedtrust.splice.admin.api.client.commands.HttpCommandException
-import org.lfdecentralizedtrust.splice.util.{SpliceRateLimitMetrics, SpliceRateLimiter}
```
martinflorian-da, line 13: Just going to pretend I understand how this passed static checks previously...

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/migration/ParticipantPartyMigrator.scala
```diff
@@ -95,7 +103,8 @@ class ParticipantPartyMigrator(
         case None =>
           logger.info(s"Importing ACS for party ids $partiesToMigrateFinal from scan")
           for {
-            _ <- importAcs(partiesToMigrateFinal, getAcsSnapshot)
+            _ <- importAcs(partiesToMigrateFinal.parties, getAcsSnapshot)
+            _ <- configProvider.clearPartiesToMigrate()
```
martinflorian-da, line 107: Right, makes sense... so I guess one of my testing ideas doesn't work then...

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorConfigProvider.scala
```diff
@@ -4,34 +4,90 @@
...
-    url: String,
-)
+import scala.concurrent.{ExecutionContext, Future}
 
-object ScanUrlInternalConfig {
-  implicit val scanUrlCodec: Codec[ScanUrlInternalConfig] = deriveCodec[ScanUrlInternalConfig]
-}
-
-class ValidatorConfigProvider(config: ValidatorInternalStore) {
+class ValidatorConfigProvider(config: ValidatorInternalStore, val loggerFactory: NamedLoggerFactory)
```
martinflorian-da, line 16:
```suggestion
class ValidatorDynamicConfigProvider(config: ValidatorInternalStore, val loggerFactory: NamedLoggerFactory)
```

Quite off-topic to this PR, more like a general "let's consider refactoring this" comment; to a casual reader it is hard to distinguish otherwise that this is not related to the (static) config from the config files.

(cc @pasindutennage-da )

apps/validator/src/test/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStoreTest.scala
martinflorian-da: You moved it to a different directory? (I'm assuming that this was intentional; it looks fine to me.)

nicu-da: Still git should show it as moved not deleted and re-added

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/migration/ParticipantPartyMigrator.scala
nicu-da: Well we can't test in integration tests because we sys.exit if the app fails to start. I will do a manual test

martinflorian-da: > we sys.exit if the app fails to start

Ugh right, we don't have info to "assert-fails-to-start". Meh.

> I will do a manual test

Thanks! The code looks perfect to me but it seems like the right thing to do.

apps/validator/src/test/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStoreTest.scala
martinflorian-da: Yeah probs it changed slightly too much or something like that. :shrug:

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorReonboardingIntegrationTest.scala
nicu-da, line 1: @martinflorian-da can you take another look? Added a test for the new functionality. In theory there's a slight chance that it will be flaky but I am confident enough that the startup is not fast enough for it to flake.

martinflorian-da: Ufff IDK, it does sound quite a bit shaky flaky to me... Things change, we'll forget what we did here, then we'll reorder app init order or change something else... and then go debug this.

I guess you run `start` where you do so it sets up the DB for you? Can't we just do that in the test, before running start?

Yes it seems a bit over the top for this particular PR, but the capability of initialing this store with arbitrary data before we start an app sounds like a handy thing to have also for other tests.

> I am confident enough

Not going to veto it but if it was my PR I think I'd lean towards not adding that test, even if the alternative was "I only tested it manually once".

nicu-da: > I guess you run start where you do so it sets up the DB for you? Can't we just do that in the test, before running start?

We could but it's not trivial

> Not going to veto it but if it was my PR I think I'd lean towards not adding that test, even if the alternative was "I only tested it manually once".

honestly SGTM, will just remove the test, I was on the fence about it but then realized that our startup is real slow but then again that can always change and we shouldn't count on it

martinflorian-da: :+1:

...it's also a dangerous precedent... Not everyone's gut feeling is as well calibrated as yours, but they might still end up copying this approach...

PR #3214, opened 2025-11-17
Added store_id to validator_internal_config

pasindutennage-da:
Adding store_id to `validator_internal_config` to ensure isolation.

### DIFF
--- a/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
+++ b/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
@@ -93,7 +93,7 @@ class DbValidatorInternalStore(
     updateAction.map(_ => ())
   }
 
-  def getConfig[T](
+  override def getConfig[T](
       key: String
   )(implicit tc: TraceContext, decoder: Decoder[T]): OptionT[Future, Decoder.Result[T]] = {
 


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/StoreDescriptorManager.scala
```diff
@@ -0,0 +1,127 @@
...
+case class StoreHasData[StoreId](
+    storeId: StoreId,
+    lastIngestedOffset: Long,
+) extends InitializeDescriptorResult[StoreId]
+case class StoreHasNoData[StoreId](
+    storeId: StoreId
+) extends InitializeDescriptorResult[StoreId]
+case class StoreNotUsed[StoreId]() extends InitializeDescriptorResult[StoreId]
+
+object StoreDescriptorManager extends StoreErrors {
```
nicu-da: should we just name this `StoreDescripttorStore`? Sounds a bit weird but at the same time that's kind of what it is

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -4,40 +4,102 @@
...
 
   val profile: JdbcProfile = storage.profile.jdbc
+  private val storeDescriptor = StoreDescriptor(
+    version = 2,
+    name = "DbValidatorInternalConfigStore",
+    party = key.validatorParty,
+    participant = participantId,
+    key = Map(
+      "validatorParty" -> key.validatorParty.toProtoPrimitive,
+      "dsoParty" -> key.dsoParty.toProtoPrimitive,
```
nicu-da: Do we use the dso party for any other validator store descriptors?

apps/validator/src/test/scala/org/lfdecentralizedtrust/splice/store/ValidatorInternalStoreTest.scala
```diff
@@ -130,31 +131,31 @@ class DbValidatorInternalStoreTest
...
 
   override protected def timeouts = new ProcessingTimeout
-  override protected def loggerFactory: NamedLoggerFactory = NamedLoggerFactory.root
 
-  private def buildDbStore(): ValidatorInternalStore = {
-    val elc = ErrorLoggingContext(
-      loggerFactory.getTracedLogger(getClass),
-      loggerFactory.properties,
-      TraceContext.empty,
+  private def buildDbStore(): Future[ValidatorInternalStore] = {
```
nicu-da: we probably want a basic test that same key cross store descriptors doesn't conflict

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/StoreDescriptorManager.scala
pasindutennage-da: Resolved: Renamed to StoreDescripttorStore

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Yes, the existing DbValidatorStore uses the following.

```
 acsStoreDescriptor = StoreDescriptor(
        version = 2,
        name = "DbValidatorStore",
        party = key.validatorParty,
        participant = participantId,
        key = Map(
          "validatorParty" -> key.validatorParty.toProtoPrimitive,
          "dsoParty" -> key.dsoParty.toProtoPrimitive,
        ),
      ),
```

apps/validator/src/test/scala/org/lfdecentralizedtrust/splice/store/ValidatorInternalStoreTest.scala
pasindutennage-da: Resolved

Added a new test

```
"CONFIG_PROVIDER for two different store descriptors should not collide" in {
      for {
        provider1 <- mkProvider("alice")
        provider2 <- mkProvider("bob")
        _ <- provider1.setScanUrlInternalConfig(scanConfig1)
        _ <- provider2.setScanUrlInternalConfig(scanConfig2)
        retrievedConfigOption1 <- provider1.getScanUrlInternalConfig.value
        retrievedConfigOption2 <- provider2.getScanUrlInternalConfig.value
      } yield {
        retrievedConfigOption1 shouldBe Some(scanConfig1)
        retrievedConfigOption2 shouldBe Some(scanConfig2)
      }
    }
  }
```

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V052__update_scan_config.sql
```diff
@@ -0,0 +1,17 @@
+ALTER TABLE validator_internal_config
+    ADD COLUMN store_id integer;
+
+ALTER TABLE validator_internal_config
+DROP CONSTRAINT uc_validator_internal_config;
+
+ALTER TABLE validator_internal_config
+    ALTER COLUMN store_id SET NOT NULL;
```
rautenrieth-da, line 8: 1. You can only do this if the table is empty, which it won't be on CILR. Either leave the column optional (you can add a TODO to make it NOT NULL in the future), or truncate the table (should be fine since validator_internal_config currently only caches the scan urls).
2. You can include the NOT NULL in the ADD COLUMN statement

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V052__update_scan_config.sql
```diff
@@ -0,0 +1,17 @@
...
+
+ALTER TABLE validator_internal_config
+    ADD CONSTRAINT uc_validator_internal_config
+        PRIMARY KEY (config_key, store_id);
```
rautenrieth-da:
```suggestion
        PRIMARY KEY (store_id, config_key);
```
since we're more likely to filter by store_id only (e.g., for listing all keys) than by config_key only.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/StoreDescriptorStore.scala
```diff
@@ -0,0 +1,127 @@
...
+          "initializeDescriptor.2",
+        )
+        .getOrRaise(
+          new RuntimeException(s"No row for $descriptor found, which was just inserted!")
+        )
+
+      _ <- storage
+        .update(
+          sql"""
+             insert into store_last_ingested_offsets (store_id, migration_id)
```
rautenrieth-da: `store_last_ingested_offsets` has nothing to do with store descriptors. Please move the two queries back to DbMultiDomainAcsStore.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -47,18 +109,23 @@ class DbValidatorInternalStore(
     updateAction.map(_ => ())
   }
 
-  override def getConfig[T: Decoder](key: String)(implicit tc: TraceContext): OptionT[Future, T] = {
+  override def getConfig[T: Decoder](key: String): OptionT[Future, T] = {
```
rautenrieth-da: I'm suprised you can remove the trace context here, since you use `logger.debug` below - what trace context does that logging statement use?

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
rautenrieth-da: Why we include it in DbValidatorStore: The validator ACS store filters for the right DSO party because someone could create fake DsoRules and similar contracts with a random DSO party. In theory you could also run two different splice networks on the same participant, each with their own DSO party.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -4,40 +4,102 @@
...
+      result <- initializedResult
+    } yield {
+      result match {
+        case StoreHasNoData(storeId) =>
+          logger.info(
+            s"ValidatorInternalStore for validator party '${key.validatorParty}' and DSO party '${key.dsoParty}' initialized with store ID $storeId."
+          )
+          internalState.updateAndGet(_.copy(storeId = Some(storeId)))
+          ()
+        case _ =>
```
rautenrieth-da: `InitializeDescriptorResult` is a small sealed trait, I'd handle all cases explicitly.

The code also seems off. The only other type is `StoreHasData`, and that's not a failure, it should say something like "store resumed with store ID" instead.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
nicu-da: Right, so it doesn't really make sense to include it here as well, it just adds confusion

rautenrieth-da: Don't we need (or don't need) it for the same reason we have it in the ACS store descriptor? The point of the store descriptors is to distinguish between "different" instances of the same app, and `DbValidatorInternalStore` and `DbValidatorStore` should use the same notion for what is the "same" validator app.

The list of scan URLs is also derived from DsoRules, so the DbValidatorInternalStore does store data for a specific DSO.

pasindutennage-da: Resolved, Removed DsoParty from the Store descriptor

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V052__update_scan_config.sql
rautenrieth-da: > You can include the NOT NULL in the ADD COLUMN statement

So that's wrong, you need two different statements like you did.

pasindutennage-da: In summary, the existing code needs no modifications?

```
ALTER TABLE validator_internal_config
    ADD COLUMN store_id integer;   --- safe because it allows null values for existing rows

ALTER TABLE validator_internal_config
DROP CONSTRAINT uc_validator_internal_config; ---safe

ALTER TABLE validator_internal_config
    ALTER COLUMN store_id SET NOT NULL; --- applies only to existing rows-- hence safe

ALTER TABLE validator_internal_config
    ADD CONSTRAINT uc_validator_internal_config
        PRIMARY KEY (store_id, config_key); --- made store_id as the first 

ALTER TABLE validator_internal_config
    ADD CONSTRAINT fk_store_id
        FOREIGN KEY (store_id)
            REFERENCES store_descriptors(id); --- safe
```

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V052__update_scan_config.sql
pasindutennage-da: Resolved: Used PRIMARY KEY (store_id, config_key)

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: I used the trace context as a class param.

But after our discussion, I revoked, and used trace context in the methods.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V052__update_scan_config.sql
pasindutennage-da: Marking as resolved, after our discussion

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/StoreDescriptorStore.scala
pasindutennage-da: Resolved: Moved to DbMultiDomainAcsStore

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved: after refactoring, this logic is no longer available

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorConfigProvider.scala
```diff
@@ -8,10 +8,11 @@ import cats.implicits.{catsSyntaxOptionId, toBifunctorOps}
 import com.digitalasset.canton.logging.{NamedLoggerFactory, NamedLogging}
 import com.digitalasset.canton.topology.PartyId
```
pasindutennage-da, line 9: Unneeded changes, not sure how it happened, but nothing is breaking

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
```diff
@@ -28,18 +24,31 @@ trait ValidatorInternalStore {
       key: String
   )(implicit tc: TraceContext, decoder: Decoder[T]): OptionT[Future, Decoder.Result[T]]
 
+  def deleteConfig(key: String)(implicit
```
pasindutennage-da, line 29: unneeded change -- happened during merging. sorry

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -4,73 +4,138 @@
...
+) extends ValidatorInternalStore
     with AcsJdbcTypes
     with NamedLogging {
 
   val profile: JdbcProfile = storage.profile.jdbc
+  private val storeDescriptor = StoreDescriptor(
+    version = 2,
+    name = "DbValidatorInternalConfigStore",
+    party = key.validatorParty,
+    participant = ParticipantId.tryFromProtoPrimitive("PAR::dummy::dummy"),
```
pasindutennage-da: Because for ValidtorInternalConfig descriptor, only the validatorPartyId is enough.

pasindutennage-da: But looks a bit ugly -- should we used the actual participantId?

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V052__update_scan_config.sql
rautenrieth-da: I think you now need to truncate the table.

@nicu-da added code that actually writes to the new store in https://github.com/hyperledger-labs/splice/pull/3201, and that change has been deployed to CILR, so the table won't be empty. From a quick look at the PR, truncating the table should be fine as long as we don't do it while someone is in the middle of a migration.

@nicu-da sounds like we should get this change in before the upcoming release?

nicu-da: We should not have any data as we write just as a temp store during recovery from key and we don't run any such thing on cilr.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -4,73 +4,139 @@
...
+) extends ValidatorInternalStore
     with AcsJdbcTypes
     with NamedLogging {
 
   val profile: JdbcProfile = storage.profile.jdbc
+  private val storeDescriptor = StoreDescriptor(
+    version = 2,
+    name = "DbValidatorInternalConfigStore",
+    party = key.validatorParty,
+    participant = participant,
```
nicu-da, line 42: let's please remove the participant id, I don't see the need for duplication with the validator party

pasindutennage-da: I agree. It is there only to satisfy the existing store descriptor description.

Do you suggest to have a dummy participant? (something like "aaa::bbb")

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V052__update_scan_config.sql
pasindutennage-da: Can I conclude that there is no need to truncate, and that the current structure is fine?

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
rautenrieth-da: The participant id is used by the UpdateHistory store, and

1. We can't change the store descriptor for UpdateHistory stores or else we lose all data
2. It makes sense to include there, as UpdateHistory is the update stream from a given participant for a given party, and stores participant-local data (offsets). However stupid and unlikely it is, if someone has two validator apps with the same validator party connected to different participants, we don't want their data to mix. I think when I added the participant id I had an actual (not just theoretical) reason to do so, but I don't remember.

Same argument for the descriptors for ACS stores.

pasindutennage-da: @rautenrieth-da @nicu-da 

I made the participat as an optional arg -- hence existing tables do not brake + for the internal validator config we use the default dummy participant name (Par::name::name)

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V052__update_scan_config.sql
rautenrieth-da: > conclude that there is no need to truncate

This migration will crash and require manual database intervention if the `validator_internal_config` table is not empty.

From what I understand, it's currently only non-empty during recovery from key. So to break their app, someone would have to start recovery from key, crash in the middle of it, then upgrade the app, then resume recovery from key. And in that case it's probably better to require manual intervention than to silently truncate the database which would break the recovery process.

So I'd say yes, don't truncate, but make sure you merge this PR before something else starts to write to the ValidatorInternalStore.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/StoreDescriptorStore.scala
```diff
@@ -0,0 +1,101 @@
...
+  *                    to the update stream that feeds the store).
+  * @param participant The participant that serves the update stream that feeds this store.
+  * @param key         A set of named values that are used to filter the update stream or
+  *                    can otherwise be used to distinguish between different instances of the store.
+  */
+case class StoreDescriptor(
+    version: Int,
+    name: String,
+    party: PartyId,
+    participant: ParticipantId = ParticipantId.tryFromProtoPrimitive("PAR::name::dummy"),
```
rautenrieth-da:
```suggestion
    participant: Option[ParticipantId],
```
and adjust the description to something like
```
  * @param participant The participant that serves the data that feeds this store, or `None` if this store does not contain participant-specific data.
```
and serialize it as null/undefined if it's undefined

Although to be honest, I'd just leave this mandatory and use the real participant id:
- The data written by https://github.com/hyperledger-labs/splice/pull/3201/ comes from the participant, so not including the participant id is questionable
- Store descriptors are immutable. If we need the participant id in the future, we won't be able to add it retroactively

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V052__update_scan_config.sql
pasindutennage-da: Thanks.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/StoreDescriptorStore.scala
pasindutennage-da: Thanks,

Made participant ID field mandatory

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -4,73 +4,139 @@
...
+    internalState
+      .get()
+      .storeId
+      .getOrElse(throw new RuntimeException("Using storeId before it was assigned"))
+
+  def initializeState()(implicit tc: TraceContext): Future[Unit] = {
+    val initializedResult: Future[Int] = StoreDescriptorStore
+      .getStoreIdForDescriptor(storeDescriptor, storage)
+    for {
+      result <- initializedResult
```
rautenrieth-da: Out of interest, is there a reason why you didn't simply write
```
result <- StoreDescriptorStore.getStoreIdForDescriptor(storeDescriptor, storage)
```
here, which would avoid the need for the `val` above?

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -4,73 +4,139 @@
...
+
     logger.debug(
-      s"saving validator config in database with key '$key' and value '$json'"
+      s"saving validator config in database with key '$key' and value '$jsonValue'"
     )
     updateAction.map(_ => ())
   }
 
-  override def getConfig[T](
+  def getConfig[T](
```
rautenrieth-da: Why did you remove the `override` here? Unless I'm missing something, it should stay.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -4,73 +4,139 @@
...
+    val queryAction = sql"""SELECT config_value
+      FROM validator_internal_config
+      WHERE config_key = $key AND store_id = $storeId
+    """.as[Json].headOption
+
+    val jsonOptionT: OptionT[FutureUnlessShutdown, Json] =
+      storage.querySingle(queryAction, "get-validator-internal-config")
+
+    val resultOptionT: OptionT[FutureUnlessShutdown, Decoder.Result[T]] =
+      jsonOptionT.map { json =>
```
rautenrieth-da, line 111:
```scala
for {
  json <- storage.querySingle(queryAction, "get-validator-internal-config")
} yield {
  val decodeResult = json.as[T]
  // ...
}
```
if you want to avoid some of the vals.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorConfigProvider.scala
```diff
@@ -8,10 +8,11 @@ import cats.implicits.{catsSyntaxOptionId, toBifunctorOps}
...
 import com.digitalasset.canton.tracing.TraceContext
+import io.circe.Codec
```
rautenrieth-da, line 11: The changes in this file are unnecessary. If possible, try to avoid them as it only leads to noise in the git log and merge conflicts for people that happen to work on the same file.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
```diff
@@ -314,7 +314,24 @@ class ValidatorApp(
                     synchronizerId =>
                       packageVetting.vetCurrentPackages(synchronizerId, amuletRules)
                   }
-              } yield ()
+
+                participantId <- participantAdminConnection.getParticipantId()
+                idValidatorParty = ParticipantPartyMigrator.toPartyId(
```
rautenrieth-da, line 319: @nicu Can you please confirm that this is correct? I don't understand validator app initialization.

For the validator app store, we construct a `ValidatorStore.Key` at https://github.com/hyperledger-labs/splice/blob/main/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala#L732, with the validator party coming from https://github.com/hyperledger-labs/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/Node.scala#L76.

Does the party id on the participant always end up being the same as what we manually construct here? I assume this approach for manually constructing the validator party was copied from ParticipantPartyMigrator. Is the reason why ParticipantPartyMigrator is manually constructing the party id because the party migration needs to run in `preInitializeAfterLedgerConnection` where we din't wait for the actual validator party yet?

The way the code is written now, if a validator app operator changes `config.validatorPartyHint`, they would lose all data in the ValidatorInternalStore. Do people know that they must never change the hint?

And an unrelated note: `Node.waitForPartyBeforePreinitialize` is always false, can we remove it?

.gitignore
```diff
@@ -1,6 +1,7 @@
 .daml
 *.dar
 !daml/dars/*
+*.iml
```
rautenrieth-da, line 4: Is this intentional? I have never seen .iml files in git before.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
nicu-da: The hint usage should be alright, during normal init we actually always check if it's set https://github.com/hyperledger-labs/splice/blob/main/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala#L370, we also have checks that it never changes after https://github.com/hyperledger-labs/splice/blob/main/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala#L411 

I do agree that it's not ideal that we have to construct it beforehand, as in theory it could not exist yet, but this hints at issues in our init steps that are a big convoluted and not that easy to fix.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Thanks. Resolved with the suggestion

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved, added override

.gitignore
pasindutennage-da: Yes, Couple of time .imls were added when I committed, so felt safe to just add in in ignore

PR #3254, opened 2025-11-21
Persisting Scan Urls in the Validator App

pasindutennage-da:
# ✨ Feature: Persist Scan URLs for Validator App Crash Recovery

This PR introduces logic to persist the most recent set of scan connection URLs used by the `ValidatorApp`. This PR improves the security of the system during crash recovery.

Fixes https://github.com/DACH-NY/canton-network-internal/issues/451

## 💡 Context and Problem

Currently, the `ValidatorApp` initializes scan connections using a predefined set of **seed URLs**. After initialization, the app may update this set to include **all possible scan apps** discovered during operation.

In a **crash recovery scenario**, the `ValidatorApp` is forced to **re-bootstrap** using only the initial, potentially stale, set of seed URLs. If the correct, current scan URLs have changed over time, relying on the old seeds can introduce **security issues** or lead to an incorrect operational state.

## ✅ Solution

This PR addresses the issue by **persisting the most recent, known-good set of scan connections**.

* Upon recovery, the `ValidatorApp` will now prioritize using the **last persisted set of scan connections** instead of defaulting back to the initial seed URLs.

### Non-Breaking Change

* An **optional configuration setting** is added to the validator scan configuration to enable this persistence feature.
* By default, **scan persistence is turned on**.

## 📋 Checklist of Changes

* Added necessary changes to `ValidatorApp` to enable the persistence of scan connection URLs.
* Implemented required modifications within `BftScanConnection` to support persistence logic.
* Added a new **integration test** to cover the new persistence and recovery behavior.

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
@@ -194,34 +194,19 @@ class BftScanConnectionIntegrationTest
       sv4ScanBackend,
     )
 
-    eventually() {
-      val allHealthy = Seq(
-        sv1Backend,
-        sv1ScanBackend,
-        sv2Backend,
-        sv2ScanBackend,
-        sv3Backend,
-        sv3ScanBackend,
-        sv4Backend,
-        sv4ScanBackend,
-      ).forall { scan =>
-        scan.httpHealth.successOption.exists(_.active)
-      }
-      allHealthy shouldBe true
-    }
-
     loggerFactory.assertEventuallyLogsSeq(SuppressionRule.LevelAndAbove(Level.INFO))(
       {
         aliceValidatorBackend.startSync()
       },
       logs => {
-        val messages = logs.map(_.message)
+        val aliceValidatorLogs = logs.filter(_.loggerName.contains("validator=aliceValidator"))
+        val messages = aliceValidatorLogs.map(_.message)
         withClue("Validator should first bootstrap with 1 and then 4 scans") {
           messages.filter(_.contains(bootstrapsWith1UrlLog)) should have length 1
           messages.filter(_.contains(bootstrapsWith4UrlsLog)) should have length 1
         }.withClue(
           s"Actual Logs: \n ${messages.filter(_.contains(bootstrapsWith1UrlLog))} \n ${messages
-              .filter(_.contains(bootstrapsWith4UrlsLog))} "
+              .filter(_.contains(bootstrapsWith4UrlsLog))}"
         )
       },
     )
@@ -246,7 +231,8 @@ class BftScanConnectionIntegrationTest
         aliceValidatorBackend.startSync()
       },
       logs => {
-        val messages = logs.map(_.message)
+        val aliceValidatorLogs = logs.filter(_.loggerName.contains("validator=aliceValidator"))
+        val messages = aliceValidatorLogs.map(_.message)
         withClue(
           "Validator should bootstrap with all the scan urls persisted to the internal store"
         ) {


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,120 @@
...
+      sv1Backend,
+      sv1ScanBackend,
+    )
+
+    eventually() {
+      val allHealthy = Seq(sv1ScanBackend).forall { scan =>
+        scan.httpHealth.successOption.exists(_.active)
+      }
+      allHealthy shouldBe true
+    }
```
nicu-da: this should be guaranteed by the startAllSync already

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,120 @@
...
+    messages.exists(_.contains(s"Validator bootstrapping with scan URI: $uri"))
+  }
+
+  override def environmentDefinition: SpliceEnvironmentDefinition =
+    EnvironmentDefinition
+      .simpleTopology4Svs(this.getClass.getSimpleName)
+      .addConfigTransforms((_, config) =>
+        ConfigTransforms.updateAllValidatorConfigs {
+          case (name, c) if name == "aliceValidator" =>
+            val dbEnabledConfig = BftScanClientConfig.Bft(
```
nicu-da: Not sure I  fully follow here, where do we enable the config?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,120 @@
...
+import com.digitalasset.canton.{HasActorSystem, HasExecutionContext}
+import org.apache.pekko.http.scaladsl.model.Uri
+import org.lfdecentralizedtrust.splice.config.ConfigTransforms
+import org.lfdecentralizedtrust.splice.integration.EnvironmentDefinition
+import org.lfdecentralizedtrust.splice.integration.tests.SpliceTests.IntegrationTest
+import org.lfdecentralizedtrust.splice.scan.admin.api.client.BftScanConnection.BftScanClientConfig
+import org.lfdecentralizedtrust.splice.util.{SvTestUtil, WalletTestUtil}
+import org.slf4j.event.Level
+
+class InternalConfigEnabledBftScanConnectionIntegrationTest
```
nicu-da: Any chance we can change this to ValidatorScanConnectionsIntegrationTest and also move any other related tests from ValidatorIntegrationTest or other test classes?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,120 @@
...
+    extends IntegrationTest
+    with WalletTestUtil
+    with SvTestUtil
+    with HasExecutionContext
+    with HasActorSystem {
+
+  override protected def runEventHistorySanityCheck: Boolean = false
+  override protected def runUpdateHistorySanityCheck: Boolean = false
```
nicu-da: Why do we need to overwrite this?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  private def existsUrl(messages: Seq[String], uri: String) = {
+    messages.exists(_.contains(s"Validator bootstrapping with scan URI: $uri"))
+  }
```
nicu-da: Nit: I find it a lot more readable to have the private methods after the tests

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    loggerFactory.assertEventuallyLogsSeq(SuppressionRule.LevelAndAbove(Level.INFO))(
+      {
+        aliceValidatorBackend.startSync()
+      },
+      logs => {
+        val messages = logs.map(_.message)
+        withClue("Validator should bootstrap with only sv1 scan") {
+          existsUrl(messages, "http://localhost:5012") &&
```
nicu-da: lets use the forAtLeast/forExactly/forNo for all log message handling as it handles failure messages a lot better

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
```diff
@@ -0,0 +1,120 @@
...
+      },
+      logs => {
+        val messages = logs.map(_.message)
+        withClue("validator should eventually refresh the scan list") {
+          messages.exists(_.contains(s"Updated scan list with 4 scans:"))
+        } should be(true).withClue(s"Actual Logs: $logs")
+      },
+    )
+
+    loggerFactory.assertEventuallyLogsSeq(SuppressionRule.LevelAndAbove(Level.INFO))(
```
nicu-da: I think I'm missing the exact assertions where we see that the scan urls in the db are being updated and that the validator uses those during startup (note: you have access to the validator app state after startup and can access the actual config provider)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -870,6 +870,7 @@ object BftScanConnection {
     protected val initialScanConnections: Seq[SingleScanConnection]
     protected val initialFailedConnections: Map[Uri, Throwable]
     protected val connectionBuilder: Uri => Future[SingleScanConnection]
+    protected val persistScanUrlsCallback: Option[Seq[(String, String)] => Future[Unit]]
```
```suggestion
    protected val scanUrlsChangedCallback: Option[Seq[(String, String)] => Future[Unit]]
```

nicu-da: Why is the input an option though?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1172,22 +1181,35 @@ object BftScanConnection {
   }
 
   private def bootstrapWithSeedNodes(
-      seedUrls: NonEmptyList[Uri],
+      seedUris: NonEmptyList[Uri],
```
nicu-da: any specific reason for this change?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1172,22 +1181,35 @@ object BftScanConnection {
...
+      lastPersistedScanUrlList: Option[Seq[DsoScan]],
+      refreshScanUrlsCallback: Option[Seq[(String, String)] => Future[Unit]],
   )(implicit
       ec: ExecutionContextExecutor,
       tc: TraceContext,
       mat: Materializer,
   ): Future[BftScanConnection] = {
     val logger = loggerFactory.getTracedLogger(getClass)
+
+    val bootstrapUris: NonEmptyList[Uri] = lastPersistedScanUrlList match {
```
nicu-da: This should be handled outside this class, the bft scan connection should have no knowledge of this fallback and the existence of the db scan urls

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1263,12 +1293,15 @@ object BftScanConnection {
           loggerFactory,
         )
 
-      case ts @ BftScanClientConfig.BftCustom(_, _, _, _, _) =>
+      case ts @ BftScanClientConfig.BftCustom(_, _, _, _, _, _) =>
         // We bootstrap with the set of provided seed-urls.
         // Since not all trusted SV seeds are provided (most likely), they will not be used in the initial scan connection checking.
         // In the future, add a new threshold for how many trusted seed-urls should be there.
 
+        val enableValidatorInternalStore = ts.enableInternalStore.getOrElse(false)
```
nicu-da: Same as above, this should not leak into the bftScanConnection

nicu-da: also `enableInternalStore` should have a default and not be optional

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -64,14 +64,14 @@ class DbValidatorInternalStore(
...
 
-  def initializeState()(implicit tc: TraceContext): Future[Unit] = {
-    for {
-      result <- StoreDescriptorStore
-        .getStoreIdForDescriptor(storeDescriptor, storage)
-    } yield {
-      internalState.updateAndGet(_.copy(storeId = Some(result)))
-      ()
-    }
+  def initialize()(implicit tc: TraceContext): Future[DbValidatorInternalStore] = {
```
nicu-da: Why do we need to do this? I would much rather keep it explicit returning unit as it much better reflects the side effects it has. 
Though as a side note why can't we have this done as part of the constructor and we need the side effect?

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
```diff
@@ -183,6 +184,36 @@ class ValidatorApp(
...
+
+    val optionTConfig = validatorConfigProvider.getScanUrlInternalConfig()
+
+    optionTConfig.map { internalConfigs =>
+      internalConfigs.map { internalConfig =>
+        (internalConfig.svName, internalConfig.url)
+      }.toList
+    }.value
+
+  }
```
nicu-da: Lets move this into a `ValidatorScanConnection` class where we can locate all the code related to scan connections

nicu-da (commented):
Thank you! Great start for the integration.

Can you also add the issue to the body so we have the connections between the PR and issue? 
And I assume afterwards we plan on enabling this by default? I suggest we start with it enabled by default in all the tests though

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved

I used the factory method `apply` to instantiate the object and to call this init()

`apply()` returns Future[DbValidatorInternalStore] , which is the object we need

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: There are instances of BftScanConnection, outside the main BftScanConnection used by the validtor App.

ScanAggregatesConnection
ScanHistoryBackfillingTrigger 

These don't need persistence (at least in the current scope)

Hence, the argument is an `Option`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Resolved: reset to seedUrls

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
pasindutennage-da: Resolved: Now by default persistence is enabled

pasindutennage-da: Resolved: Now by default persistence is enabled

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Resolved: Moved the logic to the place where we process the scan-configs

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Resolved

enableInternalStore is true by default, and not Optional

BftScanConnection now only sees a set of seedUrls

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
pasindutennage-da: Resolved: Moved to ValidatorScanConnection

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
pasindutennage-da: Resolved: removed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
pasindutennage-da: Resolved: Moved to existing test file BftScanConnectionIntegrationTest

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
pasindutennage-da: Resolved: Added to EOF

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
pasindutennage-da: > `existsUrl(messages, "http://localhost:5012")`

This checks whether a perticular URL was used at Bootstrap, confirming that it loaded the config from DB

We already checked the DB set and get in my previous unit test. Hence, here we only check if Validator correctly uses the set of URLs, that we expect it to retireve from the DB. -- we start with only 1 seed-url, but upon restart, it bootstrap with 4 urls.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
pasindutennage-da: Resolved: Removed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
nicu-da: I think we still need to process the rename. 

Also why use an option instead of defaulting to _ => Future.unit?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -978,13 +979,31 @@ object BftScanConnection {
...
+          successfullyConnectedUrls = newState.openConnections.keys.toSet
+
+          filteredScans: Seq[DsoScan] = scansInDsoRules
+            .filter(scan => successfullyConnectedUrls.contains(scan.publicUrl))
+
+          dsoScanSeq: Seq[(String, String)] = filteredScans.map(scan =>
+            (scan.svName, scan.publicUrl.toString)
+          )
+
+          _ = persistScanUrlsCallback.map(f => f(dsoScanSeq))
```
nicu-da: Did we explicitly chose to persist scans that we managed to connect to instead of all the scans? I would honestly expect to have all the scans persisted, even if we can't connect to some of them

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1180,12 +1197,16 @@ object BftScanConnection {
...
       builder: (Uri, NonNegativeFiniteDuration) => Future[SingleScanConnection],
+      refreshScanUrlsCallback: Option[Seq[(String, String)] => Future[Unit]],
   )(implicit
       ec: ExecutionContextExecutor,
       tc: TraceContext,
       mat: Materializer,
   ): Future[BftScanConnection] = {
     val logger = loggerFactory.getTracedLogger(getClass)
+
+    seedUrls.toList.foreach(uri => logger.info(s"Validator bootstrapping with scan URI: $uri"))
```
nicu-da: Can we add all the urls in a single log message? Makes debugging easier

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1238,6 +1260,8 @@ object BftScanConnection {
       clock: Clock,
       retryProvider: RetryProvider,
       loggerFactory: NamedLoggerFactory,
+      persistScanUrlsCallback: Option[Seq[(String, String)] => Future[Unit]],
+      lastPersistedScanUrlList: Future[Option[List[(String, String)]]],
```
nicu-da: Can't we pass the config provider directly here? Or at least a function that loads the urls list. I'm not a fan of passing around Futures, that only leads to pain

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1505,6 +1567,7 @@ object BftScanConnection {
           ScanAppClientConfig.DefaultAmuletRulesCacheTimeToLive,
         scansRefreshInterval: NonNegativeFiniteDuration =
           ScanAppClientConfig.DefaultScansRefreshInterval,
+        enableInternalStore: Boolean = true,
```
```suggestion
        useLastKnownConnectionsForInitialization: Boolean = true,
```

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
```diff
@@ -64,14 +64,16 @@ class DbValidatorInternalStore(
...
 
-  def initializeState()(implicit tc: TraceContext): Future[Unit] = {
-    for {
-      result <- StoreDescriptorStore
-        .getStoreIdForDescriptor(storeDescriptor, storage)
-    } yield {
-      internalState.updateAndGet(_.copy(storeId = Some(result)))
-      ()
-    }
+  private def getOrSaveStoreDescriptor()(implicit
```
nicu-da: Sorry, I probably missed it, but why do we need this to be here instead of the apply in `DbValidatorInternalStore`?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
```diff
@@ -157,4 +178,72 @@ class BftScanConnectionIntegrationTest
...
+    )
+
+    loggerFactory.assertEventuallyLogsSeq(SuppressionRule.LevelAndAbove(Level.INFO))(
+      {
+        aliceValidatorBackend.startSync()
+      },
+      logs => {
+        val messages = logs.map(_.message)
+        withClue("Validator should bootstrap with only sv1 scan") {
+          existsUrl(messages, "http://localhost:5012") &&
```
nicu-da: I think we still need to process https://github.com/hyperledger-labs/splice/pull/3254#discussion_r2555081486

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
```diff
@@ -157,4 +178,72 @@ class BftScanConnectionIntegrationTest
...
+          !existsUrl(messages, "http://localhost:5112") &&
+          !existsUrl(messages, "http://localhost:5212") &&
+          !existsUrl(messages, "http://localhost:5312")
+        } should be(true).withClue(s"Actual Logs: $logs")
+      },
+    )
+
```
nicu-da, line 213: I would still explicitly assert here on the actual config provider that the db contains the values we want.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
```diff
@@ -157,4 +178,72 @@ class BftScanConnectionIntegrationTest
...
+        )
+      },
+      logs => {
+        val messages = logs.map(_.message)
+        withClue("validator should eventually refresh the scan list") {
+          messages.exists(_.contains(s"Updated scan list with 4 scans:"))
+        } should be(true).withClue(s"Actual Logs: $logs")
+      },
+    )
+
```
nicu-da, line 213: Here as well, db assertion and correlate that with the log messages

nicu-da (commented):
Nice, I think there are only minor things to still discuss

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
pasindutennage-da: Resolved: Extended test to confirm 2 things

- Does it bootstrap with expected number of uris?
- Does the set of scan urls saved match the expected scans -- uses actual config.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Resolved: renamed to scanUrlsChangedCallback

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
```diff
@@ -157,4 +180,73 @@ class BftScanConnectionIntegrationTest
...
+      ScanUrlInternalConfig(getSvName(3), "http://localhost:5212"),
+      ScanUrlInternalConfig(getSvName(4), "http://localhost:5312"),
+    )
+
+    withClue("Persisted state should contain the expected four internal scan configurations") {
+      val maybeActualConfigs: Option[Seq[ScanUrlInternalConfig]] = persistedState.value.futureValue
+
+      maybeActualConfigs.fold {
+        fail("The persisted state for scan URLs was None (not found or not persisted).")
+      } { actualConfigs =>
```
nicu-da: no need to do this, you can just call `.value` on an option in tests

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
```diff
@@ -157,4 +180,73 @@ class BftScanConnectionIntegrationTest
...
+    }
+
+    loggerFactory.assertEventuallyLogsSeq(SuppressionRule.LevelAndAbove(Level.INFO))(
+      {
+        aliceValidatorBackend.stop()
+        aliceValidatorBackend.startSync()
+      },
+      logs => {
+        val messages = logs.map(_.message)
+        withClue("Validator should bootstrap with 4 scans only") {
```
nicu-da:
```suggestion
        withClue("Validator should bootstrap with all the scan urls persisted to the internal store") {
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
pasindutennage-da: Resolved:

val persistedState: OptionT[Future, Seq[ScanUrlInternalConfig]] =
      aliceValidatorBackend.appState.configProvider.getScanUrlInternalConfig()

    val expectedConfigs = Seq(
      ScanUrlInternalConfig(getSvName(1), "http://localhost:5012"),
      ScanUrlInternalConfig(getSvName(2), "http://localhost:5112"),
      ScanUrlInternalConfig(getSvName(3), "http://localhost:5212"),
      ScanUrlInternalConfig(getSvName(4), "http://localhost:5312"),
    )

    withClue("Persisted state should contain the expected four internal scan configurations") {
      val maybeActualConfigs: Option[Seq[ScanUrlInternalConfig]] = persistedState.value.futureValue

      maybeActualConfigs.fold {
        fail("The persisted state for scan URLs was None (not found or not persisted).")
      } { actualConfigs =>
        actualConfigs should contain theSameElementsAs expectedConfigs
      }
    }

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Resolved: Reverted to persisting all scans

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Resolved: all urls in single line

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
pasindutennage-da: Resolved: Added DB assertion (only once), and then corelated with logs

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Resolved: used useLastKnownConnectionsForInitialization

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/db/DbValidatorInternalStore.scala
pasindutennage-da: Resolved: done in DbValidatorInternalStore apply()

pasindutennage-da: Resolved: done in DbValidatorInternalStore apply()

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Resolved:  defaulting to _ => Future.unit

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Resolved: used function that loads the urls list

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
pasindutennage-da: I could not find a good way to use `forAtLeast/forExactly/forNo`

In this scenario, a lot of logs are created, not just what I wish to check. Reason being I check at level `info`

I used 

```
messages.filter(_.contains(bootstrapsWith1UrlLog)) should have length 0
messages.filter(_.contains(bootstrapsWith4UrlsLog)) should have length 2
```

Which acheives more or less the very same behavior.

Note that in this scenario, we are not looking at any failure scenario.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
pasindutennage-da: Please see my comment in the original thread.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/InternalConfigEnabledBftScanConnectionIntegrationTest.scala
nicu-da: Works for me, though out of curiosity why can't we use `forAtLeast/forExactly/forNo`? Why does it matter how many logs are created? Those do pretty much the same thing you implemented right?

nicu-da (approved):
Thank you! LGTM! Nice work

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
pasindutennage-da: Resolved: Done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BftScanConnectionIntegrationTest.scala
pasindutennage-da: Done, accepted change

PR #3278, opened 2025-11-27
republish canton images

isegall-da:
[cluster test](https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/47389)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


MAINTENANCE.md
```diff
@@ -20,11 +20,13 @@
    `nix store prefetch-file --json --hash-type sha256 https://digitalasset.jfrog.io/artifactory/canton-enterprise/canton-enterprise-<version>.tar.gz | jq -r '.hash'` and
    `nix store prefetch-file --json --hash-type sha256  https://www.canton.io/releases/canton-open-source-<version>.tar.gz | jq -r '.hash'`
 4. Update the Canton version and hashes of the oss and enterprise versions in `nix/canton-sources.json`.
-5. In case you have also made configuration changes to Canton in `simple-topology-canton.conf`, remember
+5. Update the image hashes of the Canton images in `nix/canton-sources.json` by running:
```
nicu-da: nit: can we please add this to cncluster or some script we can easily run?

martinflorian-da: +1

Maybe just a very short script in `scripts/`...

martinflorian-da: We could also make this script do steps 3 and 4 as well, while we are at it... maybe a nice beginner-ish tech debt, if you want to avoid the feature creep for this PR.

cluster/images/canton/Dockerfile
martinflorian-da, line 1: The only place I can find where we still reference this is for the docs in https://docs.dev.sync.global/deployment/console_access.html ... and these don't really have any testing (not even occasional manual testing as we open our consoles with the local canton binary from nix).

So the issue I see here is that without any testing (am I just missing it?) we might end up shipping an image that doesn't really work as expected.

So IMO:

- Let's double check if there is really a good enough reason to keep this image. For the one reason I found above: We can rewrite our docs to use the participant canton image instead. They all ship the same binary and I don't see that changing soon.
- If we do find that we want to continue shipping this image, let's start using it for something, sometimes, somehow... to make sure it keeps working.

MAINTENANCE.md
isegall-da: And I thought that I can get away without figuring out the exact jq command to actually make the substitution in place...

cluster/images/canton/Dockerfile
isegall-da: Good point. Since we're not actually building from it as a base, I'm definitely leaning toward just killing it (in splice. canton will still build it as a base, and ship it).

> We can rewrite our docs to use the participant canton image instead

WDYT between:
- use the participant canton image from splice
- use the base canton image from canton's GAR
?

martinflorian-da: Hmm...

pro canton:

- it seems much more clean

pro splice:

- packages enterprise, doesn't it? that might contain extra features also for remote consoles :shrug:
- we know it works because the link should be the same we use everywhere?
- it might be easier to figure out the right version to use here... (this is IMO the winning reason)

So I think I'm leaning towards splice.

(Also the main reason we use images here at all is that it allows us to use an easy oneliner for starting the canton process from nothing... so I don't think it's terribly important which URL we end up using.)

isegall-da: > packages enterprise

No such thing as enterprise any more. The splice one will be an identical copy of the (public) one from canton

> it might be easier to figure out the right version to use here

We have the canton version in the docs already, so shouldn't be hard to inject it here


My main pro-canton is rather against-splice which is that using the participant image might cause some confusion as in you are running a participant or something like that, but I might be over-thinking this and people just copy-paste the command regardless of the characters in it...

martinflorian-da: To be clear I'm OK with both variants... I really don't think it matters a lot.

cluster/images/canton-domain/app.conf
isegall-da, line 1: whoops, I realized that canton doesn't actually have this image. They have a canton-synchronizer directory, which I assume was this, but it's actually empty.

build-tools/bump-canton.sh
```diff
@@ -0,0 +1,45 @@
...
+
+_info "Fetching enterprise tar.gz hash..."
+enterprise_sha256=$(nix store prefetch-file --json --hash-type sha256 "https://digitalasset.jfrog.io/artifactory/canton-enterprise/canton-enterprise-${NEW_VERSION}.tar.gz" | jq -r '.hash')
+# Apparently, nix256 is a thing... While nix seems to accept both, when you have the wrong one,
+# nix shows the nix256-formatted one, so we use that one too
+# (also for backward compatibility with the format we used before this script)
+enterprise_nix256="sha256:"$(nix hash convert --to nix32 "$enterprise_sha256")
+set_value enterprise_sha256 "$enterprise_nix256"
+
+_info "Fetching open source tar.gz hash..."
```
isegall-da, line 36: Soren is looking into why these two are not identical yet. In theory, they should be now...

PR #3283, opened 2025-11-27
Helm chart changes for Validator Scan Client Config

pasindutennage-da:
**BFT Scan Client Configuration Update Helm**

Updates the validator Helm chart to support the Canton BFT Scan client configuration. This enables deployment of the BftCustom client configuration via Helm values.

Fixes scan part of https://github.com/hyperledger-labs/splice/issues/2409

**Logs**

**With bft-custom**

```
{"@timestamp":"2025-12-05T14:08:24.838Z","message":"Validator bootstrapping with 2 seed URLs: List(https://scan.sv-2-eng.scratchb.network.canton.global, https://scan.sv-3-eng.scratchb.network.canton.global)","logger_name":"o.l.s.s.a.a.c.BftScanConnection$:validator=validator_backend","thread_name":"canton-env-ec-70","level":"INFO","trace-id":"62f006ff1157ddc9f7ea0b6b39696a13","span-name":"app_init","span-id":"5f2b2e0d36f90b55"}



{"@timestamp":"2025-12-05T14:08:47.873Z","message":"Updated scan list with 2 scans: BftState(Map(https://scan.sv-2-eng.scratchb.network.canton.global -> (org.lfdecentralizedtrust.splice.scan.admin.api.client.SingleScanConnection@117f3521,Seed URL #0), https://scan.sv-3-eng.scratchb.network.canton.global -> (org.lfdecentralizedtrust.splice.scan.admin.api.client.SingleScanConnection@41fc2e14,Seed URL #1)),Map())","logger_name":"o.l.s.s.a.a.c.BftScanConnection$ConfigurationProvidedScansBft:validator=validator_backend","thread_name":"canton-env-ec-63","level":"INFO","trace-id":"62f006ff1157ddc9f7ea0b6b39696a13","span-name":"app_init","span-id":"5f2b2e0d36f90b55"}



- name: ADDITIONAL_CONFIG_BFT_SCAN
          value: |
            canton.validator-apps.validator_backend.scan-client.type = "bft-custom"
            canton.validator-apps.validator_backend.scan-client.seed-urls = ["https://scan.sv-2-eng.scratchb.network.canton.global","https://scan.sv-3-eng.scratchb.network.canton.global"]
            canton.validator-apps.validator_backend.scan-client.trusted-svs = ["Digital-Asset-Eng-2","Digital-Asset-Eng-3"]
            canton.validator-apps.validator_backend.scan-client.threshold = 2


```

**Without any scan config**

```

{"@timestamp":"2025-12-05T14:12:31.397Z","message":"retrieved validator config from database with key 'validator_scan_internal_config_key' and value '[{\"url\":\"https://scan.sv-2-eng.scratchb.network.canton.global\",\"svName\":\"Digital-Asset-Eng-2\"},{\"url\":\"https://scan.sv-3-eng.scratchb.network.canton.global\",\"svName\":\"Digital-Asset-Eng-3\"}]'","logger_name":"o.l.s.v.s.d.DbValidatorInternalStore:validator=validator_backend","thread_name":"canton-env-ec-72","level":"DEBUG","trace-id":"0123c78b61b01e6d97fc7622a83628d2","span-name":"app_init","span-id":"809b54d32a09b284"}
{"@timestamp":"2025-12-05T14:12:31.400Z","message":"Validator bootstrapping with 2 seed URLs: List(https://scan.sv-2-eng.scratchb.network.canton.global, https://scan.sv-3-eng.scratchb.network.canton.global)","logger_name":"o.l.s.s.a.a.c.BftScanConnection$:validator=validator_backend","thread_name":"canton-env-ec-72","level":"INFO","trace-id":"0123c78b61b01e6d97fc7622a83628d2","span-name":"app_init","span-id":"809b54d32a09b284"}



{"@timestamp":"2025-12-05T14:12:32.031Z","message":"Updated scan list with 4 scans: BftState(Map(https://scan.sv-2-eng.scratchb.network.canton.global -> (org.lfdecentralizedtrust.splice.scan.admin.api.client.SingleScanConnection@5a02c7ca,Seed URL #0), https://scan.sv-3-eng.scratchb.network.canton.global -> (org.lfdecentralizedtrust.splice.scan.admin.api.client.SingleScanConnection@462598ba,Seed URL #1), https://scan.sv-1.scratchb.network.canton.global -> (org.lfdecentralizedtrust.splice.scan.admin.api.client.SingleScanConnection@5f62cbab,Digital-Asset-1), https://scan.sv-2.scratchb.network.canton.global -> (org.lfdecentralizedtrust.splice.scan.admin.api.client.SingleScanConnection@46d1cf25,Digital-Asset-2)),Map())","logger_name":"o.l.s.s.a.a.c.BftScanConnection$AllDsoScansBft:validator=validator_backend","thread_name":"canton-env-ec-41","level":"INFO","trace-id":"0123c78b61b01e6d97fc7622a83628d2","span-name":"app_init","span-id":"809b54d32a09b284"}



 - name: ADDITIONAL_CONFIG_BFT_SCAN
          value: |
            canton.validator-apps.validator_backend.scan-client.type = "bft"
           canton.validator-apps.validator_backend.scan-client.seed-urls = [${_scan.admin-api.address} ]

```

### DIFF
--- a/cluster/helm/splice-validator/tests/validator_test.yaml
+++ b/cluster/helm/splice-validator/tests/validator_test.yaml
@@ -130,7 +130,7 @@ tests:
     set:
       scanClient:
         scanType: "trust-single"
-        scanAddress: "https://single.scan.url:5003"
+        scanAddress: "https://single.scan.url:5004"
     documentSelector:
       path: kind
       value: Deployment
@@ -140,7 +140,7 @@ tests:
           pattern: "canton.validator-apps.validator_backend.scan-client.type = \"trust-single\""
       - matchRegex:
           path: spec.template.spec.containers[?(@.name=='validator-app')].env[?(@.name=='ADDITIONAL_CONFIG_BFT_SCAN')].value
-          pattern: 'canton.validator-apps.validator_backend.scan-client.url = "https://single.scan.url:5003"'
+          pattern: 'canton.validator-apps.validator_backend.scan-client.url = "https://single.scan.url:5004"'
 
   - it: "configures scan-client for standard bft mode, including multiple seedUrls"
     set:


cluster/helm/splice-validator/templates/validator.yaml
```diff
@@ -105,12 +105,27 @@ spec:
           value: |
             canton.validator-apps.validator_backend.scan-client.type = "trust-single"
             canton.validator-apps.validator_backend.scan-client.url = ${_scan.admin-api.address}
-        {{ else -}}
+        { { - else - } }
         - name: ADDITIONAL_CONFIG_BFT_SCAN
           value: |
+            {{- $config := .Values.scanClientConfig }}
```
pasindutennage-da: move the trust-single here, but leave a dup for backward compatibility

cluster/helm/splice-validator/templates/validator.yaml
```diff
@@ -105,12 +105,27 @@ spec:
...
+            {{- if or $config.threshold $config.trustedSvs }}
```
pasindutennage-da: trusted svs and threshold should be checked and set atomically

pasindutennage-da: trusted svs and threshold should be checked and set atomically

cluster/helm/splice-validator/values-template.yaml
```diff
@@ -7,8 +7,8 @@ fixedTokens: false
 
 # Custom annotations and labels for deployed resources
 pod:
-  annotations: {}
-  labels: {}
+  annotations: { }
```
pasindutennage-da: why we need to change this?

cluster/helm/splice-validator/values-template.yaml
```diff
@@ -33,6 +33,11 @@ persistence:
   secretName: "postgres-secrets"
   enablePgInitContainer: true
 
+scanClientConfig:
+  useLastKnownConnectionsForInitialization: true
+  threshold: null
```
pasindutennage-da: remove defaults for unknowns

cluster/helm/splice-validator/values-template.yaml
```diff
@@ -74,7 +79,7 @@ service:
 # set the poll interval used by the UI in milliseconds (optional)
 # uiPollInterval:
 
-extraInitContainers: []
+extraInitContainers: [ ]
```
pasindutennage-da: why we need a change here>

cluster/helm/splice-validator/values.schema.json
```diff
@@ -9,6 +9,27 @@
       "type": "object",
       "description": "The authentication configuration for the application"
     },
+    "scanClientConfig": {
```
pasindutennage-da: Add a Helm test

cluster/helm/splice-validator/values.schema.json
```diff
@@ -9,6 +9,27 @@
       "type": "object",
       "description": "The authentication configuration for the application"
     },
```
pasindutennage-da, line 11: Add this config in ci-main validator 1 and check the logs 

Do we have integration tests at Pulumi level?

cluster/helm/splice-validator/values-template.yaml
pasindutennage-da, line 1: Deploy in SctachNet

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: @nicu-da 

We set the Trust single for both cases "Values.svValidator' and "Values.nonSvValidatorTrustSingleScan"

Given that do you think we should add TrustSingle to "ADDITIONAL_CONFIG_BFT_SCAN" block too?

(sorry if I have misunderstood your suggestion)

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: Resolved: Done

pasindutennage-da: Resolved

cluster/helm/splice-validator/values-template.yaml
pasindutennage-da: Resolved

cluster/helm/splice-validator/values-template.yaml
pasindutennage-da: Done

cluster/helm/splice-validator/values-template.yaml
pasindutennage-da: Done

cluster/helm/splice-validator/values.schema.json
pasindutennage-da: just check logs

cluster/helm/splice-validator/values.schema.json
pasindutennage-da: Done

cluster/helm/splice-validator/templates/validator.yaml
```diff
@@ -105,11 +105,28 @@ spec:
           value: |
             canton.validator-apps.validator_backend.scan-client.type = "trust-single"
             canton.validator-apps.validator_backend.scan-client.url = ${_scan.admin-api.address}
-        {{ else -}}
+        {{- else -}}
         - name: ADDITIONAL_CONFIG_BFT_SCAN
           value: |
+            {{- $config := .Values.scanClientConfig }}
+            {{- if eq ($config.useLastKnownConnectionsForInitialization | default true) false }}
```
nicu-da: Do we need equality checks for boolean values? 🤔

cluster/helm/splice-validator/templates/validator.yaml
nicu-da, line 103: So my suggestion was to clean up the way we configure scan. With the new options we're introducing a confusing UX by having both the top level `nonSvValidatorTrustSingleScan` and `scanAddress` and also the new `scanClient` configuration block. 

What I suggest is that we move both into the new `scanClient` block while still keeping backwards compatibility if they were set in the older versions. 

We should also update the docs and the standalone values files to reflect the new structure and also add release notes about the recommendation to switch from the old values to the new values.

cluster/helm/splice-validator/templates/validator.yaml
```diff
@@ -105,11 +105,28 @@ spec:
...
+            canton.validator-apps.validator_backend.scan-client.use-last-known-connections-for-initialization = false
```
nicu-da: This makes the assumption that the code default is true, we should just set it to the value of `useLastKnownConnectionsForInitialization` if provided

cluster/helm/splice-validator/templates/validator.yaml
```diff
@@ -105,11 +105,28 @@ spec:
...
+            {{- end }}
+            {{- $hasThreshold := $config.threshold }}
+            {{- $hasTrustedSvs := $config.trustedSvs }}
+            {{- if and $hasThreshold $hasTrustedSvs (ge (len $hasTrustedSvs) ($hasThreshold | int)) }}
+            canton.validator-apps.validator_backend.scan-client.type = "bft-custom"
+            canton.validator-apps.validator_backend.scan-client.seed-urls = [ ${_scan.admin-api.address} ]
```
nicu-da: this is actually fairly confusing, do you see a reason why not to use scanAddress directly?

cluster/helm/splice-validator/templates/validator.yaml
```diff
@@ -105,11 +105,28 @@ spec:
...
         - name: ADDITIONAL_CONFIG_BFT_SCAN
           value: |
+            {{- $config := .Values.scanClientConfig }}
+            {{- if eq ($config.useLastKnownConnectionsForInitialization | default true) false }}
+            canton.validator-apps.validator_backend.scan-client.use-last-known-connections-for-initialization = false
+            {{- end }}
+            {{- $hasThreshold := $config.threshold }}
+            {{- $hasTrustedSvs := $config.trustedSvs }}
+            {{- if and $hasThreshold $hasTrustedSvs (ge (len $hasTrustedSvs) ($hasThreshold | int)) }}
+            canton.validator-apps.validator_backend.scan-client.type = "bft-custom"
```
nicu-da, line 119: I would actually make this a filed in the scanclient configuration.

So in the values of the helm chart itself you would pass the client type and then any relevant configs.

nicu-da: This is just to improve the ux and be more explicit instead of having x amount of fields that have to be set to trigger a certain client type. 
This would also replace the need for `nonSvValidatorTrustSingleScan` in the new scanClient config block

cluster/helm/splice-validator/tests/validator_test.yaml
```diff
@@ -114,6 +114,52 @@ tests:
...
+      value: Deployment
+    asserts:
+      - matchRegex:
+          path: spec.template.spec.containers[?(@.name=='validator-app')].env[?(@.name=='ADDITIONAL_CONFIG_BFT_SCAN')].value
+          pattern: 'canton\.validator-apps\.validator_backend\.scan-client\.type = "bft"'
+      - notMatchRegex:
+          path: spec.template.spec.containers[?(@.name=='validator-app')].env[?(@.name=='ADDITIONAL_CONFIG_BFT_SCAN')].value
+          pattern: 'canton\.validator-apps\.validator_backend\.scan-client\.threshold'
+
+  - it: "turns off useLastKnownConnectionsForInitialization when set to false"
```
nicu-da: We should add a test that the old top level `nonSvValidatorTrustSingleScan` is still used when set

cluster/helm/splice-validator/values.schema.json
nicu-da: These schemas support more complex checks right? we should ensure fields set  actually make sense together. so no threshold if no trusted svs are set, no truested svs and threshold if `nonSvValidatorTrustSingleScan` is set and any other combinations that make sense.

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: I tried many ways, but the only way that works is the following.

```
{{- $config := .Values.scanClientConfig }}
            {{- if and $config (hasKey $config "useLastKnownConnectionsForInitialization") (not $config.useLastKnownConnectionsForInitialization)}}
            canton.validator-apps.validator_backend.scan-client.use-last-known-connections-for-initialization = false
            {{- end}}
```

IMO, go templates is not very smart at processing (1) variable exists AND (2) variable is set AND (3) variable is false and AND (4) equals false pass


This is my first time using Helm, so I don't confirm my above statement. Just what seems to work for now,

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: Resolved:

`{{- if and $config (hasKey $config "useLastKnownConnectionsForInitialization") (not $config.useLastKnownConnectionsForInitialization)}}`

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: Resolved: merged every scan client in to a single block, kept others for backward compatibility

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: Marking as resolved because bool checks work like this (perhaps there are other options, will improve later)

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: Refactored scan-config to a single block, and kept other for backward compatiblity.

TODO: update docs and standalone values

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: Resolved

Used specified scan-seed Urls, if the config is available

If not, for backward compatibility, kept the _scan.admin-api.address.

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: Resolved: Used type in the config, and used matching

cluster/helm/splice-validator/values.schema.json
pasindutennage-da: I added checks for scanClientConfig

It still lacks dynamic checks for threshold <= len(seedUrl) && threshold <= len(trustedSv). In my understanding, standard Json only provide array length checks for static values. To have dynamic checks we need JSONLogic or something simmilar

If my above statement is true, then I think it is ok to depend on the validator scala initailzation logic checks -- if parameters don't make sense, then the deployment will fail anyway (just not at helm pulumi level though)

cluster/helm/splice-validator/tests/validator_test.yaml
pasindutennage-da: Resolved: Added the test, and a few more

cluster/helm/splice-validator/templates/validator.yaml
```diff
@@ -108,8 +108,28 @@ spec:
         {{ else -}}
         - name: ADDITIONAL_CONFIG_BFT_SCAN
           value: |
+            {{- if .Values.scanClientConfig }}
+              {{- if eq .Values.scanClientConfig.scanType "trust-single" }}
+            canton.validator-apps.validator_backend.scan-client.type = "trust-single"
+            canton.validator-apps.validator_backend.scan-client.url = {{ index .Values.scanClientConfig.seedUrls 0 | quote }}
```
nicu-da: In this scenario I would expect that we don't ask for an array (eg if the client provides 2 items you will just use the first one and it will be confusing)

cluster/helm/splice-validator/templates/validator.yaml
```diff
@@ -108,8 +108,28 @@ spec:
         {{ else -}}
         - name: ADDITIONAL_CONFIG_BFT_SCAN
           value: |
+            {{- if .Values.scanClientConfig }}
```
nicu-da: I think we're still missing also incorporating the `scanAddress` here, which is used as the seed and single scan url right?

cluster/helm/splice-validator/values.schema.json
```diff
@@ -23,6 +63,52 @@
     }
   },
   "allOf": [
```
nicu-da, line 64: Can we also add validation here if the new scanClientConfig block is set then the previous configs (`nonSvValidatorTrustSingleScan` and `scanAddress`) are not set?

cluster/helm/splice-validator/templates/validator.yaml
```diff
@@ -108,8 +108,28 @@ spec:
...
+              {{- if eq .Values.scanClientConfig.scanType "trust-single" }}
+            canton.validator-apps.validator_backend.scan-client.type = "trust-single"
+            canton.validator-apps.validator_backend.scan-client.url = {{ index .Values.scanClientConfig.seedUrls 0 | quote }}
+              {{- else}}
+                {{- if and (hasKey .Values.scanClientConfig "useLastKnownConnectionsForInitialization") (not .Values.scanClientConfig.useLastKnownConnectionsForInitialization)}}
```
nicu-da: Sorry to come back to this but thinking about it, it doesn't actually make sense to make this a first class citizen of the helm charts. We shouldn't need to ever disable this unless somehow it's broken and then we can just set the ADDITIONAL_CONFIG env variables. So i would fully remove this one from helm

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: Resolved: used ScanAddress

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: Resolved: used ScanAddress

cluster/helm/splice-validator/values.schema.json
pasindutennage-da: Resolved: Added

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: Resolved: Removed

cluster/helm/splice-validator/templates/validator.yaml
```diff
@@ -108,8 +108,24 @@ spec:
         {{ else -}}
         - name: ADDITIONAL_CONFIG_BFT_SCAN
           value: |
+            {{- if .Values.scanClientConfig }}
```
nicu-da:
```suggestion
            {{- if .Values.scanClient }}
```

We actually don't use config in any of the helm values so it might make sense to remove it from here as well

cluster/helm/splice-validator/templates/validator.yaml
nicu-da, line 1: What I'm missing is the changes to the validator-values.yaml, we should update the documentation in that file and also update the values to use the new config format.

Ideally under the new block, if `scanClientConfig.scanAddress` is set the type should default to `bft` to preserve the same behavior as before as well.

nicu-da: And we also need to update the actual rst documentation to reflect the new configs. In theory it should auto update but we should go over the install steps for a helm validator to make sense we didn't miss something.

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: Resolved: Renamed

cluster/helm/splice-validator/values.schema.json
pasindutennage-da: Added to validator runbook, and checked logs

cluster/helm/splice-validator/values-template.yaml
pasindutennage-da: Done

cluster/helm/splice-validator/templates/validator.yaml
pasindutennage-da: Resolved:

In a future PR we will update the values

pasindutennage-da: Updated: For now only config.yaml in cluster/deployment/config.yaml is updated. To keep backward compatibility, for now nothing is changed in the existing configs

cluster/pulumi/common-validator/src/config.ts
```diff
@@ -9,9 +9,18 @@ import {
 import { clusterSubConfig } from '@lfdecentralizedtrust/splice-pulumi-common/src/config/configLoader';
 import { z } from 'zod';
 
+export const ScanClientConfigSchema = z.object({
```
nicu-da: Nit: ideally this would also represent through types only the options that are valid together as an ADT

cluster/helm/splice-validator/values.schema.json
```diff
@@ -452,6 +536,29 @@
...
+          "anyOf": [
+            {
+              "required": ["scanAddress"]
+            },
+            {
+              "required": ["nonSvValidatorTrustSingleScan"]
+            }
+          ]
+        },
+        "errorMessage": "If 'scanClient' is set, 'scanAddress' and 'nonSvValidatorTrustSingleScan' must not be set."
```
nicu-da: We might wanna expand on this a bit, as this would be a message to the user. Maybe just add that everything should be configured under the scanClient key

cluster/helm/splice-validator/tests/validator_test.yaml
```diff
@@ -126,3 +125,84 @@ tests:
           value: ledger-api-user
       - exists:
           path: spec.template.spec.containers[?(@.name=='validator-app')].env[?(@.name=='ADDITIONAL_CONFIG_DISABLE_AUTH')].value
+
+  - it: "configures scan-client for trust-single mode"
```
nicu-da, line 129: This all looks good. 
What do you think of adding one more step to the scanClient, to improve defaults? 

We can default to scanType: bft if no other type is set but the scanClient block is present. 
And in that scenario we can either accept scanAddress or seedUrls as seeds. 

This way client would be able to configure it simply as:

```
scanClient:
  scanAddress: address
```

or 

```
scanClient:
  seedUrls:
    - address 
```

cluster/pulumi/common-validator/src/config.ts
pasindutennage-da: Done, added checks in TS

pasindutennage-da: Done, added checks in TS

cluster/helm/splice-validator/values.schema.json
pasindutennage-da: Done

cluster/helm/splice-validator/tests/validator_test.yaml
pasindutennage-da: Modification 1: if type is not specified, then defaults to bft type, however with seed-url

pasindutennage-da: Modification 2: added a test

PR #3286, opened 2025-11-27
Revert "Add trigger for periodic topology snapshots"

OriolMunoz-da:
Reverts hyperledger-labs/splice#3188

I haven't been able to figure out where `cn-gcp-bucket-da-cn-devnet-da-cn-topology-snapshots` is supposed to be set

Julien is off now and tomorrow, he can restore this PR with the fix for the secret


pasindutennage-da (approved):
Blindly approving because this is only a revert

OriolMunoz-da:
force push because automatic revert of PRs doesn't include signoff (sigh)

PR #3287, opened 2025-11-28
Update "required" in JSON Ledger API Specs

giner:
"required" is updated based on property descriptions using the following script:

    # Update "required" only for properties marked "Required" or "Optional" in their description.
    yq -Y --indentless-lists '
      .components |= (
        (.. | select(type == "object" and has("properties"))) |= (
          (.required // []) as $req_orig
          | .required += (
              [
                .properties | to_entries[]
                | select( (.value.description? // "") | (contains("Required") and ( (contains("Required authorization") or contains("Required unless")) | not)) ) # only add if marked "Required", but not "Required authorization" or "Required unless"
                | .key
                | select(. as $r | $req_orig | contains([$r]) | not) # only add if not already present
              ]
            )
          | .required -= (
              [
                .properties | to_entries[]
                | select( (.value.description? // "") | contains("Optional") ) # only remove if marked "Optional"
                | .key
              ]
            )
          | del(.required | select(length == 0)) # remove empty required arrays
        )
      )
    '

The following script would remove all properties which don't have "Required" in their descriptions. I don't know which approach is right here so I'm using the one which makes less changes for now.

    # Update "required" only for properties marked "Required" and remove those which are not.
    yq -Y --indentless-lists '
      .components |= (
        (.. | select(type == "object" and has("properties"))) |= (
          (.required // []) as $req_orig
          | .required = (
              [
                .properties | to_entries[]
                | select( (.value.description? // "") | (contains("Required") and ( (contains("Required authorization") or contains("Required unless")) | not)) ) # only add if marked "Required", but not "Required authorization" or "Required unless"
                | .key
                | select(. as $r | $req_orig | contains([$r]) | not) # only add if not already present
              ]
            )
          | del(.required | select(length == 0)) # remove empty required arrays
        )
      )
    '

Note that Python-based version of `yq` (https://github.com/kislyuk/yq) was used to preserve formatting however https://github.com/mikefarah/yq will also work.

### DIFF
--- a/canton/community/ledger/ledger-json-api/src/test/resources/json-api-docs/openapi.yaml
+++ b/canton/community/ledger/ledger-json-api/src/test/resources/json-api-docs/openapi.yaml
@@ -2145,7 +2145,7 @@ components:
       type: object
       required:
       - synchronizer
-      - identityProviderId
+      - onboardingTransactions
       properties:
         synchronizer:
           description: |-
@@ -2195,11 +2195,6 @@ components:
       title: AllocatePartyRequest
       description: 'Required authorization: ``HasRight(ParticipantAdmin) OR IsAuthenticatedIdentityProviderAdmin(identity_provider_id)``'
       type: object
-      required:
-      - partyIdHint
-      - identityProviderId
-      - synchronizerId
-      - userId
       properties:
         partyIdHint:
           description: |-
@@ -2250,6 +2245,7 @@ components:
       - contractId
       - templateId
       - packageName
+      - witnessParties
       properties:
         offset:
           description: |-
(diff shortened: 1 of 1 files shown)


martinflorian-da:
@giner Unless I'm misunderstanding something here, you are proposing a change to our fork of `canton`. That most likely is not what you want. Notably the Canton binaries we package are not built from that folder.

I think you want to upstream a change to Canton / the ledger API itself?

Canton is still in the process of moving it's development fully into the open. For now we have this repo which does periodic dumps from the actual internal repo where development happens: https://github.com/digital-asset/canton

I suggest opening a PR there.

(Closing based on my understanding so far; if I misunderstood something please reopen and/or yell at me.)

PR #3288, opened 2025-11-28
add an example unit test in a Pulumi project

mblaze-da:
fixes https://github.com/DACH-NY/canton-network-internal/issues/2820

This PR introduces configuration for unit tests in Pulumi projects using the Jest testing framework. These tests are run during `make cluster/pulumi/test` which is already run in precommit hooks and in CI.

The PR is marked as DRAFT as it is missing documentation that is planned to be added after these functional changes are first reviewed.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -23,8 +23,11 @@
     - [Handling Errors in Integration Tests](#handling-errors-in-integration-tests)
     - [Connecting external tools to the shared Canton instances](#connecting-external-tools-to-the-shared-canton-instances)
     - [Testing App Upgrades](#testing-app-upgrades)
-    - [Testing from a custom canton instance]
+    - [Testing from a custom canton instance](#testing-from-a-custom-canton-instance)
   - [Deployment Tests](#deployment-tests)
+    - [Helm checks](#helm-checks)
+    - [Pulumi tests](#pulumi-tests)
+    - [Pulumi state checks](#pulumi-state-checks)
 - [CI Without Approval](#ci-without-approval)
 
 # Testing in Splice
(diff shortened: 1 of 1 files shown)


cluster/pulumi/common/src/retries.test.ts
martinflorian-da, line 1: As per CI, these still need that `// Copyright (c) 2024 ...` header. Just copy from another file. And yes, you need to make it 2024 until we fix our CI check... which we postponed to 2026 :clown_face:

cluster/pulumi/common/src/retries.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { retry } from './retries';
+
+// suppresses error logs from retry to get cleaner test output
+const consoleErrorMock = jest.spyOn(console, 'error').mockImplementation(() => {});
+afterAll(() => {
+  consoleErrorMock.mockRestore();
+});
+
+test('retry runs the action again if it fails', async () => {
```
martinflorian-da:
```suggestion
test('retry runs the action again if it fails with an exception', async () => {
```

nit

martinflorian-da (commented):
LGTM so far, thank you!

CI seems not happy though. I think the next gate is that you confirm that this gets run on `[static]`, and doesn't fail: https://github.com/hyperledger-labs/splice/actions/runs/19759028726/job/56616467560?pr=3288

Since this is kind-of an foundational PR for Pulumi unit tests, I'd suggest also pinging @nicu-da for a set of eyes at some point. This way he can't complain when some unit test annoys him later on :smiley: (he needs to do bigger Pulumi changes occasionally).

cluster/pulumi/common/src/retries.test.ts
nicu-da: Can we add a unit test for actual pulumi code? Not saying this isn't useful but this doesn't touch on pulumi resources
Eg: maybe a test for the sv runbook or something similar, that it installs a helm chart with a given set of values?

mblaze-da: We can, but my goal with this PR is not to demonstrate different kinds of tests that we can write but to introduce infrastructure for adding them. AFAIK testing Pulumi resources should be straight forward with the usage of `pulumi.runtime.setMocks`. There are some examples & tips [here](https://www.pulumi.com/docs/iac/guides/testing/unit/). The feature which motivated me to add these tests also does not involv testing resources.

nicu-da: Thanks for the explanation. works for me

martinflorian-da: > AFAIK testing Pulumi resources should be straight forward

Can we still do a "first such test" at some point soon? It's always easier when there is a reference already on `main`.

martinflorian-da:
> CI seems not happy though

It's happy now, so I have nothing left to comment at this stage :)

Also happy to just merge as if you prefer, and follow up with the docs separately.

Not approving because it still says "draft"; btw you do know that on GitHub you also have `draft` as regular PR status, right? See this button for example

image

cluster/pulumi/common/src/retries.test.ts
mblaze-da: Sure, I created https://github.com/hyperledger-labs/splice/issues/3293 so I don't forget.

martinflorian-da (approved):
LGTM, thank you!

PR #3292, opened 2025-11-28
Local store ingestion performance tests

OriolMunoz-da:
Implements local tests for the scan store and sv dso store following the [design doc](https://docs.google.com/document/d/1rvAec6BuKx61TdJ6sY07QRAUA1p_WgedDIisgczFDPI/edit?tab=t.0#heading=h.j1o9vy5fqmrz) to the extent that it was useful to test these (meaning that it's missing a lot of what is defined there, doesn't support CI runs, etc., but is good enough to see the scale of how slow things are locally)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/PerformanceRunner.scala
OriolMunoz-da, line 1: ignore this file for now

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/DownloadTxMainnet.scala
OriolMunoz-da, line 1: downloads 1h worth of mainnet updates and dumps it into a file, to be loaded in the test above

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/DbSvDsoStoreIngestionPerformanceTest.scala
```diff
@@ -0,0 +1,143 @@
...
+    val ingestionConfig = IngestionConfig()
+    val dumpFile = Paths.get("/home/oriolmunoz/mainnetdump/update_history_response.json")
+    val dump = (for {
+      json <- io.circe.parser
+        .parse(
+          Files
+            .readString(dumpFile)
+            .replace( // TODO: maybe just use the store with dsoParty=mainnet's
+              "DSO::1220b1431ef217342db44d516bb9befde802be7d8899637d290895fa58880f19accc",
+              dsoParty.toProtoPrimitive,
```
OriolMunoz-da: every test uses its own dsoParty... see TODO

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/DbSvDsoStoreIngestionPerformanceTest.scala
```diff
@@ -0,0 +1,143 @@
...
+    with HasExecutionContext {
+
+  "Ingestion performance test" in {
+    // read csv file into pekko stream
+
+    val store = mkStore()
+    store.multiDomainAcsStore.ingestionSink.initialize().futureValue
+    val timings = mutable.ListBuffer[Long]()
+    val ingestionConfig = IngestionConfig()
+    val dumpFile = Paths.get("/home/oriolmunoz/mainnetdump/update_history_response.json")
```
OriolMunoz-da: this dump is 1h of mainnet updates (see next file for how it was obtained).
It should be okay, because:
- Any (filtered-in) create causes an insert
- Any archive causes a delete, so as long as the template is included in the filter
- Txlog doens't care about whether the contract ids involved are present in the ACS store or not

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/DbSvDsoStoreIngestionPerformanceTest.scala
```diff
@@ -0,0 +1,143 @@
...
+                )
+                .toList
+            )
+          )
+          .map { _ =>
+            val after = System.nanoTime()
+            val duration = after - before
+            timings ++= Seq.fill(batch.length)(duration / batch.length)
+            val avg = timings.sum.toDouble / timings.size
+            println(
```
rautenrieth-da: Does this wait for the output to be flushed? If yes, it could add significant overhead, consider only printing at the end or only once every N iterations.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/DbSvDsoStoreIngestionPerformanceTest.scala
```diff
@@ -0,0 +1,143 @@
...
+      .zipWithIndex
+      .runWith(Sink.foreachAsync(parallelism = 1) { case (batch, index) =>
+        println(s"Ingesting batch $index of ${batch.length} elements")
+        val before = System.nanoTime()
+        store.multiDomainAcsStore.ingestionSink
+          .ingestUpdateBatch(
+            NonEmptyList.fromListUnsafe(
+              txs
+                .map(tx =>
+                  TreeUpdateOrOffsetCheckpoint.Update(tx.update.update, tx.update.synchronizerId)
```
rautenrieth-da: Consider doing this transformation outside of the loop, to lower the processing overhead.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/DbSvDsoStoreIngestionPerformanceTest.scala
```diff
@@ -0,0 +1,143 @@
...
+      .fromIterator(() => txs.iterator)
+      .batch(ingestionConfig.maxBatchSize.toLong, Vector(_))(_ :+ _)
+      .zipWithIndex
+      .runWith(Sink.foreachAsync(parallelism = 1) { case (batch, index) =>
+        println(s"Ingesting batch $index of ${batch.length} elements")
+        val before = System.nanoTime()
+        store.multiDomainAcsStore.ingestionSink
+          .ingestUpdateBatch(
+            NonEmptyList.fromListUnsafe(
+              txs
```
rautenrieth-da: Isn't this ingesting the whole `txs` list in each iteration? Shouldn't it use `batch` instead? I'm surprised nothing explodes when we reingest the same transactions over and over.

OriolMunoz-da: :facepalm: I'm surprised my performance was not god-awful

rautenrieth-da: It was probably fine because the first iteration did a lot of work, and all other iterations were no-ops due to the last ingested offset guard.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/DownloadScanAcsSnapshot.scala
OriolMunoz-da, line 1: There was no requirement for this, but I've had to use it twice last year, so I figured it might be useful...

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/SplicePerf.scala
```diff
@@ -0,0 +1,139 @@
...
+          throw e
+      } finally {
+        Await.result(as.terminate(), 1.minute)
+      }
+    }
+
+    Opts
+      .subcommand(
+        Command(
+          "download-scan-updates",
```
meiersi-da, line 49: Missing: entry in `TESTING.md` explaining how we do perf testing. No need to repeat there what one can discover from the CLI using `splice-perf --help`. However, the link to it, and the link to the design doc wrt outstanding work seems useful.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/SplicePerf.scala
```diff
@@ -0,0 +1,139 @@
...
+                  "Name of the test to run (supports globbing)",
+                  "t",
+                ),
+              Opts
+                .option[String]("config-path", "Path to the config file", "c")
+                .map(
+                  Paths.get(_)
+                ),
+              Opts
+                .option[String]("dump-path", "Path to the update history dump file", "d")
```
meiersi-da: consider making the comment more general: "Path to data dump file (e.g., ACS snapshot + update history)"

Also is it a file or a directory?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/StoreIngestionPerformanceTest.scala
```diff
@@ -0,0 +1,146 @@
...
+    val storageFactory = new StorageSingleFactory(storageConfig)
+    val storage =
+      storageFactory.tryCreate(
+        connectionPoolForParticipant = false,
+        None,
+        new WallClock(timeouts, loggerFactory),
+        None,
+        new DbStorageMetrics(
+          new DbStorageHistograms(MetricName("store", "perftest"))(new HistogramInventory),
+          NoOpMetricsFactory,
```
rautenrieth-da, line 70: Related to R6 in the design doc: Wouldn't it be useful to collect all metrics, and dump them at the end using `OnDemandMetricsReader` (or similar) to a text/json file? It's not as nice as grafana, but not too much implementation overhead and you could still manually compare two different runs.

For storage, you would get `DbStorageMetrics` with things like load on the database pool. For `HistoryMetrics`, you would get `UpdateHistory.latency`. I remember in the participant, we used to collect separate timers for data processing (e.g., code that calls `ProtobufCodec.serializeValue`) and database interactions (waiting for connection, submitting query, waiting for result, ...). This allowed us to make informed decisions on what code paths to optimize.

OriolMunoz-da: copy pasting for reference:

> If the final goal is to push metrics to grafana, then don't spend time on adding something just for local runs.

also, as per R6:

> Metrics gathering is not required for local runs.

still, for local runs we could

> we could push them to the same grafana, but tag them with local=true, user=${whoami} or something like that

PR #3294, opened 2025-11-30
Add a reaper job for pods with unknown pod status

pasindutennage-da:
Solves https://github.com/DACH-NY/canton-network-internal/issues/1418

Adds a new cluster wide reaper job to delete all broken pods.

### DIFF
--- a/cluster/stacks/scratchnets/validator-runbook/Pulumi.sv-runbook-upgrade.scratcha.yaml
+++ b/cluster/stacks/scratchnets/validator-runbook/Pulumi.sv-runbook-upgrade.scratcha.yaml
@@ -1,1 +1,0 @@
-encryptionsalt: v1:1hsaMyBViOk=:v1:pQ5JnBrSyK1WwGmG:vMoXytotkCwQl5LnldBvYC9nEKmTCA==
--- a/cluster/stacks/scratchnets/validator-runbook/Pulumi.sv-runbook-upgrade.scratchd.yaml
+++ b/cluster/stacks/scratchnets/validator-runbook/Pulumi.sv-runbook-upgrade.scratchd.yaml
@@ -1,1 +1,0 @@
-encryptionsalt: v1:zaHBu6vc6uo=:v1:G14u6usUXoE9b9RO:z1YtMcA2YreRm3rbJVbeV0yg0QBcdA==


cluster/pulumi/infra/src/index.ts
```diff
@@ -44,6 +45,8 @@ istioMonitoring(network.ingressNs, []);
 
 configureStorage();
 
+installPodCleanupCronJob(network.ingressNs.ns.metadata.name);
```
nicu-da: Why do we install it in this namespace?

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,85 @@
...
+import * as pulumi from '@pulumi/pulumi';
+import { Input } from '@pulumi/pulumi';
+
+const deleteBadPodsCommand = [
+  '/bin/sh',
+  '-c',
+  `kubectl get pods -A --field-selector=status.phase!=Running -o custom-columns=NAMESPACE:.metadata.namespace,NAME:.metadata.name,STATUS:.status.phase,REASON:.status.containerStatuses[*].state.waiting.reason | \\
+  grep -E "Error|Unknown|CrashLoopBackOff" | \\
+  awk '{print $1, $2}' | \\
+  xargs -n 2 sh -c 'kubectl delete pod -n "$0" "$1"' _ || true`,
```
nicu-da: Lets make sure we add logging for the steps and log command execution and what pods we're deleting.

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,85 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import * as k8s from '@pulumi/kubernetes';
+import * as pulumi from '@pulumi/pulumi';
+import { Input } from '@pulumi/pulumi';
+
+const deleteBadPodsCommand = [
+  '/bin/sh',
+  '-c',
+  `kubectl get pods -A --field-selector=status.phase!=Running -o custom-columns=NAMESPACE:.metadata.namespace,NAME:.metadata.name,STATUS:.status.phase,REASON:.status.containerStatuses[*].state.waiting.reason | \\
+  grep -E "Error|Unknown|CrashLoopBackOff" | \\
```
nicu-da: We should not delete CrashLookBackoff pods. Also unknown should be `containerstatusunknown`. 
Also for the error containers, it could also be status error because of other reasons that we wouldn't want to terminate. 
What we want is to delete pods with state terminated reason error and exit code 137. But the error pods should be deleted only from the multi validator namespace.

I would actually delete only from the multi validator namespace at first and then see if we need to expand it to the other namespaces.

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,85 @@
...
+  awk '{print $1, $2}' | \\
```
nicu-da: can't we print just the pod name through kubectl to avoid the need for awk as well?

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,85 @@
...
+  xargs -n 2 sh -c 'kubectl delete pod -n "$0" "$1"' _ || true`,
+];
+
+export function installPodCleanupCronJob(namespace: Input<string>): k8s.batch.v1.CronJob {
+  const name = 'bad-pod-cleaner';
+  const labels = { app: name, environment: pulumi.getStack() };
+
+  // grants cluster-level permissions (get, list, delete for pods) regardless of namespace.
```
nicu-da: This is a bit dangerous, let's start just with the multi validator namespace.

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,85 @@
...
+  '/bin/sh',
+  '-c',
+  `kubectl get pods -A --field-selector=status.phase!=Running -o custom-columns=NAMESPACE:.metadata.namespace,NAME:.metadata.name,STATUS:.status.phase,REASON:.status.containerStatuses[*].state.waiting.reason | \\
+  grep -E "Error|Unknown|CrashLoopBackOff" | \\
+  awk '{print $1, $2}' | \\
+  xargs -n 2 sh -c 'kubectl delete pod -n "$0" "$1"' _ || true`,
+];
+
+export function installPodCleanupCronJob(namespace: Input<string>): k8s.batch.v1.CronJob {
+  const name = 'bad-pod-cleaner';
```
nicu-da:
```suggestion
  const name = 'gc-error-pods';
```

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,85 @@
...
+const deleteBadPodsCommand = [
+  '/bin/sh',
+  '-c',
+  `kubectl get pods -A --field-selector=status.phase!=Running -o custom-columns=NAMESPACE:.metadata.namespace,NAME:.metadata.name,STATUS:.status.phase,REASON:.status.containerStatuses[*].state.waiting.reason | \\
+  grep -E "Error|Unknown|CrashLoopBackOff" | \\
+  awk '{print $1, $2}' | \\
+  xargs -n 2 sh -c 'kubectl delete pod -n "$0" "$1"' _ || true`,
+];
+
+export function installPodCleanupCronJob(namespace: Input<string>): k8s.batch.v1.CronJob {
```
nicu-da: lets move this to it's own namespace

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,85 @@
...
+      jobTemplate: {
+        spec: {
+          template: {
+            spec: {
+              serviceAccountName: 'default',
+              restartPolicy: 'OnFailure',
+              containers: [
+                {
+                  name: name,
+                  image: 'bitnami/kubectl:latest',
```
nicu-da: We should not use latest as docker image tags.
Also isn't bitnami not safe to use as it requires a license nowdays? We should use the `rancher` image. 

I would suggest we replace latest with the kubectl version tag. We can export an env variable from the nix shell, the same we do for PULUMI_VERSION and use it here.

nicu-da (commented):
Thank you!

cluster/pulumi/infra/src/index.ts
nicu-da: also this should be disabled by default, and have the ability to enable it though the cluster config

pasindutennage-da: Resolved: Cronjob has it's own namespace

cluster/pulumi/infra/src/maintenance.ts
pasindutennage-da: Resolved: added logs

cluster/pulumi/infra/src/maintenance.ts
pasindutennage-da: Resolved: Only multi validaotr pods deleted

cluster/pulumi/infra/src/maintenance.ts
pasindutennage-da: Resolved

Had to use JQ to pass the kubectl output for filtering

cluster/pulumi/infra/src/maintenance.ts
pasindutennage-da: Resolved: only multivalidator namespace is in scope

cluster/pulumi/infra/src/maintenance.ts
pasindutennage-da: Resolved: Done

cluster/pulumi/infra/src/index.ts
```diff
@@ -42,6 +43,8 @@ if (enableAlerts && !clusterIsResetPeriodically) {
 }
 istioMonitoring(network.ingressNs, []);
 
+deployGCPodReaper('cluster-pod-reaper', { parent: network.ingressNs.ns });
```
nicu-da: This should behind a flag and by default not be installed, we want it only in CILR

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,133 @@
...
+const targetNamespace = 'gc-pod-reaper';
+const serviceAccountName = 'gc-pod-reaper-service-account';
+const reaperImage = 'rancher/kubectl:latest';
+const schedule = '0 3 * * *'; // Run once daily at 03:00 AM UTC
+
+const deleteBadPodsCommand = [
+  '/bin/sh',
+  '-c',
+  `
+    LOG_FILE="/tmp/pod-reaper-$(date +%Y%m%d%H%M%S).log";
```
nicu-da: why do we need a log file and can't log directly to stdout? How will this works in the cluster?

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,133 @@
...
+  );
+
+  return new k8s.batch.v1.CronJob(
+    cronJobName,
+    {
+      metadata: {
+        name: cronJobName,
+        namespace: ns.metadata.name,
+        annotations: {
+          'pulumi.com/description': 'Scheduled job to automatically clean up error/stuck pods',
```
nicu-da: Hmm, what's the purpose of the annotation? Does it do something special in pulumi?

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,133 @@
...
+const deleteBadPodsCommand = [
+  '/bin/sh',
+  '-c',
+  `
+    LOG_FILE="/tmp/pod-reaper-$(date +%Y%m%d%H%M%S).log";
+    echo "--- $(date) Starting Pod Reaper ---" >> $LOG_FILE;
+    echo "Processing pods for deletion..." >> $LOG_FILE;
+
+    TARGET_PODS=$(
+      kubectl get pods -A -o json | \\
```
nicu-da: Shouldn't we also limit our search to a given namespace here?

cluster/pulumi/infra/src/maintenance.ts
pasindutennage-da: Used rancher with kubectl version tag

cluster/pulumi/infra/src/index.ts
pasindutennage-da: done: enableGCReaperJob in config.yaml

cluster/pulumi/infra/src/index.ts
pasindutennage-da: Done

enableGCReaperJob is the new flag

cluster/pulumi/infra/src/maintenance.ts
pasindutennage-da: Resolved, removed log file

cluster/pulumi/infra/src/maintenance.ts
pasindutennage-da: Resolved: Removed annotations

cluster/pulumi/infra/src/maintenance.ts
pasindutennage-da: Resolved: Only considering the provided set of namespaces (multi-validator)

cluster/deployment/scratchnetb/.envrc.vars
```diff
@@ -37,3 +37,5 @@ export APPROVE_SV_RUNBOOK=true
 export CACHE_DEV_DOCKER_REGISTRY=$DEV_DOCKER_REGISTRY
 
 export SELF_HOSTED_FLUENT_BIT=true
+
+export KUBECTL_VERSION="v1.28.8"
```
pasindutennage-da, line 41: config only passes .envrc.vars files and .yaml files. hence this is needed

cluster/deployment/scratchnetb/config.yaml
```diff
@@ -70,3 +70,4 @@ svs:
       walletUser: auth0|68c2c41e6470184569e4521e
       auth0:
         clientId: fBFpzZSRuufk9ueV2pSEkOXNYXuQtVmY
+enableGCReaperJob: true
```
pasindutennage-da: todo -- remove this and add to CILR

cluster/deployment/scratchnetb/.envrc.vars
nicu-da: Is needed in what scenario? 🤔

cluster/pulumi/common/src/config/configSchema.ts
```diff
@@ -21,6 +21,7 @@ export const ConfigSchema = z.object({
     active: defaultActiveMigration,
   }),
   persistentHeapDumps: z.boolean().default(false),
+  enableGCReaperJob: z.boolean().default(false),
```
nicu-da: Any reason to not have this as part of InfraConfigSchema?

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,179 @@
...
+const schedule = '0 * * * *'; // Run once an hour
+
+const deleteBadPodsCommand = [
+  '/bin/bash',
+  '-c',
+  `
+    apt-get update &&
+    apt-get install -y curl jq &&
+    curl -LO https://dl.k8s.io/release/${kubectlVersion}/bin/linux/amd64/kubectl &&
+    install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl;
```
nicu-da, line 25: Not a big fan of doing this in the script we run. We should have a pre made image.

Let's just use the custom splice-debug docker image we build

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,179 @@
...
+          spec: {
+            template: {
+              spec: {
+                serviceAccountName: serviceAccountName,
+                restartPolicy: 'OnFailure',
+                // Nodes in this GKE cluster are configured with Taints (cn_infra, cn_apps,
+                // gke-managed-components) that prevent standard Pods from being scheduled.
+                // These Tolerations allow the gc-pod-reaper-job to run on all available
+                // nodes, ensuring the cleanup task can execute regardless of node role.
+                tolerations: [
```
nicu-da: re-use `infraAffinityAndTolerations` please

cluster/pulumi/multi-validator/src/installNode.ts
```diff
@@ -14,6 +15,46 @@ import { installPostgres } from './postgres';
 
 export async function installNode(): Promise<void> {
   const namespace = exactNamespace('multi-validator', true);
+  const podManagementRole = new k8s.rbac.v1.Role(
```
nicu-da: We need to create this only if the gc job is enabled. Any reason not to create  this alongside the other objects? Even if it's in another namespace. We already specify the target namespace when creating the gc job.

nix/shell.nix
```diff
@@ -132,5 +132,6 @@ in pkgs.mkShell {
   PROTOC = "${pkgs.protobuf_25}";
 
   PULUMI_VERSION="${pkgs.pulumi-bin.version}";
+  KUBECTL_VERSION="${pkgs.kubectl.version}";
```
nicu-da, line 135: We're still missing passing this to the operator when we deploy it and also passing it to each stack objects from stack.ts

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -0,0 +1,179 @@
...
+        exit 1
+    fi
+
+    echo "--- Starting Cleanup Loop ---";
+
+    for NAMESPACE in $TARGET_NAMESPACES_LIST; do
+        echo "Processing namespace: $NAMESPACE";
+
+        BAD_PODS=$(
+          kubectl get pods -n "$NAMESPACE" -o json | \\
```
martinflorian-da, line 49: Did you consider just using something like this, so you don't need to `get` first?

```
kubectl delete pods --all-namespaces --field-selector=status.phase=Failed
```

nicu-da: I wouldn't pods can be failed for a whole lot of other reasons, and this shouldn't interfere with that

cluster/deployment/scratchnetb/config.yaml
pasindutennage-da: added deployment config.yaml

pasindutennage-da: added deployment config.yaml

cluster/pulumi/common/src/config/configSchema.ts
pasindutennage-da: moved into InfraConfigSchema

cluster/pulumi/infra/src/maintenance.ts
pasindutennage-da: Done

cluster/pulumi/infra/src/maintenance.ts
pasindutennage-da: Done

cluster/pulumi/multi-validator/src/installNode.ts
pasindutennage-da: Done; Moved

nix/shell.nix
pasindutennage-da: Done, 

```
KUBECTL_VERSION: {
              type: 'Literal',
              literal: {
                value: config.requireEnv('KUBECTL_VERSION'),
              },
            },
```

Issue #3006, opened 2025-10-31
sv-app: Add automation for executing weight change and withdraw instructions


dfordivam:

dfordivam:
Closing as this no longer need to be fixed for CIP-73

PR #3297, opened 2025-12-01
SV-app automation for VoteExecutionInstruction execution

dfordivam:
This PR includes

- SV-app automation for `VoteExecutionInstruction` execution and expiry
- Fixes to `ReceiveFaucetCouponTrigger` to ignore weight 0 licenses for `ValidatorLivenessActivityRecord` creation, 

- Added Integration tests to verify
  - Batched `ValidatorLicense` modification via vote
  - Check the rewards issuance with non-default `ValidatorLicense` weights
  - `UnclaimedRewards` creation with non-default weights
  - Confirm that weight 0 licenses still recorded activity without the coupon creation.

Fixes https://github.com/hyperledger-labs/splice/issues/3006


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/SvDsoStore.scala
```diff
@@ -1226,6 +1238,21 @@ object SvDsoStore {
           validator = Some(PartyId.tryFromProtoPrimitive(contract.payload.validator)),
         )
       },
+      mkFilter(VoteExecutionInstruction.COMPANION)(vi => vi.payload.dso == dso) { contract =>
+        val validatorParty = contract.payload.instruction match {
+          case changeWeight: EI_ValidatorLicenseChangeWeight =>
+            Some(PartyId.tryFromProtoPrimitive(changeWeight.validator))
+          case withdraw: EI_ValidatorLicenseWithdraw =>
+            Some(PartyId.tryFromProtoPrimitive(withdraw.validator))
+          case _ =>
```
meiersi-da, line 1247: please replace the default case with explicit matches, so that one future changes to the variant type we get compile errors pointing to updating the filter.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExecuteVoteInstructionTrigger.scala
```diff
@@ -0,0 +1,109 @@
...
+            .submit(
+              Seq(store.key.svParty),
+              Seq(store.key.dsoParty),
+              cmd,
+            )
+            .noDedup
+            .yieldUnit()
+            .map { _ =>
+              TaskSuccess(
+                show"Executed VoteExecutionInstruction for validator: $validatorParty"
```
meiersi-da: Can one see from the earlier logs what VoteExecutionInstruction it was that was executed? I believe yes, but would be good to check. In general, when writing new triggers it is valuable to at least once observe their logs on an integration test to see whether the logs are useful and match expectations.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/automation/ReceiveFaucetCouponTrigger.scala
```diff
@@ -79,6 +79,20 @@ class ReceiveFaucetCouponTrigger(
 
   override protected def completeTask(task: ReceiveFaucetCouponTrigger.Task)(implicit
       tc: TraceContext
+  ): Future[TaskOutcome] = {
+    val ReceiveFaucetCouponTrigger.Task(license, unclaimedRound) = task
+    license.payload.weight.toScala match {
+      case Some(x) if x == 0 =>
+        // weight is zero; don't exercise RecordValidatorLivenessActivity
+        Future.successful(TaskSuccess("weight is 0, liveness activity not recorded"))
```
meiersi-da:
```suggestion
        Future.successful(TaskSuccess("weight is 0, skipping the recording of liveness activity"))
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvTimeBasedOnboardingIntegrationTest.scala
```diff
@@ -225,5 +230,71 @@ class SvTimeBasedOnboardingIntegrationTest
...
+        }
+
+        // Wait for the instruction to be created
+        eventually() {
+          sv1Backend.participantClientWithAdminToken.ledger_api_extensions.acs
+            .filterJava(VoteExecutionInstruction.COMPANION)(
+              dsoParty,
+              _ => true,
+            ) should have length 1
+        }
```
meiersi-da: use `actAndCheck`. That automatically creates useful clue to debug failures.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorLicenseMetadataTimeBasedIntegrationTest.scala
```diff
@@ -197,4 +205,121 @@ class ValidatorLicenseMetadataTimeBasedIntegrationTest
       succeed
     }
   }
+
+  "expired ValidatorLivenessActivityRecord creates UnclaimedReward with correct weight" in {
```
meiersi-da, line 144: This is essentially an extended version of the test above `"unclaimed ValidatorLivenessActivityRecord contracts should be expired"`, or not?

Let's merge the two to save on CI resources.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletRewardsTimeBasedIntegrationTest.scala
```diff
@@ -96,5 +106,313 @@ class WalletRewardsTimeBasedIntegrationTest
...
+          None,
+        ),
+      )(
+        "license weight is updated",
+        _ => {
+          val licenses = getAliceLicense()
+          licenses should have length 1
+          licenses.head.data.weight.toScala.map(BigDecimal(_)) shouldBe Some(newWeight)
+        },
+      )
```
meiersi-da: You are using this vote change multiple times in your tests. I'd suggest to package it up into a utility to improve maintainability.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletRewardsTimeBasedIntegrationTest.scala
```diff
@@ -96,5 +106,313 @@ class WalletRewardsTimeBasedIntegrationTest
...
+          walletUsdToAmulet(-0.1 + faucetCouponAmountUsd),
+          walletUsdToAmulet(0.5 + faucetCouponAmountUsd),
+        ),
+      )
+    }
+
+    // This test verifies that the OpenMiningRoundSummary correctly sums
+    // ValidatorLivenessActivityRecord weights and that rewards are
+    // appropriately capped when total weights in a round are high.
+    // See: test_ValidatorLivenessWeightInRunNextIssuance for similar test in daml
```
meiersi-da: Integration tests tend to be quite expensive to run and maintain. It's not clear that this test here is required to be written as an integration test.

I'd suggest that you move the test for the correct computation of the mining round summary into the `SvDsoStoreTest`: there checking that the retrieval of validator liveness coupon weights works is what's key. 

The issuance caps are checked in the Daml tests. So I'd skip checking them again here. The likelihood of the Daml test passing, but this integration test failing due to a bug in the computation is very low.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvTimeBasedOnboardingIntegrationTest.scala
dfordivam: Fixed, internally the `modifyValidatorLicensesWithVoting` uses `actAndCheck`

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorLicenseMetadataTimeBasedIntegrationTest.scala
dfordivam: Yes, I have removed the test above this, as the new one is sufficient to confirm the unclaimed coupon expiry.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletRewardsTimeBasedIntegrationTest.scala
dfordivam: Right, I have added `modifyValidatorLicensesWithVoting`, (similar to the existing `scheduleDomainMigration` API)

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletRewardsTimeBasedIntegrationTest.scala
dfordivam: This is in fact necessary to verify that the `SummarizingMiningRoundTrigger` is properly making use of the `SvDsoStore` APIs. For checking the retrieval of validator liveness coupon weights we already have a test in `DbSvDsoStoreTest`. OTOH the daml test is checking the `runNextIssuance` which is basically a testing only API, and is not run while doing the actual issuance.
To be clear, if we revert the changes in `SummarizingMiningRoundTrigger` to use count of `ValidatorLivenessActivityRecord` (instead of weight), only this test would catch the issue.

But I have removed the test added above, as this one is checking that the "validator gets rewards proportional to license weight" also.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExecuteVoteInstructionTrigger.scala
dfordivam: Yes, doing a grep on `INFO` and `VoteExecutionInstruction` filters out most of noise and reveals the contract that triggered this. We can further narrow down it by `validatorParty`

```
148226:{"@timestamp":"2025-12-02T17:28:00.379+09:00","message":"Completing dso delegate task AssignedContract(\n  contract = Contract(\n    contractId = 00babf90a98533bb91ebcd09be304f537bde4f7045e6f9b54c31c08eff47fa8c95ca121220e66754e1fbe74b41db0e648bcf26c074228d1816495e393b1950958bb1374fab,\n    templateId = 47ccca819cef...:Splice.DSO.VoteExecution:VoteExecutionInstruction,\n    payload = Record(\n      fields = Seq(\n        PartyId(DSO-8270255a-8270255a::1220b0073dd3...),\n        EI_ValidatorLicenseChangeWeight(Record(fields = Seq(PartyId(alice-validator8270255a-1::122014f6bee1...), Numeric(10.0000000000)))),\n        Timestamp(ms = 114039996000, str = 1970-01-02T07:40:39.996Z),\n        Some(ContractId(00658a3dc14b09e0161198bbd50b175c255dacc788f9c6853229c15dc1bfaa40a7ca121220796994db0032a658d0ce4726b63c51ccee62d3467f9c7030b957227cdcb463b2))\n      )\n    ),\n    createdEventBlob = 0a03322e3112,\n    createdAt = 1970-01-01T07:40:39.996Z\n  ),\n  domain = global-domain::1220b0073dd3...\n) after waiting a delay of 38 ms","logger_name":"o.l.s.s.a.d.ExecuteVoteInstructionTrigger:ValidatorLicenseMetadataTimeBasedIntegrationTest/config=8270255a/SV=sv1","thread_name":"ValidatorLicenseMetadataTimeBasedIntegrationTest-8270255a-env-ec-979","level":"INFO","trace-id":"a4b70e31f26bc72ffb1fc0fb4c9567b8","span-name":"ExecuteVoteInstructionTrigger","span-id":"692299cbca3abf6b"}
148423:{"@timestamp":"2025-12-02T17:28:00.805+09:00","message":"Completed processing with outcome: Executed VoteExecutionInstruction for validator: alice-validator8270255a-1::122014f6bee1...","logger_name":"o.l.s.s.a.d.ExecuteVoteInstructionTrigger:ValidatorLicenseMetadataTimeBasedIntegrationTest/config=8270255a/SV=sv1","thread_name":"ValidatorLicenseMetadataTimeBasedIntegrationTest-8270255a-env-ec-993","level":"INFO","trace-id":"a4b70e31f26bc72ffb1fc0fb4c9567b8","span-name":"ExecuteVoteInstructionTrigger","span-id":"692299cbca3abf6b"}
```

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/SvDsoStore.scala
dfordivam: I tried it but it fails with the following error. Am I missing some way to make this work?

```
 match may not be exhaustive.
[error] It would fail on the following input: (x: org.lfdecentralizedtrust.splice.codegen.java.splice.dso.voteexecution.ExecutionInstruction forSome x not in (org.lfdecentralizedtrust.splice.codegen.java.splice.dso.voteexecution.executioninstruction.EI_ValidatorLicenseChangeWeight, org.lfdecentralizedtrust.splice.codegen.java.splice.dso.voteexecution.executioninstruction.EI_ValidatorLicenseWithdraw))
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
```diff
@@ -191,6 +196,93 @@ trait SvTestUtil extends TestCommon {
...
+        )(
+          "vote request has been created",
+          _ =>
+            onlyModifyValidatorLicensesVoteRequests(
+              svToCreateVoteRequest.listVoteRequests()
+            ).loneElement,
+        )
+
+        if (svsToCastVotes.nonEmpty) {
+          loggerFactory.suppressErrors {
```
meiersi-da: That's a very heavy hammer, which risks hiding actual errors. Why is this required?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
```diff
@@ -191,6 +196,93 @@ trait SvTestUtil extends TestCommon {
...
+            }
+          case _ => false
+        }
+      }
+
+    actAndCheck(timeUntilSuccess = 60.seconds)(
+      "Voting on validator license modification", {
+        val (_, voteRequest) = actAndCheck(
+          "Creating vote request",
+          eventuallySucceeds() {
```
meiersi-da: why and how can it fail in this step?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
```diff
@@ -191,6 +196,93 @@ trait SvTestUtil extends TestCommon {
...
+              case _: SRARC_ModifyValidatorLicenses => true
+              case _ => false
+            }
+          case _ => false
+        }
+      }
+
+    actAndCheck(timeUntilSuccess = 60.seconds)(
+      "Voting on validator license modification", {
+        val (_, voteRequest) = actAndCheck(
```
meiersi-da: nested actAndCheck is a code smell. Why is it required?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
```diff
@@ -191,6 +196,93 @@ trait SvTestUtil extends TestCommon {
...
+          case arc: ARC_DsoRules =>
+            arc.dsoAction match {
+              case _: SRARC_ModifyValidatorLicenses => true
+              case _ => false
+            }
+          case _ => false
+        }
+      }
+
+    actAndCheck(timeUntilSuccess = 60.seconds)(
```
meiersi-da: This is a rather long duration. Might be required though...

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExecuteVoteInstructionTrigger.scala
```diff
@@ -0,0 +1,109 @@
...
+      controller: String,
+  )(implicit tc: TraceContext): Future[TaskOutcome] = {
+    val instructionType = instruction.payload.instruction
+
+    val validatorParty = instructionType match {
+      case changeWeight: EI_ValidatorLicenseChangeWeight =>
+        PartyId.tryFromProtoPrimitive(changeWeight.validator)
+      case withdraw: EI_ValidatorLicenseWithdraw =>
+        PartyId.tryFromProtoPrimitive(withdraw.validator)
+      case _ =>
```
meiersi-da, line 65: Please use an explicit match here to make the compilation fail on future changes.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExecuteVoteInstructionTrigger.scala
meiersi-da:
```suggestion
                show"Executed vote instruction affecting validator: $validatorParty"
```

project/ignore-patterns/canton_network_test_log.ignore.txt
```diff
@@ -133,4 +133,7 @@ Circuit breaker .* tripped after .* failures.*(Command|Splice)CircuitBreakerTest
 # This is more likely in test code where we run multiple apps against the same database, all trying to create the same index.
 Index .* should be created and is invalid, dropping it
 
+# Expected as the test intentionally creates a vote for a party without ValidatorLicense
+ValidatorLicense not found for validator.*SvTimeBasedOnboardingIntegrationTest
```
meiersi-da: use `SuppressingLogger.assertLog` (or one of its variants) to check that the expected error is raised, and to avoid triggering the error check.

Search the test sources for examples.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvValidatorLicenseIntegrationTest.scala
```diff
@@ -111,4 +115,66 @@ class SvValidatorLicenseIntegrationTest
...
+      _ => {
+        val licenses = getLicenses(newParty2)
+        licenses should have length 1
+        licenses.head.data.weight.toScala shouldBe None
+      },
+    )
+
+    // Create a vote for batch modification, with both weight change and withdrawal
+    import env.executionContext
+    modifyValidatorLicensesWithVoting(
```
meiersi-da:
```suggestion
    modifyValidatorLicensesAndCheck(
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
```diff
@@ -191,6 +196,93 @@ trait SvTestUtil extends TestCommon {
     )
   }
 
+  def modifyValidatorLicensesWithVoting(
+      svToCreateVoteRequest: SvAppBackendReference,
+      svsToCastVotes: Seq[SvAppBackendReference],
```
meiersi-da, line 201: AFAIS, all your calls set this argument to `Seq.empty`. I'd then suggest to remove it and call out that the function only works for a single SV setup

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
```diff
@@ -191,6 +196,93 @@ trait SvTestUtil extends TestCommon {
...
+                  eventually() {
+                    try {
+                      sv.castVote(
+                        voteRequest.contractId,
+                        true,
+                        "url",
+                        "description",
+                      )
+                    } catch {
+                      case NonFatal(e) => fail(e)
```
meiersi-da: This looks fishy. Could it be that the problem is that you are sometimes acting on stale vote requests due to the previous vote having been applied, but not yet propagated to the next SV?

I'd suggest to refactor the code so that it loops through all SVs, and checks for each that the updated vote is visible to all SVs before advancing to the next one.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletRewardsTimeBasedIntegrationTest.scala
meiersi-da: OK. Let's keep it.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExecuteVoteInstructionTrigger.scala
meiersi-da: Just saw your other message. I'll check on my machine how to do this.

meiersi-da: OK. I understand now my confusion: we do have proper pattern matching support for protobuf `oneof`, but we don't for the Daml variants, as we use the Java codegen. Your code is thus right. Sorry for the noise.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExecuteVoteInstructionTrigger.scala
meiersi-da: thx for checking.

meiersi-da: Note also that you might want to setup `lnav` for perusing the logs with much better rendering: https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#setting-up-lnav-to-inspect-canton-and-cometbft-logs

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/SvDsoStore.scala
meiersi-da: jup. understood now that we don't have proper pattern matching support due to using the Java codegen for Daml.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExecuteVoteInstructionTrigger.scala
dfordivam: Nice!
I was not aware of this

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
dfordivam: There is one usage with multiple SVs in `SvTimeBasedOnboardingIntegrationTest`.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
dfordivam: I based the API on an existing one so not quite sure why the checks were added in the first place.
But now I have changed it to be minimal, I guess if it fails to work then subsequent guards could be implemented based on the situation.

project/ignore-patterns/canton_network_test_log.ignore.txt
dfordivam: fixed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
dfordivam: For the time being I got rid of it, let's see if it is really necessary based on CI results.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
dfordivam: I removed the nested here. But from the API caller's perspective there is still a nested `actAndCheck`. I had a look at code elsewhere and found similar nested use of `actAndCheck` (in caller and internally in API), so it should be ok here also.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
dfordivam: Removed this

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
dfordivam: Removed this

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvValidatorLicenseIntegrationTest.scala
```diff
@@ -111,4 +115,72 @@ class SvValidatorLicenseIntegrationTest
...
+      "ValidatorLicense granted",
+      _ => {
+        val licenses = getLicenses(newParty2)
+        licenses should have length 1
+        licenses.head.data.weight.toScala shouldBe None
+      },
+    )
+
+    // Create a vote for batch modification, with both weight change and withdrawal
+    import env.executionContext
```
meiersi-da:
```suggestion
```
seems like a left-over

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
```diff
@@ -191,6 +196,71 @@ trait SvTestUtil extends TestCommon {
     )
   }
 
+  def modifyValidatorLicenses(
+      svToCreateVoteRequest: SvAppBackendReference,
+      svsToCastVotes: Seq[SvAppBackendReference],
+      changes: Seq[ValidatorLicenseChange],
+  )(implicit
+      ec: ExecutionContextExecutor
```
meiersi-da:
```suggestion
      ec: ExecutionContext
```

is what we usually use. Any particular reason for using the the `ExecutionContextExecutor` ?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SvTestUtil.scala
```diff
@@ -191,6 +196,71 @@ trait SvTestUtil extends TestCommon {
...
+      "vote request has been created",
+      _ =>
+        onlyModifyValidatorLicensesVoteRequests(
+          svToCreateVoteRequest.listVoteRequests()
+        ).loneElement,
+    )
+
+    if (svsToCastVotes.nonEmpty) {
+      svsToCastVotes.parTraverse { sv =>
+        Future {
```
meiersi-da: Not sure that the futures are worth it. The all contend in modifying the same vote request. A simple for loop would suffice.

Your call though.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvValidatorLicenseIntegrationTest.scala
meiersi-da: ignore.

PR #3307, opened 2025-12-01
Update Canton to 3.5.9

martinflorian-da:
[ci]

Among other things, gets us https://github.com/DACH-NY/canton/pull/29563

Needed for https://github.com/DACH-NY/canton-network-internal/issues/2842

### DIFF
--- a/docs/src/release_notes.rst
+++ b/docs/src/release_notes.rst
@@ -7,6 +7,11 @@
 
 .. release-notes:: upcoming
 
+  - Validator
+
+    - Fix a bug introduced in 0.5.0/0.5.1 that could cause participant pruning to prune active data.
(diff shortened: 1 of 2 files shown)


docs/src/release_notes.rst
```diff
@@ -7,6 +7,11 @@
 
 .. release-notes:: upcoming
 
+  - Validator
+
+    - Fix a bug introduced in 0.5.0/0.5.1 that could cause participant pruning to prune active data.
```
isegall-da, line 12: Don't we want to mention the performance issue that triggered the expedition?

martinflorian-da: Oh yes definitely... but that doesn't need testing on CILR, we can follow up tomorrow?

martinflorian-da: (that fix was already on our main)

PR #3326, opened 2025-12-03
[ci] Remove cpu limit in istio proxy

ray-roestenburg-da:
linux and k8s already enforce fair CPU sharing through requests and scheduling, preventing any container from monopolizing or destabilizing the node. CPU's are not a rigid resource (best effort, throttling / scheduling), unlike memory which is rigid.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/expected/infra/expected.json
+++ b/cluster/expected/infra/expected.json
@@ -2269,6 +2269,17 @@
         },
         "autoscaleMax": 30,
         "autoscaleMin": 2,
+        "defaults": {
+          "global": {
+            "proxy": {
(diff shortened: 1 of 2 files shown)


ray-roestenburg-da:
The devil is always in the details, it is not recommended to do this, because of how envoy decides how many threads ti start based on limit, see https://github.com/istio/istio/issues/51456 so the "don't set cpu limits at all' doesn't always apply.. (it's probably good advice for our own containers, but not a blanket statement for everything)

PR #3333, opened 2025-12-04
implement resolved config versioning

mblaze-da:
In the current version the resolved config is generated using the `cluster/deployment/update-resolved-config` make target. In the design doc we've agreed to reuse the "expected files muscle memory". I'm willing to simply include `cluster/deployment/update-resolved-config` in `cluster/pulumi/update-expected`. The reason I didn't do this yet is that I was worried about adding new steps and new ways to fail to make targets and CI pipeline steps that already have a well established meaning. I'd like to ask you again for opinion on this matter. Is it better to avoid introducing a new user facing make target or to avoid redefining existing processes? I'll update the doc accordingly when we decide. BTW, I made it so the pre-commit hooks update those configs in a similar fashion that they fix formatting and linting issues.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -471,6 +471,15 @@ they cover. Please see the [jest "getting started"](https://jestjs.io/docs/getti
 for tips on how to write tests. The type definitions for jest come from the `@jest/globals`
 package usage of which is described [here](https://jestjs.io/docs/getting-started#type-definitions).
 
+### Versioning resolved configuration
+To make sure that changes to cluster configuration are made consciously in light of various
+modularization and reuse mechanisms in the config loader a complete and fully resolved version of
+the cluster configuration file is versioned. Such a file is generated as `config.resolved.yaml`
+for each clusters' main `config.yaml` using `make cluster/deployment/update-resolved-config -j`.
+The `cluster/pulumi/update-expected` make target also includes resolved configuration update.
+Appropriate checks are conducted in pre-commit hooks and later in CI to make sure that these
+resolved configuration files are kept in sync with the original unresolved ones.
+
 ### Pulumi state checks
 To make sure that the impact of changes in the Pulumi resource definitions is well understood we
 rely on checked in `expected` files that need to be updated whenever the expected deployment state changes.


.pre-commit-config.yaml
```diff
@@ -50,9 +50,17 @@ repos:
         language: system
         entry: "scripts/fix-ts.py"
         types_or: [javascript, jsx, ts, tsx]
-        exclude: "(/[.]prettierrc[.]cjs$|jest.config.ts$)"
-      - id: pulumi_config
-        name: check pulumi configs
+        # FIXME: The current implementation of fix-ts.py makes it impossible to format files in
```
martinflorian-da: You probably already saw that, but static checks don't allow `FIXME`s... You can create a tech debt issue and make a `TODO(#...)` for it.

cluster/pulumi/resolvedConfigEmitter.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { loadClusterYamlConfig, getMainConfigPath } from '@lfdecentralizedtrust/splice-pulumi-common/src/config/configLoader';
+import { writeFileSync } from 'fs';
+import { dump } from 'js-yaml';
+import { dirname, resolve } from 'path';
+
+const config = loadClusterYamlConfig();
+const header = '# This file is generated for preview. Do not edit directly.\n';
```
martinflorian-da:
```suggestion
const header = '# This file is generated automatically. Do not edit directly.\n';
```

bikeshed

martinflorian-da (commented):
LGTM, thanks! No you only need to convince CI as well :)

> The reason I didn't do this yet is that I was worried about adding new steps and new ways to fail to make targets and CI pipeline steps that already have a well established meaning.

Fair concern; seems like a risk we need to take... no good way to test before merging.

From a quick search it seems that we only do `update-expected` from CI (as opposed to, say `pulumi-test`) when we do backports; and there it's probably a feature that you also update the new "resolved" files.

> Is it better to avoid introducing a new user facing make target

Wait but you already have that? I'm definitely going to use just `update-resolved-config` whenever I want a quick feedback loop...

> or to avoid redefining existing processes?

It seems nicer to add this to `update-expected` (on the top level; the per-project ones obviously don't make sense). But not married to this; other people may have different opinions.

> BTW, I made it so the pre-commit hooks update those configs in a similar fashion that they fix formatting and linting issues.

SGTM

martinflorian-da (approved):
Very nice, thank you!

PR #3347, opened 2025-12-05
Fix rateLimitIsEnforced

rautenrieth-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/6581.

https://github.com/hyperledger-labs/splice/commit/984ec93755089ad74e10b193ee1b12722adae6fc switched to new endpoint, the assert should change with it.


nicu-da (approved):
thx!

PR #3386, opened 2025-12-10
Avoid races between sv reward triggers in test

rautenrieth-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/6355


nicu-da (approved):
thx!

PR #3388, opened 2025-12-10
update istio and expose istiod chart values in the configuration

mblaze-da:
According to https://github.com/DACH-NY/canton-network-internal/issues/2791 I updated istio and made it possible to override the istiod chart values from the configuration to allow for testing of the proposed tweaks. The definition of an optional `DestinationRule` for high performance GRPC will come in a separate PR.

### DIFF
--- a/cluster/deployment/mock/config.yaml
+++ b/cluster/deployment/mock/config.yaml
@@ -233,6 +233,17 @@ infra:
       spec:
         key: value
         anotherKey: anotherValue
+  istio:
+    istiodValues: # example istiod overrides
+      global:
+        proxy:
+          resources:
+            requests:
+              cpu: 500m
+              memory: 512Mi
+            limits:
+              cpu: 2000m
+              memory: 2Gi
 cluster:
   nodePools:
     apps:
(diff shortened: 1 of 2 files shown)


nicu-da (approved):
Niice! Would be nice to add a possible override to the mock cluster config.yaml to have an example of it.

cluster/pulumi/infra/src/istio.ts
```diff
@@ -69,6 +70,60 @@ function configureIstiod(
   ingressNs: k8s.core.v1.Namespace,
   base: k8s.helm.v3.Release
 ): k8s.helm.v3.Release {
+  // https://artifacthub.io/packages/helm/istio-official/istiod
+  const defaultValues = {
```
martinflorian-da, line 74: Hm it seems like we might want to move this into a yaml file as well, but happy to do it later to get this in more quickly/

martinflorian-da (approved):
Thanks a lot! Very timely

cluster/pulumi/infra/src/istio.ts
mblaze-da: That was my intention initially but I decided to hold up until we have a working example of common/default included file in both splice and internal.

martinflorian-da: It seems that for what we want to use this for short-term we can live with hardcoded defaults. Let's try to get this change on the MainNet branch next, in case we need to fix things quickly.

PR #3402, opened 2025-12-12
add destination rules for sequencer high performance gRPC

mblaze-da:
fixes https://github.com/DACH-NY/canton-network-internal/issues/2791

I am not very confident about these changes. I had some difficulties figuring out the sequencer hostname and I'm still not 100% sure I got it right. I'm also not sure if my assumptions about how to iterate through SVs and migrations are correct.

### DIFF
--- a/cluster/pulumi/infra/src/istio.ts
+++ b/cluster/pulumi/infra/src/istio.ts
@@ -641,10 +641,10 @@ function configureSequencerHighPerformanceGrpcDestinationRules(
   return [
     ...(function* () {
       for (const migration of DecentralizedSynchronizerUpgradeConfig.runningMigrations()) {
-        for (const sv of coreSvsToDeploy) {
+        for (const sv of allSvsToDeploy) {
           yield configureSequencerHighPerformanceGrpcDestinationRule(
             ingressNs,
-            sv.ingressName,
+            sv.nodeName,
             migration.id
           );
         }
@@ -655,11 +655,11 @@ function configureSequencerHighPerformanceGrpcDestinationRules(
 
 function configureSequencerHighPerformanceGrpcDestinationRule(
   ingressNs: k8s.core.v1.Namespace,
-  ingressName: string,
+  nodeName: string,
   migrationId: number
 ): k8s.apiextensions.CustomResource {
   const sequencerName = `global-domain-${migrationId}-sequencer`;
-  const ruleName = `${ingressName}-${sequencerName}-high-perf-grpc-rule`;
+  const ruleName = `${nodeName}-${sequencerName}-high-perf-grpc-rule`;
   return new k8s.apiextensions.CustomResource(ruleName, {
     apiVersion: 'networking.istio.io/v1beta1',
     kind: 'DestinationRule',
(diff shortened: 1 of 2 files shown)


cluster/pulumi/infra/src/istio.ts
```diff
@@ -635,6 +635,60 @@ function configurePublicInfo(ingressNs: k8s.core.v1.Namespace): k8s.apiextension
     : [];
 }
 
+function configureSequencerHighPerformanceGrpcDestinationRules(
+  ingressNs: k8s.core.v1.Namespace
+): Array<k8s.apiextensions.CustomResource> {
+  return [
+    ...(function* () {
+      for (const migration of DecentralizedSynchronizerUpgradeConfig.runningMigrations()) {
+        for (const sv of coreSvsToDeploy) {
```
martinflorian-da: Why not `allSvsToDeploy`? I guess it doesn't matter for MainNet (only core SVs there), but for consistency?

cluster/pulumi/infra/src/istio.ts
```diff
@@ -635,6 +635,60 @@ function configurePublicInfo(ingressNs: k8s.core.v1.Namespace): k8s.apiextension
...
+  const ruleName = `${ingressName}-${sequencerName}-high-perf-grpc-rule`;
+  return new k8s.apiextensions.CustomResource(ruleName, {
+    apiVersion: 'networking.istio.io/v1beta1',
+    kind: 'DestinationRule',
+    metadata: {
+      name: ruleName,
+      namespace: ingressNs.metadata.name,
+    },
+    spec: {
+      host: `${sequencerName}.${ingressName}.svc.cluster.local`,
```
martinflorian-da: To be clear: The intention here is to use k8s-internal addresses, not the external one we define via `VirtualService`?

If so... can't you just check on a scratchnet cluster or CILR if you got the hostname right? I'd launch a shell on some validator pod and try reaching the endpoint from there via curl or grpcurl.

(Not very confident about the networking stuff here myself...)

mblaze-da: I intended to set the internal address here. I tested it using the method you've suggested and corrected the implementation.

cluster/pulumi/infra/src/istio.ts
```diff
@@ -635,6 +635,60 @@ function configurePublicInfo(ingressNs: k8s.core.v1.Namespace): k8s.apiextension
...
+  return new k8s.apiextensions.CustomResource(ruleName, {
+    apiVersion: 'networking.istio.io/v1beta1',
+    kind: 'DestinationRule',
+    metadata: {
+      name: ruleName,
+      namespace: ingressNs.metadata.name,
+    },
+    spec: {
+      host: `${sequencerName}.${nodeName}.svc.cluster.local`,
+      trafficPolicy: {
```
martinflorian-da, line 672: @nicu-da Can you confirm that we still consider this a useful change?

nicu-da: Yup I think it's worth testing, we expect the traffic on the sequencers to increase and from what I've read this seems like the way to handle it in istio

PR #3417, opened 2025-12-16
fix GHA runner version updater workflow and enable it

mblaze-da:
These changes fix the workflow after initial implementation which was required on main to start testing. After these changes are merged the workflow will be enabled on schedule on main.


julientinguely-da (approved):
Lgtm, thanks!

Issue #767, opened 2025-06-05
Make the maximum Limit value configurable


canton-network-da:
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#12610). Original author: @OriolMunoz-da ]

## What is this about?

Currently, both PageLimit and HardLimit have a hard-coded `MaxPageSize` of 1000 elements.

## How important is this and why?

This should be made configurable. We might need to adjust it at some point in the future, or by-case.

[Original discussion](https://github.com/DACH-NY/canton-network-node/pull/12563#discussion_r1620690666)

martinflorian-da:
> this would have helped for

@cocreature volunteered to double check if this is actually true

PR #3423, opened 2025-12-17
[ci] bulk-storage support for acs snapshots

isegall-da:
Part of #3429 
Implements dumping an ACS snapshots to S3 storage, with a unit test. Not yet integrated with anything else.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
@@ -59,6 +59,7 @@ class S3BucketConnection(
         AsyncRequestBody.fromBytes(content.array()),
       )
       .asScala
+      // TODO(#3429): consider checking the checksum from the response
       .map(_ => ())
   }
 }
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
@@ -49,6 +49,7 @@ case class ZstdGroupedWeight(minSize: Long) extends GraphStage[FlowShape[ByteStr
       new ZstdDirectBufferCompressingStreamNoFinalizer(tmpBuffer, compressionLevel)
 
     def compress(input: ByteString): ByteString = {
+      // TODO(#3429): use a buffer pool to avoid allocating a new ByteBuffer for each compress call
       val inputBB = ByteBuffer.allocateDirect(input.size)
       inputBB.put(input.toArrayUnsafe())
       inputBB.flip()
@@ -75,7 +76,8 @@ case class ZstdGroupedWeight(minSize: Long) extends GraphStage[FlowShape[ByteStr
 
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
     new GraphStageLogic(shape) with InHandler with OutHandler {
-      // TODO(#3429): consider implementing a pool of tmp buffers to avoid allocating a new one for each stage
+      // TODO(#3429): consider implementing a pool of tmp buffers to avoid allocating a new one for each stage,
+      //   and moving some initialization into preStart(), otherwise we allocate even if the stream never runs or fails before starting.
       private val tmpBuffer = ByteBuffer.allocateDirect(zstdTmpBufferSize)
       private val zstd = new AtomicReference[ZSTD](new ZSTD(tmpBuffer, 3))
       private val state: AtomicReference[State] = new AtomicReference[State](State.empty())


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,111 @@
...
+    dbReadChunkSize: Int,
+    maxFileSize: Long,
+)
+
+object BulkStorageConfigs {
+  val bulkStorageConfigV1 = BulkStorageConfig(
+    1000,
+    (64 * 1024 * 1024).toLong,
+  )
+  val bulkStorageTestConfig = BulkStorageConfig(
```
ray-roestenburg-da: This can be defined in a test instead of here?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,111 @@
...
+    val s3Connection: S3BucketConnection,
+    override val loggerFactory: NamedLoggerFactory,
+)(implicit actorSystem: ActorSystem, tc: TraceContext, ec: ExecutionContext)
+    extends NamedLogging {
+
+  def getAcsSnapshotChunk(
+      migrationId: Long,
+      timestamp: CantonTimestamp,
+      after: Option[Long],
+  ): Future[(Position, ByteString)] = {
```
ray-roestenburg-da, line 64: Prefer a case class definition over a tuple, `case class AcsSnapshotChunk(position: Position, data: ByteString)`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotSource.scala
```diff
@@ -0,0 +1,73 @@
...
+import org.lfdecentralizedtrust.splice.scan.store.AcsSnapshotStore
+import org.lfdecentralizedtrust.splice.scan.store.AcsSnapshotStore.QueryAcsSnapshotResult
+import org.lfdecentralizedtrust.splice.store.PageLimit
+import org.lfdecentralizedtrust.splice.store.events.SpliceCreatedEvent
+
+import java.util.concurrent.atomic.AtomicReference
+import scala.concurrent.ExecutionContext
+import scala.util.{Failure, Success}
+
+case class AcsSnapshotSource(
```
ray-roestenburg-da: I don't think you need to write your own graphstage for this, you can do this with unfoldAsync (or statefulMapConcat but unfoldAsync is easier probably)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
```diff
@@ -0,0 +1,68 @@
...
+  // Use only for testing, when the object size is known to be small
+  def readFullObject(key: String): ByteBuffer = {
+    val obj = s3Client.getObject(GetObjectRequest.builder().bucket(bucketName).key(key).build())
+    val bytes = obj.readAllBytes()
+    val ret = ByteBuffer.allocateDirect(bytes.length)
+    ret.put(bytes)
+  }
+
+  // Writes a full object from memory into an s3 object
+  def writeFullObject(key: String, content: ByteBuffer)(implicit tc: TraceContext) = {
```
ray-roestenburg-da: would be nice to add the return type here since it's a public def, is this a blocking call?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
```diff
@@ -0,0 +1,110 @@
...
+  private object State {
+    def empty(): State = State(ByteString.empty, minSize)
+  }
+
+  class ZSTD(
+      val tmpBufferSize: Int,
+      val compressionLevel: Int = 3,
+  ) {
+
+    val tmpBuffer = ByteBuffer.allocateDirect(tmpBufferSize)
```
ray-roestenburg-da: This is allocated outside of JVM, not GC-ed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
```diff
@@ -0,0 +1,110 @@
...
+        bytes ++ bs,
+        left - bs.size,
+      )
+
+  }
+  private object State {
+    def empty(): State = State(ByteString.empty, minSize)
+  }
+
+  class ZSTD(
```
ray-roestenburg-da, line 43: I think this should be Closeable, and on stream completion / termination you need to make sure all resources are closed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
```diff
@@ -0,0 +1,110 @@
...
+import org.apache.pekko.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import org.apache.pekko.util.ByteString
+
+import java.nio.ByteBuffer
+import java.util.concurrent.atomic.AtomicReference
+
+/** A Pekko GraphStage that zstd-compresses a stream of bytestrings, and splits the output into zstd objects of size (minWeight + delta).
+  * Somewhat similar to Pekko's built-in GroupedWeight, but outputs valid zstd compressed objects.
+  */
+case class ZstdGroupedWeight(minSize: Long) extends GraphStage[FlowShape[ByteString, ByteString]] {
```
ray-roestenburg-da, line 17: preStart and postStop can probably be used to init and cleanup the compressingStream (in GraphStageLogic) (I should have made this comment on the GraphStageLogic further down).
Especially postStop important to always close the compressingStream.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
ray-roestenburg-da: I think you are likely currently leaking native memory and not cleaning up the compressingStream correctly

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
```diff
@@ -0,0 +1,110 @@
...
+  class ZSTD(
+      val tmpBufferSize: Int,
+      val compressionLevel: Int = 3,
+  ) {
+
+    val tmpBuffer = ByteBuffer.allocateDirect(tmpBufferSize)
+    val compressingStream =
+      new ZstdDirectBufferCompressingStreamNoFinalizer(tmpBuffer, compressionLevel)
+
+    def compress(input: ByteString): ByteString = {
```
ray-roestenburg-da, line 51: you likely need to add some exception handling for compression, and make sure you close the stream if it fails

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotBulkStorageTest.scala
```diff
@@ -0,0 +1,185 @@
...
+    test.andThen({ case _ => Seq("docker", "stop", "s3mock").! })
+  }
+
+  def readUncompressAndDecode(
+      s3BucketConnection: S3BucketConnection
+  )(s3obj: S3Object): Array[httpApi.CreatedEvent] = {
+    val compressed = s3BucketConnection.readFullObject(s3obj.key())
+    val uncompressed = ByteBuffer.allocateDirect(compressed.capacity() * 200)
+    compressed.flip()
+    new ZstdDirectBufferDecompressingStream(compressed).read(uncompressed)
```
ray-roestenburg-da: you probably always need to close this once done.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotBulkStorageTest.scala
```diff
@@ -0,0 +1,185 @@
...
+      "adobe/s3mock",
+    ).!
+    test.andThen({ case _ => Seq("docker", "stop", "s3mock").! })
+  }
+
+  def readUncompressAndDecode(
+      s3BucketConnection: S3BucketConnection
+  )(s3obj: S3Object): Array[httpApi.CreatedEvent] = {
+    val compressed = s3BucketConnection.readFullObject(s3obj.key())
+    val uncompressed = ByteBuffer.allocateDirect(compressed.capacity() * 200)
```
ray-roestenburg-da, line 118: This leaks

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/Limit.scala
```diff
@@ -61,6 +61,13 @@ object PageLimit {
     )
 }
 
+/** Limit with no constraints. Must not be used for production, use only for testing.
+  */
+case class UnboundLimit private (limit: Int) extends Limit
```
ray-roestenburg-da: Maybe rather do   // TODO (#767): make configurable on Limit.MaxPage so you don't need this?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,111 @@
...
+        limit = HardLimit.tryCreate(config.dbReadChunkSize),
+        Seq.empty,
+        Seq.empty,
+      )
+    } yield {
+      val encoded = snapshot.createdEventsInPage.map(event =>
+        CompactJsonScanHttpEncodings.javaToHttpCreatedEvent(event.eventId, event.event)
+      )
+      val contractsStr = encoded.map(_.asJson.noSpacesSortKeys).mkString("\n") + "\n"
+      val contractsBytes = ByteString(contractsStr)
```
ray-roestenburg-da: `contractsStr.getBytes(StandardCharSets.UTF_8)`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,111 @@
...
+      )
+      (snapshot.afterToken.fold(End: Position)(Index(_)), contractsBytes)
+    }
+
+  }
+
+  def dumpAcsSnapshot(migrationId: Long, timestamp: CantonTimestamp): Future[Unit] = {
+
+    @SuppressWarnings(Array("org.wartremover.warts.Var"))
+    var idx = 0
```
ray-roestenburg-da: This needs to be a volatile member or needs to be atomic integer or reference, you are mutating it concurrently from a future (from a thread on the executioncontext that runs the stream)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
ray-roestenburg-da: you need to at least deref the tmpBuffer at some point so that it becomes unreachable and can get deleted. (there is still the problem that you don't know when the JVM will actually free it)
there is cleanup code for this kind of stuff but that's not portable I think.  And once you use directbuffers you really need to set -XX:MaxDirectMemorySize to you know when this happens.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
```diff
@@ -0,0 +1,110 @@
...
+      tmpBuffer.flip()
+      val result = ByteString.fromByteBuffer(tmpBuffer)
+      tmpBuffer.clear()
+      compressingStream.close()
+      result
+    }
+  }
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with InHandler with OutHandler {
```
ray-roestenburg-da, line 78: So this needs s preStart to create the buffer and a postStop to deref it (and close the zstd stream)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
```diff
@@ -0,0 +1,68 @@
...
+      .build()
+    s3Client.putObject(
+      putObj,
+      RequestBody.fromBytes(content.array()),
+    )
+  }
+}
+
+object S3BucketConnection {
+  def apply(s3Config: S3Config, bucketName: String, loggerFactory: NamedLoggerFactory) = {
```
ray-roestenburg-da:
```suggestion
  def apply(s3Config: S3Config, bucketName: String, loggerFactory: NamedLoggerFactory): S3BucketConnection = {
```

ray-roestenburg-da (commented):
Going in the right direction IMHO, some initial comments. The possible memory leak, handling exceptions, and always making sure the zstd object is closed are my most important comments.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
isegall-da: The problem was that zstd asserts that the buffer is direct, but I'll definitely look into not leaking it at least.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: In general I'd agree, but this is private (that I forgot to label), and then used in unfoldAsync which expects a tuple, so the case class will then immediately need to be unwrapped into the tuple anyway.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: hmm.. I originally had it an atomicInt, but I'm mutating it from a future with parallelism 1, so convinced myself that it can be a var. But you're probably right than it's better to just be prudent and revert to that.

Just for my education, is my reasoning that parallelism 1 is good enough solid, or am I still missing something?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotSource.scala
isegall-da: Argh, sorry, this is an unused leftover. I indeed already reimplemented it with unfoldAsync.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
isegall-da: The reason I omitted the return type is that I didn't find anything useful in it, so I don't actually care about returning it. Would it be cleaner to add a `()` in the end to make that explicit?

Indeed, it is blocking, hence the `Future { }` wrapper in the call site. Added to the comment.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotBulkStorageTest.scala
isegall-da: Probably a stupid question, but once the method terminates, wouldn't this be dereferenced, thus GC'ed? If not, how do I explicitly avoid this leakage?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
ray-roestenburg-da: ok makes sense

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
ray-roestenburg-da: yes it's easier to read with ()

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
ray-roestenburg-da: the 1 does not matter you are still accessing the var from that one other thread which isn't allowed in principle. I shoot myself in the foot every time I want to be too clever or loose around these rules, as well as the fact that code changes and the next person doesn't see the special case, or starts copying a bad way of doing things.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
isegall-da: Added a retry on the whole pipeline

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/Limit.scala
isegall-da: Fair. I half-did #767 by making it an argument, but not fully configurable (e.g. by an app config, so this tackles only the use case of overriding the max per usage in code).

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,149 @@
...
+
+case class BulkStorageConfig(
+    dbReadChunkSize: Int,
+    maxFileSize: Long,
+)
+
+object BulkStorageConfigs {
+  val bulkStorageConfigV1 = BulkStorageConfig(
+    1000,
+    (64 * 1024 * 1024).toLong,
```
ray-roestenburg-da: nitpick but you can also do `64L * 1024 * 1024`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,149 @@
...
+  )
+}
+
+sealed trait Position
```
ray-roestenburg-da: nitpick: Maybe nice to put these in an object

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,149 @@
...
+      timestamp: CantonTimestamp,
+      after: Option[Long],
+  ): Future[(Position, ByteString)] = {
+    for {
+      snapshot <- acsSnapshotStore.queryAcsSnapshot(
+        migrationId,
+        snapshot = timestamp,
+        after,
+        limit = HardLimit.tryCreate(config.dbReadChunkSize),
+        Seq.empty,
```
ray-roestenburg-da, line 71: nit, since you do limit = , maybe nice to do partyIds = Seq.empty, templates = Seq.empty for clarity

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,149 @@
...
+        }
+        .via(ZstdGroupedWeight(config.maxFileSize))
+        // Add a buffer so that the next object continues accumulating while we write the previous one
+        .buffer(
+          1,
+          OverflowStrategy.backpressure,
+        )
+        .mapAsync(1) { zstdObj =>
+          val objectKey = s"snapshot_$idx.zstd"
+          Future {
```
ray-roestenburg-da: You should use a blocking threadpool for this future, since it blocks, now you're likely using the default dispatcher. sadly we can't use connector which does use proper non-blocking I/O. So you need to do 
Future{}(ec) and get the ec which is used for blocking I/O in canton or get one from pekko configured as a dispatcher.

ray-roestenburg-da: of course it depends on the ec implicit on this class but it's not visible that way that this class needs a separate threadpool for blocking I/O. maybe better just too create one threadpool fixed for this

ray-roestenburg-da: Probably should wrap it in `blocking { }` anyway (the writeFullObject). And then if it uses canton's executorservice stuff it might know what to do..

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,149 @@
...
+      ): Option[(scala.concurrent.duration.FiniteDuration, Unit)] = {
+        val prefixMsg =
+          s"RetrySourcePolicy for restart of AcsSnapshotBulkStorage with ${delay} delay:"
+        lastFailure match {
+          case Some(t) =>
+            logger.info(s"$prefixMsg Last failure: ${ErrorUtil.messageWithStacktrace(t)}")
+          case None =>
+            logger.debug(s"$prefixMsg No failure, normal restart.")
+        }
+        // Always retry (TODO(#3429): consider a max number of retries?)
```
ray-roestenburg-da: This is one of the reasons why I would think it is nice for this thing to be separate from the scan app pod. Let it crash when it can't write, recover on startup where you were, and let k8s restart the pod, this then also far easier shows up on failures in monitoring, and you don't have to crash the scan app. And the reboot effect sometimes works far better than staying in process. My two cents.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
ray-roestenburg-da: If this was a separate app/service just give it a fixed threadpool to work with.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,149 @@
...
+            idx.addAndGet(1)
+          }
+        }
+      val withKs = base.viaMat(KillSwitches.single)(Keep.right)
+      withKs.watchTermination() { case (ks, done) => (ks: KillSwitch, done) }
+    }
+
+    // TODO(#3429): tweak the retry parameters here
+    val delay = FiniteDuration(5, "seconds")
+    val policy = new RetrySourcePolicy[Unit, Int] {
```
ray-roestenburg-da, line 129: Is there a test that retry works, that writeFullObject can overwrite?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
```diff
@@ -0,0 +1,73 @@
...
+    credentials: AwsBasicCredentials,
+)
+
+class S3BucketConnection(
+    val s3Client: S3Client,
+    val bucketName: String,
+    val loggerFactory: NamedLoggerFactory,
+) extends NamedLogging {
+  // Reads the full content of an s3 object into a ByteBuffer.
+  // Use only for testing, when the object size is known to be small
```
ray-roestenburg-da, line 36:
```suggestion
  // Use only for testing, or when the object size is known to be small
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
```diff
@@ -0,0 +1,73 @@
...
+  def readFullObject(key: String): ByteBuffer = {
+    val obj = s3Client.getObject(GetObjectRequest.builder().bucket(bucketName).key(key).build())
+    val bytes = obj.readAllBytes()
+    val ret = ByteBuffer.allocateDirect(bytes.length)
```
ray-roestenburg-da: not necessary to be a direct buffer right? (I mean as part of readFullObject)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  // Writes a full object from memory into an s3 object (a blocking call)
+  def writeFullObject(key: String, content: ByteBuffer)(implicit tc: TraceContext): Unit = {
+    logger.debug(s"Writing ${content.array().length} bytes to S3 object $key")
+    val putObj: PutObjectRequest = PutObjectRequest
+      .builder()
+      .bucket(bucketName)
+      .key(key)
+      .build()
+    s3Client.putObject(
```
ray-roestenburg-da: Have you looked at [S3AsyncClient](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3AsyncClient.html) ? you can convert CompletableFuture in java to Future in Scala with FutureConverters, and I read that this is non blocking based on Netty, worth a try.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
```diff
@@ -0,0 +1,123 @@
...
+    def zstdFinish(): ByteString = {
+      tmpBuffer.flip()
+      val result = ByteString.fromByteBuffer(tmpBuffer)
+      tmpBuffer.clear()
+      compressingStream.close()
+      result
+    }
+
+    override def close(): Unit = {
+      compressingStream.close()
```
ray-roestenburg-da, line 73: Make this class own the tmpBuffer as well (let is create the buffer in constructor body and assign to private var and also add `tmpBuffer = null` here in close, then everything is nicely following AutoCloseable

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotBulkStorageTest.scala
ray-roestenburg-da: Using(new ZstdDirectBufferDecompressingStream(compressed)) { ds => ds.read(uncompressed) } now fixes it I think (if you also do the other thing I mention there and deref the buffer in close)

ray-roestenburg-da (commented):
Left some more comments, especially to try out S3AsyncClient

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
ray-roestenburg-da: (but later on I noticed that there is a aws S3AsyncClient you should just use instead and not worry about the blocking, if this thing works well)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
isegall-da: Nice, adopted that indeed.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
isegall-da: Fixed (as part of moving to the asyncClient)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
isegall-da: Hmm... I actually did mean "only for testing, when the object is small", not -or-

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: I tested it manually for now, added a TODO to add a unit test for it.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: Fixed with the asyncClient that you pointed to.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: Fixed (removed the `limit =`)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: done

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: done

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: yes, but as discussed, breaking scan up into multiple microservices has its own downsides.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/Limit.scala
ray-roestenburg-da: Nice, thanks

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
ray-roestenburg-da: ok fair, also maybe not the best comment I made on this particular PR 😄

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
ray-roestenburg-da: Ok. Maybe a slippery slope. Not sure how you are going to 'know' that it's small anyway, so never mind my change here.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
ray-roestenburg-da: Thanks!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
ray-roestenburg-da: I think it makes sense to default `partyIds: Seq[PartyId] = Seq.empty, templates: Seq[PackageQualifiedName] = Seq.empty,` on `def queryAcsSnapshot(` which is nice and clean, but not necessary in this PR

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,155 @@
...
+        s"Read ${encoded.length} contracts from ACS, to a bytestring of size ${contractsBytes.length} bytes"
+      )
+      (snapshot.afterToken.fold(End: Position)(Index(_)), contractsBytes)
+    }
+
+  }
+
+  def dumpAcsSnapshot(migrationId: Long, timestamp: CantonTimestamp): Future[Unit] = {
+
+    // TODO(#3429): currently, if this crashes half-way through, there is no indication in the S3 objects that
```
ray-roestenburg-da, line 90: Curious, why not do that right now?

ray-roestenburg-da: (or is this just a next PR that you'll work on, fine by me of course)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
```diff
@@ -0,0 +1,85 @@
...
+      .bucket(bucketName)
+      .key(key)
+      .build()
+    s3Client
+      .putObject(
+        putObj,
+        AsyncRequestBody.fromBytes(content.array()),
+      )
+      .asScala
+      .map(_ => ())
```
ray-roestenburg-da, line 63: Maybe at a later stage you can get the result and check ETag against md5 hash of what you wrote, or something like that to guarantee it's really there. (not meant now in this PR)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
```diff
@@ -0,0 +1,85 @@
...
+      bucketName: String,
+      loggerFactory: NamedLoggerFactory,
+  ): S3BucketConnection = {
+    new S3BucketConnection(
+      S3AsyncClient
+        .builder()
+        .endpointOverride(s3Config.endpoint)
+        .region(s3Config.region)
+        .credentialsProvider(StaticCredentialsProvider.create(s3Config.credentials))
+        // TODO(#3429): mockS3 and GCS support only path style access. Do we need to make this configurable?
```
ray-roestenburg-da, line 79: Only if some other S3 compatible storage does not work with this, could investigate that, or wait until someone asks for it. In any case it should be defaulted to this, if we would make it configurable, and then you would need to explicitly turn it off, IMHO.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
```diff
@@ -0,0 +1,123 @@
...
+import java.nio.ByteBuffer
+import java.util.concurrent.atomic.AtomicReference
+
+/** A Pekko GraphStage that zstd-compresses a stream of bytestrings, and splits the output into zstd objects of size (minWeight + delta).
+  * Somewhat similar to Pekko's built-in GroupedWeight, but outputs valid zstd compressed objects.
+  */
+case class ZstdGroupedWeight(minSize: Long) extends GraphStage[FlowShape[ByteString, ByteString]] {
+  require(minSize > 0, "minSize must be greater than 0")
+
+  val zstdTmpBufferSize = 10 * 1024 * 1024; // TODO(#3429): make configurable?
```
ray-roestenburg-da, line 20: I think makes sense to configure as part of BulkStorageConfig, with sensible default, so it can be tuned if necessary. (though likely this is a fine number)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    override def close(): Unit = {
+      compressingStream.close()
+    }
+  }
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with InHandler with OutHandler {
+      // TODO(#3429): consider implementing a pool of tmp buffers to avoid allocating a new one for each stage
```
ray-roestenburg-da: yes good idea, quite standard to use a pool for this (and then manage cleanup / deref there)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
```diff
@@ -0,0 +1,123 @@
...
+      private val tmpBuffer = ByteBuffer.allocateDirect(zstdTmpBufferSize)
```
ray-roestenburg-da, line 81: Doing this in preStart is better, otherwise you allocate even if the stream never runs or fails before starting.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
```diff
@@ -0,0 +1,123 @@
...
+      private val zstd = new AtomicReference[ZSTD](new ZSTD(tmpBuffer, 3))
+      private val state: AtomicReference[State] = new AtomicReference[State](State.empty())
+
+      override def postStop(): Unit = {
+        super.postStop()
+        if (zstd.get() != null) {
+          zstd.get().close()
```
ray-roestenburg-da, line 88: Since you allocate the buffer in this graphStageLogic you also need to deref it here. (or move everything to Zstd)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
```diff
@@ -0,0 +1,123 @@
...
+  class ZSTD(
+      val tmpBuffer: ByteBuffer,
+      val compressionLevel: Int = 3,
+  ) extends AutoCloseable {
+
+    val compressingStream =
+      new ZstdDirectBufferCompressingStreamNoFinalizer(tmpBuffer, compressionLevel)
+
+    def compress(input: ByteString): ByteString = {
+      val inputBB = ByteBuffer.allocateDirect(input.size)
```
ray-roestenburg-da, line 53: can you not reuse this buffer instead of creating a new one for every compression? allocateDirect is slow, GC does not see native mem so it can take a while before it gets cleaned up, in the meantime native memory starts piling up. maybe better, since you always use this from graphstagelogic, is to have a reusable direct buffer there, pass it into this method, and reuse it. (otherwise you risk running out of memory by piling up native memory while gc is not necessary, if you start streaming a lot of files)

ray-roestenburg-da: direct buffers are a PITA.

ray-roestenburg-da: https://openjdk.org/jeps/442 Project Panama in JDK 22 should make things a lot better, but likely that still would not work with this library, they'll need to update it, or we of course could consider contributing...

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: Yeah, just followup work in coming PRs, this one's already quite large.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
isegall-da: Yup, that's what I meant (check if it's needed), and that's the plan.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
isegall-da: :+1: added to the todo for the pool, since we'll revisit this code then anyway.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
isegall-da: Will be improved in a followup PR that will add a buffer pool

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
isegall-da: Added another TODO for a buffer pool here as well.

PR #3428, opened 2025-12-18
make the failure_notficiations action usable from a repo with a splice submodule

mblaze-da:
This change makes it possible to use this action from the internal repo.

### DIFF
--- a/.github/actions/tests/failure_notifications/action.yml
+++ b/.github/actions/tests/failure_notifications/action.yml
@@ -45,6 +45,6 @@ runs:
       run: |
         if [[ "${RUNNER_ENVIRONMENT}" != "self-hosted" ]]; then
           echo "Not a self-hosted runner. Installing required Python packages..."
-          python3 -m pip install git requests
+          python3 -m pip install GitPython requests
         fi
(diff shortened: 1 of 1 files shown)


.github/actions/tests/failure_notifications/action.yml
```diff
@@ -42,4 +41,10 @@ runs:
         FAILURE_NOTIFICATIONS_URL: ${{ inputs.notifications_url }}
         NOTIFICATION_SLACK_CHANNEL: ${{ inputs.slack_channel }}
       shell: bash
-      run: ./.github/actions/scripts/notification-scripts/failure_notification.py --job_subname "${{ inputs.job_subname }}"
+      # on self-hosted runners all python packages are preinstalled but on github-hosted they will be missing
+      run: |
+        if [[ "${RUNNER_ENVIRONMENT}" != "self-hosted" ]]; then
```
isegall-da, line 46: hmm.. we used to have an install-deps script I believe, that was called from failure_notification.sh, for this exact purpose, but we seemed to have killed it at some point. Guess this is fine too.

PR #3434, opened 2025-12-19
add necessary permissions to the bump_gha_runner_version workflow for failure_notifications

mblaze-da:
I'm adding this after I learned it is required for failure_notifications when working on the workflow in the internal repo.


isegall-da (approved):
Thanks

PR #3435, opened 2025-12-19
Add pperek to cluster-lock-users

pawelperek-da:
Adds pperek to cluster-lock-users.json


pawelperek-da:
Context: I was unable to run the `cncluster apply` command on a locked scratch. Stephen suggested that it might be because I'm not in this file

PR #3464, opened 2026-01-06
Backup docs: clarify that both ID and DB backups are recommended

martinflorian-da:
Patch contributed by @amarqueslee-da

[static]


julientinguely-da (approved):
thanks

PR #3467, opened 2026-01-06
Bump a few SBT memory parameters

martinflorian-da:
An attempt to improve on https://github.com/DACH-NY/cn-test-failures/issues/6681

[ci]

### DIFF
--- a/.envrc.vars
+++ b/.envrc.vars
@@ -16,7 +16,7 @@ export PUBLIC_CONFIGS_PATH=${SPLICE_ROOT}/cluster/configs/configs
 
 # Inrease code heap sizes to avoid issues
 # Defaults NonNMethodCodeHeapSize=7M,NonProfiledCodeHeapSize=122M,ProfiledCodeHeapSize=122M
-export SBT_OPTS="-XX:MaxRAMPercentage=90.0 -Xmx16G -Xms2G -Xss2M -XX:+UseG1GC -XX:NonNMethodCodeHeapSize=32M -XX:NonProfiledCodeHeapSize=256M -XX:ProfiledCodeHeapSize=256M -XX:ReservedCodeCacheSize=544M"
+export SBT_OPTS="-XX:MaxRAMPercentage=90.0 -Xmx16G -Xms4G -Xss6M -XX:+UseG1GC -XX:ReservedCodeCacheSize=1G"
 
 # Provide a simple way to get the path to `sbt-launch.jar` for IntelliJ setup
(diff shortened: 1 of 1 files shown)


isegall-da (approved):
Thanks, I don't mind giving it a try...

.envrc.vars
```diff
@@ -16,7 +16,7 @@ export PUBLIC_CONFIGS_PATH=${SPLICE_ROOT}/cluster/configs/configs
 
 # Inrease code heap sizes to avoid issues
 # Defaults NonNMethodCodeHeapSize=7M,NonProfiledCodeHeapSize=122M,ProfiledCodeHeapSize=122M
-export SBT_OPTS="-Xmx8G -Xms2G -Xss2M -XX:+UseG1GC -XX:NonNMethodCodeHeapSize=32M -XX:NonProfiledCodeHeapSize=256M -XX:ProfiledCodeHeapSize=256M -XX:ReservedCodeCacheSize=544M"
+export SBT_OPTS="-XX:MaxRAMPercentage=90.0 -Xmx16G -Xms2G -Xss2M -XX:+UseG1GC -XX:NonNMethodCodeHeapSize=32M -XX:NonProfiledCodeHeapSize=256M -XX:ProfiledCodeHeapSize=256M -XX:ReservedCodeCacheSize=544M"
```
martinflorian-da: This should ensure that it doesn't OOM on small and x-small, and for the bigger ones (where tests run, i.e., canton, sbt and sometimes frontends side by side) it should be capped to 16G

nicu-da: isn't MaxRAMPercentage conflicting with xmx?

nicu-da: also gemini recommends we just drop `-XX:NonNMethodCodeHeapSize=32M -XX:NonProfiledCodeHeapSize=256M -XX:ProfiledCodeHeapSize=256M -XX:ReservedCodeCacheSize=544M` and set `-XX:ReservedCodeCacheSize=1G` and let the jvm handle the different pool allocations

we should also start with a higher initial memory as the process of increasing it it's expensive timewise `-Xms2G`
we can also bump the `-Xss` to 6M, it's per thread so yolo

martinflorian-da: > isn't MaxRAMPercentage conflicting with xmx?

AFIU it's "lowest one wins"; so this should mean 90%, but never above 16. Judging by the logs I saw it seems like it's working just like that.

> we just drop

Yeah I didn't bother trying to understand what these were for...

Let me revise this line with you suggestions and start another `[ci]` run :shrug:

PR #3470, opened 2026-01-07
Minting delegation automation and HTTP APIs

dfordivam:
Fixes https://github.com/hyperledger-labs/splice/issues/3469

This implements the backend changes for the minting delegation support.

1. Adds HTTP APIs to wallet-app for listing proposals + delegations, and performing actions on them.

2. adds automation in `ExternalPartyWalletService` for performing minting for the external-party (by the `validatorParty` as the delegate)

Note: The minting of `ValidatorRewardCoupons` require that the external-party has created a `ValidatorRight` contract with it as the validator. Otherwise the minting of `ValidatorRewardCoupons` will not happen.

3. Adds two trigger to auto-reject expired `MintingDelegationProposal` and `MintingDelegation`

4. Adds a simtime integration test for above.

For 1 and 3 we could have tested via a walltime integration test, but it seems better to have a single test and keep the testing code in one place.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/ExternalPartyWalletStore.scala
```diff
@@ -131,6 +180,28 @@ object ExternalPartyWalletStore {
...
+          co.payload.dso == dso &&
+          co.payload.validator == externalParty
+        }(co =>
+          ExternalPartyWalletAcsStoreRowData(co, rewardCouponRound = Some(co.payload.round.number))
+        ),
+        mkFilter(UnclaimedActivityRecord.COMPANION) { co =>
+          co.payload.dso == dso &&
+          co.payload.beneficiary == externalParty
+        }(ExternalPartyWalletAcsStoreRowData(_)),
+        // ValidatorRight needed for collecting ValidatorRewardCoupons via MintingDelegation
```
meiersi-da:
```suggestion
        // ValidatorRight needed for collecting ValidatorRewardCoupons issued from traffic purchases via MintingDelegation
```

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/db/DbExternalPartyWalletStore.scala
meiersi-da:
```suggestion
        version = 3,
```
needs to be bumped as you changed the filter. Bumping it ensures that existing deployments reingest the ACS for this store.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/db/DbExternalPartyWalletStore.scala
```diff
@@ -67,4 +84,69 @@ class DbExternalPartyWalletStore(
...
+        BigDecimal,
+    )
+  ]] = listSortedRewardCoupons(
+    validatorCodegen.ValidatorLivenessActivityRecord.COMPANION,
+    issuingRoundsMap,
+    _.optIssuancePerValidatorFaucetCoupon.toScala.map(BigDecimal(_)),
+    limit,
+  )
+
+  private def listSortedRewardCoupons[C, TCid <: ContractId[?], T](
```
meiersi-da, line 105: This seems like a literal copy of the method in the `DbUserWalletStore`. It's non-trivial code. Any chance to share it? I think it would be OK for the `DbExternalPartyWalletStore` to import a static method from `DbuUserWalletStore`

meiersi-da: Also asking as recently there was some talk about adjusting some of the code in that method: https://github.com/hyperledger-labs/splice/issues/767#issuecomment-3714528813

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RejectInvalidMintingDelegationProposalTrigger.scala
```diff
@@ -0,0 +1,130 @@
...
+      ],
+      reason: String,
+  )(implicit tc: TraceContext): Future[TaskOutcome] = {
+    val beneficiary = proposal.payload.delegation.beneficiary
+
+    connection
+      .submit(
+        actAs = Seq(endUserParty),
+        readAs = Seq.empty,
+        update = proposal.exercise(_.exerciseMintingDelegationProposal_Reject()),
```
meiersi-da: it would be nice to push the reason here.

meiersi-da: so the proposer knows why the proposal was rejected

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RejectInvalidMintingDelegationProposalTrigger.scala
```diff
@@ -0,0 +1,130 @@
...
+
+/** Trigger to reject invalid MintingDelegationProposal
+  * Current support for running a minting delegation is limited to the following
+  * - The beneficiary is an onboarded external-party
+  * - The delegate is the validator party
+  * Proposals not satisfying both these conditions will be auto rejected
+  *
+  * Additionally if there are multiple proposals from the same beneficiary, this
+  * trigger will reject all the existing proposals when a new proposal is received
+  */
```
meiersi-da: Not sure we need this automation, and want to maintain it. The proposals anyways need to be accepted manually, so rejecting the few that are wrong manually seems totally OK. It also makes the process more predictable.

I'd suggest to remove this trigger.

meiersi-da: It is also not clear to why only the validator operator party can be used as a delegate. It seems that it would be OK, and more uniform, if any onboarded wallet user could accept minting delegations, and thus instruct its automation to execute them. wdyt?

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/UserWalletAutomationService.scala
```diff
@@ -194,6 +196,15 @@ class UserWalletAutomationService(
       )
     )
   }
+
+  registerTrigger(
```
meiersi-da, line 198: Missing: expiry trigger for expired minting delegations and proposals

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,668 @@
...
+              env.environment.clock.now.plus(Duration.ofDays(1)).toInstant,
+            ).create,
+          )
+
+        advanceRoundsAndVerifyBalanceIncrease()
+      }
+
+      // For ValidatorRewardCoupon, we need ValidatorRight for beneficiary
+      // And also need to pause the validator's own reward collection
+      // which would normally mint this coupon
```
meiersi-da: Why is that? Is the beneficiary setup by default with a `ValidatorRight`? Can you change the setup to avoid that and match the real case?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,668 @@
...
+              readAs = Seq.empty,
+              update = new ValidatorRewardCoupon(
+                dsoParty.toProtoPrimitive,
+                beneficiaryParty.party.toProtoPrimitive,
+                BigDecimal(100.0).bigDecimal,
+                issuingRound4.round,
+              ).create,
+            )
+
+          advanceRoundsAndVerifyBalanceIncrease()
```
meiersi-da: I expect these `advanceRoundsAndVerifyBalanceIncrease` to be rather expensive in test duration. Why not create all coupons in one go, and check that they all get collected?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,668 @@
...
+
+      // Transfer a few amulets to the external party
+      // The MintingDelegationCollectRewardsTrigger should merge them
+      (1 to amuletMergeLimit + 2).foreach { i =>
+        aliceValidatorWalletClient.transferPreapprovalSend(
+          beneficiaryParty.party,
+          10.0,
+          s"transfer-$i",
+        )
+      }
```
meiersi-da, line 571: consider combining this test with the one for rewards collection. It seems that testing that merging works with reward collection can be done by creating enough amulet up front together with creating the coupons.

A final step can then test that merging works even if there are no coupons.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
```diff
@@ -0,0 +1,337 @@
...
+
+    if (hasRewardsToCollect || shouldMergeAmulets) {
+      val amuletsToMerge = if (shouldMergeAmulets) {
+        // Merge the smallest amounts first
+        // we do +1 here to maintain exactly 'mergeLimit' amulets after the mint
+        amulets
+          .sortBy(a =>
+            BigDecimal(SpliceUtil.currentAmount(a.payload, openRound.payload.round.number))
+          )
+          .take(amulets.size - mergeLimit.toInt + 1)
```
meiersi-da, line 170: Once the party has `mergeLimit` amulet, every new amulet creation will lead to another merge. That seems like a potential waste of traffic. What would it take to only merge when there are more than 2x the mergeLimit of amulet? Thus always merging a batch of `mergeLimit` size?

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/admin/http/HttpWalletHandler.scala
```diff
@@ -1278,4 +1279,217 @@ class HttpWalletHandler(
...
+              )
+              .noDedup
+              .withSynchronizerId(proposal.domain)
+              .yieldResult()
+          } yield (result.exerciseResult.mintingDelegationCid, sameBeneficiaryDelegations.drop(1)),
+          logger,
+        )
+        // Reject any remaining delegations outside the retry block
+        // In normal circumstances we should never have more than one active
+        // delegation, so the following cleanup is mainly for some edge cases.
```
meiersi-da: Not sure about this. It helps a bit, but is not crash-fault tolerant. If we do want this guarantee that eventually there are no duplicates, then I'd rather implement a deduplication trigger. It would just reject the older ones.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
dfordivam: We have discussed this recently on design-doc to have this as a limitation of minting delegation.
How/when is the `ValidatorRight` created by the beneficiary is not specified/discussed by us. Ideally this should be done by some helper script which creates the `ValidatorRight` while the delegation proposal is being created. I can add that later

> ValidatorRewardCoupon rewards of the delegee are minted, only when the delegee also created their own `ValidatorRight` contract listing themselves as the validator.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
dfordivam: I did that earlier but then changed to testing one coupon at a time. This approach is more robust as it confirm each coupon kind is being minted. 
Simply doing a "balance has increased" check, when minting multiple coupons is somewhat problematic. For example I missed to capture the `ValidatorRewardCoupon` minting race issue with the validator's own trigger when all coupons were minted together.
Another approach is to check the amount of balance increase, which is a bit more tedious to do, and I did not see it being done in other rewards integration tests.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/admin/http/HttpWalletHandler.scala
dfordivam: I thought that this scenario would be so rare that a deduplication trigger would not be worth.

Having multiple delegations requires that there are multiple proposals from same beneficiary, that the extra proposals did not get auto-rejected, AND the user accepted two or more in quick succession such that the acs store did not get a chance to see the contract just created earlier.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RejectInvalidMintingDelegationProposalTrigger.scala
dfordivam: I thought that we added a limitation in design-doc for this to be only the validator party. But the limitation was wrt the beneficiary list support. So yes this can be implemented.
Do we see any other issues in allowing all local users to start a minting delegation service?

dfordivam: And the reason we have a reject trigger is because we discussed that the user should see the proposals from only the onboarded parties in UI, otherwise it complicates how we show to the user that the party is not onboarded.

The rejection of duplicate proposals is mostly an afterthought since we had a trigger already.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/db/DbExternalPartyWalletStore.scala
dfordivam: I will look into this, importing from DbUserWalletStore seems to have its own complications.

Another idea is to move these apis to a common "DbTransferInputStore" (and also implement the sql queries for the https://github.com/hyperledger-labs/splice/blob/723e6bc84cd060dd8549d4e32b38b92dd5aef26d/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala#L68 and https://github.com/hyperledger-labs/splice/blob/723e6bc84cd060dd8549d4e32b38b92dd5aef26d/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala#L98

meiersi-da: That sounds like a good option too. Just want to avoid that fixes are applied to only one of the two instances of this code.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RejectInvalidMintingDelegationProposalTrigger.scala
meiersi-da: Couldn't you annotate the listing response with whether the party is locally hosted or not? This way the hosting resolution can be done in the backend.

meiersi-da: > I thought that we added a limitation in design-doc for this to be only the validator party. But the limitation was wrt the beneficiary list support. So yes this can be implemented.
Do we see any other issues in allowing all local users to start a minting delegation service?

It means that all local parties can rely on the validator node automating actions for them. This is already the case with their reward minting, so it doesn't feel like a big addition. The benefit is that the validator operator party becomes less special, which I intuit is preferable.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
meiersi-da: What I was going after is this line

> // And also need to pause the validator's own reward collection

I was thinking that we might want to change the setup function of the external party such that it doesn't create a `ValidatorRight`. That said: it's not worth the effort, as the test is good enough as-is. What might make sense is to clarify in the comment that our setup function also sets-up the unwanted `ValidtorRight`, which is why we deactivate it.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
meiersi-da: One thing that many tests do is to check the tx history using https://github.com/hyperledger-labs/splice/blob/723e6bc84cd060dd8549d4e32b38b92dd5aef26d/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletTxLogTestUtil.scala#L39

There you do get the distinction between the different kinds of rewards. You'd still have to compute the amounts for validator rewards as there activeness and `ValidatorRewardCoupon`s are not distinguished. However the amount computation should be straightforward when the round is not overbooked.

Sorry for asking about this, but AFAIR each of these kind of round increases adds ~15s to test duration time. Our CI time and cost really does matter.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/admin/http/HttpWalletHandler.scala
meiersi-da: Agreed that it is rare when you have your auto-reject trigger running. What I'm reacting to is that there are these edge cases. Crashes, races, etc. leading to duplicates existing nevertheless. I'd rather either go fully manual or fully automatic. Code with edge cases will cause support load on the ops teams due to these rare edge cases.

meiersi-da: Your call whether you want to go fully manual or fully automatic.

Independently in both cases the code doing the collection cannot assume that there are no duplicate `MintingDelegation`s. It should thus just pick one of the existing ones. Probably the one that was created latest. Otherwise, there will be unnecessary contention and traffic spend.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RejectInvalidMintingDelegationProposalTrigger.scala
dfordivam: > Couldn't you annotate the listing response with whether the party is locally hosted or not? This way the hosting resolution can be done in the backend.

I am not sure if I understand this.
Assuming that we removed this trigger.
If we receive a proposal from a party which has not been onboarded, should we show that in UI with the "accept" button disabled?
ie we keep the proposal but annotate the "list proposals" API to indicate that the party is not onboarded?

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/admin/http/HttpWalletHandler.scala
dfordivam: > guarantee that eventually there are no duplicates

I am leaning towards a "fully manual", ie we don't provide a guarantee of not having duplicates. If we do end up in a situation with duplicates, the user will see them in the UI and will be able to clean it up manually. So a deduplication trigger is not necessary.

Now when we accept a new proposal, and we know about an existing delegation, it makes sense to archive the old in `acceptMintingDelegationProposal`. (I assume you agree on this)

Regarding this additional code to reject any other active delegations; I see this as a nice to have code, but without providing a guarantee that there will be no duplicates. Having this does not seem worse than not having it.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RejectInvalidMintingDelegationProposalTrigger.scala
meiersi-da: > If we receive a proposal from a party which has been onboarded, should we show that in UI with the "accept" button disabled?
ie we keep the proposal but annotate the "list proposals" API to indicate that the party is not onboarded?

Yes; and have the UI show that the party is not hosted on the node. Note that the party hosting relationship can change over time. So even if upon accepting the minting delegation the party was hosted locally, it might no longer be hosted locally later.

Thus I would suggest to check the hosting relationship as part of the minting delegation reward collection trigger. Log at info level that you are skipping a delegation for an unhosted party.

The operator can then use the UI to find the delegations for which parties are no longer hosted locally and reject them.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/admin/http/HttpWalletHandler.scala
meiersi-da: > > guarantee that eventually there are no duplicates
> 
> I am leaning towards a "fully manual", ie we don't provide a guarantee of not having duplicates. If we do end up in a situation with duplicates, the user will see them in the UI and will be able to clean it up manually. So a deduplication trigger is not necessary.

Fully manual makes sense.

> Now when we accept a new proposal, and we know about an existing delegation, it makes sense to archive the old in `acceptMintingDelegationProposal`. (I assume you agree on this)

Yes that's fine with me.

> Regarding this additional code to reject any other active delegations; I see this as a nice to have code, but without providing a guarantee that there will be no duplicates. Having this does not seem worse than not having it.

It's a fine line to thread. I would not add it as:
1. it gives a false sense of security ==> devs might assume that they can assume that there really are no duplicates. I believe your code in the actual trigger currently does so.
2. it is extra code that is hard to test and can be wrong ==> might cause unnecessary debugging or support overhead

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/db/DbExternalPartyWalletStore.scala
dfordivam: I have bumped this, and version of DbUserWalletStore also ( 9ff02a5c95), as both would need to reingest the ACS.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RejectInvalidMintingDelegationProposalTrigger.scala
dfordivam: Removed the reject trigger

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RejectInvalidMintingDelegationProposalTrigger.scala
dfordivam: Removed the reject trigger. And changed the APIs to show the hosting status.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/UserWalletAutomationService.scala
dfordivam: Added expiry triggers for both

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
dfordivam: Updated the comment here 685657c45156eea0ed0e2fe355dab37d9840f7da

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
dfordivam: Fixed the test to create all coupons together and verify the balance increase

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
dfordivam: Fixed, merged this in previous test

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
dfordivam: Done, now we merge at hitting 2x the limit

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/admin/http/HttpWalletHandler.scala
dfordivam: Ok, I have simplified the API and removed the code for rejecting other active delegations

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/db/DbExternalPartyWalletStore.scala
dfordivam: Regarding this I have not done it in this PR, as this is getting quite big on its own. Perhaps this could be tackled in a subsequent PR?

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
```diff
@@ -0,0 +1,344 @@
...
+  }
+
+  private def collectRewardsAsDelegate()(implicit tc: TraceContext): Future[Boolean] = {
+    for {
+      delegations <- store.multiDomainAcsStore.listContracts(
+        MintingDelegation.COMPANION,
+        Limit.DefaultLimit,
+      )
+
+      // We expect either none, or only one active delegation per beneficiary
```
meiersi-da:
```suggestion
      // In the steady state there is at most one active delegation per beneficiary.
      // Thus if there are multiple ones, we can just pick one of them.
```

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
```diff
@@ -0,0 +1,344 @@
...
+        ValidatorRewardCoupon,
+      ]],
+      appRewardCoupons: Seq[Contract[
+        AppRewardCoupon.ContractId,
+        AppRewardCoupon,
+      ]],
+      unclaimedActivityRecords: Seq[Contract[
+        UnclaimedActivityRecord.ContractId,
+        UnclaimedActivityRecord,
+      ]],
```
meiersi-da, line 257: Please add a TODO (and issue) to integrate with @jose-velasco-ieu 's work on CIP-82 . The one of the two lands on `main` as a second one will have to execute that. It's just another kind of coupon that should be automatically collected.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
```diff
@@ -0,0 +1,344 @@
...
+    val unclaimedActivityRecordInputs: Seq[TransferInput] =
+      couponsData.unclaimedActivityRecords.map { record =>
+        new InputUnclaimedActivityRecord(record.contractId): TransferInput
+      }
+
+    val amuletInputs: Seq[TransferInput] = amuletsToMerge.map { amulet =>
+      new InputAmulet(amulet.contractId): TransferInput
+    }
+
+    livenessInputs ++ validatorCouponInputs ++ appCouponInputs ++ unclaimedActivityRecordInputs ++ amuletInputs
```
meiersi-da: Is there a place where you take into account that there is an upper limit on the number of inputs can be given to an `Amulet_Transfer`?

compare to: https://github.com/hyperledger-labs/splice/blob/6b82685b780b431d5277d4d6e2b77ffa58349601/apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/treasury/TreasuryService.scala#L767-L777

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/db/DbExternalPartyWalletStore.scala
meiersi-da: That sounds fine to me. Can you create an issue for it?

nix/canton-sources.json
```diff
@@ -3,5 +3,5 @@
   "tooling_sdk_version": "3.3.0-snapshot.20250415.13756.0.vafc5c867",
   "daml_release": "v3.3.0-snapshot.20250417.0",
   "enterprise_sha256": "sha256:0zipb0jdblr9rpmrn2hcfqvc0yks45v7dps3n8ai82hsnvx7g011",
-  "oss_sha256": "sha256:1nh1xn77bbj8s24kx3wk7scm17jpvic7sszabzwpfcn6mzbvbxir"
+  "oss_sha256": "sha256:1hkdfnacir1gin2g7pcrg1wv7f7smcsh0n7hnfxym5adn177zz97"
```
meiersi-da: Why did this change?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,654 @@
...
+          (validatorRewardAmount * BigDecimal(issuingRound.issuancePerValidatorRewardCoupon)) +
+          unclaimedActivityAmount
+
+      val (openRoundsForFee, _) = sv1ScanBackend.getOpenAndIssuingMiningRounds()
+      val openRoundForFee = openRoundsForFee.head
+      val createFee = BigDecimal(
+        openRoundForFee.payload.transferConfigUsd.createFee.fee
+      ) / BigDecimal(openRoundForFee.payload.amuletPrice)
+
+      actualIncrease should be >= (expectedTotalReward - createFee)
```
meiersi-da: Consider running with zero transfer fees to avoid this fee logic. Example:

https://github.com/hyperledger-labs/splice/blob/6b82685b780b431d5277d4d6e2b77ffa58349601/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanIntegrationTest.scala#L54-L58

meiersi-da (approved):
Very nice! Thank you!

Approving with suggestions.

nix/canton-sources.json
dfordivam: There was a hash mismatch yesterday, and unfortunately there is another hash mismatch on today's main also.
For local testing I had to modify it, but I guess this is not needed for CI and can be removed, as it will use enterprise canton

```
 'https://www.canton.io/releases/canton-open-source-3.4.11-snapshot.20260114.17499.0.v227d6c45.tar.gz':
         specified: sha256:1zkfw5zshwzi5kcm0k1p40f8cl47qh36n7xa3fbg4mr76lw1k0wm
         got:       sha256:1lgqvr3fmfdpzh50cp0q2abhc6sscjr135y8m1bd9i36kdjyvgml
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
dfordivam: Fixed

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
dfordivam: fixed in 36c0f830ac94467b2e15ba897606abc3421b0834

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
dfordivam: Added a TODO and issue https://github.com/hyperledger-labs/splice/issues/3554

PR #3485, opened 2026-01-08
Bump CPU limit on GHA runners

martinflorian-da:
[ci]

Another attempt at improving on https://github.com/DACH-NY/cn-test-failures/issues/6681

See the last few comments there.

I thought about picking a bigger runner just for the ones that failed. But:

- it seems like arbitrary large ones could fail atm
- we don't need more memory, just more CPU
- I don't see a significant downside in bumping limits; I'm keeping requests the same, so shouldn't impact costs, and it seems like we get those CPU usage spikes rarely enough to not interfere with other workloads... will monitor the situation ofc

Open to reverting this again if we do find a better theory / solution... Specifically I still have no idea why we sometimes need more CPU.

Maybe we're getting CPU throttled for some weird k8s reason (that should improve by raising limits). Maybe we should configure more verbose logging on the runners and that will tell us. But honestly if this bump fixes it I'm happy enough to not dig deeper...

### DIFF
--- a/cluster/pulumi/gha/src/runners.ts
+++ b/cluster/pulumi/gha/src/runners.ts
@@ -56,7 +56,7 @@ const runnerSpecs = [
         memory: '10Gi',
       },
       limits: {
-        cpu: '4',
+        cpu: '6',
         memory: '10Gi',
       },
     },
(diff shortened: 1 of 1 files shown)


martinflorian-da:
> fwiw it's fairly easy to see in metrics if we get throttled

@moritzkiefer-da 

Well the throttling metric says we are getting throttled, but I don't understand the direction of the causality here with "100% CPU Usage". Trying to learn about this right now...

image
 
 Also AFAIU bumping limits *is* a plausible way to avoid throttling.

PR #3493, opened 2026-01-08
Fix release notes for 0.5.5

martinflorian-da:
Part of https://github.com/DACH-NY/canton-network-internal/issues/3045

[static]

### DIFF
--- a/docs/src/release_notes.rst
+++ b/docs/src/release_notes.rst
@@ -50,7 +50,7 @@
 
     - Added an optional ``excludeDebugFields``` boolean to the request body of allocation and transfer endpoints for the Token Standard component
       (``splice-api-token-allocation-v1`` and ``splice-api-token-transfer-instruction-v1``).
-      Clients can now set this to true to omit debug information from the response.
+      Clients can now set this to true to omit debug information from the response in order to save on bandwidth.
 
 .. release-notes:: 0.5.4
 


docs/src/release_notes.rst
```diff
@@ -48,9 +44,12 @@
...
 
-    - Added an optional ``excludeDebugFields``` boolean to the request body of allocation and transfer endpoints for the Token Standard component.
-      (``splice-api-token-allocation-v1`` and ``splice-api-token-transfer-instruction-v1``
+  - Validator
+
+    - Added support for picking a custom name for new parties created when onboarding users via the `/v0/admin/users` API. See :ref:`docs <validator-users>`.
 
+    - Added an optional ``excludeDebugFields``` boolean to the request body of allocation and transfer endpoints for the Token Standard component
+      (``splice-api-token-allocation-v1`` and ``splice-api-token-transfer-instruction-v1``).
       Clients can now set this to true to omit debug information from the response.
```
isegall-da:
```suggestion
      Clients can now set this to true to omit debug information from the response in order to save on bandwidth.
```

(we've been asked about this, hence https://github.com/hyperledger-labs/splice/commit/af257b00426182800877e42186fa1a8f14b32476. Basically just applying the same improvement here as well)

PR #3495, opened 2026-01-08
Support dumping updates to bulk storage

isegall-da:

As discussed, the design is a bit different from the current dump of ACS snapshots, in the sense that it exposes the `next()` method that is expected to be repeatedly called. This fits better our overall design for long-running tasks, where we break them into small chunks, and from a periodic trigger make progress incrementally. 

While ACS snapshots will probably be smaller than whole update segments, I will nevertheless refactor it to work similarly to this, for consistency and ease of implementation of the higher level component that uses these.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala
@@ -797,8 +797,6 @@ object CompactJsonScanHttpEncodings {
   def apply() = new CompactJsonScanHttpEncodings(RemoveFieldLabels.value, RemoveFieldLabels.record)
 }
 
-
-
 // A lossless, but harder to process, encoding. Should be used only for backfilling Scan.
 case object ProtobufJsonScanHttpEncodings extends ScanHttpEncodings {
   import org.lfdecentralizedtrust.splice.util.ValueJsonCodecProtobuf


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -0,0 +1,184 @@
...
+    private val liveSource: Source[ByteString, SourceQueueWithComplete[ByteString]] =
+      Source.queue[ByteString](100, overflowStrategy = OverflowStrategy.backpressure)
+    private val (queue, initStream) = liveSource
+      .toMat(Sink.asPublisher(fanout = false))(org.apache.pekko.stream.scaladsl.Keep.both)
+      .run()
+    private val s3ObjIdx = new AtomicInteger(0)
+    private val position = new AtomicReference[Option[(Long, CantonTimestamp)]](None)
+
+    val pipeline: Future[Done] = Source
+      .fromPublisher(initStream)
```
rautenrieth-da: Why are we using `Source.fromPublisher(liveSource.toMat().run())` instead of just `liveSource`? Both are of type `Source`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -0,0 +1,184 @@
...
+        val updatesStr = encoded.map(_.asJson.noSpacesSortKeys).mkString("\n") + "\n"
+        val updatesBytes = ByteString(updatesStr.getBytes(StandardCharsets.UTF_8))
+        logger.debug(
+          s"Read ${encoded.length} updates from DB, to a bytestring of size ${updatesBytes.length} bytes"
+        )
+        queue.offer(updatesBytes).map {
+          case QueueOfferResult.Enqueued => ()
+          case QueueOfferResult.Dropped =>
+            throw new RuntimeException(
+              "Queue overflow. This should not happen with OverflowStrategy.backpressure"
```
rautenrieth-da: That doesn't sound right. The queue has an input buffer of fixed length, and there's no limit on how fast you  can call `next()`. If the trigger calling `next()` is _much_ faster than the S3 upload, won't this overflow the queue?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotBulkStorageTest.scala
```diff
@@ -86,81 +69,46 @@ class AcsSnapshotBulkStorageTest extends StoreTest with HasExecutionContext with
...
-      createdEvent.signatories.asJava,
-      createdEvent.observers.asJava,
-      createdEvent.createdAt.toInstant,
-      /* acsDelta = */ false,
-      /* representativePackageId = */ templateId.getPackageId,
-    )
-
   }
 
+//  // Similar to CompactJsonScanHttpEncodings.httpToJavaCreatedEvent, but does preserve field names, as we want to assert on their equality.
```
rautenrieth-da: Do we need this commented out code?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/HasS3Mock.scala
```diff
@@ -0,0 +1,109 @@
...
+  def getS3BucketConnectionWithInjectedErrors(
+      loggerFactory: NamedLoggerFactory
+  ): S3BucketConnection = {
+    val s3BucketConnection: S3BucketConnection = getS3BucketConnection(loggerFactory)
+    val s3BucketConnectionWithErrors = Mockito.spy(s3BucketConnection)
+    var failureCount = 0
+    val _ = doAnswer { (invocation: InvocationOnMock) =>
+      val args = invocation.getArguments
+      args.toList match {
+        case (key: String) :: _ if key.endsWith("2.zstd") =>
```
rautenrieth-da: Is this hardcoding that the mock will fail exactly once when writing any object with a key that ends with `2.zstd`? That seems quite arbitrary for a generic mock, can we move at least the hardcoded value to a function argument?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/HasS3Mock.scala
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  def readUncompressAndDecode[T](
+      s3BucketConnection: S3BucketConnection,
+      decoder: String => Either[io.circe.Error, T],
+  )(s3obj: S3Object)(implicit ec: ExecutionContext, tag: reflect.ClassTag[T]): Array[T] = {
+    val bufferAllocator = PooledByteBufAllocator.DEFAULT
+    val compressed = s3BucketConnection.readFullObject(s3obj.key()).futureValue
+    val compressedDirect = bufferAllocator.directBuffer(compressed.capacity())
+    val uncompressedDirect = bufferAllocator.directBuffer(compressed.capacity() * 200)
```
rautenrieth-da, line 64: Why 200? Is this assuming that our data is compressed by a factor of at most 200? I assume that's fine because all test data is deterministic so that the test would deterministically crash if that assumption is wrong?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -0,0 +1,184 @@
...
+    extends NamedLogging {
+
+  class UpdateHistorySegmentBulkStorage(
+      val fromMigrationId: Long,
+      val fromTimestamp: CantonTimestamp,
+      val toMigrationId: Long,
+      val toTimestamp: CantonTimestamp,
+  ) {
+    private val liveSource: Source[ByteString, SourceQueueWithComplete[ByteString]] =
+      Source.queue[ByteString](100, overflowStrategy = OverflowStrategy.backpressure)
```
rautenrieth-da: Is there any particular reason why the queue has a buffer of 100 UpdateHistory chunks? Shouldn't the zstd compression be much faster than the db queries, so that the queue is almost always empty?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala
```diff
@@ -671,8 +671,8 @@ object ScanHttpEncodings {
   }
 }
 
-// A lossy, but much easier to process, encoding. Should be used for all endpoints not used for backfilling Scan.
-case object CompactJsonScanHttpEncodings extends ScanHttpEncodings {
+class CompactJsonScanHttpEncodingsWithOrWithoutFieldLabels(withFieldLabels: Boolean)
```
rautenrieth-da: Why do we need the new encoding? As far as I can see, it's only used in test code.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/HasS3Mock.scala
```diff
@@ -0,0 +1,109 @@
...
+    val uncompressedNio = uncompressedDirect.nioBuffer(0, uncompressedDirect.capacity())
+    compressedDirect.writeBytes(compressed)
+    Using(new ZstdDirectBufferDecompressingStream(compressedDirect.nioBuffer())) {
+      _.read(uncompressedNio)
+    }
+    uncompressedNio.flip()
+    val allContractsStr = StandardCharsets.UTF_8.newDecoder().decode(uncompressedNio).toString
+    val allContracts = allContractsStr.split("\n")
+    compressedDirect.release()
```
rautenrieth-da: Is there some sort of `Using` that can be used on the buffers for better resource cleanup? Not very important, it's only test code.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/UpdateHistoryBulkStorageTest.scala
```diff
@@ -0,0 +1,168 @@
...
+        segment.watchCompletion().futureValue
+        for {
+          s3Objects <- bucketConnection.s3Client
+            .listObjects(
+              ListObjectsRequest.builder().bucket("bucket").build()
+            )
+            .asScala
+          allUpdates <- mockStore.store.getUpdatesWithoutImportUpdates(
+            None,
+            HardLimit.tryCreate(2200, 2200),
```
rautenrieth-da:
```suggestion
            HardLimit.tryCreate(segmentSize, segmentSize),
```
?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/UpdateHistoryBulkStorageTest.scala
```diff
@@ -0,0 +1,168 @@
...
+            afterIsInclusive = true,
+          )
+        } yield {
+          val objectKeys = s3Objects.contents.asScala.sortBy(_.key())
+          objectKeys should have length 2
```
rautenrieth-da, line 93: Possible other checks:
- check file size of first object 
- check number of updates in allUpdatesFromS3 (should be `segmentSize`?)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala
ray-roestenburg-da: can we have two different encodings instead of the flag? (if it is needed, or just use the other encoding in tests if needed)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala
```diff
@@ -754,36 +754,50 @@ case object CompactJsonScanHttpEncodings extends ScanHttpEncodings {
...
-      case list: javaApi.DamlList => javaApi.DamlList.of(list.toList(withoutFieldLabels))
-      case tmap: javaApi.DamlTextMap => javaApi.DamlTextMap.of(tmap.toMap(withoutFieldLabels))
-      case gmap: javaApi.DamlGenMap =>
-        javaApi.DamlGenMap.of(gmap.toMap(withoutFieldLabels, withoutFieldLabels))
-      case opt: javaApi.DamlOptional =>
-        javaApi.DamlOptional.of(opt.getValue.map(withoutFieldLabels))
-      case variant: javaApi.Variant =>
-        new javaApi.Variant(variant.getConstructor, withoutFieldLabels(variant.getValue))
-      case _ => value
+  private def maybeWithoutFieldLabels(value: javaApi.Value): javaApi.Value = {
```
ray-roestenburg-da: Would really appreciate just a different encoder, results in cleaner separate code

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -0,0 +1,184 @@
...
+
+class UpdateHistoryBulkStorage(
+    val config: BulkStorageConfig,
+    val updateHistory: UpdateHistory,
+    val s3Connection: S3BucketConnection,
+    override val loggerFactory: NamedLoggerFactory,
+)(implicit actorSystem: ActorSystem, tc: TraceContext, ec: ExecutionContext)
+    extends NamedLogging {
+
+  class UpdateHistorySegmentBulkStorage(
```
ray-roestenburg-da: why is this class an instance level inner class? Prefer if they are side by side, not nested

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -0,0 +1,184 @@
...
+      val toMigrationId: Long,
+      val toTimestamp: CantonTimestamp,
+  ) {
+    private val liveSource: Source[ByteString, SourceQueueWithComplete[ByteString]] =
+      Source.queue[ByteString](100, overflowStrategy = OverflowStrategy.backpressure)
+    private val (queue, initStream) = liveSource
+      .toMat(Sink.asPublisher(fanout = false))(org.apache.pekko.stream.scaladsl.Keep.both)
+      .run()
+    private val s3ObjIdx = new AtomicInteger(0)
+    private val position = new AtomicReference[Option[(Long, CantonTimestamp)]](None)
```
ray-roestenburg-da: do s3ObjIdx and position need to be changed atomically together or is it fine if they are updated independently? (haven't looked further yet at the code but came to mind, maybe my comment does not make sense)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
ray-roestenburg-da: +1 I don't think this is necessary

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala
isegall-da: I don't understand.
The existing encoding (`CompactJsonScanHttpEncodings`) is unchanged, the only thing I did was introduce a class "above" it that parameterizes whether to keep the labels or not, and in test case add an encoding that reuses that with labels. So we do have two different encodings. 
Not sure I understand what are you proposing instead? Definitely don't want to copy-paste the entire encoding and duplicate the entire code just to change one piece of behavior.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala
isegall-da: See above. That's exactly what I did, just refactored the reusable parts (99% of it) to a shared class between the two encoders.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: Because I don't know pekko, and as soon as something works I don't touch it again :D 
Will try that, thanks.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotBulkStorageTest.scala
isegall-da: Argh, of course not, meant to delete it. Sorry.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/HasS3Mock.scala
isegall-da: Note that this is just a refactoring, moving the existing code as it is. Didn't want to overload this PR with even more changes. I think I'll just move this part back to where it was, didn't end up reusing it anyway.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/HasS3Mock.scala
isegall-da: yup, exactly. Just empirically sufficed and everything should be deterministic.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: Hmmm.. I think you're right. This is copy-pasted from the ACS snapshot one, but the same comment applies there as well.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala
ray-roestenburg-da: I would prefer something like:

```
object RemoveFields {
  def record(value: javaApi.DamlRecord): javaApi.DamlRecord = recordWithoutFieldLabels(value)
  def value(value: javaApi.Value): javaApi.Value = valueWithoutFieldLabels(value)
  /** Recursively removes all field labels from a value.
   * ValueJsonCodecCodegen returns values with field labels, but we generally don't store field labels in databases.
   * The labels are removed to make values comparable.
   */
  private def valueWithoutFieldLabels(value: javaApi.Value): javaApi.Value = {
    value match {
      case record: javaApi.DamlRecord => recordWithoutFieldLabels(record)
      case list: javaApi.DamlList => javaApi.DamlList.of(list.toList(valueWithoutFieldLabels))
      case tmap: javaApi.DamlTextMap => javaApi.DamlTextMap.of(tmap.toMap(valueWithoutFieldLabels))
      case gmap: javaApi.DamlGenMap =>
        javaApi.DamlGenMap.of(gmap.toMap(valueWithoutFieldLabels, valueWithoutFieldLabels))
      case opt: javaApi.DamlOptional =>
        javaApi.DamlOptional.of(opt.getValue.map(valueWithoutFieldLabels))
      case variant: javaApi.Variant =>
        new javaApi.Variant(variant.getConstructor, valueWithoutFieldLabels(variant.getValue))
      case _ => value
    }
  }
  private def recordWithoutFieldLabels(value: javaApi.DamlRecord): javaApi.DamlRecord = {
    val fields = value.getFields.asScala.toList
    val fieldsWithoutLabels = fields.map { f =>
      new javaApi.DamlRecord.Field(valueWithoutFieldLabels(f.getValue))
    }
    new javaApi.DamlRecord(fieldsWithoutLabels.asJava)
  }
}
object CompactJsonScanHttpEncodings {
  def apply() = CompactJsonScanHttpEncodings(identity, identity)
}
object CompactJsonScanHttpEncodingsWithoutFields {
  def apply() = CompactJsonScanHttpEncodings(RemoveFields.value, RemoveFields.record)
}
// A lossy, but much easier to process, encoding. Should be used for all endpoints not used for backfilling Scan.
case class CompactJsonScanHttpEncodings(
  transformValue: javaApi.Value => javaApi.Value = identity, 
  transformRecord: javaApi.DamlRecord => javaApi.DamlRecord = identity
) extends ScanHttpEncodings {
  import org.lfdecentralizedtrust.splice.util.ValueJsonCodecCodegen
  override def encodeContractPayload(
      event: javaApi.CreatedEvent
  )(implicit elc: ErrorLoggingContext): Json =
    ValueJsonCodecCodegen
      .serializableContractPayload(event)
      .fold(
        err => {
          elc.error(s"Failed to encode contract payload: $err")
          encodeValueFallback(err, event.getArguments)
        },
        tryParseJson,
      )

  override def encodeChoiceArgument(
      event: javaApi.ExercisedEvent
  )(implicit elc: ErrorLoggingContext): Json =
    ValueJsonCodecCodegen
      .serializeChoiceArgument(event)
      .fold(
        err => {
          elc.error(s"Failed to encode choice argument: $err")
          encodeValueFallback(err, event.getChoiceArgument)
        },
        tryParseJson,
      )

  override def encodeExerciseResult(
      event: javaApi.ExercisedEvent
  )(implicit elc: ErrorLoggingContext): Json =
    ValueJsonCodecCodegen
      .serializeChoiceResult(event)
      .fold(
        err => {
          elc.error(s"Failed to encode exercise result: $err")
          encodeValueFallback(err, event.getExerciseResult)
        },
        tryParseJson,
      )

  override def decodeContractPayload(
      templateId: javaApi.Identifier,
      json: Json,
  ): javaApi.DamlRecord =
    ValueJsonCodecCodegen
      .deserializableContractPayload(templateId, json.noSpaces)
      .fold(
        error =>
          throw new RuntimeException(
            s"Failed to decode contract payload '${json.noSpaces}': $error"
          ),
        transformRecord,
      )

  override def decodeChoiceArgument(
      templateId: javaApi.Identifier,
      interfaceId: Option[javaApi.Identifier],
      choice: String,
      json: Json,
  ): javaApi.Value =
    ValueJsonCodecCodegen
      .deserializeChoiceArgument(templateId, interfaceId, choice, json.noSpaces)
      .fold(
        error =>
          throw new RuntimeException(
            s"Failed to decode choice argument '${json.noSpaces}': $error"
          ),
        transformValue,
      )

  override def decodeExerciseResult(
      templateId: javaApi.Identifier,
      interfaceId: Option[javaApi.Identifier],
      choice: String,
      json: Json,
  ): javaApi.Value =
    ValueJsonCodecCodegen
      .deserializeChoiceResult(templateId, interfaceId, choice, json.noSpaces)
      .fold(
        error =>
          throw new RuntimeException(s"Failed to decode choice result '${json.noSpaces}': $error"),
        transformValue,
      )
}
```
This extracts the code without introducing the boolean, which I would really prefer.
(I hate booleans and changing the behavior in a class, I rather have instances per behavior, more 'single responsibility', and extract behavior in functions or other types which are delegated to )

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1031,16 +1031,18 @@ class UpdateHistory(
   private def afterFilters(
       afterO: Option[(Long, CantonTimestamp)],
       includeImportUpdates: Boolean,
+      afterIsInclusive: Boolean,
```
ray-roestenburg-da, line 1034: Since I hate booleans and boolean arguments specifically, I would prefer something that sets up the behavior in a function for instance, that ends up in one of these sql fragments, and it gets decided which to use further up.. or that you have separate method defs for these so when you read it you know immediately what it does. Sorry for going a bit hard on the readability nitpicky side.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala
ray-roestenburg-da: I added a suggestion, see https://github.com/hyperledger-labs/splice/pull/3495/changes/BASE..72684a376459bfdc9decc867a22ba9f9059f60ef#r2682727851

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -0,0 +1,184 @@
...
+
+    /** Should be called repeatedly. Each call attempts to fetch `config.dbReadChunkSize` updates from the DB and push
+      * them to the stream. Once the segment is finished, will close the stream and must not be called again
+      *
+      * @return
+      *   Result.NotReady: not enough updates in the DB yet, did not push any data to the stream.
+      *                    The caller should retry later (after some time, so that more updates will be ingested)
+      *
+      *   Result.NotDone:  Updates were read and pushed to the stream, but the segment has not finished yet.
+      *                    More data could be available in the DB already, so the called usually wants to call next() again immediately.
```
ray-roestenburg-da:
```suggestion
      *                    More data could be available in the DB already, so the caller usually wants to call next() again immediately.
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -0,0 +1,184 @@
...
+      *
+      *   Result.Done:     Done reading and pushing updates to this segment. The pipeline is closed, and next() must not be called again
+      *                    for this segment.
+      */
+    def next(): Future[Result.Result] = {
```
ray-roestenburg-da: you can probably do this as a stream, return a source and let someone run it if you like instead of having to poll next() and sequence the future properly, this way might be more convenient.

ray-roestenburg-da: Maybe something like `Source.unfoldAsync`, and put the position in the stream, as a tuple. You then also don't need the position AtomicReference

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/HasS3Mock.scala
```diff
@@ -0,0 +1,109 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+import scala.sys.process.*
+import scala.util.Using
+
+trait HasS3Mock extends NamedLogging with FutureHelpers with EitherValues with BaseTest {
+
+  // TODO(#3429): consider running s3Mock container as a service in GHA instead of starting it here
+  def withS3Mock[A](test: => Future[A])(implicit ec: ExecutionContext): Future[A] = {
+    // start s3Mock in a contaner
+    "docker run -p 9090:9090 -e COM_ADOBE_TESTING_S3MOCK_STORE_INITIAL_BUCKETS=bucket -e debug=true -d --rm --name s3mock adobe/s3mock".!
```
ray-roestenburg-da, line 25: maybe look into testcontainers way of doing this instead of running commands

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/HasS3Mock.scala
```diff
@@ -0,0 +1,109 @@
...
+    val uncompressedNio = uncompressedDirect.nioBuffer(0, uncompressedDirect.capacity())
+    compressedDirect.writeBytes(compressed)
+    Using(new ZstdDirectBufferDecompressingStream(compressedDirect.nioBuffer())) {
+      _.read(uncompressedNio)
+    }
+    uncompressedNio.flip()
+    val allContractsStr = StandardCharsets.UTF_8.newDecoder().decode(uncompressedNio).toString
+    val allContracts = allContractsStr.split("\n")
+    compressedDirect.release()
+    uncompressedDirect.release()
```
ray-roestenburg-da: At least do try { } finally { release }

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
rautenrieth-da: I assume the `next()` API is designed so that it's convenient to call from a `PollingTrigger`. Similar to [HistoryBackfilling.backfill()](https://github.com/hyperledger-labs/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/HistoryBackfilling.scala#L62) for example.

We also have `SourceBasedTrigger`, but that one assumes you have a Source of tasks, where each task is processed individually using `processTaskWithRetry()`, so not sure how well it maps to the processing pipeline here.

ray-roestenburg-da: yeah or break out of the trigger model, this is more like the ingestion which just runs, a service that runs a stream

isegall-da: WDYM "just runs"? What other mechanism other than a polling trigger do we have for a service that runs in the background and periodically makes progress? And why break out of the trigger model that works just fine?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/HasS3Mock.scala
isegall-da: nice, forgot that we had that. Adding as an alternative in the TODO for now

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
isegall-da: I tend to agree, but note that this is going away as soon as part of fixing the < vs <= thing, which is my next step, so I don't want to overengineer this for this intermediate step.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: No, they're independent

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: > Prefer if they are side by side, not nested

Why's that? The segment uses all the `val`s of the general storage object, so you'll need to either duplicate them, or expose those, and pass the general object as an argument to the segment one, which IMHO is less nice than nesting.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala
isegall-da: Nice, I like it.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: Is there something better than a `queue` to use here then, or should it just be a queue of size 2 or something like that?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: Hmmm... But I still need `liveSource.toMat.run()` in order to have the `queue` object that I can push stuff to in `next()`, don't I? So while I can perhaps change `initStream` here to `liveSource`, not much else would change, would it?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/HasS3Mock.scala
isegall-da: Adding a comment

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/HasS3Mock.scala
isegall-da: Added try..finally as Ray suggested

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
ray-roestenburg-da: I don't even know why you have two classes TBH. Could just be implemented in the UpdateHistoryBulkStorage. But if you really want a separate class, then I would not nest it, it makes it more clear how the two interact, and in case the code gets bigger, easier in two files. nesting the classes makes them share rather private things, I would prefer a more defined interaction where every class does one specific thing.

ray-roestenburg-da: Actually the outer class doesn't really do anything, you could have just implemented it as `UpdateHistorySegmentBulkStorage`, and just pass it all the args it needs, which would make more sense to me

ray-roestenburg-da: For instance (sorry for pasting this in like this)
```
// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
// SPDX-License-Identifier: Apache-2.0

package org.lfdecentralizedtrust.splice.scan.store.bulk

import com.digitalasset.canton.data.CantonTimestamp
import com.digitalasset.canton.logging.{NamedLoggerFactory, NamedLogging}
import com.digitalasset.canton.tracing.TraceContext
import org.apache.pekko.actor.ActorSystem
import org.apache.pekko.stream.{OverflowStrategy, QueueOfferResult}
import org.apache.pekko.stream.scaladsl.{Sink, Source, SourceQueueWithComplete}
import org.apache.pekko.util.ByteString
import org.lfdecentralizedtrust.splice.http.v0.definitions
import org.lfdecentralizedtrust.splice.scan.admin.http.ScanHttpEncodings
import org.lfdecentralizedtrust.splice.store.{HardLimit, TreeUpdateWithMigrationId, UpdateHistory}

import java.nio.ByteBuffer
import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
import io.circe.syntax.*
import org.apache.pekko.Done

import java.nio.charset.StandardCharsets

class UpdateHistorySegmentBulkStorage(
    config: BulkStorageConfig,
    updateHistory: UpdateHistory,
    s3Connection: S3BucketConnection,
    fromMigrationId: Long,
    fromTimestamp: CantonTimestamp,
    toMigrationId: Long,
    toTimestamp: CantonTimestamp,
    override val loggerFactory: NamedLoggerFactory,
)(implicit actorSystem: ActorSystem, tc: TraceContext, ec: ExecutionContext) extends NamedLogging {
  private val liveSource: Source[ByteString, SourceQueueWithComplete[ByteString]] =
    Source.queue[ByteString](100, overflowStrategy = OverflowStrategy.backpressure)
  private val (queue, initStream) = liveSource
    .toMat(Sink.asPublisher(fanout = false))(org.apache.pekko.stream.scaladsl.Keep.both)
    .run()
  private val s3ObjIdx = new AtomicInteger(0)
  private val position = new AtomicReference[Option[(Long, CantonTimestamp)]](None)

  private val pipeline: Future[Done] = Source
    .fromPublisher(initStream)
    .via(ZstdGroupedWeight(config.maxFileSize))
    // Add a buffer so that the next object continues accumulating while we write the previous one
    .buffer(
      1,
      OverflowStrategy.backpressure,
    )
    .mapAsync(1) { case ByteStringWithTermination(zstdObj, isLast) =>
      val objectKey = if (isLast) s"updates_${s3ObjIdx}_last.zstd" else s"updates_$s3ObjIdx.zstd"
      // TODO(#3429): For now, we accumulate the full object in memory, then write it as a whole.
      //    Consider streaming it to S3 instead. Need to make sure that it then handles crashes correctly,
      //    i.e. that until we tell S3 that we're done writing, if we stop, then S3 throws away the
      //    partially written object.
      for {
        _ <- s3Connection.writeFullObject(objectKey, ByteBuffer.wrap(zstdObj.toArrayUnsafe()))
      } yield {
        s3ObjIdx.addAndGet(1)
        ()
      }
    }
    .runWith(Sink.ignore)

  private def encodeAndOfferToQueue(updates: Seq[TreeUpdateWithMigrationId]): Future[Unit] = {
    if (updates.nonEmpty) {
      val encoded = updates.map(update =>
        ScanHttpEncodings.encodeUpdate(
          update,
          definitions.DamlValueEncoding.CompactJson,
          ScanHttpEncodings.V1,
        )
      )
      val updatesStr = encoded.map(_.asJson.noSpacesSortKeys).mkString("\n") + "\n"
      val updatesBytes = ByteString(updatesStr.getBytes(StandardCharsets.UTF_8))
      logger.debug(
        s"Read ${encoded.length} updates from DB, to a bytestring of size ${updatesBytes.length} bytes"
      )
      queue.offer(updatesBytes).map {
        case QueueOfferResult.Enqueued => ()
        case QueueOfferResult.Dropped =>
          throw new RuntimeException(
            "Queue overflow. This should not happen with OverflowStrategy.backpressure"
          )
        case QueueOfferResult.Failure(ex) => throw ex
        case QueueOfferResult.QueueClosed => throw new RuntimeException("Queue closed")
      }
    } else Future.successful(())
  }

  /** Should be called repeatedly. Each call attempts to fetch `config.dbReadChunkSize` updates from the DB and push
    * them to the stream. Once the segment is finished, will close the stream and must not be called again
    *
    * @return
    *   Result.NotReady: not enough updates in the DB yet, did not push any data to the stream.
    *                    The caller should retry later (after some time, so that more updates will be ingested)
    *
    *   Result.NotDone:  Updates were read and pushed to the stream, but the segment has not finished yet.
    *                    More data could be available in the DB already, so the called usually wants to call next() again immediately.
    *
    *   Result.Done:     Done reading and pushing updates to this segment. The pipeline is closed, and next() must not be called again
    *                    for this segment.
    */
  def next(): Future[Result.Result] = {
    for {
      updates <- position
        .get()
        .fold(
          updateHistory.getUpdatesWithoutImportUpdates(
            Some((fromMigrationId, fromTimestamp)),
            HardLimit.tryCreate(config.dbReadChunkSize),
            afterIsInclusive = true,
          )
        )(after =>
          updateHistory.getUpdatesWithoutImportUpdates(
            Some((after._1, after._2)),
            HardLimit.tryCreate(config.dbReadChunkSize),
            afterIsInclusive = false,
          )
        )

      // TODO(#3429): Figure out the < vs <= issue
      updatesInSegment = updates.filter(update =>
        update.migrationId < toMigrationId ||
          update.migrationId == toMigrationId && update.update.update.recordTime < toTimestamp
      )
      _ <-
        if (updatesInSegment.length < updates.length || updates.length == config.dbReadChunkSize) {
          logger.debug(s"Adding ${updatesInSegment.length} updates to the queue")
          encodeAndOfferToQueue(updatesInSegment)
        } else {
          logger.debug(
            s"Not enough updates yet (queried for ${config.dbReadChunkSize}, found ${updates.length}), nothing to do"
          )
          Future.successful(())
        }
    } yield {
      if (updatesInSegment.length < updates.length) {
        queue.complete()
        Result.Done
      } else if (updatesInSegment.length < config.dbReadChunkSize) {
        Result.NotReady
      } else {
        val last = updatesInSegment.lastOption.getOrElse(
          throw new RuntimeException("No updates added, unexpectedly")
        )
        position.set(Some((last.migrationId, last.update.update.recordTime)))
        Result.NotDone
      }
    }
  }

  def watchCompletion(): Future[Unit] =
    pipeline.map(_ => ())
}


object UpdateHistorySegmentBulkStorage {
  def apply(
      config: BulkStorageConfig,
      updateHistory: UpdateHistory,
      s3Connection: S3BucketConnection,
      fromMigrationId: Long,
      fromTimestamp: CantonTimestamp,
      toMigrationId: Long,
      toTimestamp: CantonTimestamp,
  )(implicit actorSystem: ActorSystem, tc: TraceContext, ec: ExecutionContext) = new UpdateHistorySegmentBulkStorage(
    config: BulkStorageConfig,
    updateHistory: UpdateHistory,
    s3Connection: S3BucketConnection,
    fromMigrationId,
    fromTimestamp,
    toMigrationId,
    toTimestamp,
    loggerFactory: NamedLoggerFactory
  )
}

case object Result {
  sealed trait Result
  case object Done extends Result
  case object NotDone extends Result
  case object NotReady extends Result
}
```

isegall-da: @ray-roestenburg-da I think I'm missing something obvious, but can you help me figure out what's wrong in https://github.com/hyperledger-labs/splice/pull/3495/changes/3a990969cf3d2653c0efbe8c7e25727a95dfa9e3 ? 

It fails to compile with 
```
[error] /home/itai/cn-1/splice/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala:516:65: type mismatch;
[error]   CompactJsonScanHttpEncodings.type|org.lfdecentralizedtrust.splice.scan.admin.http.ScanHttpEncodings
[error]       case definitions.DamlValueEncoding.members.CompactJson => CompactJsonScanHttpEncodings
```

isegall-da: (whoops, wrong place for the comment, sorry, this is for your proposal above for the encodings...)

isegall-da: > Actually the outer class doesn't really do anything

yet... The idea is for it to manage the segments. I personally don't have a problem with nested classes that share private things, that's exactly their point. But I don't feel strongly about it, don't mind moving it out.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
ray-roestenburg-da: If you do source.unfoldAsync you don't need the queue, and you can just have a stream from source to sink and run it in this class, and have it retry on failure, add a killswitch on it to stop the stream and have this be a service you just start in the app as a BackgroundService like for instance the UpdateIngestionService.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
ray-roestenburg-da: You build the stream, materialize and run it, with a kill switch to stop it, you don't need a trigger here at all, this is just a background service. It makes progress based on how you setup the stream, it keeps getting the updates as quickly as it can (or if you want to throttle you can do that in the stream). pekko stream has a very elegant programming model to do this. My two cents.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: Yeah, that's more or less what I did in the ACS snapshots one, but that goes against how the entire app is designed with triggers, so I think makes it much more confusing.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: But this is just for one segment. Something needs to manage the different segments, and intermingle them with ACS snapshots correctly

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: I ended up removing the surrounding "manager", collapsing this into just a segment class for now. Will see where and how to implement the manager in a future step.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: Spoke with @ray-roestenburg-da  about this. We basically have two design options:
1. Implement this along the lines of the backfilling code (the direction taken in this PR), i.e. expose a "next()" method, and drive progress from a polling trigger that hits "next()" repeatedly (potentially on different objects, i.e. on different segments, different ACS snapshot dumps etc).
2. More "Pekko-native", using a materialized stream that just make progress in the background, feeds updates&snapshots as a source, and makes progress outside of the triggers infra (this is the direction I originally took in the ACS snapshots dump part in https://github.com/hyperledger-labs/splice/pull/3423).

I'm not fully decided on either, so for now I am going to leave this open, leave the two pieces as they are, and as I start working on wiring everything together make my mind on which to use. 

@moritzkiefer-da you tend to have good intuition on these things. If I go down (2), which seems quite elegant actually, albeit admittedly a bit inconsistent with how most of our other background processes work, will I regret it?

ray-roestenburg-da: @rautenrieth-da ScanVerdictStoreIngestion is also a SourceBasedTrigger, which is basically reading from mediator and ingesting into the db, which is conceptually not very different, so I think a SourceBasedTrigger is also an option

rautenrieth-da: Ah that's a good comparison. That would mean the trigger source would be the whole pipeline for reading update history chunks from the db and encoding/compressing them into byte arrays, and the trigger task would be the writing of the S3 object. I like that model. We might have to change the code a bit, as right now (AFAICT) the pipeline in `UpdateHistoryBulkStorage` exports only one segment of the history - somehow the pipeline would have to be aware of what segment each update belongs to (I was assuming the orchestration of segments would come in an upcoming PR).

On the other hand, the source based triggers are probably less robust than a polling trigger that repeats "read state -> decide action -> execute action" forever. If a task fails after 2-3min of retries, it will be skipped and you'll get missing files in S3. Using some hack like `maxRetries = Int.MaxValue` to avoid skipping tasks doesn't sound like a good idea. Also, if the SourceBasedTrigger source closes unexpectedly, it won't be restarted.

The ledger ingestion service uses a double retry loop that restarts the whole source if it completes unexpectedly, maybe we'd need something more similar to that.

ray-roestenburg-da: FYI The ScanVerdictStoreIngestion uses a restartSource that manages errors (like closing of the source etc). If you see the code it's a little shoehorned into the trigger, cause you could have just written one stream for this. (but it works well enough) Where is this logic for "If a task fails after 2-3min of retries, it will be skipped "? With that in mind, I think it would be ideal to just write this as a background service and do specifically what you want there, not to plug it into something just because it's there, and then having to worry about these kind of assumptions for what makes sense for triggers. I would prefer to reuse some functionality for handling errors and restarts etc by composition, by inserting stream stages in the pipeline (like restartSource, and put that error handling from ingestion service in a stage / stream operator for instance), instead of inheriting this from a base class and possibly getting unwanted behavior.

rautenrieth-da: > Where is this logic

`processTaskWithRetry()` [retries tasks](https://github.com/hyperledger-labs/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/TaskbasedTrigger.scala#L107), by default [up to 35x with <5sec delays](https://github.com/hyperledger-labs/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/RetryFor.scala#L24) before giving up.

[Failed tasks log warnings and record metrics](https://github.com/hyperledger-labs/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/TaskbasedTrigger.scala#L199), but other than that the [pekko pipeline ignores whether tasks are successful or not](https://github.com/hyperledger-labs/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SourceBasedTrigger.scala#L59).

ray-roestenburg-da: In these kinds of pipelines I think that's ok since it always needs the 'offset' to continue so that is safeguarded in the source unfold, the offset would not progress if the task fails.. but even that mental gymnastics are unnecessary if you just have a "managed pipeline" in a background service. In any case I've given more than 2 cents now LOL

Issue #3488, opened 2026-01-08
Implement version override per SV


mblaze-da:
Enable overriding version per SV through YAML configuration

PR #3507, opened 2026-01-09
implement version override per SV

mblaze-da:
fixes https://github.com/hyperledger-labs/splice/issues/3488


martinflorian-da (commented):
Looks very reasonable so far, thank you! Not entirely opposed to merging this without `sv-canton` for now, for smaller PRs. Up to you... (not approving in case you decide to make a bigger PR).

In any case: I'd strongly suggest doing a hdm test before merging anything related to versions...

Also some thoughts on `sv-canton`, for when you get to it:

- you probably want to ignore the override when deploying legacy or upgrade nodes, and apply it only on active; not sure if this will ever matter in practice, but semantics-wise it seems most reasonable that way?
- it *might* be a candidate for a unit test to make sure the override gets applied to all deployed releases where it matters (participant and `splice-global-domain`...); you'll probably be in a better position to judge whether it's worth it

Issue #3059, opened 2025-11-03
Support configurable threshold for trusted sequencers in ValidatorDecentralizedSynchronizerConfig


pasindutennage-da:
### Background

In the current validator implementation, `ValidatorDecentralizedSynchronizerConfig` accepts `sequencerNames` as an optional parameter. `sequencerNames` specifies the set of trusted sequencers the validator's participant will connect to.

### Problem

Current implementation does not support configurable threshold specifying how many trusted sequencers should be used at a time.

### Proposed Solution

Extend the `ValidatorDecentralizedSynchronizerConfig` to allow threshold parameter, that specifies the number of sequencer connections to connect. 

Changes are required in `ValidatorDecentralizedSynchronizerConfig`, `DomainConnector`, and `ValidatorSequencerConnectionIntegrationTest` classes.

PR #3512, opened 2026-01-09
Support configurable threshold for trusted sequencers

pasindutennage-da:
Fixes https://github.com/hyperledger-labs/splice/issues/3059

### DIFF
--- a/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
+++ b/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
@@ -160,6 +160,10 @@ class DomainConnector(
                       case None =>
                         Thresholds.sequencerConnectionsSizeThreshold(nonEmptyConnections.size)
                     }
+
+                  // max(threshold, sequencerConnectionsSizeThreshold) is used to ensure that even in small configurations
+                  // (e.g., N=3, f+1=1) we have enough retries to make sure that the request is eventually processed by a live sequencer.
+
                   val amplificationFactor = PositiveInt.tryCreate(
                     Math.max(
                       threshold.unwrap,
@@ -168,6 +172,7 @@ class DomainConnector(
                         .unwrap,
                     )
                   )
+
                   SequencerConnections.tryMany(
                     nonEmptyConnections.forgetNE,
                     threshold,


apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceConfig.scala
```diff
@@ -700,10 +700,20 @@ object SpliceConfig {
...
             ConfigValidationFailed(
-              "Configuration error: `url` and `sequencerNames` cannot both be specified for the global domain."
+              "Configuration error: `url` and `sequencerNames/threshold` are mutually exclusive parameters."
+            ),
+          )
+          _ <- Either.cond(
+            !(conf.domains.global.sequencerNames.isDefined && conf.domains.global.threshold.isDefined &&
+              conf.domains.global.sequencerNames
+                .map(x => x.length)
+                .getOrElse(0) > conf.domains.global.threshold.getOrElse(0)),
```
pasindutennage-da: >=

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/config/ValidatorAppConfig.scala
```diff
@@ -104,7 +104,10 @@ case class ValidatorDecentralizedSynchronizerConfig(
       */
     sequencerNames: Option[NonEmptyList[String]] = None,
 
-    // TODO(#3059): Add a threshold parameter to configure the number of sequencers that must be connected
+    /** An optional parameter to specify the BFT threshold for the domain connections.
+      * If not specified, length(sequencerNames) / 3 +1 will be used.
+      */
+    threshold: Option[Int] = None,
```
nicu-da: as they only make sense together, let's wrap them up in a case class so that the types also show the intent

pasindutennage-da: Done

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
```diff
@@ -153,11 +154,26 @@ class DomainConnector(
...
+                      case Some(config) => PositiveInt.tryCreate(config.threshold)
+                      case None =>
+                        Thresholds.sequencerConnectionsSizeThreshold(nonEmptyConnections.size)
+                    }
+
+                  val amplificationFactor = PositiveInt.tryCreate(
+                    Math.max(
+                      threshold.unwrap,
+                      Thresholds
+                        .sequencerSubmissionRequestAmplification(nonEmptyConnections.size)
```
nicu-da, line 171: Just for my knowledge, do you remember why we tie amplification to the threshold? I understand it not being higher than total number of sequencers, but that's basically the number of retries so why does it change with the number of sequencers?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
```diff
@@ -63,30 +69,56 @@ class ValidatorSequencerConnectionIntegrationTest
         sv4ScanBackend,
       )
 
+      // We must wait for the sequencers to "mature" in Scan before starting the validator.
+      // WITHOUT THIS: Alice queries Scan and finds that although 4 sequencers are listed,
+      // some (like SV3) have an 'availableAfter' timestamp in the future.
+      // Alice's DomainConnector correctly ignores these "not yet available" sequencers.
+      // If Alice only finds 1 valid sequencer but is configured with a trust threshold of 2,
+      // she will crash on startup to prevent operating with insufficient trust.
```
nicu-da: This does make me wonder if we should instead have a retry before it crashes?

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
pasindutennage-da: Amplification ensures we contact enough sequencers simultaneously to meet the trust threshold immediately. This prevents the system from waiting for a "patience" timeout before asking additional sequencers for required responses
.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
pasindutennage-da: which `retry` are you referring? Are you implying, in the validatorApp, before crash is triggered, we retry fetching the scans again?

In production, i think this will rarely happen, because there is a minute scale time difference between spawning a sequencer and spawning a validator. However, this is a timing assumption bug, that should be fixed, (perhaps outside this PR, not sure)

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
nicu-da:
```
/** Configures the submission request amplification. Amplification makes sequencer clients send
  * eligible submission requests to multiple sequencers to overcome message loss in faulty
  * sequencers.
  *
  * @param factor
  *   The maximum number of times the submission request shall be sent.
  * @param patience
  *   How long the sequencer client should wait after an acknowledged submission to a sequencer to
  *   observe the receipt or error before it attempts to send the submission request again (possibly
  *   to a different sequencer).
  */

```

This is the doc for the amplification, maybe I'm misunderstand it but are you suggesting that it triggers `factor` number of submissions at the same time?

The sequencer threshold represents the number of identical reads, and as sequencers are all replicates state machines a single write will produce reads on all of them.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
nicu-da:
```
      // she will crash on startup to prevent operating with insufficient trust.
```

at what point does it crash? If it's in the domain connector because we don't have enough sequencers yet, we should keep retrying for a while until we at least meet the threshold

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
martinflorian-da: Yup we definitely don't do writes in parallel, that would waste traffic on the happy path.

> do you remember why we tie amplification to the threshold

I'm not entirely sure tbh, but I'd just keep doing it here for consistency...

I'm reasonably sure that in practice we don't get to amplify more than a few times anyway before something else times out. So probably all values beyond 3 don't move the needle significantly anyway.

nicu-da: Why would we keep it here as well? We should actually fix it and just set it to 2 or 3

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
martinflorian-da: @pasindutennage-da Can you elaborate what exactly is crashing here? Is it Alice's participant, after you gave it an unsound sequencer connections config? (If so: That seems a bit questionable, should it not just reject the config change instead? We might want to open a Canton issue about this...) Or do you mean that the validator app will crash? (If so: I agree with @nicu-da that we should be able to retry/wait our way out of that...)

In the latter case I'd probably lean towards "let's just fix this in this PR or the very next one, while the context is fresh". In the former it's probably better to keep things as they are, track this as an upstream issue and accept that participants with wrong (/ too "fresh") configs will crash loop for a while...

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
martinflorian-da: > Why would we keep it here as well?

To not add scope creep to this issue/PR...

> We should actually fix it and just set it to 2 or 3

Maybe you're right... but I'd track / prioritize this separately.

There is always some risk when changing things that deep, and there is no pressure to fix this as it doesn't really cause issues for users... or am I missing something?

nicu-da: Sure, not an issue it's just an extremely confusing behavior imo

pasindutennage-da: For now, I reset the writeAmplification value to what it was before; f+1

I created a tech debt issue, and assigned my self

https://github.com/hyperledger-labs/splice/issues/3537

Please add anything relevant to that issue.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
pasindutennage-da: The following is the behavior, without the explicit (synchronization) health check in the test.

```
aliceValidator app initialization: Initialization failed, so exiting; check the application logs for details
java.lang.RuntimeException: aliceValidator app initialization: Initialize app failed
Caused by: java.lang.RuntimeException: aliceValidator app initialization: Initialize node failed
Caused by: java.lang.RuntimeException: aliceValidator app initialization: Ensuring decentralized synchronizer registered failed
Caused by: java.lang.IllegalArgumentException: requirement failed: sequencerTrustThreshold cannot be greater than number of sequencer connections. Found threshold of 2 and 1 sequencer connections
```

```
private def extractValidConnections(
      sequencers: Seq[DsoSequencer],
      domainTime: CantonTimestamp,
      migrationId: Long,
  )(implicit tc: TraceContext): Seq[GrpcSequencerConnection] = {
    // sequencer connections will be ignore if they are with a invalid Alias, empty url or not yet available (`before availableAfter`)

    logger.info(s"Available sequencers before filtering: ${sequencers}")
    logger.info(s"Domain Time: ${domainTime}")

    val validConnections = sequencers
      .collect {
        case DsoSequencer(sequencerMigrationId, _, url, svName, availableAfter)
            if migrationId == sequencerMigrationId && url.nonEmpty && !domainTime.toInstant
              .isBefore(availableAfter) =>
          for {
            sequencerAlias <- SequencerAlias.create(svName)
            grpcSequencerConnection <- GrpcSequencerConnection.create(
              url,
              None,
              sequencerAlias,
            )
          } yield grpcSequencerConnection
      }
      .collect { case Right(conn) =>
        conn
      }
    logger.info(s"Available sequencers after filtering: ${validConnections}")
    validConnections
  }
}
```

```
Available sequencers before filtering: 

sv1:: http://localhost:5108,Digital-Asset-2,           2026-01-14  T 10:24:34.387185
sv2:: http://localhost:5208,Digital-Asset-Eng-2,   2026-01-14  T 10:26:57.438164
sv3:: http://localhost:5308,Digital-Asset-Eng-3,   2026-01-14  T 10:27:00.372067

Domain Time: 2026-01-14 T 10:26:01.771226

Available sequencers after filtering

Sequencer 'Digital-Asset-2', endpoints = http://localhost:5108)) 
```

**Summary**

After fetching sequencer connections from scan, validator checks whether the timestamp of sequencer connection is before the domain time stamp (any future timestamped connections are dropped). This causes threshold check failure.

**Why we did not see this behavior before**

Before, we did tests with default threshold 1, and with very high chance at least one sequencer has a timestamp satisfying the condition. With threshold set to 2 (my test case), now, the threshold logic fails.

**Fix**

`extractValidConnections` should be wrapped in a retry loop, such that if it fails, validator should wait some milliseconds and retry.

**Next step**: 

Allthough the issue comes after my introduction of threshold param, it is not a direct consequence of the threshold param.

Option 1: close this PR with the explicit health check in the test, and solve the retry in a different PR
Option 2: do it in same PR

martinflorian-da: Thanks a lot for your thorough analysis, @pasindutennage-da ! Can we just extend `additionalCodes` here so that it retries on `IllegalArgumentException`? https://github.com/hyperledger-labs/splice/blob/1932c338de715b54684fe5bececbbf96b5d36e35/apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala#L130

I'm not sure if the short retry duration there will be enough to un-flake the test though. After reading the comments there and thinking a bit about it though, I wouldn't bother with retrying longer. We'd be making the UX worse (failing more slowly on actual errors) just for speeding up our test slightly... seems like a wrong thing to do. Keeping the explicit wait in the test might be the better way then.

pasindutennage-da: Thanks

Adding discussion I had with @nicu-da 

The easiest fix is to "busy-wait" in the validator side; if condition is not met, then update the domainTime to time.Now() and recheck. (and do it for some number of retries)

However, for all practical purposes in deployment, this would never be a possible execution, hence, we would not see this race condition in practice (validators will never start before the network is up for some time)

For now, decided to just have a readiness wait in the test. 

Created an issue https://github.com/hyperledger-labs/splice/issues/3544, for future tracking.

martinflorian-da:
```suggestion
      // her validator app will crash on startup to prevent operating with insufficient trust.
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
```diff
@@ -63,30 +69,56 @@ class ValidatorSequencerConnectionIntegrationTest
...
 
       val sv1Url = getPublicSequencerUrl(sv1Backend)
       val sv2InitialUrl = getPublicSequencerUrl(sv2Backend)
-      val initialExpectedUrls = Set(sv1Url, sv2InitialUrl)
+      val sv3Url = getPublicSequencerUrl(sv3Backend)
+      val initialExpectedUrls = Set(sv1Url, sv2InitialUrl, sv3Url)
 
       withClue("Validator should connect to the filtered list of sequencers from the config") {
-        eventually(60.seconds, 1.second) {
+        eventually(120.seconds, 1.second) {
```
martinflorian-da, line 99: Do you need this change even with the wait above? (not opposed to adding it for safety)

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
```diff
@@ -63,30 +69,56 @@ class ValidatorSequencerConnectionIntegrationTest
...
       val newSv2Url = "http://" + newSv2Address
 
       setSequencerUrl(sv2Backend, newSv2Url).futureValue
 
-      val updatedExpectedUrls = Set(sv1Url, newSv2Address)
+      val updatedExpectedUrls = Set(sv1Url, newSv2Address, sv3Url)
 
       withClue("Validator should eventually see the updated sequencer URL from the ledger") {
-        eventually(60.seconds, 1.second) {
+        eventually(120.seconds, 1.second) {
```
martinflorian-da, line 121: Do you need this change even with the wait above? (not opposed to adding it for safety)

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/config/ValidatorAppConfig.scala
```diff
@@ -114,6 +108,18 @@ case class ValidatorDecentralizedSynchronizerConfig(
...
 }
 
+case class ValidatorTrustedSynchronizerConfig(
+    /** static list of trusted sequencer names to connect to.
+      * sequencerNames is mutually exclusive with `url`.
+      */
+    sequencerNames: NonEmptyList[String],
+
+    /** parameter to specify the BFT threshold for the domain connections.
+      * If not specified, length(sequencerNames) / 3 +1 will be used.
```
martinflorian-da: missing an `f` somewhere?

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
```diff
@@ -153,11 +154,19 @@ class DomainConnector(
...
+                        Thresholds.sequencerConnectionsSizeThreshold(nonEmptyConnections.size)
+                    }
+
                   SequencerConnections.tryMany(
                     nonEmptyConnections.forgetNE,
-                    Thresholds.sequencerConnectionsSizeThreshold(nonEmptyConnections.size),
+                    threshold,
                     submissionRequestAmplification = SubmissionRequestAmplification(
-                      Thresholds.sequencerSubmissionRequestAmplification(nonEmptyConnections.size),
+                      Thresholds
```
martinflorian-da: I still think your original solution (to use the custom threshold if it exists) was more consistent and likely also more long-term correct.

`Thresholds.sequencerSubmissionRequestAmplification` essentially infers `f` from the size of the set. It seems much more reasonable to me to infer it from the custom threshold.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
pasindutennage-da: Fixed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
pasindutennage-da: Had 1/10 cases, where 60 was not enough

Since, it is a max wait (not a min wait), it doesn't affect the CI time, IMO. So tempted to keep it

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
pasindutennage-da: Same reason as above - 1/10 times the test needs more time

Also, not a min wait, hence does not really affect CI time

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/config/ValidatorAppConfig.scala
pasindutennage-da: Fixed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorSequencerConnectionIntegrationTest.scala
martinflorian-da: perfectly good call; these timeouts also have a slight documentation purpose (how long does this usually take?); but more than happy to err on the side of reducing chance of flakes

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
pasindutennage-da: Resolved

PR #3522, opened 2026-01-12
CIP-82 (Scala): Add MergeUnclaimedDevelopmentFundCouponsTrigger

jose-velasco-ieu:
## Summary

- Add a new trigger `MergeUnclaimedDevelopmentFundCouponsTrigger` to automatically merge `UnclaimedDevelopmentFundCoupon` contracts once the total count of coupons exceeds a configured threshold * 2.
- Add a new config field `unclaimedDevelopmentFundCouponsThreshold` to `SvAppBackendConfig` (default = 10).
- Add a new function `listUnclaimedDevelopmentFundCoupons` to `SvDsoStore`.
- Add a new intregration test DevelopmentFundCouponIntegrationTest which test (for the moment) the merging of `UnclaimedDevelopmentFundCoupon`s. (Note: this test will grow as additional functionalities are introduced in upcoming PRs)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
```diff
@@ -0,0 +1,107 @@
...
+  protected def retrieveTasks()(implicit
+      tc: TraceContext
+  ): Future[Seq[MergeUnclaimedDevelopmentFundCouponsTask]] = {
+    val threshold = svConfig.numUnclaimedDevelopmentFundCouponsThreshold
+    val limit = PageLimit.tryCreate(threshold * 2)
+    store.listUnclaimedDevelopmentFundCoupons(limit).map { unclaimedDevelopmentFundCoupons =>
+      if (unclaimedDevelopmentFundCoupons.length > threshold) {
+        Seq(MergeUnclaimedDevelopmentFundCouponsTask(unclaimedDevelopmentFundCoupons))
+      } else {
+        Seq()
```
meiersi-da: I only just now realized that this strategy of merging all coupons into a single coupon does not interact well with external parties preparing transaction and taking some time until they sign the prepared transaction. It can easily be the case that the prepared transaction then references stale contract-ids and fails.

wdyt about using the following strategy:
- when there are more than `threshold` coupons, then merge the smallest `threshold` amount of them down to a single coupon.

This ensures that the fund manager can always use the largest coupon to fund the an allocation w/o contending with the merge automation.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
```diff
@@ -0,0 +1,107 @@
...
+) extends PollingParallelTaskExecutionTrigger[MergeUnclaimedDevelopmentFundCouponsTask]
+    with SvTaskBasedTrigger[MergeUnclaimedDevelopmentFundCouponsTask] {
+
+  private val store = svTaskContext.dsoStore
+
+  protected def retrieveTasks()(implicit
+      tc: TraceContext
+  ): Future[Seq[MergeUnclaimedDevelopmentFundCouponsTask]] = {
+    val threshold = svConfig.numUnclaimedDevelopmentFundCouponsThreshold
+    val limit = PageLimit.tryCreate(threshold * 2)
```
meiersi-da: Why `* 2`, doesn't `+ 1` suffice?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
```diff
@@ -0,0 +1,60 @@
...
+        )(config)
+      )
+      .addConfigTransform((_, config) =>
+        ConfigTransforms.updateInitialTickDuration(NonNegativeFiniteDuration.ofMillis(500))(config)
+      )
+
+  "UnclaimedDevelopmentFundCoupons are merged" in { implicit env =>
+    actAndCheck(
+      "Advance 10 rounds", {
+        Range(0, 10).foreach(_ => advanceRoundsByOneTickViaAutomation())
```
meiersi-da: I suspect this takes quite some time. wdyt about saving some test time by running with `threshold = 3` ?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
```diff
@@ -0,0 +1,60 @@
...
+import org.lfdecentralizedtrust.splice.config.ConfigTransforms.{
+  ConfigurableApp,
+  updateAutomationConfig,
+}
+import org.lfdecentralizedtrust.splice.integration.EnvironmentDefinition
+import org.lfdecentralizedtrust.splice.sv.automation.delegatebased.AdvanceOpenMiningRoundTrigger
+import org.lfdecentralizedtrust.splice.util.{TriggerTestUtil, WalletTestUtil}
+
+@org.lfdecentralizedtrust.splice.util.scalatesttags.SpliceDsoGovernance_0_1_21
+@org.lfdecentralizedtrust.splice.util.scalatesttags.SpliceAmulet_0_1_15
```
meiersi-da: I don't think that one is required, as dso-governance 0.1.21 implies amulet 0.1.15

meiersi-da: fine to keep though. your call.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -363,6 +363,8 @@ case class SvAppBackendConfig(
     // If true, we check that topology on mediator and sequencer is the same after
     // a migration. This can be a useful assertion but is very slow so should not be enabled on clusters with large topology state.
     validateTopologyAfterMigration: Boolean = false,
+    // The threshold above which unclaimed development fund coupons will be merged.
+    numUnclaimedDevelopmentFundCouponsThreshold: Int = 10,
```
meiersi-da:
```suggestion
    unclaimedDevelopmentFundCouponsThreshold: Int = 10,
```

nit: a threshold is sort of always a number or not?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
jose-velasco-ieu: It makes sense.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
jose-velasco-ieu: You are right.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
jose-velasco-ieu: Are you referring to the case where the fund manager is an external party?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceUtil.scala
```diff
@@ -225,7 +225,7 @@ object SpliceUtil {
     Some(damlDecimal(2.85)).toJava,
 
     // developmentFundPercentage
-    Optional.empty(),
+    Some(damlDecimal(0.05)).toJava,
```
jose-velasco-ieu: @meiersi-da 
I added this to be able to generate some `UnclaimedDevelopmentFundCoupon` needed for my integration test `DevelopmentFundCouponIntegrationTest`; however, this is causing other tests to fail. For example:

- `SvTimeBasedRewardCouponIntegrationTest`  ---> "144590.6514460000 was not greater than or equal to 152007.0015220700 (WalletTestUtil.scala:85)"
- `ScanWithGradualStartsTimeBasedIntegrationTest` ---> "Message: 296079.2686453800 was not greater than or equal to 311124.0030441400"

Any suggestions on how to approach this?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
jose-velasco-ieu: done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
jose-velasco-ieu: done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
jose-velasco-ieu: removed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
jose-velasco-ieu: @meiersi-da 
Just to make sure I understood your suggestion correctly, let me check it with a concrete example.

Assume:
- threshold = 3
- We currently have the following unclaimed coupons (amounts):
[5, 8, 10, 20, 50, 100]

With the proposed strategy, my understanding is:
- Since we have more than threshold coupons, we do not merge all of them.
- Instead, we take the threshold smallest coupons (5, 8, 10) and merge only those into a single new coupon (23).
- The larger coupons (20, 50, 100) remain untouched, so external parties can safely keep using (for example) the largest one without risking stale contract-ids.

After the merge, we would end up with:
[23, 20, 50, 100]

Is this the intended behavior you had in mind?

meiersi-da: > Are you referring to the case where the fund manager is an external party?

Yes. External parties are the new default that we want to make work. Eventually we might even phase out internal parties, but for now they'll likely remain used in niche cases.

The main constraint to design for is that external parties have much higher delays between preparing and submitting a transaction. The token standard requires asset registries to support up to 24h.

> Thanks for the concrete example.

I was going for the slightly simpler approach (but didn't describe it well), which results in exactly two coupons after the merge, and always merges at least `threshold` many coupons in one go. So in your example, we'd merge `[5, 8, 10, 20, 50]` into `[93]` and we are left with `[93, 100]`. Over-time this results in typically half of the funds being available as the largest coupon.

If we want more funds to be available in stable form, then we can keep more remaining fund coupons. A simple option would be to
1. merge when there are `>= 2 * threshold` coupons
2. merge the `threshold` smallest ones down to a single coupon
Thus after a merge there are always `threshold + 1` many coupons available

This is what you did in your example. It seems that this option is preferable from a UX point of view, as it makes contention unlikely even when spending up to 90% of the funds.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceUtil.scala
meiersi-da: don't change the default. Instead make it configurable `developmentFundPercentage : Option[Decimal] = None` and only configure the non-default value in your test. 

See how a similar construction is used to run some tests with `zeroTransferFees` using this config value (https://github.com/hyperledger-labs/splice/blob/c04209f0f3d5150c5c5c3730681a5cf38dcfe67e/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala#L95) and a corresponding config transform (https://github.com/hyperledger-labs/splice/blob/c04209f0f3d5150c5c5c3730681a5cf38dcfe67e/apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/ConfigTransforms.scala#L299-L300)

jose-velasco-ieu: Thank you! Very helpful.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
jose-velasco-ieu: I’m wondering whether the default value of 10 for `unclaimedDevelopmentFundCouponsThreshold` still makes sense with the new strategy.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceUtil.scala
jose-velasco-ieu: done

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
jose-velasco-ieu: done

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
```diff
@@ -0,0 +1,111 @@
...
+) extends PollingParallelTaskExecutionTrigger[MergeUnclaimedDevelopmentFundCouponsTask]
+    with SvTaskBasedTrigger[MergeUnclaimedDevelopmentFundCouponsTask] {
+
+  private val store = svTaskContext.dsoStore
+  private val threshold = svConfig.unclaimedDevelopmentFundCouponsThreshold
+
+  protected def retrieveTasks()(implicit
+      tc: TraceContext
+  ): Future[Seq[MergeUnclaimedDevelopmentFundCouponsTask]] = {
+    val limit = PageLimit.tryCreate(2 * threshold + 1)
```
meiersi-da:
```suggestion
    val limit = PageLimit.tryCreate(2 * threshold)
```
suffices given you are checing `>=`

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
```diff
@@ -0,0 +1,111 @@
...
+            s"Merged unclaimed development fund coupons into contract ${cid.exerciseResult.result.unclaimedDevelopmentFundCouponCid.contractId}"
+          )
+        })
+        .getOrElse(TaskSuccess(s"Not enough unclaimed development fund coupons to merge"))
+    }
+  }
+}
+
+case class MergeUnclaimedDevelopmentFundCouponsTask(
+    contracts: Seq[
```
meiersi-da: consider splitting into the contracts you merge and the ones kept to ease debugging in case the merge contends in an unexpected way

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  protected def retrieveTasks()(implicit
+      tc: TraceContext
+  ): Future[Seq[MergeUnclaimedDevelopmentFundCouponsTask]] = {
+    val limit = PageLimit.tryCreate(2 * threshold + 1)
+    store.listUnclaimedDevelopmentFundCoupons(limit).map { unclaimedDevelopmentFundCoupons =>
+      if (unclaimedDevelopmentFundCoupons.length >= 2 * threshold) {
+        Seq(
+          MergeUnclaimedDevelopmentFundCouponsTask(
+            unclaimedDevelopmentFundCoupons.sortBy(_.payload.amount)
```
meiersi-da: The test does not test that the smallest coupons are merged first, as all coupons are of the same size. At least add a comment here stating why we sort.

meiersi-da: Leaving it out of the test is OK AFAIS. Code should be fine 🤞

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
```diff
@@ -0,0 +1,111 @@
...
+      } else {
+        Seq()
+      }
+    }
+  }
+
+  protected def isStaleTask(
+      unclaimedDevelopmentFundCouponsTask: MergeUnclaimedDevelopmentFundCouponsTask
+  )(implicit tc: TraceContext): Future[Boolean] = store.multiDomainAcsStore.containsArchived(
+    unclaimedDevelopmentFundCouponsTask.contracts.map(_.contractId)
```
meiersi-da: Given that you only merge the first `threshold` ones, I wouldn't consider the trigger stale when any of the non-merged ones change.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
meiersi-da: actually, perhaps better 
```suggestion
    contractsToMerge: Seq[
```
only store the contracts that should be merged

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
meiersi-da: I think it does. It ensures that merges happen not too often, and it makes more of the funds holdings available in the non-merged coupons.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
jose-velasco-ieu: I can enhance the current test by advancing a few rounds more after the first merge to force a second merge where I check that the smallest coupons are merge.

meiersi-da: Or just create a few coupons directly using a variant of https://github.com/hyperledger-labs/splice/blob/6b82685b780b431d5277d4d6e2b77ffa58349601/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/WalletTestUtil.scala#L896

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
jose-velasco-ieu: done

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
jose-velasco-ieu: Done — I did it by advancing a few rounds.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
jose-velasco-ieu: With the new strategy where we only return the contracts to merge 

```
protected def retrieveTasks()(implicit
      tc: TraceContext
  ): Future[Seq[MergeUnclaimedDevelopmentFundCouponsTask]] = {
    val threshold = svConfig.unclaimedDevelopmentFundCouponsThreshold
    val limit = PageLimit.tryCreate(2 * threshold)
    store.listUnclaimedDevelopmentFundCoupons(limit).map { unclaimedDevelopmentFundCoupons =>
      if (unclaimedDevelopmentFundCoupons.length >= 2 * threshold) {
        Seq(
          MergeUnclaimedDevelopmentFundCouponsTask(
            // Merge the `threshold` smallest coupons (by amount) to keep larger coupons stable and
            // reduce contention with externally prepared transactions referencing contract-ids.
            unclaimedDevelopmentFundCoupons.sortBy(_.payload.amount).take(threshold)
          )
        )
      } else {
        Seq()
      }
    }
  }
```

I think that the `isStaleTask` function is correct. wdyt?
```
  protected def isStaleTask(
      unclaimedDevelopmentFundCouponsTask: MergeUnclaimedDevelopmentFundCouponsTask
  )(implicit tc: TraceContext): Future[Boolean] = store.multiDomainAcsStore.containsArchived(
    unclaimedDevelopmentFundCouponsTask.contractsToMerge.map(_.contractId)
  )
```

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
meiersi-da: OK. I usually try to use advancing rounds sparingly, as they cost quite a bit of test time.

meiersi-da: Fine for you to use it here.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/MergeUnclaimedDevelopmentFundCouponsTrigger.scala
meiersi-da: yes.

PR #3523, opened 2026-01-12
Added more conditions for GC-Reaper-Job pod selection

pasindutennage-da:
Extending support to delete pods in the mult-validator name space, under "evicted" status.

Fixes https://github.com/DACH-NY/canton-network-internal/issues/1418

### DIFF
--- a/cluster/pulumi/infra/src/maintenance.ts
+++ b/cluster/pulumi/infra/src/maintenance.ts
@@ -36,7 +36,7 @@ const deleteBadPodsCommand = [
               (.status.reason == "Evicted") or
               (.status.containerStatuses[]?.state.terminated? | .reason == "Error" and .exitCode == 137) or
               (.status.initContainerStatuses[]?.state.waiting?.reason == "ContainerStatusUnknown")
-            ) | .metadata.name' | sort -u | xargs
+            ) | .metadata.name' | sort -u
         );
         if [ -z "$BAD_PODS" ]; then
             echo "No bad pods found in $NAMESPACE. Skipping.";


cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -31,11 +31,12 @@ const deleteBadPodsCommand = [
...
-                  .metadata.name) //
-              (select(.status.containerStatuses[]?.state.terminated? |
-                  .reason == "Error" and .exitCode == 137) |
-                  .metadata.name)'
+              select(
+              (.status.phase == "Unknown") or
+              (.status.reason == "Evicted") or
+              (.status.containerStatuses[]?.state.terminated? | .reason == "Error" and .exitCode == 137) or
+              (.status.initContainerStatuses[]? | (.state.terminated?.reason == "Error") or (.state.waiting?.reason == "ContainerStatusUnknown"))
+            ) | .metadata.name' | sort -u
```
nicu-da: Isn't this too lax? All the or statements will just delete all the pods with unknown or evicted or error and exit code 137 no matter if the reason is error or unknown

pasindutennage-da: Refined the rule to check for reason
For evicted, no reason is checked

cluster/pulumi/infra/src/maintenance.ts
```diff
@@ -31,11 +31,12 @@ const deleteBadPodsCommand = [
...
-                  .metadata.name) //
-              (select(.status.containerStatuses[]?.state.terminated? |
-                  .reason == "Error" and .exitCode == 137) |
-                  .metadata.name)'
+              select(
+              (.status.phase == "Unknown" and .status.reason == "ContainerStatusUnknown") or
+              (.status.reason == "Evicted") or
+              (.status.containerStatuses[]?.state.terminated? | .reason == "Error" and .exitCode == 137) or
+              (.status.initContainerStatuses[]?.state.waiting?.reason == "ContainerStatusUnknown")
+            ) | .metadata.name' | sort -u | xargs
```
martinflorian-da:
```suggestion
            ) | .metadata.name' | sort -u
```

Looks like a mistake... or at least you should use something more explicit if you want to just join all lines into one?

martinflorian-da (approved):
Looks reasonable, thanks! How are you testing / sanity checking this?

cluster/pulumi/infra/src/maintenance.ts
pasindutennage-da: Fixed. thank you

pasindutennage-da:
> Looks reasonable, thanks! How are you testing / sanity checking this?

I manually log into CILR, check the logs of the last run cronjob, and see if it has deleted. Then verify using kubectl list pods

Issue #3293, opened 2025-11-29
Create a reference test suite for Pulumi that asserts resource definition


mblaze-da:
With the recently introduced unit tests based on `jest` we still need a good reference for how to write assertions on Pulumi resource definitions.

PR #3530, opened 2026-01-13
implement two reference tests with Pulumi resource assertions

mblaze-da:
fixes https://github.com/hyperledger-labs/splice/issues/3293

I'm going to be explaining this on a CWT today (13.01.2026).

### DIFF
--- a/cluster/pulumi/canton-network/src/clusterVersion.test.ts
+++ b/cluster/pulumi/canton-network/src/clusterVersion.test.ts
@@ -33,6 +33,7 @@ test('version service should return the correct version', async () => {
 
   const versionService = installClusterVersion();
 
+  // check the version service spec for the correct version endpoint configuration
   const spec =
     'spec' in versionService
       ? (versionService.spec as pulumi.Output<{
--- a/cluster/pulumi/gha/src/runners.test.ts
+++ b/cluster/pulumi/gha/src/runners.test.ts
@@ -10,15 +10,15 @@ import { installRunnerScaleSets } from './runners';
 jest.mock('./config', () => ({
   __esModule: true,
   ghaConfig: {
-    githubRepo: 'dummy',
-    runnerVersion: 'dummy',
-    runnerHookVersion: 'dummy',
+    githubRepo: 'https://dummy-gh-repo.com',
+    runnerVersion: '1.2',
+    runnerHookVersion: '1.1',
   },
 }));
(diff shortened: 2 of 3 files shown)


mblaze-da:
I'm tagging you @nicu-da because I remember you asking about these when I first introduced unit tests in TS.

cluster/pulumi/gha/src/runners.test.ts
```diff
@@ -0,0 +1,90 @@
...
+import * as pulumi from '@pulumi/pulumi';
+import { expect, jest, test } from '@jest/globals';
+import { collectResources } from '@lfdecentralizedtrust/splice-pulumi-common/src/test';
+
+import { installRunnerScaleSets } from './runners';
+
+jest.mock('./config', () => ({
+  __esModule: true,
+  ghaConfig: {
+    githubRepo: 'dummy',
```
martinflorian-da: nit: consider using different values so we're testing that we're setting the right thing in the right place

cluster/pulumi/gha/src/runners.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  const mockController = {
+    __pulumiResource: true,
+    name: 'mock-controller',
+  } as unknown as k8s.helm.v3.Release;
+
+  const [, resources] = await collectResources(() => {
+    installRunnerScaleSets(mockController);
+  });
+
+  for (const resource of resources) {
```
martinflorian-da, line 65: Consider using comments to describe what exactly you're asserting here; it's not that easy to read.

(I'm assuming that jest doesn't have something like the `clue`s we use for scala tests?)

cluster/pulumi/canton-network/src/clusterVersion.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      switch (args.token) {
+        default:
+          return args.inputs;
+      }
+    },
+  });
+
+  const versionService = installClusterVersion();
+
+  const spec =
```
martinflorian-da, line 37: This might also benefit from some comments; why did we bother with a spec here (in case others are considering a similar testing strategy)

martinflorian-da (approved):
Nice stuff, thanks a lot!

cluster/pulumi/gha/src/runners.test.ts
mblaze-da: AFAIK, it doesn't

PR #3533, opened 2026-01-13
Align ACS semantics with those of the ledger API

isegall-da:
Part of https://github.com/hyperledger-labs/splice/issues/3527

No change required in `/vX/updates` and the queries that support it, those were already doing `>`, which is correct.

Note that there's a similar inconsistency in backfilling logic. Fixing that got a bit more involved, so I'm (at least) moving it into a separate PR, where I'll make my mind if it's worth changing. Tracking in #3527 

Also note that the corresponding fix in scan_txlog exposed that unclaimed activity records are not handled there correctly, and fixing that also got more involved than I'd like to invest. I am going to argue that we should significantly narrow the scope of scan_txlog, which would make fixing it throw-away work, so for now I just disabled the failing check. It's not more broken than it has been so far, we just didn't notice because the `<` semantics caused us to never actually parse in scan_txlog the last update in every test.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedActivityRecordIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedActivityRecordIntegrationTest.scala
@@ -32,7 +32,7 @@ class UnclaimedActivityRecordIntegrationTest
     with TriggerTestUtil
     with WalletTestUtil {
 
-  // TODO(#3527): scan_txlog does not handle unclaimed activity records correctly, so this is currently disabled.
+  // TODO(#3549): scan_txlog does not handle unclaimed activity records correctly, so this is currently disabled.
   //  We should decide whether we want to fix scan_txlog for UnclaimedActivityRecords, or narrow its scope enough that we won't care
   //  (i.e. don't try to be complete, don't use it to assert on scan's aggregates, etc)
   override protected def runUpdateHistorySanityCheck: Boolean = false


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -91,8 +92,8 @@ class AcsSnapshotStore(
         sql"""
           where #$tableAlias.history_id = $historyId
             and #$tableAlias.migration_id = $migrationId
-            and #$tableAlias.record_time >= $from -- this will be >= MinValue for the first snapshot, which includes ACS imports
-            and #$tableAlias.record_time < $until
+            and #$tableAlias.record_time #$gtFrom $from -- this will be >= MinValue for the first snapshot, which includes ACS imports, otherwise >
+            and #$tableAlias.record_time <= $until
```
isegall-da, line 96: Is it really as simple as changing only this? 
@ray-roestenburg-da @rautenrieth-da Please review also for completeness. Did I miss anything (other than the getUpdatesBefore, which is used only for backfilling, and moved to a separate PR, as explained in the comment on this PR) ?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedActivityRecordIntegrationTest.scala
```diff
@@ -32,6 +32,11 @@ class UnclaimedActivityRecordIntegrationTest
     with TriggerTestUtil
     with WalletTestUtil {
 
+  // TODO(#3527): scan_txlog does not handle unclaimed activity records correctly, so this is currently disabled.
```
rautenrieth-da: Can we add a specific issue for this? #3527 is for aligning ACS semantics, and handling UnclaimedActivityRecords sounds unrelated.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -91,8 +92,8 @@ class AcsSnapshotStore(
         sql"""
           where #$tableAlias.history_id = $historyId
             and #$tableAlias.migration_id = $migrationId
-            and #$tableAlias.record_time >= $from -- this will be >= MinValue for the first snapshot, which includes ACS imports
-            and #$tableAlias.record_time < $until
+            and #$tableAlias.record_time #$gtFrom $from -- this will be >= MinValue for the first snapshot, which includes ACS imports, otherwise >
```
rautenrieth-da, line 95: If `lastSnapshot` is None, this whole line will simplify to `true`, as there are no updates with a record time before the ACS import ones. Perhaps it would be more readable to only include the lower bound on record time if we actually need it? Leaving it as is works as well.

Sometimes including a column in a filter affects the query plan even if the filter is technically useless, so not all useless expressions can be blindly removed, but that should not be an issue here.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
rautenrieth-da: The code looks fine to me.

I'm surprised we didn't have to change any unit test. It would be nice to have a test in `AcsSnapshotStoreTest` where we have updates at the exact boundaries of snapshots.

I'm also surprised we didn't have to change any comment in the scan OpenAPI definition, but you could argue that it's implicitly clear that "the ACS at record time T" includes contracts with a record time of exactly T so there's no need to document it.

How about a release notes entry? We expect this change to not be observable in production, but maybe some users will need to adjust their testing code, like we did in `scripts/scan-txlog/scan_txlog.py`

Added a comment regarding getUpdatesBefore at https://github.com/hyperledger-labs/splice/issues/3527#issuecomment-3748629127

OriolMunoz-da: lgtm too

> I'm surprised we didn't have to change any unit test. It would be nice to have a test in AcsSnapshotStoreTest where we have updates at the exact boundaries of snapshots.

+1 for adding one and making sure it breaks with the old code
(and also agree with everything else Robert suggests)

ray-roestenburg-da: +1

isegall-da: Yes, definitely planning on adding a unit test. Was just too excited with finally making CI green that I pushed it to "non draft" prematurely :)

isegall-da: Re release notes, I commented on my first PR explicitly (that I then killed). I'm actually leaning toward not mentioning this. We never explicitly said anywhere that the semantics is <, so I'm considering this almost a bugfix for a bug that was just never actually triggered so far.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnclaimedActivityRecordIntegrationTest.scala
isegall-da: Fair, I'll move it out.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/AcsSnapshotStoreTest.scala
```diff
@@ -204,14 +204,14 @@ class AcsSnapshotStoreTest
...
-            timestamp1.minusSeconds(2L),
+            timestamp1,
           )
           _ <- store.insertNewSnapshot(None, DefaultMigrationId, timestamp1)
           // t2
           _ <- ingestCreate(
             updateHistory,
             omr2,
-            timestamp2.minusSeconds(1L),
+            timestamp2,
```
isegall-da, line 214: @rautenrieth-da @OriolMunoz-da @ray-roestenburg-da 
FYI re unit test: this change fails on main, passes now 👏

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
isegall-da: I can't judge myself the risk, and with no scale testing in place, not sure how to evaluate it, so will need to rely on your judgement here. How convinced are you with "should not be an issue here" ?

isegall-da: @rautenrieth-da 

I'm gonna take this part: "Leaving it as is works as well", and just merge as-is for now. If you feel like we should change this to slightly improve readability here, happy to do that in a separate PR.

PR #3541, opened 2026-01-14
Add UpdateHistory performance test

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/3513

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/StoreIngestionPerformanceTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/StoreIngestionPerformanceTest.scala
@@ -51,9 +51,18 @@ abstract class StoreIngestionPerformanceTest(
       .withNewTraceContext(this.getClass.getName) { implicit tc =>
         for {
           _ <- store.ingestionSink.initialize()
+          _ <- sanityCheckTables(storage) { count =>
+            Option.when(count != 0)(
+              s"Expected table to be empty before ingestion, but found $count rows."
+            )
+          }
           txs = loadTxsFromDump()
           _ <- ingestAll(store, txs)
-          _ <- sanityCheckTables(storage)
+          _ <- sanityCheckTables(storage) { count =>
+            Option.when(count == 0)(
+              s"Expected table to be non-empty after ingestion, but no rows were inserted."
+            )
+          }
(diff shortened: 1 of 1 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/StoreIngestionPerformanceTest.scala
```diff
@@ -141,6 +142,34 @@ abstract class StoreIngestionPerformanceTest(
       })
   }
 
+  protected val tablesToSanityCheck: Seq[String]
+
+  /** Make sure that there's at least one row, as otherwise things might not be wired up properly.
+    */
+  private def sanityCheckTables(storage: DbStorage)(implicit tc: TraceContext) = {
```
OriolMunoz-da: TODO: add same check but that number of rows is 0 for test startup, as otherwise the test also does nothing

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/StoreIngestionPerformanceTest.scala
```diff
@@ -141,6 +142,34 @@ abstract class StoreIngestionPerformanceTest(
...
+        .value
+        .failOnShutdownToAbortException("Should not be shutting down.")
+        .flatMap {
+          case Some(count) if count > 0 =>
+            logger.info(s"Table $tableName has $count rows, sanity check passed.")
+            Future.unit
+          case err =>
+            Future.failed(
+              new IllegalStateException(
+                s"Sanity check failed for table $tableName. There are $err rows."
```
ray-roestenburg-da: nitpick this will possibly print "There are None rows." (unlikely looking at the query) or "There are Some(0) rows." Could just do `.head` in query  or just print 'There are no rows'

ray-roestenburg-da (approved):

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/StoreIngestionPerformanceTest.scala
rautenrieth-da: That sounds like a useful check, are you going to add it in this PR?

PR #3552, opened 2026-01-14
support querying next snapshot after a timestamp

isegall-da:
Will be useful for streaming ACS snapshots to S3

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTimeBasedIntegrationTest.scala
@@ -460,7 +460,7 @@ class ScanTimeBasedIntegrationTest
         migrationId,
       )
       snapshot1 should not be None
-      snapshot1.value.toInstant shouldBe > (startTime.toInstant)
+      snapshot1.value.toInstant shouldBe >(startTime.toInstant)
       snapshot1
     }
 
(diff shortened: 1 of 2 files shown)


apps/scan/src/main/openapi/scan.yaml
```diff
@@ -445,6 +445,45 @@ paths:
...
+            description: |
+              The endpoint will return the record time of the first snapshot after this parameter.
+        - name: "migration_id"
+          in: "query"
+          required: true
+          schema:
+            type: integer
+            format: int64
+            description: |
+              The endpoint will return the record time of the first snapshot for this migration id or larger.
```
rautenrieth-da, line 471: This is different from the `/v0/state/acs/snapshot-timestamp` endpoint, which only looks at snapshots from exactly the given migration id.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -68,6 +68,34 @@ class AcsSnapshotStore(
...
+    val sameMig = select ++ sql""" from acs_snapshot
+            where snapshot_record_time > $after
+              and migration_id = $migrationId
+              and history_id = $historyId """ ++ orderLimit
+    val largerMig = select ++ sql""" from acs_snapshot
+            where migration_id > $migrationId
+              and history_id = $historyId """ ++ orderLimit
+
+    val query =
+      sql"select * from ((" ++ sameMig ++ sql") union all (" ++ largerMig ++ sql")) all_queries order by snapshot_record_time asc limit 1"
```
rautenrieth-da, line 88: A bit complicated for a table with only thousands of rows, but should make optimal use of the available primary key index.

isegall-da:
> why are we adding a new HTTP endpoint?

Mostly so that I can call it in the integration test TBH. 
But I'd say "why not"? Do we have any reason to believe that people want to query "before" more than they want "after"?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
isegall-da: yup, that was my thinking exactly

apps/scan/src/main/openapi/scan.yaml
isegall-da: TBH, I think that the semantics in the existing one is less useful. People don't really care about migration IDs, it's almost an implementation detail. They just want the previous/next snapshot from a specific timestamp.

isegall-da:
> > why are we adding a new HTTP endpoint?
> 
> Mostly so that I can call it in the integration test TBH. But I'd say "why not"? Do we have any reason to believe that people want to query "before" more than they want "after"?

@rautenrieth-da @ray-roestenburg-da I'm going to merge this as-is. Please shout if you think that having this endpoint is a mistake, we have until the next release to remove it easily if so.

PR #3570, opened 2026-01-15
stream ACS snapshot dumps to S3

isegall-da:
Currently dumps *all* snapshots. Computing which snapshots to actually dump, and which to skip, is tracked as a followup step in the [tracking issue](https://github.com/hyperledger-labs/splice/issues/3429).

Fixes https://github.com/hyperledger-labs/splice/issues/3617

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorageTest.scala
+++ b/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorageTest.scala
@@ -117,10 +117,12 @@ class AcsSnapshotBulkStorageTest
 
         clue("Add another snapshot to the store, it is also dumped") {
           store.addSnapshot(CantonTimestamp.tryFromInstant(Instant.ofEpochSecond(20)))
-          probe.expectNext(2.minutes).value shouldBe (0, CantonTimestamp.tryFromInstant(
+          val next = probe.expectNext(2.minutes)
+          next.value shouldBe (0, CantonTimestamp.tryFromInstant(
             Instant.ofEpochSecond(20)
           ))
           probe.expectNoMessage(10.seconds)
+          next.killSwitch.shutdown()
         }
 
         succeed


isegall-da:
ping @ray-roestenburg-da @rautenrieth-da can I have a review pls?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorageTest.scala
```diff
@@ -0,0 +1,248 @@
...
+      withS3Mock {
+        val store = new MockAcsSnapshotStore()
+        val s3BucketConnection = getS3BucketConnectionWithInjectedErrors(loggerFactory)
+        val probe = new AcsSnapshotBulkStorage(
+          bulkStorageTestConfig,
+          store.store,
+          s3BucketConnection,
+          loggerFactory,
+        ).getSource
+          .runWith(TestSink.probe[WithKillSwitch[(Long, CantonTimestamp)]])
```
ray-roestenburg-da, line 108: nice

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -45,91 +26,95 @@ class AcsSnapshotBulkStorage(
...
+                logger.debug("No new snapshot available, sleeping...")
+                Future.successful(Some(((lastMigrationId, lastTimestamp), None)))
+              }
+          }
+      }
+      .collect { case Some((migrationId, timestamp)) => (migrationId, timestamp) }
   }
 
-  def dumpAcsSnapshot(migrationId: Long, timestamp: CantonTimestamp): Future[Unit] = {
+  /** *
```
ray-roestenburg-da:
```suggestion
  /**
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -45,91 +26,95 @@ class AcsSnapshotBulkStorage(
...
+  private def mksrc(
+      start: (Long, CantonTimestamp)
+  ): Source[(Long, CantonTimestamp), (KillSwitch, Future[Done])] = start match {
+    case (startMigrationId, startAfterTimestamp) =>
+      logger.debug(
+        s"Starting ACS snapshot dump source, from migration $startMigrationId, timestamp $startAfterTimestamp"
+      )
+      val base =
+        getAcsSnapshotTimestampAfter(startMigrationId, startAfterTimestamp)
+          .via(SingleAcsSnapshotBulkStorage(config, acsSnapshotStore, s3Connection, loggerFactory))
```
ray-roestenburg-da: nice!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/SingleAcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,214 @@
...
+      s3Connection: S3BucketConnection,
+      loggerFactory: NamedLoggerFactory,
+  )(implicit
+      actorSystem: ActorSystem,
+      tc: TraceContext,
+      ec: ExecutionContext,
+  ): Flow[(Long, CantonTimestamp), (Long, CantonTimestamp), NotUsed] =
+    Flow[(Long, CantonTimestamp)].flatMapConcat {
+      case (migrationId: Long, timestamp: CantonTimestamp) =>
+        new SingleAcsSnapshotBulkStorage(
```
ray-roestenburg-da, line 183: style nitpick, you could do SingleAcsSnapshotBulkStorage.getSource and just pass the args to that method, you don't really need to create a new instance of SingleAcsSnapshotBulkStorage and then call it's one method

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/SingleAcsSnapshotBulkStorage.scala
```diff
@@ -0,0 +1,214 @@
...
+          migrationId,
+          timestamp,
+          config,
+          acsSnapshotStore,
+          s3Connection,
+          loggerFactory,
+        ).getSource
```
ray-roestenburg-da, line 190: You get the source (getSource) which return Source[String, ..], then you fold over it and just emit migration timestamp, this is not complete right? (maybe make a note of that) or am I missing something?

ray-roestenburg-da: Maybe I missed the TODOs in getSource. I think it would make sense for the write to s3 to be a flow that emits (migrationid, timestamp) that was written, but you're probably planning this for later?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorageTest.scala
```diff
@@ -0,0 +1,248 @@
...
+
+  "AcsSnapshotBulkStorage" should {
+    "successfully dump a single ACS snapshot" in {
+      withS3Mock {
+        val store = new MockAcsSnapshotStore().store
+        val timestamp = CantonTimestamp.now()
+        val s3BucketConnection = getS3BucketConnectionWithInjectedErrors(loggerFactory)
+        for {
+          _ <- SingleAcsSnapshotBulkStorage
+            .asSource(
```
ray-roestenburg-da, line 52: Looking at this test, maybe it's nice to separate the code into a source that gives you acs snapshots (from the store) and a flow to write the s3 stuff, and then connect them in the code. (you could also unit test just the flow for instance without having an actual source that gives you real acs snapshots from the db)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/SingleAcsSnapshotBulkStorage.scala
isegall-da: You mean to move the two lines below into the source itself?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/SingleAcsSnapshotBulkStorage.scala
isegall-da: But it has other private methods that access these arguments, so moving them to be arguments of `getSource` would require passing these around to other private methods as well. Not sure what you gain here...

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/SingleAcsSnapshotBulkStorage.scala
isegall-da: Guess that makes sense, will do

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/SingleAcsSnapshotBulkStorage.scala
ray-roestenburg-da: yeah I think that's more straightforward, and slightly more functional style, just methods calling methods and if needed passing along arguments. From perspective of the user of the function, the user just calls the function with the args, the user does not need to create an instance just to call one method, so from usage perspective it's simpler, just call the function on `object SingleAcsSnapshotBulkStorage`

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorageTest.scala
isegall-da: As things are right now, wrapping the write to s3 as a flow is 2 lines of code, so I don't see great value in adding a class for that. Having said that, there's this TODO where I do plan on doing basically that: https://github.com/hyperledger-labs/splice/blob/b1eb83a09d88ceb8673ab4e1f7eeca3fe0cbf9b6/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala#L94-L97

ray-roestenburg-da: you can add a couple of methods on an object (don't need a class) to create the separate sources and flows and then use those somewhere compose them source.via(flow). Not a huge deal.
I'm used to doing something like:
```
object Bulk {
  def acsFromScan(...): Source[.., ..]
  def acsToS3(...): Flow[..,..,..] 
}
```
And somewhere you compose the whole stream `acsFromScan().via(acsToS3)...` 
then you can test both separately, and use where you like. just FYI. in a test you can do `Source(Vector(acs1,...)).via(acsToS3)` .. etc

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/SingleAcsSnapshotBulkStorage.scala
isegall-da: Hmmm... still not sure I'm convinced :) 
This is hidden in the object's `apply`, so the user is unaware of the fact that an object is created behind the scenes. The user indeed just calls a method on the object, so that's the UX you propose. Isn't it?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorageTest.scala
isegall-da: ACK, will do as part of that TODO

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/SingleAcsSnapshotBulkStorage.scala
ray-roestenburg-da: yes I'm proposing `Object.method`, no apply, just a method on an object, it's very easy to find. It's not super important 😄

isegall-da: ACK, no strong feelings here. The only difference is that a user needs to type also `.method` after `Object` when using this, but if that seems more intuitive for you, I'm fine with that.

PR #3595, opened 2026-01-19
Use GetContract as fallback for token standard context endpoints on Scan

OriolMunoz-da:
As per https://github.com/DACH-NY/canton-network-internal/issues/2817#issuecomment-3760837095

Fetches from ledger if not found for:
- AmuletAllocation
- TransferInstruction
- LockedAmulet
(I didn't find any other applicable ones)

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ledger/api/LedgerClient.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ledger/api/LedgerClient.scala
@@ -201,7 +201,7 @@ private[environment] class LedgerClient(
   def getContract(contractId: ContractId[?], queryingParties: Seq[PartyId])(implicit
       tc: TraceContext
   ): Future[Option[CreatedEvent]] = {
-    for {
+    (for {
       stub <- withCredentialsAndTraceContext(contractServiceStub)
       contract <- stub.getContract(
         new lapi.contract_service.GetContractRequest(
@@ -209,7 +209,10 @@ private[environment] class LedgerClient(
           queryingParties.map(_.toProtoPrimitive),
         )
       )
-    } yield contract.createdEvent
+    } yield contract.createdEvent).recover {
+      case e: StatusRuntimeException if e.getStatus.getCode == io.grpc.Status.Code.NOT_FOUND =>
+        None
+    }
   }
 
   def updates(


canton/community/ledger-api/src/main/protobuf/com/daml/ledger/api/v2/contract_service.proto
OriolMunoz-da, line 1: copied from https://github.com/DACH-NY/canton/blob/release-line-3.4/community/ledger-api/src/main/protobuf/com/daml/ledger/api/v2/contract_service.proto

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
```diff
@@ -52,11 +52,11 @@ import scala.jdk.CollectionConverters.*
 
 trait MultiDomainAcsStore extends HasIngestionSink with AutoCloseable with NamedLogging {
   protected def storeName: String
-  protected def storeParty: String
+  def storeParty: PartyId
 
   protected implicit lazy val mc: MetricsContext = MetricsContext(
     "store_name" -> storeName,
-    "store_party" -> storeParty,
+    "store_party" -> storeParty.toString, // using .toString for historical reasons
```
OriolMunoz-da, line 59: ported from the underlying definition of storeParty

I could use `toProtoPrimitive`, but then we'd have two metric labels. lmk if that's fine

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/ContractFetcher.scala
```diff
@@ -0,0 +1,113 @@
...
+import com.digitalasset.canton.logging.NamedLoggerFactory
+import com.digitalasset.canton.time.Clock
+import com.digitalasset.canton.tracing.TraceContext
+import org.lfdecentralizedtrust.splice.environment.BaseLedgerConnection
+import org.lfdecentralizedtrust.splice.store.MultiDomainAcsStore.ContractCompanion
+import org.lfdecentralizedtrust.splice.util.Contract
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait ContractFetcher {
```
OriolMunoz-da: better name desired, I don't want to imply that this is supposed to use for all contracts

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
```diff
@@ -124,4 +129,36 @@ class UpdateIngestionService(
...
   startIngestion()
+
+  @SuppressWarnings(Array("org.wartremover.warts.Var"))
+  @volatile
+  private var waitForResumePromise = Promise.successful(())
+
+  /** Note that any in-flight events being processed when `pause` is called will still be processed.
+    * For test purposes.
+    */
+  def pause(): Future[Unit] = blocking {
```
OriolMunoz-da, line 141: necessary for tests

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardFetchFallbackIntegrationTest.scala
```diff
@@ -0,0 +1,82 @@
...
+              )
+            )
+          )
+        )(config)
+      )
+  }
+
+  "Token Standard Transfers should" should {
+
+    "TransferInstruction context can be fetched from Scan even if it's not yet ingested into the store" in {
```
OriolMunoz-da, line 34: this test wants the fallback to be enabled and the validity to be long enough, whereas [this assertion](https://github.com/hyperledger-labs/splice/blob/f0d4cd52e330522fa25592dc4538331d317e690d/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardTransferIntegrationTest.scala#L435) wants to see an archival immediately, which is in opposition to what this test wants, so I had to move this into a different file

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
moritzkiefer-da: can we slap the @ViibleForTesting stuff on it or does that not work here?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -115,7 +115,7 @@ final class DbMultiDomainAcsStore[TXE](
   import profile.api.jdbcActionExtensionMethods
 
   override lazy val storeName = acsStoreDescriptor.name
-  override lazy val storeParty = acsStoreDescriptor.party.toString
+  override lazy val storeParty = acsStoreDescriptor.party
```
moritzkiefer-da, line 118: is this just a noop or why is it ok to change this

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
moritzkiefer-da: I think toString is ok, we generally shorten these identifiers in metrics

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/ContractFetcher.scala
moritzkiefer-da: Maybe "ChoiceContextContractFetcher" to make the use clear (at least for now).

I'd also add some comments explaining the rop issues at the top and how this helps

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/ContractFetcher.scala
```diff
@@ -0,0 +1,111 @@
...
+        companion: C
+    )(id: ContractId[?])(implicit
+        companionClass: ContractCompanion[C, TCid, T],
+        traceContext: TraceContext,
+    ): Future[Option[Contract[TCid, T]]] =
+      OptionT(store.multiDomainAcsStore.lookupContractById(companion)(id))
+        .map(_.contract)
+        .orElse(
+          OptionT(fallbackLedgerClient.getContract(id, Seq(store.multiDomainAcsStore.storeParty)))
+            // `getContract` may return a contract that was archived (and thus missing from the store).
```
moritzkiefer-da:
```suggestion
            // `getContract` will return archived contracts until they have been pruned (and thus missing from the store).
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/ContractFetcher.scala
```diff
@@ -0,0 +1,111 @@
...
+              logger.debug(s"Falling back to ledger for contract $javaCreatedEvent")
+              companionClass
+                .fromCreatedEvent(companion)(javaCreatedEvent)
+            }
+        )
+        .value
+  }
+
+  case class StoreContractFetcherWithLedgerFallbackConfig(
+      enabled: Boolean = false,
```
moritzkiefer-da: How about we keep this on by default? My reasoning would be:

1. We do actually want other SVs to enable it for testing. Otherwise for example our devnet perf tests are not gonna change much.
2. We still have the dev->test->mainnet test cycle. So we should find most stuff before mainnet explodes hopefully.
3. Asking SVs to set a config flag if it goes badly is pretty reasonable (cmopared to having to ask 600 valdiators to set it).

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardFetchFallbackIntegrationTest.scala
```diff
@@ -0,0 +1,82 @@
...
+              ContractFetcher.StoreContractFetcherWithLedgerFallbackConfig(
+                enabled = true
+              )
+            )
+          )
+        )(config)
+      )
+  }
+
+  "Token Standard Transfers should" should {
```
moritzkiefer-da: how about also testing it for allocations given that you changed it there as well?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/ContractFetcher.scala
OriolMunoz-da: sounds good to me
just to double-check, you proposed disabled by default here: https://github.com/DACH-NY/canton-network-internal/issues/2817#issuecomment-3760837095
were you referring to something else?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: only used in https://github.com/hyperledger-labs/splice/pull/3595#discussion_r2708312705

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/ContractFetcher.scala
moritzkiefer-da: > just to double-check, you proposed disabled by default here: https://github.com/DACH-NY/canton-network-internal/issues/2817#issuecomment-3760837095

That was just my past self being an idiot.

ray-roestenburg-da (approved):
Thanks!

PR #3603, opened 2026-01-20
POC: Daml for SV staking mechanics

meiersi-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/daml/splice-amulet-test/daml/Splice/Scripts/UnitTests/Amulet/SvStaking.daml
+++ b/daml/splice-amulet-test/daml/Splice/Scripts/UnitTests/Amulet/SvStaking.daml
@@ -78,6 +78,7 @@ sv1StakingState : StakerState
 sv1StakingState = StakerState with
   config = StakerConfig with relockRatio = 0.8
   lifetimeRewards = 1e9
+  exemptedLifetimeRewards = 0.0
   mintingAllowance = 0.0
   unconstrainedStake = 800e6
   stakes = []
@@ -119,6 +120,7 @@ sv2StakingState : StakerState
 sv2StakingState = StakerState with
   config = StakerConfig with relockRatio = 0.4
   lifetimeRewards = 1e9
+  exemptedLifetimeRewards = 0.0
   mintingAllowance = 0.0
   unconstrainedStake = 400e6
   stakes = []
@@ -174,6 +176,7 @@ example3_stakingState = StakingState with
     , (staker1, StakerState with
         config = StakerConfig with relockRatio = 0.5
         lifetimeRewards = 500e6
+        exemptedLifetimeRewards = 0.0
         mintingAllowance = 0.0
         unconstrainedStake = 300e6
         stakes = []
@@ -181,6 +184,7 @@ example3_stakingState = StakingState with
     , (staker2, StakerState with
         config = StakerConfig with relockRatio = 0.2
         lifetimeRewards = 200e6
+        exemptedLifetimeRewards = 0.0
         mintingAllowance = 0.0
         unconstrainedStake = 100e6
         stakes = []
@@ -199,6 +203,7 @@ example3_expected_newStaker1StakingState : StakerState
 example3_expected_newStaker1StakingState = StakerState with
     config = StakerConfig with relockRatio = 0.5
     lifetimeRewards = 500e6 + staker1Allowance
+    exemptedLifetimeRewards = 0.0
     mintingAllowance = 0.5 * staker1Allowance
     unconstrainedStake = 300e6 + 0.5 * staker1Allowance
     stakes = []
@@ -210,6 +215,7 @@ example3_expected_newStaker2StakingState : StakerState
 example3_expected_newStaker2StakingState = StakerState with
     config = StakerConfig with relockRatio = 0.2
     lifetimeRewards = 200e6 + staker2Allowance
+    exemptedLifetimeRewards = 0.0
     mintingAllowance = 0.8 * staker2Allowance
     unconstrainedStake = 100e6 + 0.2 * staker2Allowance
     stakes = []
@@ -237,3 +243,4 @@ test_example3 = do
 
 -- TODO: add an example where some stakes are unlocked or freshly locked
 -- TODO: add an example with multiple SVs
+-- TODO: add an example with exempted lifteime rewards
--- a/daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
+++ b/daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
@@ -33,6 +33,11 @@ data StakerConfig = StakerConfig with
 data StakerState = StakerState with
     config :  StakerConfig -- ^ Configuration chosen by the staker.
     lifetimeRewards : Decimal -- ^ Total lifetime rewards earned by the staker.
+    exemptedLifetimeRewards : Decimal
+      -- ^ Portion of lifetime rewards that are exempted from staking requirements.
+      -- MAY be negative to reflect an exemption for future rewards.
+      --
+      -- Determined using SV voting.
     mintingAllowance : Decimal -- ^ Unused minting allowance for SV rewards.
     unconstrainedStake : Decimal -- ^ Amount of stake that is not currently locked or constrained.
     stakes : [Stake] -- ^ Stakes with constraints.
@@ -112,9 +117,11 @@ updateMintingAllowances context stakingState =
           None -> pure (sv, (0.0, weight))
           Some stakerState -> do
             let stakedAmount = getStakedAmount stakerState
+            let consideredLifetimeRewards =
+                  max 0.0 (stakerState.lifetimeRewards - stakerState.exemptedLifetimeRewards)
             let stakedRatio
-                  | stakerState.lifetimeRewards == 0.0 = 1.0
-                  | otherwise = min 1.0 (stakedAmount / stakerState.lifetimeRewards)
+                  | consideredLifetimeRewards == 0.0 = 1.0
+                  | otherwise = min 1.0 (stakedAmount / consideredLifetimeRewards)
             let earnedRatio = getEarnedRatio stakingState.config.stakingWeightSchedule stakedRatio
             let earnedWeight = weight * earnedRatio
             let unearnedWeight = weight - earnedWeight


daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
```diff
@@ -0,0 +1,234 @@
...
+data StakerState = StakerState with
+    config :  StakerConfig -- ^ Configuration chosen by the staker.
+    lifetimeRewards : Decimal -- ^ Total lifetime rewards earned by the staker.
+    exemptedLifetimeRewards : Decimal
+      -- ^ Portion of lifetime rewards that are exempted from staking requirements.
+      -- MAY be negative to reflect an exemption for future rewards.
+      --
+      -- Determined using SV voting.
+    mintingAllowance : Decimal -- ^ Unused minting allowance for SV rewards.
+    unconstrainedStake : Decimal -- ^ Amount of stake that is not currently locked or constrained.
```
moritzkiefer-da, line 42: I don't think I understand what this is. You say it's the amount not locked up. But then you add the relock amount to it?

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
```diff
@@ -0,0 +1,234 @@
+-- Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+-- | Staking for the SV rewards tranche Splice Amulet.
```
moritzkiefer-da, line 4: There seem to be a lot of implicit assumptions in this code that are not spelled out which make it quite hard for me to follow how the high-level minting and governance flow change. I think I got it now but let's spell it out more explicitly. Here is what I understood:

1. All SV weights are now tracked through governance including beneficiary weights. So no more "GSF has a very large weight and how they split it up is technically up to them and only defined through local configuration".
2. SV Reward Minting (and third party staking minting) does not work through creating coupons and minting them once per round. Instead you track the minting allowance and people can withdraw from that at their discretion.
3. For SVs we do want them to do something per round to get rewards so the current mechanism through activity records is replaced by `skippedSvs` which is filled in through some undetermined mechanism.
4. Once per round the SVs call `updateMintingAllowances` through a confirmation based process (likely with some SVs only committing to a hash and sharding it across SVs or something if the number of stakers grows large).
5. Not fully clear to me how the ghost SVs translate. I guess you have some dummy party that accrues a balance that the SVs can mint from instead of minting from unclaimed rewards?

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
meiersi-da: ah sorry... its staked amount that is neither being "onboarded" nor "offboarded". It's definitely locked. Will fix the comment.

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
meiersi-da: Thanks for the quick turnaround @moritzkiefer-da ; and the summary below. I've answered inline; and will ping @bame-da for a review before adjusting the code to inline your commentary in an appropriate form.

> There seem to be a lot of implicit assumptions in this code that are not spelled out which make it quite hard for me to follow how the high-level minting and governance flow change. I think I got it now but let's spell it out more explicitly. Here is what I understood:
> 
> 1. All SV weights are now tracked through governance including beneficiary weights. So no more "GSF has a very large weight and how they split it up is technically up to them and only defined through local configuration".

Yes. I overlooked the requirement to now track all of them on-ledger; and thus have not spelled it out. I also realize that this will very likely require tracking the state of escrowed rewards from milestone-based SV rewards. 

I'd expect that these are tracked in a separate `escrowedMintingAllowance`. What's unclear is whether they immediately count as `lifetimeRewards` or only once they get released form escrow. I suspect the latter.

@bame-da: what's your view?

> 2. SV Reward Minting (and third party staking minting) does not work through creating coupons and minting them once per round. Instead you track the minting allowance and people can withdraw from that at their discretion.

Yes.

> 3. For SVs we do want them to do something per round to get rewards so the current mechanism through activity records is replaced by `skippedSvs` which is filled in through some undetermined mechanism.

Yes, the strawman I have in mind is a comparison between the last status update and the round closing time to show that they have been active after `roundClosesAt - 2 * tickDuration`.

> 4. Once per round the SVs call `updateMintingAllowances` through a confirmation based process (likely with some SVs only committing to a hash and sharding it across SVs or something if the number of stakers grows large).

Yes. Also the computation needs to be refactored a bit to take the form `Summarize --> offLedgerSummary -->  parallel updateStakerSate`.

> 5. Not fully clear to me how the ghost SVs translate. I guess you have some dummy party that accrues a balance that the SVs can mint from instead of minting from unclaimed rewards?

See above. I'd make the `escrowedMintingAllowance` explicit.

daml/splice-amulet-test/daml/Splice/Scripts/UnitTests/Amulet/SvStaking.daml
```diff
@@ -0,0 +1,246 @@
...
+        actualWeightRatio = getEarnedRatio defaultStakingWeightSchedule stakedRatio
+
+-- Default config and context
+-----------------------------
+
+defaultStakingConfig : StakingConfig
+defaultStakingConfig = StakingConfig with
+  stakingDelay = days 30
+  unlockDelay = days 365
+  minStakeAmount = 100_000.0
```
bame-da, line 60: Why is this needed? For a fresh SV, the first round they mint, 80% may be less than 100k. Later on, nobody is going so stake anywhere between 0 and 20% of their historic total.

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
```diff
@@ -0,0 +1,234 @@
...
+    relockRatio : Decimal -- ^ Ratio of unlocked stake that is automatically relocked. Most be in [0.0, 1.0].
+  deriving (Eq, Show)
+
+-- | State of an individual staker.
+data StakerState = StakerState with
+    config :  StakerConfig -- ^ Configuration chosen by the staker.
+    lifetimeRewards : Decimal -- ^ Total lifetime rewards earned by the staker.
+    exemptedLifetimeRewards : Decimal
+      -- ^ Portion of lifetime rewards that are exempted from staking requirements.
+      -- MAY be negative to reflect an exemption for future rewards.
```
bame-da, line 38: I don't understand this. Is this an absolute amount? Why would this be negative? Wouldn't it be > lifetimeRewards if it takes into account future rewards?

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
```diff
@@ -0,0 +1,234 @@
...
+      --
+      -- Determined using SV voting.
+    mintingAllowance : Decimal -- ^ Unused minting allowance for SV rewards.
```
bame-da, line 41: Isn't this just `1.0 - relockRatio`? Under what circumstances would it be different?

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
bame-da: Isn't this always the quantity in an associated LockedAmulet - the amounts in the `stakes`? If so, is it a good idea to keep this amount in the StakerState?

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
```diff
@@ -0,0 +1,234 @@
...
+    unconstrainedStake : Decimal -- ^ Amount of stake that is not currently locked or constrained.
+    stakes : [Stake] -- ^ Stakes with constraints.
```
bame-da, line 43: For consistency, name `constrainedStakes` and `ConstrainedStake` or some such?
It would make clearer that total stake = constrained + unconstrained. Right now, one needs to infer that as the naming could suggest that `stakes` is inclusive of `unconstrainedStake`.

bame-da: Actually, I'm not sure whether it is or isn't inclusive... The data type of `Stake` doesn't make sense to me.

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
```diff
@@ -0,0 +1,234 @@
...
+  deriving (Eq, Show)
+
+-- | An individual stake with its associated metadata.
+data Stake = Stake with
```
bame-da, line 47: This data type doesn't make sense to me.
Suppose I stake 100.0 at T1, then accumulate a bunch of stake through relock, then stake another 100.0 at T2, then accumulate a bunch more through relock to end up at a total of 250.0. I now request to unlock 51.0 at T3...

What will the `stakes` field contain? I'd imagine I have two elements with amount 100.0 and stakedAt T1 and T2. But I'd also expect an element with amount 51.0 and requestedUnlockAt T3. But what's the stakedAt of that last element?

Or if the idea is that at times T1 + stakingDelay and T2 + stakingDelay the elements disappear from `stakes`, and I can only unlock from `unconstrainedStake`, then I will never have both `stakedAt` and `requestedUnlockAt` set at the same time so maybe it should be a sum type like

```
data StakingConstraint
  = SCStakedAt Time
  | SCRequestedUnlockAt Time
```

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
meiersi-da: Yes, apologies. The difference `liftetimeRewards - exemptedLifetimeRewards` may be negative. Will fix.

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
meiersi-da: This was meant as a technical optimization to avoid having to track many unlocked stakes, which haven't been reclaimed. It's though something to reconsider when doing the proper implementation that deals with the in- and out-flow of `(Locked)Amulet`.

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
meiersi-da: The current code uses this field to track both unminted rewards and unlocked stakes.

The reason for tracking a minting allowance is to avoid having the SVs mint for a coin owner. I realize though that the totals of unlocked stake and unminted coin should be tracked separately.

Even then the total unminted coin is different from `1.0 - relockRatio`. 

Side note: the state shown here is state that's tracked with `signatory dso` only. There's then a second process that mirrors that `dso` state to the actual parties. The split for that is due to the fact that observers can block a transaction if they haven't vetted the required packages. We do not want the DSO to get blocked, which is why we opt to do this kind of two-step issuance.

The alternative would be to make certain packages required to connect to a synchronzier.

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
meiersi-da: > Or if the idea is that at times T1 + stakingDelay and T2 + stakingDelay the elements disappear from stakes, and I can only unlock from unconstrainedStake.

That was the core idea. The reason for having both fields is that a staker might still request to unlock before the `stakedAt` constraint runs out. I'd have imagined that the unlocking works by ordering the stakes for which no unlocked has yet been requested in reverse order of `stakedAt`, and adding the `requestedUnlockAt` constraint.

The last stake encountered in this process might get split into two stakes, as would be required in your example.

daml/splice-amulet/daml/Splice/Amulet/SvStaking.daml
```diff
@@ -0,0 +1,234 @@
...
+            Some (earnedWeight, _) ->
+              let totalAllowance = earnedWeight * mintingAllowancePerSvWeight
+              in
+                  addTotalAllowance totalAllowance stakerState
+
+            None ->
+              -- Third-party staker: issue from staking pool for their part of the total pool stake
+              -- Note that an SV staker that is offboarded from the SVs will become a third-party staker.
+              let stakedAmount = getStakedAmount stakerState
+                  totalAllowance = stakingPoolMintingAllowance * (stakedAmount / totalThirdPartyStake)
```
meiersi-da, line 171: This is likely not how it should work. The third-party stake should be weighted as well.

PR #3604, opened 2026-01-20
Avoid stale button reference in SvFrontendIntegrationTest

martinflorian-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/6828

[ci]

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvFrontendIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvFrontendIntegrationTest.scala
@@ -1084,7 +1084,7 @@ class SvFrontendIntegrationTest
             _ => checkNewVoteRequestInProgressTab(previousVoteRequestsInProgress),
           )
 
-          val (_, reviewButton) = actAndCheck(
+          actAndCheck(
             "sv1 operator creates a new vote request with a long expiration time", {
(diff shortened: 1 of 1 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvFrontendIntegrationTest.scala
```diff
@@ -1097,6 +1097,8 @@ class SvFrontendIntegrationTest
             },
           )
           val requestId = eventually() {
+            // find the review button again because the DOM may have been updated
+            val reviewButton = checkNewVoteRequestInProgressTab(previousVoteRequestsInProgress + 1)
```
pawelperek-da, line 1101: huh, I have a feeling that a function named `checkNewVoteRequestInProgressTab` probably shouldn't return the `reviewButton` but that's beyond the scope of this task 😛

martinflorian-da: I tend to agree :see_no_evil: Overall 1. when fixing a flake I just want to do it as quickly and with as few changes as possible and 2. every time I see a selenium test I wonder whether we shouldn't just kill it altogether in favor of vite + testing only the APIs :see_no_evil: (Not saying that we should do it here, but it's my excuse for being less motivated on these tests...)

PR #3615, opened 2026-01-20
[ci] Add storeDescriptorUserVersion config to scan.

### LINKED ISSUE #3534: Add to scan a config value that allows triggering reingestion of txlog
We made changes to parsing of votes, that makes "old" SVs (e.g. DA-2) not show votes before Nov'24 in scan&SV UIs.
The easiest way to fix that is to trigger re-ingestion/reparsing of txlog. We don't want an endpoint in Scan that triggers it as that is unauth'ed, so adding a config parameter for the store ID seems like the easiest.

ray-roestenburg-da:
Fixes #3534 

(I think this is what @rautenrieth-da meant)

- [x] Add helm value and test
- [x] Add docs
- [x] Thinking of a way to test this

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
+++ b/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
@@ -2123,6 +2123,7 @@ class DbScanStoreTest
 
     "force re-ingestion of acs" in {
       for {
+        // ingestion in these store tests is simulated by directly interacting with the ingestion sink (dummyDomain.create)
         // create store, ingest an update with aliceValidatorLicense
         store <- mkStore()
         _ <- dummyDomain.create(aliceValidatorLicense)(store.multiDomainAcsStore)
@@ -2131,7 +2132,9 @@ class DbScanStoreTest
         )
         // create store again but now with new storeDescriptor userVersion, ingest an update with aliceValidatorLicense again
         storeReingest <- mkStore(dsoParty = dsoParty, acsStoreDescriptorUserVersion = Some(1L))
-        // ingestUpdate would fail on the same storeId for aliceValidatorLicense without a new user version.
+        // Below 'dummyDomain.create' would fail on the same storeId for aliceValidatorLicense without a new user version.
+        // there is a unique constraint on (store_id, migration_id, contract_id) in the acs table (acs_store_template_sid_mid_cid) that would be violated.
+        // Successfully creating the license again here proves that the store has switched to a new store descriptor.
         _ <- dummyDomain.create(aliceValidatorLicense)(storeReingest.multiDomainAcsStore)
         resultAfter <- storeReingest.getValidatorLicenseByValidator(
           Vector(alice)
@@ -2162,13 +2165,16 @@ class DbScanStoreTest
         voteRequestContracts = mkVoteRequests() :+ activeVoteRequest
         _ <- assertListOfAllPastVoteRequestResults(voteRequestContracts, store)
         _ <- dummyDomain.create(activeVoteRequest)(store.multiDomainAcsStore)
-        // create the store with a new txLogStoreDescriptorUserVersion and check that it has acs entries but no txLog entries of the previous store
+        // create the store with a new txLogStoreDescriptorUserVersion and
+        // check that it has acs entries but no txLog entries of the previous store
+        // this proves that the store descriptor has changed and a new storeId is used.
         storeReingest <- mkStore(
           dsoParty = dsoParty,
           txLogStoreDescriptorUserVersion = Some(1L),
         )
       } yield {
         // new store should not have txLog entry from previous store
+        // because ingestion in these store tests is simulated by directly interacting with the ingestion sink
         storeReingest
           .listVoteRequestResults(
             Some("AddSv"),
@@ -2180,7 +2186,7 @@ class DbScanStoreTest
           )
           .futureValue
           .toList should have size 0
-        // should have the acs entry
+        // should have the active acs entry
         storeReingest.listVoteRequests().futureValue.toList should contain(activeVoteRequest)
       }
     }


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
```diff
@@ -59,6 +59,7 @@ case class ScanAppBackendConfig(
     txLogBackfillBatchSize: Int = 100,
     bftSequencers: Seq[BftSequencerConfig] = Seq.empty,
     cache: ScanCacheConfig = ScanCacheConfig(),
+    storeDescriptorUserVersion: Option[Long] = None,
```
rautenrieth-da: I would let users configure the userVersion of  `acsStoreDescriptor` and `txLogStoreDescriptor` independently. Otherwise it's impossible to bump one store without the other, and they are independent stores.

ray-roestenburg-da: right that makes sense

ray-roestenburg-da: I'll also add the config to the sv and validator apps, makes sense

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
```diff
@@ -1958,12 +1971,17 @@ class DbScanStoreTest
...
     )(parallelExecutionContext, implicitly, implicitly)
 
     for {
       _ <- store.multiDomainAcsStore.testIngestionSink.initialize()
-      _ <- store.multiDomainAcsStore.testIngestionSink
-        .ingestAcs(nextOffset(), Seq.empty, Seq.empty, Seq.empty)
+      _ <-
+        if (!skipIngestAcs) {
+          store.multiDomainAcsStore.testIngestionSink
+            .ingestAcs(nextOffset(), Seq.empty, Seq.empty, Seq.empty)
```
rautenrieth-da:
```suggestion
      _ <- initializeResult match {
          case ResumeAtOffset(_) => Future.unit
          case InitializeAcsAtLatestOffset => store.multiDomainAcsStore.testIngestionSink.ingestAcs(...)
          case InitializeAcsAtOffset(_) => store.multiDomainAcsStore.testIngestionSink.ingestAcs(...)
```

With `initializeResult` coming from the above `initialize()` call. This way you don't need the new boolean parameter (I know someone who doesn't like them).

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
```diff
@@ -2069,4 +2087,77 @@ class DbScanStoreTest
...
+    }
+  }
+
+  "Changing the txLogStoreDescriptorUserVersion" should {
+    "force re-ingestion of txLog" in {
+      for {
+        store <- mkStore()
+        _ <- assertListOfAllPastVoteRequestResults(store)
+        // create the store with a new txLogStoreDescriptorUserVersion and check that it logs that TxLog backfilling will start restoring previous entries
+        storeReingest <- loggerFactory.assertLogsSeq(SuppressionRule.Level(Level.INFO))(
```
rautenrieth-da: Do we need the asserts on the log here? IMO it would have been enough to check that for `storeReingest`, an acs-based query returns something (like you do in `force re-ingestion of acs`) while a txlog-based query returns nothing (like you do below).

ray-roestenburg-da: yeah I can leave it out, I thought it does really prove that the tx log is going to be refilled, which is why I liked it.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
ray-roestenburg-da: LOL yes super ironic! 😄

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
ray-roestenburg-da: Left it out and added the acs check.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
ray-roestenburg-da: done

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
```diff
@@ -2069,4 +2113,76 @@ class DbScanStoreTest
...
+      for {
+        // create store, ingest an update with aliceValidatorLicense
+        store <- mkStore()
+        _ <- dummyDomain.create(aliceValidatorLicense)(store.multiDomainAcsStore)
+        result <- store.getValidatorLicenseByValidator(
+          Vector(alice)
+        )
+        // create store again but now with new storeDescriptor userVersion, ingest an update with aliceValidatorLicense again
+        storeReingest <- mkStore(dsoParty = dsoParty, acsStoreDescriptorUserVersion = Some(1L))
+        // ingestUpdate would fail on the same storeId for aliceValidatorLicense without a new user version.
```
isegall-da: Not sure I understand what this comment refers to, did you want to assert that something fails?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
```diff
@@ -2069,4 +2113,76 @@ class DbScanStoreTest
...
   }
+  "Changing the acsStoreDescriptorUserVersion" should {
+    val alice = userParty(443)
+    val aliceValidatorLicense = validatorLicense(
+      alice,
+      dsoParty,
+      Some(new FaucetState(new Round(0), new Round(1000), 0L)),
+    )
+
+    "force re-ingestion of acs" in {
```
isegall-da, line 2124: Not sure I see where this test checks that something was actually re-ingested.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
```diff
@@ -2069,4 +2113,76 @@ class DbScanStoreTest
...
+              Optional.empty(),
+            )
+          ),
+      )
+      for {
+        store <- mkStore()
+        voteRequestContracts = mkVoteRequests() :+ activeVoteRequest
+        _ <- assertListOfAllPastVoteRequestResults(voteRequestContracts, store)
+        _ <- dummyDomain.create(activeVoteRequest)(store.multiDomainAcsStore)
+        // create the store with a new txLogStoreDescriptorUserVersion and check that it has acs entries but no txLog entries of the previous store
```
isegall-da: Don't we wanna test that it actually does start reingesting? Currently IIUC, we're only checking that the new store is empty.

docs/livepreview.sh
```diff
@@ -10,7 +10,7 @@ if [[ $# -eq 1 ]] && [[ "$1" == "--skip-daml" ]]; then
   echo "Skipping generation of DAML docs as requested."
 else
   # Generate DAML docs
-  (cd "$SPLICE_ROOT"; sbt --client --batch damlBuild)
+  (cd "$SPLICE_ROOT"; sbt --batch damlBuild)
```
isegall-da, line 13: unintentional?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
ray-roestenburg-da: the ingestUpdate happens from the dummyDomain.create (it's about the line below the comment). without a new user version, the store id would be the same, and the create would fail (unique constraint `scan_acs_store_store_id_migration_id_contract_id_idx`). Not failing, and being able to read the ingested update, shows that the store descriptor has changed and in turn a new store id has been created. (it's not the most clear test, I agree with that)

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
ray-roestenburg-da: Its the one validator license that's added again, as explained on the other comment.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
ray-roestenburg-da: I discussed this earlier with Robert, I had some log assertions of getting into that state, but decided it was not necessary. https://github.com/hyperledger-labs/splice/pull/3615#discussion_r2726956921

docs/livepreview.sh
ray-roestenburg-da: Intentional. `./docs/livepreview.sh` does not work with that flag when I already have a running sbt in another terminal, `damBuild` can't be found cause I'm in a sub-project for testing, `apps-scan` for instance, which is just annoying. It does work if no sbt is already running, which sort of defeats the purpose, or you happen to be in the right project for `damlBuild` to be available I guess, which is just dumb IMHO. Also, `--client` has been removed in sbt `2.x` (I wonder why LOL). too esoteric a problem to really figure out..

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
isegall-da: > without a new user version, the store id would be the same, and the create would fail

Sounds like we should do exactly that in the test then.

> it's not the most clear test, I agree with that

I agree, let's try to make it clearer. At the very least, with explicit comments, ideally by changing the test (and if needed, making the implementation more testable). If this fails a month from now, nobody, including you, would be able to understand what this test does and why and what broke.

docs/livepreview.sh
isegall-da: OK. I definitely cannot claim to understand sbt well enough to have an opinion, but if you're sure this fixes something and does not break anything else, then I'm fine.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
ray-roestenburg-da: I'll add some comments. re-ingestion on store descriptor change has been long proven to work, this proves that user version modifies the store descriptor. These store tests don't test ingestion, they simulate ingestion through the ingestion sink. The comments should be clear enough for what's tested. I'll create a separate issue to make this code more testable in general and easier to verify.

ray-roestenburg-da: https://github.com/hyperledger-labs/splice/issues/3718

PR #3621, opened 2026-01-20
refactor: extract KeyValueStore to common

isegall-da:
This factors out of the validator, to common, the key-value storage functionality from ValidatorInternalStore, and also applies some renaming to make it a more generic key-value and not specifically for "configurations".

Motivation:
Add a key-value store to Scan (specifically, in the near-term, to store progress of the bulk-storage dumps)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/KeyValueStoreTest.scala
+++ b/apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/KeyValueStoreTest.scala
@@ -23,9 +23,17 @@ class KeyValueStoreTest
       participant = mkParticipantId("participant"),
       key = Map(),
     )
+  private val storeDescriptor2 =
+    StoreDescriptor(
+      version = 2,
+      name = "DbMultiDomainAcsStoreTest",
+      party = dsoParty,
+      participant = mkParticipantId("participant"),
+      key = Map(),
+    )
 
-  private def mkStore() = KeyValueStore(
-    storeDescriptor,
+  private def mkStore(descriptor: StoreDescriptor) = KeyValueStore(
+    descriptor,
     storage,
     loggerFactory,
   )
@@ -39,7 +47,7 @@ class KeyValueStoreTest
 
     "set and get a payload successfully" in {
       for {
-        store <- mkStore()
+        store <- mkStore(storeDescriptor)
         _ <- store.setValue(configKey, configValue)
         retrievedValue <- store.getValue[String](configKey).value
       } yield {
@@ -49,7 +57,7 @@ class KeyValueStoreTest
 
     "return None for a non-existent key" in {
       for {
-        store <- mkStore()
+        store <- mkStore(storeDescriptor)
         retrievedValue <- store.getValue[String]("non-existent-key").value
       } yield {
         retrievedValue shouldBe None
@@ -58,7 +66,7 @@ class KeyValueStoreTest
 
     "update an existing payload" in {
       for {
-        store <- mkStore()
+        store <- mkStore(storeDescriptor)
         _ <- store.setValue(configKey, configValue)
         _ <- store.setValue(configKey, otherValue)
         retrievedValue <- store.getValue[String](configKey).value
@@ -69,7 +77,7 @@ class KeyValueStoreTest
 
     "handle multiple different keys independently" in {
       for {
-        store <- mkStore()
+        store <- mkStore(storeDescriptor)
         _ <- store.setValue(configKey, configValue)
         _ <- store.setValue(otherKey, otherValue)
 
@@ -83,7 +91,7 @@ class KeyValueStoreTest
 
     "delete single key" in {
       for {
-        store <- mkStore()
+        store <- mkStore(storeDescriptor)
         _ <- store.setValue(configKey, configValue)
         _ <- store.setValue(otherKey, otherValue)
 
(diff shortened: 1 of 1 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/KeyValueStore.scala
```diff
@@ -0,0 +1,80 @@
...
+import com.digitalasset.canton.lifecycle.CloseContext
+import com.digitalasset.canton.logging.{ErrorLoggingContext, NamedLoggerFactory, NamedLogging}
+import com.digitalasset.canton.resource.DbStorage
+import com.digitalasset.canton.tracing.TraceContext
+import io.circe.{Decoder, Encoder}
+import org.lfdecentralizedtrust.splice.store.db.{DbValidatorInternalStore, StoreDescriptor}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+case class KeyValueStoreDbTableConfig(
```
isegall-da: I have some doubts about this:

My motivation is that I don't like the current names of the table and columns, they are too "config" specific, while at least in scan it won't be for configs (I claim that in the validator it is no longer that either..).

Alternatives are:
1. Just accept the names, reuse them in scan
2. Rename the table and columns in validator app
3. Make this configurable, as I do here (in which case, I'm sure there's a nicer Scala'y way, to put this config class in the DB instance, and generalize it here, but I wasn't sure how...)

Would doing (2) make more sense? Or does it make sense to make this configurable? (in which case, I'd be happy for a nicer alternative for how)

isegall-da: Option (3) does allow in theory in the future to have multiple key-value tables for the same app, but not sure we'll ever want to actually do that.

isegall-da: OTOH I realize now that our flyway migration is not app-specific, so we basically have the same tables for scan and validators, so now I'm actually leaning much stronger to option (2)...

isegall-da: Please ignore, was talking to myself, and convinced myself to change this :D

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
isegall-da, line 1: I'm thinking now that maybe this should just be added as another store in DbAppStore instead of having this class (and a similar one in Scan very soon). Thoughts?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbKeyValueStore.scala
```diff
@@ -1,50 +1,47 @@
...
 
-  override def setConfig[T](key: String, value: T)(implicit
+  override def setValue[T](key: String, value: T)(implicit
       tc: TraceContext,
       encoder: Encoder[T],
   ): Future[Unit] = {
     val jsonValue: Json = value.asJson
 
-    val action = sql"""INSERT INTO validator_internal_config (config_key, config_value, store_id)
+    val action = sql"""INSERT INTO key_value_store (key, value, store_id)
```
ray-roestenburg-da, line 41: changing the table name, what about existing data? (does this matter for the validator store version?)

rautenrieth-da: We're only renaming a table and some columns/constraints, that shouldn't affect existing data, no?

isegall-da: The table itself is also renamed accordingly in flyway, so data would be preserved

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/KeyValueStoreTest.scala
```diff
@@ -0,0 +1,104 @@
...
+  )
+
+  "KeyValueStore" should {
+
+    val configKey = "key1"
+    val configValue = "jsonPayload1"
+    val otherKey = "key2"
+    val otherValue = "jsonPayload2"
+
+    "set and get a payload successfully" in {
```
rautenrieth-da, line 48: Suggestion for one more test: create two stores with different descriptors, check that they don't affect each others data.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
rautenrieth-da: I don't understand the suggestion, you want to add a generic `def keyValues: KeyValueStore` to `trait AppStore`?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbKeyValueStore.scala
ray-roestenburg-da: thanks missed that (flyway)

nicu-da:
> > This makes me wonder how do we treat such releases that make rollbacks not possible without a restore from backup though
> 
> Doesn't that apply to any change that adds a new flyway migration? We have 50 migrations already and haven't called them out in the release notes before. Shouldn't we just make it the default expectation that you can't downgrade to a previous release without restoring from backup?

Additions are not an issue for rollbacks, and i suspect most of our migrations are additions and not mutations/deletions. We have occasionally rolled back at least on our own nodes so I think we should start giving more awareness to these forward only changes.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/store/ValidatorInternalStore.scala
isegall-da: Something like that. Mostly noticed that all other instantiations of StoreDescriptor are all in one place (e.g. for scan, it's in DbScanStore), so was wondering whether we want to add it somewhere that each app just needs to define the StoreDescriptor for it. But that whole setup with AcsStores, TxLogStores, etc is so complex as it is, that I'm no longer thinking this is a great idea...

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/KeyValueStoreTest.scala
isegall-da: 👍 
(note that these tests are not new, it's just refactoring of the existing ones, separating the generic key-store ones from the ones specific for validator data, but I'll add a test nevertheless)

isegall-da:
> Hm, [this line](https://github.com/hyperledger-labs/splice/blob/main/canton/community/ledger/ledger-api-core/src/main/scala/com/digitalasset/canton/platform/store/FlywayMigrations.scala#L59) says we explicitly configure flyway to complain if the database contains future migrations

So we basically never allowed downgrade between versions that contain any flyway migration?

Issue #3562, opened 2026-01-15
Run store performance tests on CI


OriolMunoz-da:
(non-exhaustive) list of requirements:
- [ ] CloudSQL instance on the same region/zone as the GH runners (to avoid latency)
- [ ] Run on GHA as we don't want CircleCI anymore
- [ ] Run all performance tests `run *` option)

PR #3634, opened 2026-01-21
Deploy CloudSQL instance for performance tests

OriolMunoz-da:
Part of https://github.com/hyperledger-labs/splice/issues/3562

I expect things to break (missing permissions). After this is deployed:
- Confirm that the env vars are readable within a GHA job
- Add the job to run the performance tests

### DIFF
--- a/cluster/pulumi/gha/src/performanceTests.ts
+++ b/cluster/pulumi/gha/src/performanceTests.ts
@@ -1,3 +1,5 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
 import { CloudPostgres, ExactNamespace } from '@lfdecentralizedtrust/splice-pulumi-common';
 
 export function createCloudSQLInstanceForPerformanceTests(


cluster/pulumi/gha/src/performanceTests.ts
```diff
@@ -0,0 +1,25 @@
...
+  return new CloudPostgres(
+    ghaNamespace,
+    'performance-test-db',
+    'performance-test-db',
+    'performance-test-db-secret',
+    {
+      enabled: true,
+      maintenanceWindow: { day: 2, hour: 8 },
+      protected: false,
+      tier: 'db-custom-2-7680', // same as devnet & testnet as of Jan 2026
```
OriolMunoz-da, line 17: the actual configuration is available in the internal repo, not in this one...
that being said, it's likely we'll increase the DB size, not make it smaller, so this should be fine as baseline

cluster/pulumi/gha/src/runners.test.ts
```diff
@@ -15,13 +15,17 @@ jest.mock('./config', () => ({
...
 }));
+class FakeCloudPostgres extends pulumi.Resource {}
 jest.mock('@lfdecentralizedtrust/splice-pulumi-common', () => ({
   __esModule: true,
   appsAffinityAndTolerations: {},
   DOCKER_REPO: 'https://dummy-docker-repo.com',
   HELM_MAX_HISTORY_SIZE: 42,
   imagePullSecretByNamespaceNameForServiceAccount: () => [],
   infraAffinityAndTolerations: {},
+  CloudPostgres: function CloudPostgres() {
```
OriolMunoz-da, line 26: test below breaks without this

cluster/pulumi/gha/src/runners.ts
```diff
@@ -754,12 +775,17 @@ function installPodMonitor(runnersNamespace: Namespace) {
...
+const GHA_NAMESPACE_NAME = 'gha-runners';
 export function installRunnerScaleSets(controller: k8s.helm.v3.Release): void {
-  const runnersNamespace = new Namespace('gha-runners', {
+  const runnersNamespace = new Namespace(GHA_NAMESPACE_NAME, {
     metadata: {
-      name: 'gha-runners',
+      name: GHA_NAMESPACE_NAME,
     },
   });
+  const exactNs: ExactNamespace = {
```
isegall-da, line 785: One day I'll understand why we have this ExactNamespace all over the place...

OriolMunoz-da: you mean this? https://github.com/hyperledger-labs/splice/blob/b60b1dd3334da65597c27df6476cf7738e11781f/cluster/pulumi/common/src/namespace.ts#L5

PR #3635, opened 2026-01-21
Avoid scan rate limiting in `TokenStandardCliTestDataTimeBasedIntegrationTest`

martinflorian-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/6839

[ci]

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardCliTestDataTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardCliTestDataTimeBasedIntegrationTest.scala
@@ -690,7 +690,8 @@ class TokenStandardCliTestDataTimeBasedIntegrationTest
         )
       def replaceStringsInJson(viewValue: Json) = {
         val current = viewValue.spaces2SortKeys
-        val amuletRulesId = eventuallySucceeds()(sv1ScanBackend.getAmuletRules().contractId.contractId)
+        val amuletRulesId =
+          eventuallySucceeds()(sv1ScanBackend.getAmuletRules().contractId.contractId)
         val allContracts =
           "\"([0-9a-fA-F]{138})\"".r.findAllIn(current).matchData.map(_.group(1)).toSeq
 


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardCliTestDataTimeBasedIntegrationTest.scala
```diff
@@ -105,6 +107,16 @@ class TokenStandardCliTestDataTimeBasedIntegrationTest
           _.withPausedTrigger[CollectRewardsAndMergeAmuletsTrigger]
         )(config)
       )
+      .addConfigTransforms((_, config) =>
+        // The test iteslf may hit scan hard, but this test is not about testing rate limiting...
```
moritzkiefer-da:
```suggestion
        // The test itself may hit scan hard, but this test is not about testing rate limiting...
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardCliTestDataTimeBasedIntegrationTest.scala
```diff
@@ -678,7 +690,7 @@ class TokenStandardCliTestDataTimeBasedIntegrationTest
         )
       def replaceStringsInJson(viewValue: Json) = {
         val current = viewValue.spaces2SortKeys
-        val amuletRulesId = sv1ScanBackend.getAmuletRules().contractId.contractId
+        val amuletRulesId = eventually()(sv1ScanBackend.getAmuletRules().contractId.contractId)
```
moritzkiefer-da: what is the eventually for? If it's for handling the command failure from it not existing (not sure why it would not exist?) don't you need an `eventuallySucceeds`?

martinflorian-da: Ah I meant to comment on this line before marking ready, sorry...

> what is the eventually for?

Just a cheap extra safety against flakes... this is the thing that failed previously. I don't care here if scan has a random hiccup.

> don't you need an eventuallySucceeds

Yes, thanks for catching that!

rautenrieth-da: The [original issue](https://github.com/DACH-NY/cn-test-failures/issues/6839) was that this call failed due to rate limiting. Why is only this call being rate limited? Can we cache the amulet rules to reduce the amount of calls?

martinflorian-da: It's just a dumb call from out of the test itself... I wouldn't overthink this tbh; we're not testing rate limiting here and the thing that is causing too much load on scan is not part of our app logic; just the test logic.

martinflorian-da: I mean we could implement caching inside of the test but the gain should just be a minor test duration improvement... doesn't seem worth it to me.

rautenrieth-da (approved):

PR #3637, opened 2026-01-21
[dnm] test perftest cloudsql connectivity

OriolMunoz-da:


.github/workflows/performance_tests.yml
```diff
@@ -0,0 +1,45 @@
...
+        id: setup
+        uses: ./splice-shared-gha/.github/actions/tests/common_test_setup
+        with:
+          test_name: oss
+          target: 'oss'
+
+      - name: Env vars
+        shell: bash
+        run: |
+          echo "Checking connection to Splice services... Connecting to '$PERFORMANCE_TESTS_DB_HOST' with user '$PERFORMANCE_TESTS_DB_USER'..."
```
OriolMunoz-da, line 43: @isegall-da can I get you to double-check what I did?

https://github.com/hyperledger-labs/splice/actions/runs/22406758775/job/64868657025?pr=3637#step:5:2 indicates that the environment variables (from [this PR](https://github.com/hyperledger-labs/splice/pull/3814/changes#diff-770bf17dcb4d122c764186e610403210e4a04231d33243e5d439e840a25abebbR541)) are not actually accessible here, contrary from what we expected

am I missing something? apart from setting these as GH secrets, what other options do I have?

OriolMunoz-da: AI says

> The ACTIONS_RUNNER_CONTAINER_HOOK_TEMPLATE (/pod.yaml) is responsible for applying the environment variables to the workflow pod. Verify that this template correctly includes the PERFORMANCE_TESTS_DB_* variables.

does that sound right?

isegall-da: Ha?! But that's exactly what we did!!

I would try two things:
- putting a very long sleep here, then opening a shell on the runner pod to see if the env vars are there.
- mounting it as a file (e.g. via a configmap) in the runner instead of env vars, and reading the file here. Ugly, but maybe will work?!

isegall-da: Given that this is credentials to a test DB, I don't really mind them being in a GH secret either.

OriolMunoz-da: > Ha?! But that's exactly what we did!!

We didn't, we set it as an environment variable next to that one, as opposed to inside the pod.yaml

Checking both pods:
```
runner@self-hosted-k8s-small-r559s-runner-ndtmx:~$ echo $PERFORMANCE_TESTS_DB_USER
cnadmin
```
but:
```
ci@self-hosted-k8s-small-r559s-runner-ndtmx-workflow:/__w/splice/splice$ echo $PERFORMANCE_TESTS_DB_USER
```

I'm not sure I understand what the difference is, more so because the run reports:
```
Runner name: 'self-hosted-k8s-small-r559s-runner-ndtmx'
Runner group name: 'Default'
Machine name: 'self-hosted-k8s-small-r559s-runner-ndtmx'
```

PR #3653, opened 2026-01-22
add support for migrating postgres pvcs to hyperdisk

nicu-da:
[static]

3 step migration process:

1. Deploy with hyperdisk support enabled and migrating true. This takes a snapshot of the old pvc configures postgres to recreate it with new values for the pvc template that specifies the dataSource as the created snapshot, thus bootstrapping from the old data
2. Deploy with hyperdisk support enabled, migrating false and recreate postgres releases set to true. This removes the snapshot and the pvc template dataSource so that in the future if you delete the pvc it doesn't recreate it from the snapshot
3. Deloy with hyperdisk support enabled only 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/common/src/postgres.ts
+++ b/cluster/pulumi/common/src/postgres.ts
@@ -282,7 +282,7 @@ export class SplicePostgres extends pulumi.ComponentResource implements Postgres
         aliases: [{ name: logicalNameAlias, type: 'kubernetes:helm.sh/v3:Release' }],
         dependsOn: [passwordSecret],
         ...((supportsHyperdisk &&
-          // during the migration we first delete the stateful set, which keeps the old pvcs, and the recreate with the new pvcs
+          // during the migration we first delete the stateful set, which keeps the old pvcs (stateful sets always keep the pvcs), and then recreate with the new pvcs
           // the stateful sets are immutable so they need to be recreated to force the change of the pvcs
           hyperdiskSupportConfig.hyperdiskSupport.migrating) ||
         spliceConfig.pulumiProjectConfig.replacePostgresStatefulSetOnChanges


cluster/helm/splice-postgres/templates/postgres.yaml
```diff
@@ -105,6 +105,10 @@ spec:
           storage: {{ $.Values.db.volumeSize }}
       storageClassName: {{ $.Values.db.volumeStorageClass }}
       volumeMode: Filesystem
+      {{- with .Values.db.dataSource }}
+      dataSource:
+        {{- toYaml . | nindent 8 }}
```
martinflorian-da, line 110: you are a brave man... I would have done a helm test here just to be sure I counted the whitespaces right (I think I'd still soft recommend it)

cluster/pulumi/common/src/postgres.ts
```diff
@@ -224,6 +227,35 @@ export class SplicePostgres extends pulumi.ComponentResource implements Postgres
 
     // an initial database named cantonnet is created automatically (configured in the Helm chart).
     const smallDiskSize = clusterSmallDisk ? '240Gi' : undefined;
+    const supportsHyperdisk =
+      hyperdiskSupportConfig.hyperdiskSupport.enabled && !useInfraAffinityAndTolerations;
+    let hyperdiskMigrationValues = {};
```
martinflorian-da, line 232: any reason not do `const hyperdiskMigrationValues = xyz ? { ... } : { }` like we usually do?

cluster/pulumi/common/src/postgres.ts
```diff
@@ -242,11 +281,19 @@ export class SplicePostgres extends pulumi.ComponentResource implements Postgres
       {
         aliases: [{ name: logicalNameAlias, type: 'kubernetes:helm.sh/v3:Release' }],
         dependsOn: [passwordSecret],
+        ...((supportsHyperdisk &&
+          // during the migration we first delete the stateful set, which keeps the old pvcs, and the recreate with the new pvcs
```
martinflorian-da:
```suggestion
          // during the migration we first delete the stateful set, which keeps the old pvcs (stateful sets always keep the pvcs), and then recreate with the new pvcs
```

martinflorian-da: > which keeps the old pvcs

I'm not sure I fully understand this. Does it keep the PVCs or just their contents, via the snapshot?

If the former: Why, exactly? Do we have some retain-on-delete on them? Do we expect these to get garbage collected once we kill the old node pool?

If the latter: How do we know again that the snapshot will be created before the PVC disappears? I guess because Pulumi always deletes things after creating everything it was asked to delete?

(I'm obviously confused as usual; the relevant point for this PR is probably: Let's add more text to this comment...)

cluster/pulumi/common/src/postgres.ts
```diff
@@ -224,6 +227,35 @@ export class SplicePostgres extends pulumi.ComponentResource implements Postgres
...
+        `pg-data-${xns.logicalName}-${instanceName}-snapshot`,
+        {
+          apiVersion: 'snapshot.storage.k8s.io/v1',
+          kind: 'VolumeSnapshot',
+          metadata: {
+            name: `pg-data-${instanceName}-snapshot`,
+            namespace: xns.logicalName,
+          },
+          spec: {
+            volumeSnapshotClassName: 'dev-vsc',
```
martinflorian-da, line 244: offtopic, just ranting: that name...

cluster/helm/splice-postgres/templates/postgres.yaml
nicu-da: I did a full manual test to migrate a scratch

cluster/pulumi/common/src/postgres.ts
nicu-da: have to create the snapshot in  the if

cluster/pulumi/common/src/postgres.ts
nicu-da: statefulsets always keep the pvcs, you have to explicitly drop them.
snapshot is created before because anyway pulumi creates a dependency between the snapshot creation and stateful set recreation

cluster/pulumi/common/src/postgres.ts
nicu-da: have no idea who created that, it's in all the clusters :shrug: not managed by us

cluster/pulumi/common/src/postgres.ts
martinflorian-da: ah I see; this sounds like a reason then...

want to add a brief comment? (nit)

cluster/pulumi/common/src/postgres.ts
martinflorian-da: I see, thanks! So we'll kill the old PVC with the node pool then? (Will that just work?)

> Let's add more text to this comment..

:pray:

nicu-da: Open for suggestions :smile:, we do mention that we keep the old pvcs, not the snapshots

cluster/pulumi/common/src/postgres.ts
nicu-da: I don't see the value of the comment tbh, it would comment on the syntax of typescript not add any extra details about the why

cluster/pulumi/common/src/postgres.ts
martinflorian-da: Updated my suggestion above; that would have helped me not ask so many questions.

Then maybe add a sentence about when/how the old PVCs die?

cluster/pulumi/common/src/postgres.ts
martinflorian-da: Actually, why can't you write this in functional style again?

Would something like this work?

```
const hyperdiskMigrationValues = (xyz ? (() => { new xyz; return {abc}) : (() => {}))()
```

(all of this is nit; I just stumbled about the inconsistency so figured other readers might too; we very rarely use mutable variables)

cluster/pulumi/common/src/postgres.ts
nicu-da: > Then maybe add a sentence about when/how the old PVCs die?

But that's fully orthogonal right? It sounds like we're going in the steps of documenting how stateful sets work and how to deal with pvcs. Especially for this process I see no value in that as once we migrate I plan on purging everything.

martinflorian-da: > Especially for this process I see no value in that as once we migrate I plan on purging everything

ACK fair point.

cluster/pulumi/common/src/postgres.ts
nicu-da: We do use  mutable variables in  tyepscript when needed. 
I find that the mutable state with the if is much cleaner to read compared to an anonymous function that gets called in the same line. It took me a few reads to figure out the nested complexity there.

PR #3660, opened 2026-01-22
Make shellcheck cover our `.source` files

martinflorian-da:
...and fix them so it passes.

This is to avoid issues such as https://github.com/hyperledger-labs/splice/pull/3650

[static]

I tried tweaking `.pre-commit-config.yaml` instead but it didn't quite work / ended up way too complex.

With only two affected files, adding a line on top and adding `+x` seems like a reasonable workaround...

### DIFF
--- a/build-tools/lib/libcli.source
+++ b/build-tools/lib/libcli.source
@@ -1,11 +1,14 @@
-# -*- shell-script -*-
+#!/usr/bin/env bash
+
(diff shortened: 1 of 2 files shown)


isegall-da:
> That and chmod a+x

Ah, I missed that part. Ugly as hell, but if it works...

PR #3662, opened 2026-01-23
Increase backup wait time and delete timeouts in node-backup

OriolMunoz-da:
Attempts to fix maybe https://github.com/DACH-NY/cn-test-failures/issues/7017

### DIFF
--- a/cluster/scripts/node-backup.sh
+++ b/cluster/scripts/node-backup.sh
@@ -56,7 +56,7 @@ function wait_for_pvc_backup() {
       break
     else
       (( i++ )) && (( i > WAIT_FOR_BACKUP_RETRIES )) && {
-        kubectl delete volumesnapshot -n "$namespace" "$backupName";
+        # kubectl delete volumesnapshot -n "$namespace" "$backupName";
         _error "Timed out waiting for backup of $description PVC";
       }
       sleep 5
@@ -153,7 +153,7 @@ function wait_for_cloudsql_backup() {
       break
     else
       (( i++ ))&& (( i > WAIT_FOR_BACKUP_RETRIES )) && {
-        gcloud sql backups delete "$id" --instance "$db_id" --quiet;
+        # gcloud sql backups delete "$id" --instance "$db_id" --quiet;
         _error "Timed out waiting for backup of $description db";
       }
       sleep 5


cluster/scripts/node-backup.sh
```diff
@@ -148,7 +152,10 @@ function wait_for_cloudsql_backup() {
       _info "Backup of $description ready! Backup ID: $id "
       break
     else
-      (( i++ ))&& (( i > 300 )) &&_error "Timed out waiting for backup of $description db"
+      (( i++ ))&& (( i > WAIT_FOR_BACKUP_RETRIES )) && {
```
martinflorian-da, line 155: Did this time out too?

OriolMunoz-da: no, but I'd rather mirror the behavior between both

cluster/scripts/node-backup.sh
```diff
@@ -54,7 +55,10 @@ function wait_for_pvc_backup() {
       _info "Backup of $description PVC ready!"
       break
     else
-      (( i++ )) && (( i > 300 )) && _error "Timed out waiting for backup of $description PVC"
+      (( i++ )) && (( i > WAIT_FOR_BACKUP_RETRIES )) && {
+        kubectl delete volumesnapshot -n "$namespace" "$backupName";
```
martinflorian-da: How do we know we're allowed to `delete` here at all? If an operation is running then this will just fail too, won't it? But I guess that's fine because then we'll still retry?

martinflorian-da: > then we'll still retry

Hm or will this script just die, because we're not capturing the error code of the `kubectl delete`?

OriolMunoz-da: hopefully it works and it gets deleted, but otherwise you're correct, it will `exit 1` and retry

martinflorian-da: Hm and assuming the `delete` never works... will we loop forever here?

OriolMunoz-da: `_error` has an `exit 1` inside, so either kubectl delete fails or _error exits

cluster/scripts/node-backup.sh
martinflorian-da: I like the consistency but sentences that contain both `delete` and `backup` always make scared and look more closely...

cluster/scripts/node-backup.sh
martinflorian-da: I guess it does... so fine then, not veto-ing if you feel confident about this.

PR #3690, opened 2026-01-26
make cluster config access outside of Pulumi use resolved config

mblaze-da:
These changes are an attempt to address the problem with accessing cluster configuration outside of Pulumi projects as described in https://github.com/DACH-NY/canton-network-internal/issues/3270.

### DIFF
--- a/build-tools/cncluster
+++ b/build-tools/cncluster
@@ -1469,7 +1469,7 @@ subcommand_whitelist[debug_shell]='Spin up a debug container on the cluster, and
 function subcmd_debug_shell() {
   _cluster_must_exist
 
-  VERSION_NUMBER=${OVERRIDE_VERSION:-$(yq ".synchronizerMigration.active.version" < config.yaml)}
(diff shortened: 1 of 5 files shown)


build-tools/cncluster
```diff
@@ -1469,7 +1469,7 @@ subcommand_whitelist[debug_shell]='Spin up a debug container on the cluster, and
 function subcmd_debug_shell() {
   _cluster_must_exist
 
-  VERSION_NUMBER=${OVERRIDE_VERSION:-$(yq ".synchronizerMigration.active.version" < config.yaml)}
+  VERSION_NUMBER=${OVERRIDE_VERSION:-$("${SPLICE_ROOT}/cluster/scripts/get-resolved-config.sh" | yq ".synchronizerMigration.active.version")}
```
martinflorian-da, line 1472: This has the side affect of updating the resolved config... but I guess that is fine enough.

PR #3717, opened 2026-01-28

canton-network-da:


mblaze-da:
> @mblaze-da Is it safe to assume that this PR should be ignored by anyone but you?

@martinflorian-da yes, sorry.

PR #3529, opened 2026-01-13
Restrict amulet configs to not allow fees

moritzkiefer-da:
[ci]

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #3720, opened 2026-01-28
Don't deduct holding fees when deciding what to merge

martinflorian-da:
[ci]

This should be the main fix for https://github.com/DACH-NY/cn-test-failures/issues/6669

I'll keep the issue open to track:

- confirmation via dedicated test (hope to add that soon after this PR)
- mentioning fix in release notes (only if I manage to confirm via test before it gets packaged)
- confirmation at customer (in a few weeks once this hits MainNet)

Going to leave the bigger cleanups to #2257 (I started a bit with that but turns out there is a lot to clean up...)

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
@@ -18,7 +18,7 @@ import scala.concurrent.Future
 
 trait TransferInputStore extends AppStore with LimitHelpers {
 
-  /** List all nonzero amulets owned by a user in descending order */
+  /** List all amulets owned by a user in descending order */
   // TODO(#2257): we don't really need the round here anymore
   def listSortedAmuletsAndQuantity(
       submittingRound: Long,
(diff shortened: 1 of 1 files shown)


martinflorian-da:
> > Going to leave the bigger cleanups to #2257 (I started a bit with that but turns out there is a lot to clean up...)
> 
> Note that we already have a full cleanup in #3529. I just can't merge because I need the CIP to go through first …

Oh nice!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
```diff
@@ -30,10 +31,11 @@ trait TransferInputStore extends AppStore with LimitHelpers {
     .map(c =>
       (
         SpliceUtil
-          .currentAmount(c.payload, submittingRound),
+          .currentAmount(c.payload, submittingRound, false),
```
moritzkiefer-da:
```suggestion
          .currentAmount(c.payload, submittingRound, deductHoldingFees = false),
```
noone understands booleans

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
```diff
@@ -30,10 +31,11 @@ trait TransferInputStore extends AppStore with LimitHelpers {
...
         c,
       )
     )
+    // TODO(#2257): consider making this filter an assertion with critical failure instead; amulet should always be > 0
```
moritzkiefer-da: I'd honestly just delete it instead. But happy to not do this in this PR

martinflorian-da: if you think deleting is fine enough then let's just do it here...

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
martinflorian-da: neither do I :)

martinflorian-da: fixing

PR #3730, opened 2026-01-28
Namespace NPM packages

pawelperek-da:
Wraps `common-typeface-lato`, `common-typeface-termina` and `splitwell-frontend` packages with `@lfdecentralizedtrust` namespace and adds static checks

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


scripts/check-npm-package-names.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
ray-roestenburg-da: 2026

scripts/check-npm-package-names.sh
```diff
@@ -0,0 +1,54 @@
...
+# SPDX-License-Identifier: Apache-2.0
```
ray-roestenburg-da: can you do this without bash (cause I hate bash) did you test this on linux and mac 😄 and on  various shells like fish and zsh

scripts/check-npm-package-names.sh
isegall-da: Nope, that doesn't work, we're stuck in 2024. See https://github.com/hyperledger-labs/splice/pull/3677

scripts/check-npm-package-names.sh
isegall-da: This does ` /usr/bin/env bash` so guaranteed to be bash, not fish or zsh.

Also I'm mot sure we need mac, this runs in ci which is linux.

Having said that, we did agree to try to reduce bash usage...

pawelperek-da: I checked the script on macos and it seems to behave the same as on linux, I can change it to some more maintainable language though

ray-roestenburg-da: yeah python please

ray-roestenburg-da: Apologies for my trolling review, I've just had too many of these scripts not work locally for some reason (and I would appreciate if scripts just work everywhere), and I hate bash.

pawelperek-da: Oh trust me, I like me a good sysadmining and I find bash insane to write in. Modern replacements like `nushell` are mildly better but actual programming languages are miles ahead so I if we can do that I'm all for it. I just wanted to blend in with the other CI scripts we have 😅

Issue #2456, opened 2025-09-29
Incremental update of ACS snapshot in AcsSnapshotTrigger


ray-roestenburg-da:
Compute and persist the ACS snapshot continuously, mutating the table, such that we only store the most recent ACS snapshot in the table.
Subscribe to the update stream, based on the creates and archives you see mutate this table, and every 3h, copy that into the acs snapshot table.

moritzkiefer-da:
> You cannot do this for backfilling migration ids as there you're not subscribed to any update stream

yep I think for older migration the queries seem fine. If you want to do better in theory you could wire up backfilling to snapshots somehow. I think that may make sense eventually but I would treat this separately.

> You might be able to skip the copy step if you just put everything in the acs_snapshot_data table directly, but avoid serving it. Namely:

ah right that might be nicer.

OriolMunoz-da:
> yep I think for older migration the queries seem fine.

Is it because we know that eventually :tm: they won't be executed anymore, or why is it acceptable for those?

PR #3733, opened 2026-01-29
Implement incremental acs snapshots

rautenrieth-da:
Fixes https://github.com/hyperledger-labs/splice/issues/2456

### DIFF
--- a/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerTest.scala
+++ b/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerTest.scala
@@ -653,6 +653,42 @@ class AcsSnapshotTriggerTest
         )
         .futureValue shouldBe RetrieveTaskForBackfillingMigrationResult.Done
     }
+
+    "use the backfillingMigrationIdO value" in {
+      // A snapshot with recordTime < targetRecordTime
+      val incrementalSnapshot = IncrementalAcsSnapshot(
+        snapshotId = 1L,
+        historyId = 1L,
+        tableName = AcsSnapshotStore.IncrementalAcsSnapshotTable.Backfill.tableName,
+        recordTime = snapshotTime2.minusSeconds(100L),
+        migrationId = migration1,
+        targetRecordTime = snapshotTime2,
+      )
+      val recordTimeRange = DomainRecordTimeRange(snapshotTime1, snapshotTime3)
+
+      // Update the incremental snapshot with 30sec worth of data.
+      AcsSnapshotBackfillingTrigger
+        .retrieveTaskForBackfillingMigration(
+          backfillingMigrationIdO = Some(migration1),
+          currentMigrationId = migration4,
+          isHistoryBackfilled = returnForMigration(migration1 -> true),
+          getIncrementalSnapshot = () => Future.successful(Some(incrementalSnapshot)),
+          getLatestSnapshot = unused1,
+          getMinRecordTime = returnForMigration(migration1 -> Some(recordTimeRange.min)),
+          getMaxRecordTime = returnForMigration(migration1 -> Some(recordTimeRange.max)),
+          getPreviousMigrationId = unused1,
+          storageConfig = storageConfig,
+          updateInterval = java.time.Duration.ofSeconds(30L),
+          logger = logger,
+        )
+        .futureValue shouldBe RetrieveTaskForBackfillingMigrationResult.Task(
+        migration1,
+        AcsSnapshotTriggerBase.UpdateIncrementalSnapshotTask(
+          snapshot = incrementalSnapshot,
+          updateUntil = incrementalSnapshot.recordTime.plusSeconds(30L),
+        ),
+      )
+    }
   }
 
   private def storageConfig = ScanStorageConfig(


apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V054__incremental_acs_snapshots.sql
```diff
@@ -0,0 +1,75 @@
...
+    add constraint acs_incremental_snapshot_data_next_pk
+       primary key (history_id, contract_id);
+
+-- Needed because ACS snapshots are ordered by creation time
+create index acs_incremental_snapshot_data_next_ca_ci
+    on acs_incremental_snapshot_data_next (history_id, created_at, contract_id);
+
+-- Incremental ACS snapshot for backfilling
+create table acs_incremental_snapshot_data_backfill
+(
```
OriolMunoz-da: nit: does it make sense to define this as `create table acs_incremental_snapshot_data_backfill like acs_incremental_snapshot_data_next`? they're the same after all

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -406,13 +410,420 @@ class AcsSnapshotStore(
...
+      from acs_incremental_snapshot
+      where history_id = $historyId
+        and table_name = ${table.tableName}
+    """.as[IncrementalAcsSnapshot].headOption
+  }
+
+  /** The state of an incremental snapshot.
+    *
+    * @param table The table name of the incremental snapshot.
+    *              MUST be a compile time constant, as it is included directly in the SQL statement.
```
OriolMunoz-da, line 441: this is not true here

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1859,6 +1859,27 @@ class UpdateHistory(
     }
   }
 
+  /** Returns the record time of the first non-import update (if any) on the given migration. */
+  def getMinUpdateRecordTime(
```
OriolMunoz-da: this is unused

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
```diff
@@ -0,0 +1,313 @@
...
+                // No updates exist for the current migration, so nothing to do.
+                logger.info(
+                  s"No updates other than ACS imports found. Retrying snapshot creation later."
+                )
+                None
+            }
+        }
+      case Some(snapshot) =>
+        if (snapshot.migrationId != migrationId) {
+          // Incremental snapshot contains data from a wrong migration, delete it and start over.
```
OriolMunoz-da: "wrong migration" is necessarily the previous migration after a HDM (or next, in the case of rollbacks)?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
```diff
@@ -0,0 +1,313 @@
...
+          Future.successful(
+            Some(AcsSnapshotTriggerBase.DeleteIncrementalSnapshotTask(snapshot))
+          )
+        } else {
+          assert(!snapshot.recordTime.isAfter(snapshot.targetRecordTime))
+          if (snapshot.recordTime == snapshot.targetRecordTime) {
+            // Incremental snapshot is complete, copy it to historical storage
```
OriolMunoz-da: I cannot tell from what I've reviewed so far, how the recordTime increments to become equal to targetRecordTime, so I'd suggest documenting (or link to somewhere) here

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,178 @@
...
+
+  private def retrieveTaskForPastMigrationId(migrationIdToBackfill: Long)(implicit
+      tc: TraceContext
+  ): Future[Option[AcsSnapshotTriggerBase.Task]] = {
+    retrieveTaskForMigration(
+      migrationId = migrationIdToBackfill,
+      historyIngestedUntil = Some(CantonTimestamp.MaxValue),
+    ).flatMap {
+      // `retrieveTaskForMigration()` will never stop creating snapshots for the given migration id.
+      // For backfilling however, we need to check when we have reached the end of the current migration.
```
OriolMunoz-da:
```suggestion
      // For backfilling however, we need to check when we have reached the end of the currently backfilling migration.
```
I interpret the "current" one is the "live" one, which is not what you mean here I think

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,178 @@
...
+      // We do this by aborting any task that would start work on a snapshot with a target record time
```
OriolMunoz-da: nit
```suggestion
      // We do this by skipping any task that would start work on a snapshot with a target record time
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -406,13 +410,420 @@ class AcsSnapshotStore(
...
+    *                          to a historical snapshot.
+    */
+  def initializeIncrementalSnapshot(
+      table: IncrementalAcsSnapshotTable,
+      initializeFrom: AcsSnapshot,
+      targetRecordTime: CantonTimestamp,
+  )(implicit tc: TraceContext): Future[Unit] = {
+    assert(targetRecordTime.isAfter(initializeFrom.snapshotRecordTime))
+    val statement = for {
+      snapshotId <- sql"""
```
OriolMunoz-da, line 499: snapshotId is unused

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -406,13 +410,420 @@ class AcsSnapshotStore(
...
+      history_id,
+      create_id,
+      contract_id,
+      created_at,
+      unlocked_amulet_balance,
+      locked_amulet_balance,
+      template_id,
+      stakeholders
+    """
+    val copyFromUpdateHistorySourceColumns: SQLActionBuilder =
```
OriolMunoz-da, line 851: it looks like the amulet balances are 0 for any non-amulet row, can you add a comment documenting that?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -406,13 +410,420 @@ class AcsSnapshotStore(
...
+    storage.queryAndUpdate(
+      withIncrementalSnapshotIdempotencyCheck(table, statement, Some(snapshot)),
+      "deleteIncrementalSnapshot",
+    )
+  }
 }
 
 object AcsSnapshotStore {
 
+  final case class IncrementalAcsSnapshotTable(val tableName: String)
```
OriolMunoz-da: if you make the constructor private, you won't need to require the string to be a constant, it will be enforced by the fact that table names can only be created in the companion object

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -406,13 +410,420 @@ class AcsSnapshotStore(
...
+      nextSnapshotTargetRecordTime: CantonTimestamp,
+  )(implicit tc: TraceContext): Future[Unit] = {
+    logger.debug(
+      s"Saving incremental snapshot ${snapshot.snapshotId} at ${snapshot.recordTime}"
+    )
+    assert(snapshot.tableName == table.tableName)
+    assert(snapshot.historyId == historyId)
+    assert(snapshot.recordTime == snapshot.targetRecordTime)
+    val statement = for {
+      // Note: Only one client can write to acs_snapshot_data at a time, enforced via advisory locks.
```
OriolMunoz-da, line 626: I might be missing something, but `withExclusiveSnapshotDataLock`  is only called in old code
So I am not sure how you prevent AcsSnapshotTrigger and AcsSnapshotBackfillingTrigger from inserting into acs_snapshot_data at the same time

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -406,13 +410,420 @@ class AcsSnapshotStore(
...
+        where row_id > $max_row_id_before
+      """.as[(Long, Long)].head
+
+      (unlocked_amulet_balance, locked_amulet_balance) <- sql"""
+        select
+            sum(s.unlocked_amulet_balance) AS unlocked_amulet_balance,
+            sum(s.locked_amulet_balance) AS locked_amulet_balance
+        from #${table.tableName} s
+        where history_id = $historyId
+      """.as[(BigDecimal, BigDecimal)].head
```
OriolMunoz-da, line 657: remind me again, why can't this (min, max, amulet balance) be done with a returning on the insert above?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -406,13 +410,420 @@ class AcsSnapshotStore(
...
+          delete
+          from #${table.tableName} as s
+          using update_history_exercises as e
+          where s.contract_id = e.contract_id
+            and e.history_id = $historyId
+            and s.history_id = $historyId
+            and migration_id = ${snapshot.migrationId}
+            and e.record_time > ${snapshot.recordTime}
+            and e.record_time <= $targetRecordTime
+            and e.consuming
```
OriolMunoz-da:
```suggestion
          where s.history_id = $historyId
            and s.contract_id = e.contract_id
            and e.history_id = $historyId
            and e.migration_id = ${snapshot.migrationId}
            and e.record_time > ${snapshot.recordTime}
            and e.record_time <= $targetRecordTime
            and e.consuming
```
ordered them as the indexes are, and added `e.` to disambiguate

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerTest.scala
```diff
@@ -37,536 +43,399 @@ class AcsSnapshotTriggerTest
...
-          Future.successful(
-            Seq(
-              TreeUpdateWithMigrationId(
-                UpdateHistoryResponse(treeUpdate(now.plusSeconds(1800L)), dummyDomain),
-                1L,
-              )
-            )
-          )
+    "updating" should {
+      "update an incremental snapshot that is old" in new AcsSnapshotTriggerTestScope() {
```
OriolMunoz-da: it's unclear to me what "old" means here

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
OriolMunoz-da: see suggestion below

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
rautenrieth-da: Whenever we switch migrations. For the forward direction (AcsSnapshotTrigger) it's after a HDM. For the backward direction (AcsSnapshotBackfillingTrigger) it's when we're done backfilling one migration and switch to the next.

Reminder to self: add tests for the forward direction in AcsSnapshotTriggerTest.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
rautenrieth-da: A few lines below:
```scala
val updateUntil = snapshot.recordTime
  .plus(context.config.pollingInterval.asJava)
  .min(snapshot.targetRecordTime) // <-- see here
```

I'll improve the comments.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
rautenrieth-da: Good catch, I completely missed it!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
rautenrieth-da: min/max is calculated from the destination table (acs_snapshot_data), amulet balance is calculated from the source table (acs_incremental_snapshot_data_next). I couldn't find a way to do it in one statement.

Also, both queries complete instantly (<100ms), while the copying takes 30min or so (note that I didn't remove any unused indexes in this PR).  I'd rather have separate statements that are easy to understand for us and the query planner.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerTest.scala
rautenrieth-da:
```suggestion
      "update an incremental snapshot that is not near its target time" in new AcsSnapshotTriggerTestScope() {
```

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__incremental_acs_snapshots.sql
```diff
@@ -0,0 +1,75 @@
+-- Incremental ACS snapshot for ingestion
+create table acs_incremental_snapshot_data_next
```
ray-roestenburg-da, line 27:
```suggestion
create table next_acs_incremental_snapshot_data
```

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__incremental_acs_snapshots.sql
```diff
@@ -0,0 +1,75 @@
+-- Incremental ACS snapshot for ingestion
```
ray-roestenburg-da, line 23: maybe add comment here about what next means (it's a scratch / working space for computing an ACS snapshot)

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__incremental_acs_snapshots.sql
ray-roestenburg-da: Or maybe a better name for the temporary nature of this table, if I understand that correctly

ray-roestenburg-da: ah wait, it's a naming for 'one of two tables', next or backfil. ok this is fine

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V054__incremental_acs_snapshots.sql
```diff
@@ -0,0 +1,75 @@
...
+-- State of the incremental ACS snapshots
+create table acs_incremental_snapshot
+(
+    snapshot_id                     bigint generated by default as identity primary key,
+    history_id                      bigint not null references update_history_descriptors (id),
+
+    -- We maintain two incremental ACS snapshots in parallel, one for the ingestion
+    -- and one for backfilling. Each incremental ACS snapshot gets its own table.
+    -- This column indicates which table this snapshot is for.
+    table_name                      text   not null,
```
ray-roestenburg-da: Is this table_name value used in SQL? or could you just as well had a boolean or short field for next or backfil? (I was thinking of SQL injection but maybe far fetched)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
ray-roestenburg-da: no need for `val` before tableName

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
```diff
@@ -0,0 +1,313 @@
...
+            // First, make sure we don't process too much. The target time for the update is computed
+            // from the state of the incremental snapshot (and not from the current time).
+            val updateUntil = snapshot.recordTime
+              .plus(context.config.pollingInterval.asJava)
+              .min(snapshot.targetRecordTime)
+
+            val now = context.clock.now
+
+            // Next, make sure we don't process too little.
+            if (now.isBefore(updateUntil)) {
```
ray-roestenburg-da: Nice to have: a test checking this logic, or extract this in a separate method to easy unit test

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V054__incremental_acs_snapshots.sql
```diff
@@ -0,0 +1,75 @@
+-- Incremental ACS snapshot for ingestion
+create table acs_incremental_snapshot_data_next
+(
+    -- In production, we have a separate table for each scan instance so this
+    -- column will be the same for all rows. However, in tests we can have multiple
+    -- scan instances writing to the same table, so we need to include it.
+    history_id   bigint      not null references update_history_descriptors (id),
```
ray-roestenburg-da: This table and the backfil one should have a foreign key to the acs_incremental_snapshot table, as discussed

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerTest.scala
```diff
@@ -37,536 +43,399 @@ class AcsSnapshotTriggerTest
...
+        val snapshot = IncrementalAcsSnapshot(
+          snapshotId = 1L,
+          historyId = 1L,
+          tableName = AcsSnapshotStore.IncrementalAcsSnapshotTable.Next.tableName,
+          recordTime = now.minusSeconds(1801L),
+          migrationId = currentMigrationId,
+          targetRecordTime = now.minusSeconds(1800L),
+        )
+        previousIncrementalSnapshot(snapshot)
+        updatesBetween(currentMigrationId, now.minusSeconds(6000L), now.minusSeconds(1L))
```
ray-roestenburg-da: maybe some more constants in these tests as we talked about, if that makes it easier to understand.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1872,6 +1872,46 @@ class UpdateHistory(
...
+        select min(record_time) as min_record_time, max(record_time) as max_record_time
+        from #$table
+        where history_id = $historyId
+        and migration_id = $migrationId
+        and record_time > ${CantonTimestamp.MinValue}
+      """
+    }
+    storage
+      .query(
+        (sql"""
```
ray-roestenburg-da, line 1893: Don't remember, do you have a query plan or this?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,215 @@
...
+
+  private val isDone: AtomicBoolean = new AtomicBoolean(false)
+
+  private[automation] def isDoneBackfillingAcsSnapshots =
+    isDone.get()
+
+  override def retrieveTasks()(implicit
+      tc: TraceContext
+  ): Future[Seq[AcsSnapshotTriggerBase.Task]] = {
+    if (isDone.get()) {
```
ray-roestenburg-da: isDoneBackfillingAcsSnapshots?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,215 @@
...
+          case Right(Some(task)) =>
+            Future.successful(Seq(task))
+        }
+    }
+  }
+
+}
+
+object AcsSnapshotBackfillingTrigger {
+  def retrieveTaskForBackfillingMigration(
```
ray-roestenburg-da, line 104: nice

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,215 @@
...
+          migrationIdToBackfillO <- getPreviousMigrationId(earliestKnownBackfilledMigrationId)
+          task <- migrationIdToBackfillO match {
+            case Some(migrationIdToBackfill) =>
+              // There is a migration before the earliest known backfilled migration,
+              // check if we can start backfilling on it.
+              logger.debug(
+                s"Earliest backfilled migration is $earliestKnownBackfilledMigrationId, looking tasks in migration $migrationIdToBackfill."
+              )
+              for {
+                proposedTaskO <- AcsSnapshotTriggerBase.retrieveTaskForMigration(
```
ray-roestenburg-da: Minor naming issue: reads slightly odd that a method is called on an object called ...Base, maybe this function can be moved to AcsSnapshotTrigger object?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTrigger.scala
```diff
@@ -3,286 +3,120 @@
...
+  private def taskBeyondUpdateHistoryEnd(
+      task: AcsSnapshotTriggerBase.Task,
+      lastIngestedRecordTimeO: Option[CantonTimestamp],
+  ): Boolean = {
+    task match {
+      case UpdateIncrementalSnapshotTask(_, updateUntil) =>
+        // Don't move beyond the last ingested record time, otherwise we'll miss updates
+        // between `lastIngestedRecordTime` and `updateUntil`.
+        lastIngestedRecordTimeO.forall(updateUntil.isAfter)
+      case InitializeIncrementalSnapshotFromImportUpdatesTask(_, _, _) =>
```
ray-roestenburg-da: case _ => false ? (anything other than UpdateIncrementalSnapshotTask)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
rautenrieth-da: `AcsSnapshotTrigger` and `AcsSnapshotBackfillingTrigger` are independent triggers, all common functionality is implemented in their common ancestor `AcsSnapshotTriggerBase`. Wouldn't it be more odd to use a method from a sibling class?

ray-roestenburg-da: There's nothing really "base" about the object is what I meant, the object is not an ancestor. They are both acs snapshot triggers, I would not call them independent at least not in how the code is written,  AcsSnapshotTrigger.Task would have made more sense to me in that sense, but then you would have to rename the AcsSnapshotTrigger to something else, which is maybe more annoying.. anyway, not a big issue

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,215 @@
...
+      case Some(incrementalSnapshot) =>
+        getRecordTimeRange(incrementalSnapshot.migrationId).flatMap {
+          case None =>
+            logger.error(s"Incremental snapshot without update range: $incrementalSnapshot")
+            Future.successful(Right(None))
+          case Some(range) =>
+            if (incrementalSnapshot.targetRecordTime.isAfter(range.max)) {
+              // Incremental snapshot exists, but is beyond the end of the migration.
+              // Delete it, and let the next iteration pick up the previous migration to start backfilling on.
+              Future.successful(
```
OriolMunoz-da: I could use a comment as to _when_ this can happen

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
```diff
@@ -0,0 +1,295 @@
...
+                    )
+                  case None =>
+                    // No updates exist for the current migration, so nothing to do.
+                    logger.info(
+                      s"No updates other than ACS imports found. Retrying snapshot creation later."
+                    )
+                    None
+                }
+            }
+          case Some(snapshot) =>
```
OriolMunoz-da:
```suggestion
          case Some(incrementalSnapshot) =>
```

makes my reviewing life easier

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
```diff
@@ -0,0 +1,295 @@
...
+            if (snapshot.migrationId != migrationId) {
+              // Incremental snapshot contains data from a wrong migration, delete it and start over.
+              Future.successful(
+                Some(AcsSnapshotTriggerBase.DeleteIncrementalSnapshotTask(snapshot))
+              )
+            } else {
+              // Note: the code below makes sure that `recordTime` never moves past `targetRecordTime`.
+              assert(!snapshot.recordTime.isAfter(snapshot.targetRecordTime))
+              if (snapshot.recordTime == snapshot.targetRecordTime) {
```
OriolMunoz-da: (thinking out loud to make sure I'm getting this right)
snapshot.recordTime gets incremented in the branch below (copy-pasting):

```
                // Incremental snapshot is not yet complete.
                // The intent is that we process around 30sec (the update interval) of updates per iteration.
                // The target time for the update is therefore computed
                // from the state of the incremental snapshot (and not from the current time).
                val updateUntil = snapshot.recordTime
                  .plus(updateInterval)
                  .min(snapshot.targetRecordTime) // Don't move past the target record time.
```

which sets it to every 30s
and then it's stored in store.updateIncrementalSnapshot
The `.min(snapshot.targetRecordTime) // Don't move past the target record time.` will make sure that `snapshot.recordTime == snapshot.targetRecordTime` at some point (and never more)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
```diff
@@ -0,0 +1,295 @@
...
+                  )
+                )
+              } else {
+                // Incremental snapshot is not yet complete.
+                // The intent is that we process around 30sec (the update interval) of updates per iteration.
+                // The target time for the update is therefore computed
+                // from the state of the incremental snapshot (and not from the current time).
+                val updateUntil = snapshot.recordTime
+                  .plus(updateInterval)
+                  .min(snapshot.targetRecordTime) // Don't move past the target record time.
```
OriolMunoz-da: there's this requirement:
```
    * checking that the returned task does not move beyond the end of the migration:
    * - AcsSnapshotTrigger should check that the snapshot does not move beyond the last ingested record time.
    * - AcsSnapshotBackfillingTrigger should check that the snapshot does not move beyond the end of the migration.
```
does it make sense to apply it here, by passing the limit of each trigger also as argument?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,215 @@
...
+          storageConfig = storageConfig,
+          updateInterval = context.config.pollingInterval.asJava,
+          logger: TracedLogger,
+        )
+        .flatMap {
+          case Left(Done) =>
+            logger.info(
+              s"Backfilling incremental ACS snapshots is complete, this trigger won't do any work again."
+            )
+            isDone.set(true)
```
ray-roestenburg-da: So it is ok for concurrent execution of retrieveTasks? hypothetically many calls can enter isDone.get() and get false, and then here it's set to true possibly many times (especially if retrieveTaskForBackfillingMigration takes long).

ray-roestenburg-da: Would you want a call to retrieveTasks to do something if another call to retrieveTasks is still busy?

ray-roestenburg-da: or is this taking advantage of the fact that this kind of trigger does only one thing at a time, and we don't have to worry about it... that makes me wonder about why this isDone is an atomic, so I'm a little confused, which is probably my problem :-) (oh wait no, this is a PollingParallelTaskExecutionTrigger)

rautenrieth-da: `retrieveTasks` will never be called concurrently due to the way a PollingParallelTaskExecutionTrigger works. I could have used a `val` instead of an `AtomicReference`, the latter just avoids a `@SuppressWarnings` and `@volatile`.

ray-roestenburg-da: I think if we are that sure, I would rather have a private var and a suppresswarning and no volatile? by default when I see an atomicref I try to find concurrency bugs (so yeah I'm probably the problem)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,215 @@
...
+  SaveIncrementalSnapshotTask,
+  UpdateIncrementalSnapshotTask,
+}
+import org.lfdecentralizedtrust.splice.scan.config.ScanStorageConfig
+import org.lfdecentralizedtrust.splice.util.DomainRecordTimeRange
+
+import java.util.concurrent.atomic.AtomicBoolean
+import scala.concurrent.{ExecutionContext, Future}
+
+class AcsSnapshotBackfillingTrigger(
```
ray-roestenburg-da, line 27: Minor:
```suggestion
class IncrementalAcsSnapshotBackfillingTrigger(
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTrigger.scala
```diff
@@ -3,286 +3,120 @@
...
+  SaveIncrementalSnapshotTask,
+  UpdateIncrementalSnapshotTask,
+}
 import org.lfdecentralizedtrust.splice.scan.config.ScanStorageConfig
+import org.lfdecentralizedtrust.splice.util.DomainRecordTimeRange
 
-import java.util.concurrent.atomic.AtomicReference
 import scala.concurrent.{ExecutionContext, Future}
 
 class AcsSnapshotTrigger(
```
ray-roestenburg-da, line 26: `IncrementalAcsSnapshotTrigger` ?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
```diff
@@ -0,0 +1,295 @@
...
+
+abstract class AcsSnapshotTriggerBase(
+    store: AcsSnapshotStore,
+    updateHistory: UpdateHistory,
+    protected val context: TriggerContext,
+)(implicit
+    ec: ExecutionContext,
+    tracer: Tracer,
+    mat: Materializer,
+    // we always return 1 task, so PollingParallelTaskExecutionTrigger in effect does nothing in parallel
```
ray-roestenburg-da, line 41: Where is this guaranteed? in retrieveTasks? so it's more of a rule for extending this class? Because this class does not define it, so this statement feels incorrect here, it depends on the subclasses right?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
```diff
@@ -0,0 +1,295 @@
...
+) extends PollingParallelTaskExecutionTrigger[AcsSnapshotTriggerBase.Task] {
```
ray-roestenburg-da, line 42: Also, can we extend from another Trigger than the `PollingParallel...` so it is clear it is not parallel? (maybe define a PollingSequentialTaskExecutionTrigger than ensures it's never parallel)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
ray-roestenburg-da: super nitpick:

if(isDone.get() || !updateHistory.isReady) {
  Future.successful(Seq.empty)
} else {
  ...
}

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,215 @@
...
+      logger: TracedLogger,
+  )(implicit
+      tc: TraceContext,
+      ec: ExecutionContext,
+  ): Future[Either[Done, Option[AcsSnapshotTriggerBase.Task]]] = {
+    getIncrementalSnapshot().flatMap {
+      case Some(incrementalSnapshot) =>
+        getRecordTimeRange(incrementalSnapshot.migrationId).flatMap {
+          case None =>
+            logger.error(s"Incremental snapshot without update range: $incrementalSnapshot")
```
ray-roestenburg-da: Does it make sense to also log the migrationId for which there is no update data?

ray-roestenburg-da: oh wait that's in the incrementalSnapshot, never mind

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
ray-roestenburg-da: Agree, when does it happen that "Incremental snapshot exists, but is beyond the end of the migration."

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,215 @@
...
+            Future.successful(Right(None))
+          case Some(range) =>
+            if (incrementalSnapshot.targetRecordTime.isAfter(range.max)) {
+              // Incremental snapshot exists, but is beyond the end of the migration.
+              // Delete it, and let the next iteration pick up the previous migration to start backfilling on.
+              Future.successful(
+                Right(
```
ray-roestenburg-da: weird that this does not fit on one line

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,215 @@
...
+                  updateInterval = updateInterval,
+                  getIncrementalSnapshot = getIncrementalSnapshot,
+                  getLatestSnapshot = getLatestSnapshot,
+                  getRecordTimeRange = getRecordTimeRange,
+                  logger = logger,
+                )
+                rangeO <- getRecordTimeRange(migrationIdToBackfill)
+                actualTask <- (proposedTaskO, rangeO) match {
+                  case (Some(proposedTask), Some(range))
+                      if taskBeyondMigrationEnd(proposedTask, range.max) =>
```
ray-roestenburg-da: strange why this is not behind / part of the case (formatting)? especially if you look at the long line after it

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,215 @@
...
+      case None =>
+        // There is no incremental snapshot. Find the right place to start backfilling.
+        for {
+          migrationIdToBackfillO <- getPreviousMigrationId(earliestKnownBackfilledMigrationId)
+          task <- migrationIdToBackfillO match {
+            case Some(migrationIdToBackfill) =>
+              // There is a migration before the earliest known backfilled migration,
+              // check if we can start backfilling on it.
+              logger.debug(
+                s"Earliest backfilled migration is $earliestKnownBackfilledMigrationId, looking tasks in migration $migrationIdToBackfill."
```
ray-roestenburg-da:
```suggestion
                s"Earliest backfilled migration is $earliestKnownBackfilledMigrationId, looking for tasks in migration $migrationIdToBackfill."
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
```diff
@@ -0,0 +1,215 @@
...
+                  migrationId = migrationIdToBackfill,
+                  isHistoryBackfilled = isHistoryBackfilled,
+                  storageConfig = storageConfig,
+                  updateInterval = updateInterval,
+                  getIncrementalSnapshot = getIncrementalSnapshot,
+                  getLatestSnapshot = getLatestSnapshot,
+                  getRecordTimeRange = getRecordTimeRange,
+                  logger = logger,
+                )
+                rangeO <- getRecordTimeRange(migrationIdToBackfill)
```
ray-roestenburg-da: would it make sense to getRecordTimeRange first, and if there is some range, then retrieveTaskForMigration? I don't completely get why you have a proposedTask, and then sometimes not use it

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
```diff
@@ -0,0 +1,295 @@
...
+  }
+}
+
+object AcsSnapshotTriggerBase {
+
+  /** Determines the next task to perform in order to update the incremental snapshot for the given migration.
+    *
+    * This method treats the given migration as having no end. The caller is responsible for
+    * checking that the returned task does not move beyond the end of the migration:
+    * - AcsSnapshotTrigger should check that the snapshot does not move beyond the last ingested record time.
```
ray-roestenburg-da: I would have put these comments in AcsSnapshotTrigger and AcsSnapshotBackfillingTrigger, this base doesn't "know" about the usages.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
ray-roestenburg-da: +1 on this

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotBackfillingTrigger.scala
rautenrieth-da: That's logged as part of the IncrementalSnapshot case class.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
rautenrieth-da: Implemented this suggestion

PR #3743, opened 2026-01-29
add config resolution to the _update_cluster_config script

mblaze-da:
fixes https://github.com/DACH-NY/cn-test-failures/issues/7065

The problem that these changes solve was that CI in some cases does indeed modify the config and later reads the field that it modified. So the assumption in `get-resolved-config.sh` the config resolution in CI can be skipped was wrong. The proposed fix extends the config modification command with immediate config resolution.


martinflorian-da:
Let's launch a ciperiodic hdm test right after merging, to confirm the fix!

PR #3751, opened 2026-01-29
grafana: Make all dashboards use the default timezone

giner:
It's just useful to have the same time zone on all dashboards so that there no need to think about it when switching from one dashboard to another while trying to correlate events.

This exercise can be repeated any time by running:
```shell
cd cluster/pulumi/infra/grafana-dashboards
find -name "*.json" | xargs -l sh -c 'jq ".timezone = \"\"" "$1" | sponge "$1"' --
```

Before the change:
```shell
find -name "*.json" | xargs jq .timezone | sort | uniq -c
     28 ""
      2 "browser"
     15 "UTC"
```

*Note*: When a new dashboard is created timezone settings are copied from default_timezone of Grafana settings (which is "browser" unless changed). Because of that for every new dashboard we create timezone should be explicitly reset either manually or by running the snippet above.


isegall-da:
@nicu-da 
Can you take a look please? All our dashboards are in UTC already, aren't they? Or did I just configure that locally for myself? If one can do the latter than pushing utc to everyone might not be the right thing. WDYT?

giner:
> if people then have other dashboards that are not utc it become a confusing mix.
a confusing mix is the current state :)

The main goal of this change is to bring consistency, so either removal or explicit default are fine. The reason why I thought of explicit UTC:
- logs are usually in UTC
- we often share screenshots with each other with no information about the zone, explicit UTC would make most of screenshots have UTC regardless of where they are coming from

Thoughts?

PR #3761, opened 2026-01-30
add config resolution to remaining hdm commands that edit the migration config

mblaze-da:
add config resolution to remaining hdm commands that edit the migration config


moritzkiefer-da (approved):
thx didn't check if there are others that are missing but I guess we'll learn soon enough :)

PR #3773, opened 2026-01-31
grafana: Add Grafana dashboards check to pre-commit

giner:
This check verifies Grafana dashboards and updates them if necessary.
There is only one check at this point:
- timezone is set to default

More checks can be added later if needed.

The check runs from:
- pre-commit hook for changed dashboard files
- CI for all dashboard files


isegall-da (approved):
Nice, thank you

PR #3791, opened 2026-02-02
Bumped confirmation rate to 25

pasindutennage-da:


nicu-da:
shouldn't this just be the removal after we bump splice?

PR #3805, opened 2026-02-03
Add random_page_cost flag to cloudsql in pulumi

OriolMunoz-da:
This is already manually applied to mainnet (where it matters), so I'd say we don't need to expedite it's deployment.

Two notes:
- It's now the default for all apps (not just sequencer). Shout if you disagree
- It's made configurable just in case things break, though I find that unlikely

### DIFF
--- a/cluster/pulumi/common/src/config/cloudSql.ts
+++ b/cluster/pulumi/common/src/config/cloudSql.ts
@@ -14,7 +14,10 @@ export const CloudSqlConfigSchema = z.object({
   protected: z.boolean(),
   tier: z.string(),
   enterprisePlus: z.boolean(),
-  flags: z.record(z.string()).default({}),
+  flags: z.record(z.string()).default({
+    random_page_cost: '1.1',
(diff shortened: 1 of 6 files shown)


cluster/pulumi/common/src/config/cloudSql.ts
```diff
@@ -14,6 +14,7 @@ export const CloudSqlConfigSchema = z.object({
   protected: z.boolean(),
   tier: z.string(),
   enterprisePlus: z.boolean(),
+  flags: z.record(z.string()).default({}),
```
moritzkiefer-da: feels a bit confusing that we allow arbitrary flags but then we ignore all but random_page_cot. I think I would have expected that either they all get applied in `databaseFlags` or we limit it to the one we care about. I think I like the generic option of applying all flags. Seems like nice extensibility if we need it again.

moritzkiefer-da: and then also move the default in here instead of in `postgres.ts`

PR #3806, opened 2026-02-03
Persist external transaction hash to db

jagathweerasinghe-da:
[static]

Signed-off-by: Jagath Weerasinghe <jagath.weerasinghe@digitalasset.com>

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UpdateHistoryIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UpdateHistoryIntegrationTest.scala
@@ -51,7 +51,8 @@ class UpdateHistoryIntegrationTest
       onboardExternalParty(aliceValidatorBackend, Some("externalParty1"))
     val ledgerBeginOffsetSv1 = sv1Backend.participantClient.ledger_api.state.end()
     val ledgerBeginOffsetAlice = aliceValidatorBackend.participantClient.ledger_api.state.end()
-    val tapAmountUsd = com.digitalasset.daml.lf.data.Numeric.assertFromString("33." + "3".repeat(10))
+    val tapAmountUsd =
+      com.digitalasset.daml.lf.data.Numeric.assertFromString("33." + "3".repeat(10))
     val transferAmount =
       com.digitalasset.daml.lf.data.Numeric.assertFromString("11." + "1".repeat(10))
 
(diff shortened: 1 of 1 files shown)


apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V056__update_history_for_hash.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE update_history_transactions ADD COLUMN external_transaction_hash bytea;
+
+-- Creating the 'updt_hist_tran_hi_eth' index on a production-sized database (~150M rows) can take ~300s.
+-- To avoid long migration waits, the index is created asynchronously by the application. The SQL is left commented for reference.
+
+-- create index updt_hist_tran_hi_eth on update_history_transactions (history_id, external_transaction_hash) where external_transaction_hash is not null;
```
ray-roestenburg-da, line 5: I would rather remove this and only point to SqlIndexInitializationTrigger where this is done

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/StoreTest.scala
```diff
@@ -1281,6 +1289,7 @@ abstract class StoreTest extends AsyncWordSpec with BaseTest {
       commandId: String = "",
       recordTime: Instant = defaultEffectiveAt,
       updateId: String = nextUpdateId(),
+      externalTransactionHash: ByteString = getRandomByteString(),
```
ray-roestenburg-da: Can you write a store test which shows that the transaction hash has been written as expected?

ray-roestenburg-da: I'm also wondering of the random hash is useful, if any test does some deep equality on Transaction (hopefully not) this could cause annoying issues?

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V056__update_history_for_hash.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE update_history_transactions ADD COLUMN external_transaction_hash bytea;
+
+-- Creating the 'updt_hist_tran_hi_eth' index on a production-sized database (~150M rows) can take ~300s.
+-- To avoid long migration waits, the index is created asynchronously by the application. The SQL is left commented for reference.
```
ray-roestenburg-da:
```suggestion
-- To avoid long migration waits, the index is created asynchronously in the SqlIndexInitializationTrigger.
```

PR #3812, opened 2026-02-03
[ci] removed small leftover

ray-roestenburg-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
@@ -84,6 +84,20 @@ import java.time.Instant
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters.*
 
+class DbScanTxLogStoreConfig(loggerFactory: NamedLoggerFactory)
+    extends TxLogStore.Config[TxLogEntry] {
+  override val parser: org.lfdecentralizedtrust.splice.scan.store.ScanTxLogParser =
+    new ScanTxLogParser(
+      loggerFactory
+    )
+  override def entryToRow: org.lfdecentralizedtrust.splice.scan.store.TxLogEntry => Option[
+    org.lfdecentralizedtrust.splice.scan.store.db.ScanTables.ScanTxLogRowData
+  ] =
+    ScanTables.ScanTxLogRowData.fromTxLogEntry
+  override def encodeEntry = TxLogEntry.encode
+  override def decodeEntry = TxLogEntry.decode
+}
+
 object DbScanStore {
   type CacheKey = java.lang.Long // caffeine metrics function demands AnyRefs
   type CacheValue = BigDecimal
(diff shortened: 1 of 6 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanTables.scala
```diff
@@ -225,9 +225,6 @@ object ScanTables extends AcsTables {
           ScanTxLogRowData(
```
moritzkiefer-da: why do we need to decode BalanceChangeTxLogEntries at all? Could this not just not produce an entryt so map it to `None` instead of whatever it is now?

ray-roestenburg-da: Just being safe in case this would somehow occur, otherwise it will trip with a `txEncodingFailed()` (RuntimeException). The code expects always to create an entry: `def fromTxLogEntry(record: TxLogEntry): ScanTxLogRowData = {` which is maybe not ideal. But of course I can remove BalanceChangeTxLogEntry from the TxlogEntry generated classes, though this just felt a bit risky, if I'm not fully aware if this can happen in the future somehow, reading old data or whatever. It just didn't feel worth it, this just makes sure the columns don't get written.

ray-roestenburg-da: I guess it can't happen, the scan txlog parser will not create this type of entry anymore, so I could also just fail the encoding (just handle it like  `case _ => `, cause it's unexpected)

moritzkiefer-da: Sorry that wasn.t what I meant. I meant parse it but throw it away intead of turning it into an empty entry so make the function return an Option and make this one none. We still have the historic balance change tx log entries so I don't think you can assume we won't encounter it and the unexpected case probably results in a warning or something.

ray-roestenburg-da: yeah Ok I can return `Option` (as far as I know this code is only used for encoding entries and since the source of that is ScanTxLogParser, it should be fine, but famous last words so I'll just change to option)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
```diff
@@ -151,21 +165,7 @@ class DbScanStore(
 
   override lazy val txLogConfig: org.lfdecentralizedtrust.splice.store.TxLogStore.Config[
     org.lfdecentralizedtrust.splice.scan.store.TxLogEntry
-  ] {
-    val parser: org.lfdecentralizedtrust.splice.scan.store.ScanTxLogParser;
-    def entryToRow
-        : org.lfdecentralizedtrust.splice.scan.store.TxLogEntry => org.lfdecentralizedtrust.splice.scan.store.db.ScanTables.ScanTxLogRowData
-  } = new TxLogStore.Config[TxLogEntry] {
```
ray-roestenburg-da, line 158: removed the unnecessary structural refinement, and just created a concrete type for it

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanTables.scala
ray-roestenburg-da: Fixed

PR #3816, opened 2026-02-03
Make Docker compose deployments log at INFO by default

martinflorian-da:
Came up in a support interaction: https://digitalasset.atlassian.net/jira/servicedesk/projects/DA/queues/custom/374/DA-3500

[ci]


docs/src/release_notes_upcoming.rst
```diff
@@ -13,6 +13,9 @@
 
     - postgres-exporter: disabled exporting the settings table, to workaround `an issue of postgres-exporter <https://github.com/prometheus-community/postgres_exporter/issues/1240>`__.
 
+    - Splice apps and Canton components deployed via Docker compose now log at ``INFO`` level by default instead of ``DEBUG``.
+      In case you do want to change this, export the ``LOG_LEVEL`` environment variable before running ``./start.sh``. e.g., ``export LOG_LEVEL="DEBUG"; ./start.sh``.
```
isegall-da, line 17: Should we mention this in the actual docs too, or is that there already?

martinflorian-da: IMO it's too hacky... If there is need for this we should make it an inch more user-friendly first IMO... and then happy to add to docs :shrug:

isegall-da: Fair. I don't see that we have that for k8s either (we do, however, have "When launched locally, splice-node will create a log/ directory located at the root of the repository and log into canton_network.log. Canton logs into canton.log." which is like 3 years outdated since we removed any docs for running locally :facepalm:

PR #3821, opened 2026-02-04
grafana, cosmetic: Print message on dashboards check success

giner:
It's useful to see that the script has actually run.  Plus some rewording.


stas-sbi:
@martinflorian-da should I rebase? The CI error doesn't seem to be related to my changes.

moritzkiefer-da:
@stas-sbi yes that would be great the fix is in main.

PR #3845, opened 2026-02-04
Support dumping multiple segments to bulk-storage

isegall-da:
Fixes #3866 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
@@ -63,7 +63,7 @@ class UpdateHistoryBulkStorage(
 
   /** Gets the very first updates segment for this network after genesis
     * May return None if unknown yet. This could happen if no updates have been ingested,
-    * so we do not know the genesis record time yet. The caller should then sleep and retry.
+    * so we do not know the genesis record time yet. The caller should then schedule a retry.
     */
   private def getFirstSegmentFromGenesis: Future[Option[UpdatesSegment]] =
     for {
@@ -108,7 +108,7 @@ class UpdateHistoryBulkStorage(
             logger.info(s"Dumping next updates segment: $next")
             Future.successful(Some((Some(next), Some(next))))
           case None =>
-            logger.debug(s"Next segment after $current not known yet, sleeping...")
+            logger.debug(s"Next segment after $current not known yet, scheduling next attempt...")
             after(5.seconds, actorSystem.scheduler)(
               Future.successful(Some((current, None)))
             )


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1237,6 +1237,22 @@ class UpdateHistory(
     }
   }
 
+  def getLowestMigrationForRecordTime(
+      recordTime: CantonTimestamp
+  )(implicit tc: TraceContext): Future[Option[Long]] = {
+    // Including migration >= 0 to make sure it hits the indices
+    val filter = NonEmptyList.of(sql"migration_id >= 0 and record_time > ${recordTime}")
```
ray-roestenburg-da, line 1244: did you test query plan?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -0,0 +1,152 @@
...
+import org.apache.pekko.stream.scaladsl.{Keep, RestartSource, Source}
+import org.lfdecentralizedtrust.splice.scan.config.ScanStorageConfig
+import org.lfdecentralizedtrust.splice.scan.store.ScanKeyValueProvider
+import org.lfdecentralizedtrust.splice.store.{HardLimit, TimestampWithMigrationId, UpdateHistory}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration.*
+
+class UpdateHistoryBulkStorage(
+    val config: ScanStorageConfig,
```
ray-roestenburg-da, line 21: if these are all vals you might as well use case class

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -0,0 +1,152 @@
...
+          getSegmentEndAfter(
+            TimestampWithMigrationId(first.update.update.recordTime, first.migrationId)
+          ).map(Some(_))
+      }
+    } yield {
+      segmentEnd.map(UpdatesSegment(TimestampWithMigrationId(CantonTimestamp.MinValue, 0), _))
+    }
+
+  /** Gets the segment from which this app should start dumping, e.g. after a restart.
+    * May return None if unknown yet. The caller should then sleep and retry.
```
ray-roestenburg-da, line 83: nit: reschedule instead of sleep

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -0,0 +1,152 @@
...
+      actorSystem: org.apache.pekko.actor.ActorSystem,
+  ) = {
+    Source
+      .unfoldAsync(Option.empty[UpdatesSegment]) { current =>
+        getNextSegment(current).flatMap {
+          case Some(next) =>
+            logger.info(s"Dumping next updates segment: $next")
+            Future.successful(Some((Some(next), Some(next))))
+          case None =>
+            logger.debug(s"Next segment after $current not known yet, sleeping...")
```
ray-roestenburg-da: scheduling next ..:

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
isegall-da: It's the same query that we use for all queries that use `afterFilters`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: IDK, doesn't feel right, it's more a class than a named tuple, which is how I at least think of "case class". It also won't change much other than not having to type these "val"s, would it?

ray-roestenburg-da: yes nothing more than that. are all the vals needed outside of the class? it's just a bit peculiar to need all the constructor args as values later on for me.

isegall-da: This is probably my scala ignorance here...
No, most of these are private (which I realize now that I did not bother labeling as such), but required outside of the constructor in future calls to methods. Is not what "val" actually means here?

ray-roestenburg-da: you can just leave out the val. adding val makes it public

PR #3863, opened 2026-02-05
Stream contracts instead of ingesting them as a single Seq

OriolMunoz-da:


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1100,13 +1096,75 @@ final class DbMultiDomainAcsStore[TXE](
...
-    override def ingestAcs(
+
+    def ingestAcsStreamInBatches(
+        source: Source[Seq[BaseLedgerConnection.ActiveContractsItem], NotUsed],
+        offset: Long,
+    )(implicit
+        tc: TraceContext,
+        mat: Materializer,
+    ): Future[Unit] = {
+      // TODO: this is still loading everything into memory
```
OriolMunoz-da: To be handled in a different PR because otherwise it gets out of hand

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -2195,94 +2278,6 @@ object DbMultiDomainAcsStore {
...
 
-  case class TxLogEvent(
-      eventId: String,
-      synchronizerId: SynchronizerId,
-      acsContractId: Option[ContractId[?]],
-  )
-
-  /** Like [[IngestionSummary]], but with all fields mutable to simplify collecting the content from helper methods */
-  @SuppressWarnings(Array("org.wartremover.warts.Var"))
-  case class MutableIngestionSummary(
```
OriolMunoz-da, line 2206: moved to separate file as-is, except for dropping updateId which was always None (last instance must've been before batching was introduced for update ingestion)

OriolMunoz-da, line 2206: moved to separate file as-is, except for dropping updateId which was always None (last instance must've been before batching was introduced for update ingestion)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -2195,94 +2278,6 @@ object DbMultiDomainAcsStore {
     )
   }
 
-  case class TxLogEvent(
```
OriolMunoz-da, line 2198: unused

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
```diff
@@ -111,18 +111,15 @@ class UpdateIngestionService(
...
-      (acs, incompleteOut, incompleteIn) <- connection.activeContracts(filter, offset)
-      _ <- ingestionSink.ingestAcs(
-        offset,
-        acs,
-        incompleteOut,
-        incompleteIn,
-      )
-    } yield ()
+    ingestionSink.ingestAcsStreamInBatches(
+      batchSource(connection.activeContracts(filter, offset)),
```
rautenrieth-da: What if the app crashes during the stream?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
```diff
@@ -105,50 +104,47 @@ class BaseLedgerConnection(
...
-    for {
-      responseSequence <- activeContractsRequest runWith Sink.seq
-      contractStateComponents = responseSequence
-        .map(_.contractEntry)
-      active = contractStateComponents.collect {
+    activeContractsRequest
+      .map(_.contractEntry)
+      .map[Option[BaseLedgerConnection.ActiveContractsItem]] {
+        case lapi.state_service.GetActiveContractsResponse.ContractEntry.Empty =>
+          None
```
rautenrieth-da, line 118: This is just to make the compiler happy and should never happen in practice, right?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1100,13 +1096,75 @@ final class DbMultiDomainAcsStore[TXE](
...
+        source: Source[Seq[BaseLedgerConnection.ActiveContractsItem], NotUsed],
+        offset: Long,
+    )(implicit
+        tc: TraceContext,
+        mat: Materializer,
+    ): Future[Unit] = {
+      // TODO: this is still loading everything into memory
+      val summaryState = MutableIngestionSummary.empty
+      for {
+        // If we crash halfway through we want to start over with the whole ACS ingestion.
```
rautenrieth-da: It would be better if the app made a little bit of progress after a restart in a crash loop, but since we had to restart before as well, it's fine.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
rautenrieth-da: I see a `Source` that is used in `Source.runWith(Sink.foreachAsync(...))`, that looks like stream processing to me, where are we loading everything into memory?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1130,85 +1188,103 @@ final class DbMultiDomainAcsStore[TXE](
...
             )
-        } yield {
+        } yield ()
+      }
+    }
+
+    private def markAcsIngestedAsOf(offset: Long, summaryState: MutableIngestionSummary)(implicit
+        traceContext: TraceContext
+    ): Future[Unit] = {
+      if (hasFinishedAcsIngestion) {
```
rautenrieth-da: Wouldn't it make sense to check this _before_ we attempt to ingest the ACS? Otherwise you will end up with an error message and corrupt data.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
```diff
@@ -749,7 +751,30 @@ object MultiDomainAcsStore extends StoreErrors {
         acs: Seq[ActiveContract],
         incompleteOut: Seq[IncompleteReassignmentEvent.Unassign],
         incompleteIn: Seq[IncompleteReassignmentEvent.Assign],
-    )(implicit traceContext: TraceContext): Future[Unit]
+    )(implicit mat: Materializer, traceContext: TraceContext): Future[Unit] = {
```
rautenrieth-da, line 756: The whole `ingestAcs` method is now not needed anywhere except in tests, right?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
```diff
@@ -111,18 +111,15 @@ class UpdateIngestionService(
...
-      // TODO(#863): stream contracts instead of ingesting them as a single Seq
-      (acs, incompleteOut, incompleteIn) <- connection.activeContracts(filter, offset)
-      _ <- ingestionSink.ingestAcs(
-        offset,
-        acs,
-        incompleteOut,
-        incompleteIn,
-      )
-    } yield ()
+    ingestionSink.ingestAcsStreamInBatches(
```
rautenrieth-da, line 111: The other call site in this file is `ingestionSink.ingestAcs(participantBegin, Seq.empty, Seq.empty, Seq.empty)`. Is the plan to replace it by `ingestionSink.ingestAcsStreamInBatches(Source.empty)`?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1100,13 +1096,75 @@ final class DbMultiDomainAcsStore[TXE](
...
+        tc: TraceContext,
+        mat: Materializer,
+    ): Future[Unit] = {
+      // TODO: this is still loading everything into memory
+      val summaryState = MutableIngestionSummary.empty
+      for {
+        // If we crash halfway through we want to start over with the whole ACS ingestion.
+        // If the offset is the same between calls, this isn't strictly necessary, as `on conflict do nothing` prevents any issues.
+        // But if the offset changes between calls (which can happen in InitializeAcsAtLatestOffset), then we might miss some archival events,
+        // and thus we need to ingest the entire ACS from the beginning.
```
rautenrieth-da:
```suggestion
        // and thus we need to ingest the entire ACS from the beginning.
        // Since we are not using any long-running transaction spanning the entire ACS ingestion
        // process, ACS queries could return incomplete data shorty after application start.
        // This is fine because all clients are expected to use [[waitUntilAcsIngested()]] to avoid
        // reading ACS data before it has finished ingesting.
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: this is referring to the MutableIngestionSummary, see for example:
https://github.com/hyperledger-labs/splice/blob/21ebcb7cc1c5224ac1b6e76898e5bc24db913b42/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala#L1650

So this PR makes it so that the UpdateIngestionService doesn't load everything to memory... but the summary still does because of that

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1100,13 +1096,87 @@ final class DbMultiDomainAcsStore[TXE](
...
+      } else {
+        // TODO: this is still loading everything into memory
+        val summaryState = MutableIngestionSummary.empty
+        for {
+          // If we crash halfway through we want to start over with the whole ACS ingestion.
+          // If the offset is the same between calls, this isn't strictly necessary, as `on conflict do nothing` prevents any issues.
+          // But if the offset changes between calls (which can happen in InitializeAcsAtLatestOffset), then we might miss some archival events,
+          // and thus we need to ingest the entire ACS from the beginning.
+          // Since we are not using any long-running transaction spanning the entire ACS ingestion
+          // process, ACS queries could return incomplete data shorty after application start.
```
ray-roestenburg-da:
```suggestion
          // process, ACS queries could return incomplete data shortly after application start.
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1100,13 +1096,87 @@ final class DbMultiDomainAcsStore[TXE](
...
+        Future.failed(
+          new RuntimeException(
+            s"ACS was already ingested for store $acsStoreId, cannot ingest again"
+          )
+        )
+      } else {
+        // TODO: this is still loading everything into memory
+        val summaryState = MutableIngestionSummary.empty
+        for {
+          // If we crash halfway through we want to start over with the whole ACS ingestion.
```
moritzkiefer-da, line 1115: hm, is this worth doing over just having a single transaction for the ACS? If we had incremental ACS resumption at an offset (planned in Canton but not there) I see the value in splitting it up but without that the chance that the offset hasn't moved seems small enough that I'm not sure you ever hit this codepath and it adds additional complexity.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1100,13 +1096,87 @@ final class DbMultiDomainAcsStore[TXE](
...
+          // But if the offset changes between calls (which can happen in InitializeAcsAtLatestOffset), then we might miss some archival events,
+          // and thus we need to ingest the entire ACS from the beginning.
+          // Since we are not using any long-running transaction spanning the entire ACS ingestion
+          // process, ACS queries could return incomplete data shorty after application start.
+          // This is fine because all clients are expected to use [[waitUntilAcsIngested()]] to avoid
+          // reading ACS data before it has finished ingesting.
+          _ <- clearDataForCurrentMigrationId()
+          _ <- source.runWith(
+            Sink.foreachAsync[Seq[BaseLedgerConnection.ActiveContractsItem]](parallelism = 1) {
+              batch =>
```
moritzkiefer-da: Are we not missing batching logic here similar to what we have for ingest updates? Or am I missing something that batches it? iirc the ledger API will give you one stream element per contract.

OriolMunoz-da: this _is_ batched: the insert into the acs_store table will insert all of `batch.collect { case ActiveContractsItem.ActiveContract(contract) => contract }` in a single insert

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: > hm, is this worth doing over just having a single transaction for the ACS?

how does one do that without still loading the entire ACS into memory with Slick? e.g., in DbStorage:
```
def bulkOperation[A](
      statement: String,
      values: immutable.Iterable[A], <--- whole list
      profile: Profile,
      transactional: Boolean = true,
  )
```

building a transactional DBIO will also end up with everything in memory as inserts

short of using raw JDBC, I'm not sure how to do that

OriolMunoz-da: I guess with `withPinnedSession`...?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
moritzkiefer-da: Ye but isn't `batch` a single stream element from the source which as I mentioned above is usually one contract? I think we need to batch across stream elements

rautenrieth-da: There is 
```scala
private def batchSource[T](source: Source[T, NotUsed]): Source[Vector[T], NotUsed] =
    source.batch(config.ingestion.maxBatchSize.toLong, Vector(_))(_ :+ _)
```
in UpdateIngestionService

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
moritzkiefer-da: oh I thought there was some variant where you can just call beginTransaction/endTransaction. Don't know the slick API so maybe it doesn't exist nicely dunno.

moritzkiefer-da: there is a `withTransaction` is that not usable?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
moritzkiefer-da: ah thanks that's the part I was missing. All good then thanks!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: for anyone reading, that's from an older version of Slick https://scala-slick.org/doc/2.0.3/connection.html#transactions

OriolMunoz-da: So because:
- I can get a session, but I cannot pass it down to each DBIO or run it explicitly with it
- I also cannot build a DBIO without effectively loading every insert statement into memory (back  to square 0)

We could of course fall back to JDBC, but since this is implemented will keep it as-is for now

OriolMunoz-da: So because:
- I can get a session, but I cannot pass it down to each DBIO or run it explicitly with it
- I also cannot build a DBIO without effectively loading every insert statement into memory (back  to square 0)

We could of course fall back to JDBC, but will keep it as-is for now

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: s.a. https://github.com/hyperledger-labs/splice/pull/3863#discussion_r2782770689

PR #3880, opened 2026-02-06
Added memory limits for load tester

pasindutennage-da:
Fixes https://github.com/DACH-NY/canton-network-internal/issues/3389

### DIFF
--- a/cluster/pulumi/common/src/config/loadTesterConfig.ts
+++ b/cluster/pulumi/common/src/config/loadTesterConfig.ts
@@ -4,22 +4,7 @@ import { clusterYamlConfig } from '@lfdecentralizedtrust/splice-pulumi-common/sr
 import util from 'node:util';
 import { z } from 'zod';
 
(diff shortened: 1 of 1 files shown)


nicu-da (approved):
Nice, thanks! 
Do we also plan on adding pulumi support to modify them from the cluster config.yaml if needed?

cluster/pulumi/common/src/config/loadTesterConfig.ts
```diff
@@ -4,6 +4,23 @@ import { clusterYamlConfig } from '@lfdecentralizedtrust/splice-pulumi-common/sr
 import util from 'node:util';
 import { z } from 'zod';
 
+export const ResourceRequirementsSchema = z
```
nicu-da: We already have this, check `K8sResourceSchema`

pasindutennage-da: @nicu-da 

Thanks. Changed.

PR #3892, opened 2026-02-06
Use = any(array) instead of in (a,b,...)

ray-roestenburg-da:
Fixes #3665 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/DbVotesStoreQueryBuilder.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/DbVotesStoreQueryBuilder.scala
@@ -4,6 +4,7 @@
 package org.lfdecentralizedtrust.splice.store
 
 import cats.data.NonEmptyList
+import com.daml.ledger.javaapi.data.codegen.ContractId
 import com.digitalasset.canton.config.CantonRequireTypes.String3
 import com.digitalasset.canton.logging.NamedLogging
 import com.digitalasset.canton.resource.DbStorage.Implicits.BuilderChain.toSQLActionBuilderChain
@@ -104,13 +105,15 @@ trait DbVotesAcsStoreQueryBuilder extends AcsQueries with LimitHelpers with Name
   ): SqlStreamingAction[Vector[
     AcsQueries.SelectFromAcsTableResult
   ], AcsQueries.SelectFromAcsTableResult, Effect.Read] = {
-    val voteRequestTrackingCidsSql = sqlValueList(trackingCids)
+    val cids: Seq[ContractId[?]] = trackingCids
+    val voteRequestTrackingCidsSql =
+      inClause(trackingCidColumnName, cids)
     selectFromAcsTable(
       acsTableName,
       acsStoreId,
       domainMigrationId,
       VoteRequest.COMPANION,
-      where = (sql""" #$trackingCidColumnName in """ ++ voteRequestTrackingCidsSql).toActionBuilder,
+      where = voteRequestTrackingCidsSql,
       orderLimit = sql"""limit ${sqlLimit(limit)}""",
     )
   }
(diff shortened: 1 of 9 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/DbVotesStoreQueryBuilder.scala
```diff
@@ -104,13 +105,15 @@ trait DbVotesAcsStoreQueryBuilder extends AcsQueries with LimitHelpers with Name
   ): SqlStreamingAction[Vector[
     AcsQueries.SelectFromAcsTableResult
   ], AcsQueries.SelectFromAcsTableResult, Effect.Read] = {
-    val voteRequestTrackingCidsSql = inClause(trackingCids)
+
+    val voteRequestTrackingCidsSql =
+      inClause(trackingCidColumnName, trackingCids.map(_.asInstanceOf[ContractId[?]]))
```
OriolMunoz-da: is the asInstanceOf really required here?

OriolMunoz-da (approved):
thanks

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/DbVotesStoreQueryBuilder.scala
ray-roestenburg-da: Let me check to be sure

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsQueries.scala
```diff
@@ -279,8 +281,21 @@ trait AcsQueries extends AcsJdbcTypes {
       .getOrElse(SQLActionBuilderChain(sql""))
   }
 
-  protected def inClause[V: SetParameter](seq: Iterable[V]): SQLActionBuilderChain =
-    sql"(" ++ sqlCommaSeparated(seq.map(v => sql"$v")) ++ sql")"
+  protected def inClause[V: ClassTag](
```
rautenrieth-da, line 287: Why are we not using [DbStorage.toInClause](https://github.com/DACH-NY/canton/blob/main/community/base/src/main/scala/com/digitalasset/canton/resource/Storage.scala#L822), which looks almost identical? @OriolMunoz-da do you remember perhaps?

rautenrieth-da (approved):
Thanks!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsQueries.scala
OriolMunoz-da: either:
- unaware that it existed
- was introduced after we created `inclause`, since it's in the canton fork

but from what I can see, it doesn't use `= any(x)` so we don't want to use it?

moritzkiefer-da: > but from what I can see, it doesn't use = any(x) so we don't want to use it?

It does in Canton main and these days also 3.4 (but we didn't bump our fork yet).

If we can use it, that seems preferable to me instead of rolling our own.

ray-roestenburg-da: I thought it was simple enough to just use our own, the canton main introduces a whole lot of other stuff I'm not sure we need (The `ToDbPrimitive` amongst other things)

ray-roestenburg-da: Of course can wait for bump of fork if that's preferable and then update accordingly.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/DbVotesStoreQueryBuilder.scala
ray-roestenburg-da: Fixed.

PR #3899, opened 2026-02-09
Prevent ingestion summary from loading ACS into memory

OriolMunoz-da:


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1134,13 +1133,25 @@ final class DbMultiDomainAcsStore[TXE](
...
+                ).map { _ =>
+                  val newAcsSize = summaryState.acsSizeDiff + acsSizeSoFar
+                  val summary = summaryState
+                    .toIngestionSummary(
+                      synchronizerIdToRecordTime = Map.empty,
+                      offset = offset,
+                      newAcsSize = newAcsSize,
+                      metrics = metrics,
+                    )
+                  handleIngestionSummary(summary)
```
OriolMunoz-da, line 1145: this should also fix the fact that our metrics don't get updated until the ACS is ingested (so you cannot see progress)

ray-roestenburg-da (approved):
Thanks

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
rautenrieth-da: Are the contracts in the ACS ordered by their creation time? `handleIngestionSummary` is being used in `DbSvDsoStore` to update the last open/issuing round, and if the contracts are not ordered then those metrics will jump around. The other metrics updated in `toIngestionSummary` should be fine.

rautenrieth-da (approved):

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
OriolMunoz-da: gah, good catch!... but wasn't that also the case before? Before it passed the entire ACS as a Seq, whereas now it passes the stream, so the result is the same
(I'm still trying to figure out the answer to your question)

moritzkiefer-da: Isn't that why we have ethe `waitUntilAcsIngested` thing for places that do care about having seen the full ACS?

OriolMunoz-da: that's not what we're talking about here, but this: https://github.com/hyperledger-labs/splice/blob/53fd17ce7611e7bda8851d2d88ea93a7594418b5/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala#L113

moritzkiefer-da: ah I see, I'm not sure I'd worry all that much about those metrics tbh. Reingesting the full ACS is rare and this will fix itself once you get the first round change after the ingestion.

PR #3923, opened 2026-02-11
DRAFT: Make SV app auto-enable free confirmation responses

meiersi-da:
Fixes #3726

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/ReconcileDynamicDomainParametersTrigger.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/ReconcileDynamicDomainParametersTrigger.scala
@@ -19,7 +19,6 @@ import org.lfdecentralizedtrust.splice.sv.util.SvUtil
 import org.lfdecentralizedtrust.splice.util.AmuletConfigSchedule
 import com.digitalasset.canton.admin.api.client.data.DynamicSynchronizerParameters as ConsoleDynamicSynchronizerParameters
 import com.digitalasset.canton.time.{PositiveFiniteDuration, PositiveSeconds}
-import com.digitalasset.canton.config.NonNegativeFiniteDuration
 import com.digitalasset.canton.config.RequireTypes.{NonNegativeLong, PositiveInt}
 import com.digitalasset.canton.data.CantonTimestamp
 import com.digitalasset.canton.protocol.DynamicSynchronizerParameters
@@ -30,6 +29,7 @@ import com.digitalasset.canton.tracing.TraceContext
 import io.opentelemetry.api.trace.Tracer
 import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
 import org.apache.pekko.stream.Materializer
+import org.lfdecentralizedtrust.splice.sv.config.SvAppBackendConfig
 
 import java.time.Instant
 import scala.concurrent.{ExecutionContext, Future}
@@ -42,8 +42,7 @@ class ReconcileDynamicSynchronizerParametersTrigger(
     override protected val context: TriggerContext,
     store: SvDsoStore,
     participantAdminConnection: ParticipantAdminConnection,
-    preparationTimeRecordTimeTolerance: NonNegativeFiniteDuration,
-    mediatorDeduplicationTimeout: NonNegativeFiniteDuration,
+    config: SvAppBackendConfig,
 )(implicit
     override val ec: ExecutionContext,
     mat: Materializer,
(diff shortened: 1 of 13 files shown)


meiersi-da:
@moritzkiefer-da : could I have your review on this. It seems to me that we don't need an extra test, as the approach here:
1. tests that we can move from `false` to `true`
2. tests that all txs go through with the flag set to `true`

I'd rely on Thibault's testing that the confirmation responses are actually free. I'd rely on the code not distinguishing anything about the flag, to be sure that we can switch back from `true` to `false`. wdyt?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/ReconcileDynamicDomainParametersTrigger.scala
```diff
@@ -177,10 +180,12 @@ class ReconcileDynamicSynchronizerParametersTrigger(
       amuletConfig: AmuletConfig[USD],
       synchronizerConfig: Option[SynchronizerConfig],
       preparationTimeRecordTimeToleranceTarget: Option[InternalNonNegativeFiniteDuration],
+      enableHeuristicFreeConfirmationResponses: Boolean,
   ): DynamicSynchronizerParameters = {
     val domainFeesConfig = amuletConfig.decentralizedSynchronizer.fees
     // Make sure that the bootstrap script for the upgrade domain is aligned with any changes made to the
     // dynamic domain parameters here to prevent the soft synchronizer upgrade test from failing
+    // FIXME: look into what's required for this!
```
moritzkiefer-da: not sure what this comment is referring to

meiersi-da: The alignment of the bootstrap script for the upgrade domain. It's not clear to me whether I need to adjust it or not. I suspect not, as the free confirmation responses will just be enabled again.

moritzkiefer-da:
> Ah good point. I did manually investigate the transactions. But I'll add the assertion to the test that has the positive assertion.

Ah I think I see how it works now. Not sure I like that approach though tbh. Always bootstrapping with the "wrong" config only to then change it later feels off to me. Why not just have one test (repurposing the record time tolerance one is totally fine) that does that deliberately (by changing the config at the start) but use the config for bootstrapping as well?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/ReconcileDynamicDomainParametersTrigger.scala
moritzkiefer-da: I don't think the bootstrap script is the part you're missing. It's Sv1Initializer and missing that does feel off to me.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/sv1/SV1Initializer.scala
```diff
@@ -456,6 +456,7 @@ class SV1Initializer(
       PositiveFiniteDuration.tryOfSeconds(
         sv1Config.initialSynchronizerFeesConfig.baseRateBurstWindow.duration.toSeconds
       ),
+      freeConfirmationResponses = config.enableFreeConfirmationResponses,
```
meiersi-da, line 459: @moritzkiefer-da : I've now changed the initializer to always set it the same way as the trigger.

meiersi-da: The PR has a green CI. I'm keeping it in draft until we've aligned on the roll-out strategy.

Could you do another review?

docs/src/release_notes_upcoming.rst
```diff
@@ -7,22 +7,45 @@
...
-release-notes:: Upcoming
+.. release-notes:: Upcoming
 
-- Daml
+    .. important::
 
-  - Optimize the number of views in the automation run by the SV app to convert ``FeaturedAppActivityMarker`` contracts into ``AppRewardCoupon`` contracts.
+        **SV node operators must only deploy this release on networks where
+        all other SV nodes run Splice 0.5.10 or higher!**
+        Otherwise the sequencers will disagree on the traffic cost of confirmation responses and fork the ledger.
```
meiersi-da, line 16: @moritzkiefer-da : how much of a problem do you see this for all the deployments that we need to do? Is it better to wait another week so that the chance of an old release sticking around is lower, or are you comfortable going for this release?

moritzkiefer-da: I'm ok with putting it in this release. I assume you double checked that the Canton change really is in 0.5.10? (I'm fairly confident but definitely feels worth checking again)

PR #3928, opened 2026-02-11
Don't crash a test if taking a debug screenshot fails

rautenrieth-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/6802

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/FrontendIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/FrontendIntegrationTest.scala
@@ -443,11 +443,21 @@ trait FrontendTestCommon extends TestCommon with WebBrowser with CustomMatchers 
   protected def screenshot()(implicit webDriver: WebDriverType): Unit = {
     clue("Saving screenshot") {
       val fullScreen = webDriver.findElement(By.tagName("body"))
-      val screenshotFile = fullScreen.getScreenshotAs(OutputType.FILE)
-      val time = Calendar.getInstance.getTime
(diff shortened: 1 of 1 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/FrontendIntegrationTest.scala
```diff
@@ -443,11 +443,21 @@ trait FrontendTestCommon extends TestCommon with WebBrowser with CustomMatchers
...
+        val screenshotFile = fullScreen.getScreenshotAs(OutputType.FILE)
+        val time = Calendar.getInstance.getTime
+        val timestamp = new SimpleDateFormat("yy-MM-dd-H:m:s.S").format(time)
+        val filename = Paths.get("log", s"screenshot-${timestamp}.png").toString
+        FileUtils.copyFile(screenshotFile, new File(filename))
+      } catch {
+        case e: org.openqa.selenium.WebDriverException =>
+          // We still want to be notified of the failure through the log checker,
+          // we just don't want to stop generating debug info if the screenshot fails.
+          logger.warn("Failed to take screenshot", e)
```
moritzkiefer-da, line 459: that will still fail CI. Do we want that? I guess yes because CI fails anyway when we get to this point?

rautenrieth-da: The idea was to not crash the whole test and continue collecting debug info (e.g., the page source that is dumped next) to understand the actual root cause. https://github.com/hyperledger-labs/splice/pull/3929 might either fix the error or at least provide better errors to find the actual cause. If not, then we can demote the warning the first time it happens again.

PR #3929, opened 2026-02-11
Update Selenium

rautenrieth-da:
We saw two failures to capture screenshots recently. These could be due to the page being rendered at the time, or due to an outdated driver. The bump from [Selenium 4.12 to 4.21](https://github.com/SeleniumHQ/selenium/blob/trunk/java/CHANGELOG) includes the following

```
v4.19.0
======
* [bidi] Update the capture screenshot APIs to include all parameters and remove scroll parameter (#13743)

v4.17.0
======
* Fix the ScreenshotException message

v4.14.0
======
* Add browsing context's capture box and element screenshot commands
```

### DIFF
--- a/project/ignore-patterns/canton_network_test_log.ignore.txt
+++ b/project/ignore-patterns/canton_network_test_log.ignore.txt
@@ -139,3 +139,6 @@ Simulated S3 write error
 # Warnings in s3Mock log (which we redirect to our logs), that we don't care about
 \[s3Mock\].*Ignoring option --illegal-access=warn
 Ryuk has been disabled
+
+# Happened after upgrading Selenium when calling `webDriver.quit()`, looks benign
+"message":"connection closed locally","logger_name":"o.o.s.remote.http.WebSocket"


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/FrontendIntegrationTest.scala
```diff
@@ -156,7 +155,7 @@ trait FrontendTestCommon extends TestCommon with WebBrowser with CustomMatchers
 
   val frontendNames: Seq[String] = Seq()
 
-  type WebDriverType = WebDriver with TakesScreenshot with JavascriptExecutor with WebStorage
```
rautenrieth-da, line 159: WebStorage is deprecated in 4.21 and removed in a later release (can't update to latest because scalatest is behind)

rautenrieth-da:
Yes, pushed too soon, Selenium removed a whole bunch of functionality that can be replaced by just executing javascript. Some of the modified methods are only called exceptionally. I have checked the scripts manually but there's still a small change something will break.

moritzkiefer-da (approved):

PR #3936, opened 2026-02-11
avoid parsing all dars in unit tests

isegall-da:
Speeds up a unit test that creates a single contract from >1.5 minute to 13 seconds on my machine.
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/StoreTestBase.scala
+++ b/apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/StoreTestBase.scala
@@ -104,6 +104,7 @@ abstract class StoreTestBase
   protected val maliciousPackageId = "maliciouspackageid"
 
   // Looks up the package name from the package ID in dars.lock, to avoid having to parse all DARs just to find this mapping
+  // TODO(#3937): this is quite hacky. What we should really do is just auto-generate DarResources instead of deriving it from DARs at runtime.
   protected def tryGetPackageNameFromDarLock(identifier: Identifier): Option[String] = {
     val path = Paths.get("daml/dars.lock")
     Files


apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/StoreTestBase.scala
```diff
@@ -102,24 +103,38 @@ abstract class StoreTestBase
   protected val dummyHoldingPackageId = DummyHolding.TEMPLATE_ID.getPackageId
   protected val maliciousPackageId = "maliciouspackageid"
 
+  // Looks up the package name from the package ID in dars.lock, to avoid having to parse all DARs just to find this mapping
```
isegall-da, line 106: Happy for other ideas.
I initially thought of modifying DarResources such that in unit tests we can initialize it with only the relevant subset of dars (and then `lookupFromResources` will parse only that small subset), but this was much simpler, and seems good enough for the purpose of speeding up unit tests.

moritzkiefer-da: hm feels very hacky, have you checked how fast it actually is if we do speed up decoding? We want that anyway. Given how large you observed the slowdown to be it's also gonna matter in integration tests.

I think the real option for getting it down to zero is to just generate all of DarResources.scala instead of dynamically decoding any dars.

I'm ok with this change but let's please still create an issue to fix this properly and add a TODO here.

isegall-da: > feels very hacky

Definitely agree. 

> just generate all of DarResources.scala

I actually like the auto-generation option the best. In the end, DarResources is just a bunch of strings per dar, feels a bit silly to parse them from the dars on every run. I briefly considered just doing it now, but think I'll go with this hack for now, and create an issue to do it properly.

PR #3960, opened 2026-02-13
Add support for configuring multiple physical synchronizers in the sv app

nicu-da:
This allows us in further steps to communicate with both syncs at the same time. 

[ci]

part of https://github.com/hyperledger-labs/splice/issues/3950 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
@@ -6,6 +6,7 @@ package org.lfdecentralizedtrust.splice.sv.config
 import com.digitalasset.canton.SynchronizerAlias
 import com.digitalasset.canton.config.*
 import com.digitalasset.canton.config.RequireTypes.{
+  NonNegativeInt,
   NonNegativeLong,
   NonNegativeNumeric,
   PositiveInt,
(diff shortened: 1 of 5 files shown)


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -290,7 +290,9 @@ case class SvAppBackendConfig(
     onboarding: Option[SvOnboardingConfig] = None,
     initialAmuletPriceVote: Option[BigDecimal] = None,
     cometBftConfig: Option[SvCometBftConfig] = None,
-    localSynchronizerNode: Option[SvSynchronizerNodeConfig],
+    // TODO(#564): read this from scan
+    currentPhysicalSynchronizerId: Option[Int] = Some(0),
```
moritzkiefer-da: nit: this is not the physical synchronizer id, it's the serial of the physical synchronizer id

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -290,7 +290,9 @@ case class SvAppBackendConfig(
...
+    localSynchronizerNodes: Map[Int, SvSynchronizerNodeConfig],
```
moritzkiefer-da: add a comment what the Int is.
Also is there a reason why we use Int instead of NonNegativeInt like Canton seems to do for the serial?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
nicu-da: Good point, renaming

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
nicu-da: Yeah let me change it and actually add an alias type

PR #3964, opened 2026-02-16
DO NOT MERGE: Daml code for implementing traffic-based app rewards

meiersi-da:
@moritzkiefer-da : the code is now fully finished. If you find some time next week to do a final review, then that would be great.

Added "DO NOT MERGE" to title, as we want to merge this into the right feature branch once the implementation has progressed far enough.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


daml/splice-amulet/daml/Splice/Amulet/RewardAccountingV2.daml
```diff
@@ -0,0 +1,150 @@
...
+
+Notes on design:
+- roll-out
+  - enable dry-run of RAV_V2
+    ==> app reward calculation will be simulated, but no app reward coupons will be issued
+  - enable RAV_V2 for real
+    ==> all open rounds at this point will use RAV_V2
+    ==> activity markers will be dropped by trigger
+    ==> featured app rights
+
```
meiersi-da: SV onboarding: SVs keep track of the first record time as of which they are creating activity records. The confirmation trigger for the root hash will check whether the root hash is available on the SV's scan instance; and if not, it will just abstain from voting.

Thus we absorb the new SV into the unavailability budget. This should work for all transitions except for onboarding the second SV. For that case, the trigger should read the root hash from the other SV's scan, and confirm it on-ledger to avoid the workflow getting stuck.

meiersi-da (commented):
@moritzkiefer-da : I have a sketch of what I believe are all the choices and templates to make the switch to traffic-based app rewards possible at the Daml level. Some simple work is delayed via FIXMEs in the code.

Could I ask you for a review to check alignment, in particular:
* highlight implementation choices that you disagree with / consider fishy
* highlight aspects / workflows that are missing or that you are not sure how they'll work

You don't have to work too hard on the latter part. I expect that we also find them as part of writing up and reviewing the full design.

daml/splice-amulet/daml/Splice/Amulet/CryptoHash.daml
```diff
@@ -0,0 +1,111 @@
...
+data Hash = Hash with value : Text
+  deriving (Eq, Show)
+
+-- | Compute the hash of a record.
+hashRecord : [Hash] -> Hash
+hashRecord = hashListInternal . map (.value)
+
+-- | Compute the hash of an upgraded record so that it agrees with the old record hash
+-- when ignoring trailing None fields.
+hashUpgradedRecord : [Hash] -> [Optional Hash] -> Hash
```
moritzkiefer-da, line 34: how do you expect to use the hashUpgradedRecord and hashUpgradeVariant functions? They are completely unused atm.

daml/splice-amulet/daml/Splice/Amulet/RewardAccountingV2.daml
moritzkiefer-da: > Thus we absorb the new SV into the unavailability budget. This should work for all transitions except for onboarding the second SV. For that case, the trigger should read the root hash from the other SV's scan, and confirm it on-ledger to avoid the workflow getting stuck.


If we need that case anyway, what's the argument for not using that for all newly onboarded SVs instead of only the second one? I'm also not sure it's true that it only matters for the second one, if I onboard 3 new SVs during a round which is how basically all our tests work you have the same issue for all 3?

daml/splice-amulet/daml/Splice/AmuletConfig.daml
```diff
@@ -47,6 +47,22 @@ data TransferConfigV2 unit = TransferConfigV2 with
     maxNumLockHolders : Int
   deriving (Eq, Show)
 
+-- | Version of the scheme that is used to calculate rewards.
+-- FIXME: consider using an integer for a more seamless transition to future versions.
+data RewardCalculationVersion
+  = RCV_V1 -- ^ Purely on-ledger reward accounting as described in the whitepaper plus the CIPs before CIP-104.
```
moritzkiefer-da: nit: I feel like the abbreviation here isn't all that helpful and just spelling it out to RewardCalculationVersion_V1 is probably better

daml/splice-amulet/daml/Splice/ExternalPartyAmuletRules.daml
```diff
@@ -64,6 +77,56 @@ template ExternalPartyAmuletRules
            description
          pure (ExternalPartyAmuletRules_CreateTransferCommandResult cmd)
 
+    -- FIXME: convert into api package
+    nonconsuming choice ExternalPartyAmuletRules_MintWithBeneficiaries : ExternalPartyAmuletRules_MintWithBeneficiariesResult
```
moritzkiefer-da: Have you thought about how the reward leaderboard in scan needs to change? Ofc my usual preferred option is to remove it but at least for now we don't have a timeline for that afaik.

daml/splice-amulet/daml/Splice/AmuletRules.daml
```diff
@@ -609,6 +644,75 @@ template AmuletRules
...
+        beneficiaries : [Party]
+      observer beneficiaries
+      controller dso
+      do
+        require "at least one coupon" (not (null rewardCouponCids))
+
+        -- archive coupons
+        coupons <- forA rewardCouponCids $ \cid -> do
+          coupon <- fetchChecked (ForDso with dso) cid
+          _ <- exercise cid RewardCouponV2_DsoExpire
```
moritzkiefer-da: should we make this choice return the payload so we don't need the extra fetch first?

daml/splice-amulet/daml/Splice/AmuletRules.daml
```diff
@@ -609,6 +644,75 @@ template AmuletRules
...
+        hiddenRewardCouponCids : [ContractId RewardCouponV2]
+        beneficiaries : [Party]
+      observer beneficiaries
+      controller dso
+      do
+        require "at least one coupon" (not (null hiddenRewardCouponCids))
+
+        -- unhide coupons
+        results <- forA hiddenRewardCouponCids $ \cid -> do
+          coupon <- fetchAndArchive (ForDso with dso) cid
```
moritzkiefer-da, line 710: probably not new in this PR but I don't think I understand what logic we use to decide whether to go for a bare archive like we do here and handle any checks in the choice vs something like RewardCouponV2_DsoExpire where we have a choice on the coupon.

daml/splice-amulet/daml/Splice/Amulet.daml
```diff
@@ -373,6 +375,37 @@ template AppRewardCoupon
...
+-- that are based on on off-ledger calculations.
+-- See Splice.Amulet.RewardAccountingV2 for details.
+template RewardCouponV2
+  with
+    dso : Party
+    beneficiary : Party -- ^ The party that can claim the reward.
+    round : Round -- ^ The round for which this reward can be claimed.
+    amount : Decimal -- ^ The amount of reward to be claimed. Denominated in Amulet.
+    expiresAt : Time -- ^ Time-based expiry. Used to determine when the reward can no longer be claimed.
+    hidden : Bool
```
moritzkiefer-da: Nit: The name is probably gonna raise some questions given that everything DSO visible is public.

Wonder if something very explicit like "beneficiaryIsObserver" is preferable

daml/splice-amulet/daml/Splice/ExternalPartyAmuletRules.daml
moritzkiefer-da: What's the argument why we also want to do minting for external parties in the same change? iirc the CIP actually suggested to still use MintingDelegation. Just to save another Daml version increment once we do want to add it?

moritzkiefer-da (commented):
Nice thanks!

daml/splice-amulet/daml/Splice/ExternalPartyAmuletRules.daml
meiersi-da: This is not meant for external parties, but for decentralized apps that want to have their ultimate beneficiaries mint.

It's missing the API package though. I've implemented it on `ExternalPartyAmuletRules` because that is the more stable contract.

daml/splice-amulet/daml/Splice/AmuletRules.daml
meiersi-da: I think we still need the check against the `dso` party or not?

daml/splice-amulet/daml/Splice/AmuletRules.daml
meiersi-da: hmm... good point. I'm currently going by gut feel. I'll add a topic for our 1-1 to align on a reasonable rule; and decide where to document it.

daml/splice-amulet/daml/Splice/ExternalPartyAmuletRules.daml
meiersi-da: My intention was to list the actual featured app party on the leaderboard; i.e. ignore this attribution here from the minting.

daml/splice-amulet/daml/Splice/AmuletRules.daml
moritzkiefer-da: Sure but you can check that on the result? No strong opinion whether it's worth it but given that Daml model efficiency has become a bit more of a topic it seems at least worth considering eliminating duplicate nodes where we don't need it.

daml/splice-amulet/daml/Splice/Amulet/CryptoHash.daml
meiersi-da: might remove or add a comment.

The intended use is to list the trailing optional fields in the `[Optional Hash]` argument, which guarantees that the hashes agree across the versions where the fields were added and when they were not added.

daml/splice-amulet/daml/Splice/ExternalPartyAmuletRules.daml
moritzkiefer-da: That's not what I meant. The leaderboard currently works by aggregating data across a round after the round closes and colects all the rewards minted in that round (it attributes the reward to the open round for which you mint not the round of the activity record).

This doesn't really work with minting against ExternalPartyConfigState as the round is already closed before you exercise this choice.

daml/splice-amulet/daml/Splice/Amulet/RewardAccountingV2.daml
meiersi-da: Yeah concurrent onboarding is another problem. You are right. Let's build it such that we don't need to absorb anything into our fault-tolerance budget. 

The easiest option would then be for the newly onboarded SVs to configure the Scan URLs from which to do a BFT read for onboarding. This way we don't need logic to distinguish between the ones that are fully operational and the ones that are onboarding. In our tests, we'd just configure that to only point to SV1.

daml/splice-amulet/daml/Splice/ExternalPartyAmuletRules.daml
moritzkiefer-da: ah right ofc

daml/splice-amulet/daml/Splice/Amulet/RewardAccountingV2.daml
moritzkiefer-da: > The easiest option would then be for the newly onboarded SVs to configure the Scan URLs from which to do a BFT read for onboarding. This way we don't need logic to distinguish between the ones that are fully operational and the ones that are onboarding. In our tests, we'd just configure that to only point to SV1.

I wonder if we can piggy back on how we do scan backfilling approach where each SV reports what they have available and we use that to adjust the thresholds dynamically instead of introducing a different set of urls to backfill from.

daml/splice-amulet/daml/Splice/Amulet.daml
meiersi-da: fixed

daml/splice-amulet/daml/Splice/AmuletRules.daml
meiersi-da: These duplicate nodes only have a computation cost or not? The actual data only sends the root node, the input contracts, and the hash over the produced transaction. So it seems to me we would not be saving bandwidth on the global sunchronizer. We'd also spend a bit of extra storage on the Scan DB, as it persists the result of exercise nodes, and that one would be wider.

So, I'm inclined to keep the most direct implementation here. The one optimization I opted to do consistently is that batch choices make all informees choice observers so we save on views and GSync throughput.

daml/splice-amulet/daml/Splice/AmuletRules.daml
```diff
@@ -471,6 +511,11 @@ template AmuletRules
               transferConfigUsd = configUsd.transferConfig
               issuanceConfig = getValueAsOf newOpenRoundIssuingFor configUsd.issuanceCurve
               tickDuration
+              rewardConfig
```
meiersi-da, line 514: @dfordivam : note that I've switched to tracking all parameters required for reward computation on the `OpenMiningRound` contract. So there's no need anymore to track `AmuletRules` in the TCS.

daml/splice-amulet/daml/Splice/Amulet.daml
```diff
@@ -545,6 +584,45 @@ requireAmuletExpiredForAllRounds externalPartyConfigState0Cid externalPartyConfi
...
+rewardCouponV2_assignBeneficiariesImpl
+  : RewardCouponV2
+  -> Api.RewardAssignmentV1.RewardCoupon_AssignBeneficiaries
+  -> Update Api.RewardAssignmentV1.RewardCoupon_AssignBeneficiariesResult
+rewardCouponV2_assignBeneficiariesImpl coupon arg = do
+  let newBeneficiaries = arg.newBeneficiaries
+  let totalNewBeneficiaryAmount = sum $ map (.amount) arg.newBeneficiaries
+  require ("No more than " <> show maxNumNewCouponV2Beneficiaries <> " beneficiaries assigned")
+    (length arg.newBeneficiaries <= maxNumNewCouponV2Beneficiaries)
+  require "Total amount assigned to new beneficiaries below coupon amount" (totalNewBeneficiaryAmount <= coupon.amount)
```
moritzkiefer-da: I wonder if it wouldn't be less confusing for this to be an equal. I guess the counter argument is that maybe it's annoying to do the math exactly on the client side?

daml/splice-amulet/daml/Splice/Amulet.daml
```diff
@@ -545,6 +584,45 @@ requireAmuletExpiredForAllRounds externalPartyConfigState0Cid externalPartyConfi
...
+
+  -- return remainder to original beneficiary
+  let remainingAmount = coupon.amount - totalNewBeneficiaryAmount
+  originalBeneficiaryRemainderCid <-
+    if (remainingAmount > 0.0)
```
moritzkiefer-da: should we have a min amount below which we don't even create this. Or maybe at least see if the beneficiary is also in the arguments and merge with that instead of creating one tiny coupon contract and one "normal" one because some client-side math wasn't exact.

daml/splice-amulet/daml/Splice/Amulet.daml
```diff
@@ -372,6 +372,42 @@ template AppRewardCoupon
          return AppRewardCoupon_DsoExpireResult with
            ..
 
+-- | A coupon for receiving rewards, which is currently only used for
+-- traffic-based app rewards.
+--
+-- Designed to be flexible so we could extend it to include other kinds of rewards
+-- that are based on on off-ledger calculations.
+-- See Splice.Amulet.RewardAccountingV2 for details.
+template RewardCouponV2
```
moritzkiefer-da, line 380: Nit: Should we call it ActivityRecord instead. We did use that terminology when we had to introduce a new template for ValidatorLivenessActivityRecord.

daml/splice-amulet/daml/Splice/AmuletRules.daml
```diff
@@ -1178,6 +1316,23 @@ summarizeAndConsumeInputs csum dso sender inps = do
         changeToHoldingFeesRate = s.changeToHoldingFeesRate
         totalDevelopmentFundAmount = (+ coupon.amount) <$> s.totalDevelopmentFundAmount
 
+    summarizeAndConsumeInput _round s (InputRewardCouponV2 couponCid) = do
+      coupon <- fetchAndArchive forOwner couponCid
+      assertWithinDeadline "RewardCouponV2.expiresAt" coupon.expiresAt
```
moritzkiefer-da, line 1321: What's our rule for when we hard enforce this vs rely on expiry automation? For example expired amulets as transfer inputs we we allow. Not opposed to hard enforcement, just trying to understand it.

daml/splice-amulet/daml/Splice/AmuletRules.daml
```diff
@@ -1284,22 +1439,24 @@ summarizeTransfer sender openRoundNumber amuletPrice transferConfigAmulet inp pr
...
+          when featured $
+            forA_ (Map.toList groupedBeneficiaries) $ \(beneficiary, weight) ->
+              void $ create FeaturedAppActivityMarker
+                with
+                  dso = csum.dso
+                  provider
+                  beneficiary = beneficiary
+                  weight = weight
+        else do
+          require "beneficiaries are unset if issueAppRewards is false" (optional True null beneficiaries)
```
moritzkiefer-da, line 1458: Hm, this seems like a relatively brutal switchover. I wonder if it would be less disruptive to just ignore them for a bit and only make it a hard error later. Of course ideally people rely on the package version feature support to set this which makes this a non-issue but I wouldn't be surprised if a lot of apps end up hardcoding this (maybe even in Daml) and I'm a bit wary of dealing with the fallout of this.

daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestRewardAccountingV2.daml
```diff
@@ -0,0 +1,130 @@
...
+
+import Splice.Testing.Registries.AmuletRegistry.Parameters (defaultAmuletConfig)
+
+
+-- | Test that reward accounting can be driven via DsoRules.
+--
+-- We focus on the core logic only, as the extensive functional testing is done
+-- directly on `AmuletRules` in `TestRewardAccountingV2`, and the choices in DsoRules
+-- are pass-through choices to `AmuletRules` that don't have any additional logic.
+test_reward_accounting_v2 : Script ()
```
moritzkiefer-da: maybe add a test that the archival choice for the dry runs works?

moritzkiefer-da (approved):
Great work thanks!

daml/splice-amulet/daml/Splice/Amulet.daml
meiersi-da: I was thinking of how to make it easy to write the chaining code that shares rewards among more than 20 beneficiaries. However, I realize now that retrieving the head element is as good as retrieving the change from the optional return value.

I'll switch to a hard-equal which also takes care of your second comment.

daml/splice-amulet/daml/Splice/Amulet.daml
meiersi-da: Hmm... I'm not sure the ROI is there. That rename is at least a day of work as it pulls through both the code here and the design. I do clearly distinguish the activity records, an then the computation of minting allowances in the design.

daml/splice-amulet/daml/Splice/AmuletRules.daml
meiersi-da: The reasoning here was to limit the impact on the total supply calculation for CC (and thus indirectly its market price) caused by expired, but non-expirable coupons. For amulet that's not a problem as all the ones that are not expired are factored in.

However for coupons that matters, as they are only counted once they were minted. With this approach here, expired coupons do not have to be counted as potential supply increase.

daml/splice-amulet/daml/Splice/Amulet.daml
moritzkiefer-da: alright I don't really care about the name so your choice.

daml/splice-amulet/daml/Splice/AmuletRules.daml
moritzkiefer-da: makes sense

daml/splice-amulet/daml/Splice/AmuletRules.daml
meiersi-da: Can you double-check the code. I definitely didn't intend a hard switch-over, and I believe the code is proper, as the `issueAppRewards` config is kept as-is. What changes for a switched round is the top-level guard in Line 1443: it becomes false and Line 1459 result in a `pure ()`.

daml/splice-amulet-test/daml/Splice/Scripts/TestRewardAccountingV2.daml
```diff
@@ -0,0 +1,535 @@
...
+    processRewardsCids = map fst processRewards
+    calculateRewardsCids = map fst calculateRewards
+
+  -- no left-over processing contracts
+  [] <- query @CalculateRewardsV2 app.dso
+  [] <- query @ProcessRewardsV2 app.dso
+
+  -- check that no coupons were created
+  [] <- query @RewardCouponV2 app.dso
+  pure ()
```
meiersi-da, line 194: @moritzkiefer-da : skipping stuck dry-run test is here

daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestRewardAccountingV2.daml
meiersi-da: I have a test at the `AmuletRules` level here: https://github.com/hyperledger-labs/splice/pull/3964/changes/BASE..edc394a3e4f30960ebfb419ed551ee4f44be8e7b#r2872841450

Would you add one at the DSO level as well, given that the DSO choice is a simple forwarding choice, and will be tested by the integration test for the API call to unstuck a dry-run?

daml/splice-amulet-test/daml/Splice/Scripts/TestRewardAccountingV2.daml
moritzkiefer-da: yes but that does not test the dso governance choice. imho having a test that you don't exercise at all in the test is a bit dodgy. Too easy to get the controllers wrong or something like that.

daml/splice-amulet/daml/Splice/AmuletRules.daml
moritzkiefer-da: ah sorry I misread the assertion, looks good

daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestRewardAccountingV2.daml
moritzkiefer-da: yes I would add the dso level one. I think the inverse is ok: only have the dso one and not the amulet rules one. But having untested choices is imho dodgy and relying on integration tests doesn't work well with our common approach of first changing the Daml code and then doing integration tests as a separate step. I guess in this case this may not happen given that you have a feature branch but I'd rather still follow the rule that each choice is used in a daml script test

meiersi-da: Fair point. Fixed.

daml/splice-amulet-test/daml/Splice/Scripts/TestRewardAccountingV2.daml
meiersi-da: fixed.

daml/splice-amulet/daml/Splice/Amulet.daml
meiersi-da: switched to testing equality of input and output amounts

daml/splice-api-reward-assignment-v1/daml/Splice/Api/RewardAssignmentV1.daml
```diff
@@ -0,0 +1,68 @@
...
+        -- There MUST be at most `(view this).maxNumNewBeneficiaries` new beneficiaries in the list.
+        -- The purpose of this limit is to ensure that traffic cost of
+        -- creating coupons guards the overhead of tracking the created coupons
+        -- for the DSO party. Applications that need to assign to more
+        -- beneficiaries can do multiple assignments in a single Daml
+        -- transaction, as long as they respect the limit for each assignment.
+      extraArgs : ExtraArgs
+        -- ^ Extra arguments for extensibility. Set to empty, unless needed for specific implementations.
+    observer map (.beneficiary) newBeneficiaries
+    controller (view this).beneficiary
```
bame-da: This makes RewardCoupons transferrable. It looks like a classic bearer instrument redeemable with the `dso`. I worry that this could be an issue for some stakeholders.

daml/splice-amulet/daml/Splice/Amulet/RewardAccountingV2.daml
```diff
@@ -0,0 +1,124 @@
...
+import Splice.Amulet
+import Splice.Types
+import Splice.Util
+
+-- | State contract tracking the need to calculate and confirm the app reward amounts for the given round.
+template CalculateRewardsV2 with
+    dso : Party
+    round : Round
+    rewardCouponTimeToLive : RelTime -- ^ The duration after which the rewards for this round should expire.
+    dryRun : Bool -- ^ Whether to only simulate the confirmation without creating any ProcessRewardsV2 contracts.
```
dfordivam: This should be "any app reward contracts" right, like the comment below

meiersi-da: Yes indeed.

daml/splice-amulet/daml/Splice/AmuletRules.daml
```diff
@@ -382,6 +398,8 @@ template AmuletRules
         now <- getTime
         let configUsd = getValueAsOf now configSchedule
         let tickDuration = configUsd.tickDuration
+        let trafficPrice = Some configUsd.decentralizedSynchronizer.fees.extraTrafficPrice
```
dfordivam, line 401: This fixes the downgrade issues I hit on integ test

```
         let tickDuration = configUsd.tickDuration
-        let trafficPrice = Some configUsd.decentralizedSynchronizer.fees.extraTrafficPrice
         let rewardConfig = configUsd.rewardConfig
+        let trafficPrice =
+              if isSome rewardConfig
+                then Some configUsd.decentralizedSynchronizer.fees.extraTrafficPrice
+                else None -- Keep OpenMiningRound downgradable to prior versions for as long as possible
```

meiersi-da: ugh. sorry for missing that one. please apply as a commit here and/or on your branch.

daml/splice-amulet/daml/Splice/AmuletRules.daml
```diff
@@ -445,19 +466,38 @@ template AmuletRules
...
+              round = roundToArchive.round
+              rewardCouponTimeToLive = rewardConfig.rewardCouponTimeToLive
+              dryRun = False
+
+          when (useTrafficBasedAppRewards rewardConfig.dryRunVersion) $ do
+            void $ create CalculateRewardsV2 with
+              dso
+              round = roundToArchive.round
+              rewardCouponTimeToLive = rewardConfig.rewardCouponTimeToLive
+              dryRun = True
```
dfordivam, line 491: It is possible to have two `CalculateRewardsV2` created per round if both `mintingVersion` and `dryRunVersion` are set. Is this scenario intentionally allowed?
or we don't expect it to happen, ie the config would be carefully modified to not allow it? 
or is it that we don't want to allow this to ever happen in this choice, and should change this code to generate only one, ie `dryRun = False` only and ignore `dryRunVersion`   @meiersi-da

meiersi-da: As per our discussion this morning: yes it is. We event want them to be handled by separate trigger instances, so that they can be independently paused via the `SvAppConfig`.

daml/splice-amulet/daml/Splice/Round.daml
dfordivam, line 55: @meiersi-da I believe we want to pass the `trafficPrice` and `rewardConfig` to `SummarizingMiningRound`? I guess that was the intention of adding it to `OpenMiningRound`; to have this data available in SummarizingMiningRoundTrigger?

meiersi-da: What would you expect that trigger to do with this information? Right now the code is organized such that the `Calculate/ProcessRewardsV2` workflow runs to the side of the `SummarizingRound` workflow.

dfordivam: In `SummarizingMiningRoundTrigger` I need to determine the rewardConfig.mintingVersion for the round, and the CC totals for `featuredAppRewardCoupons` based on total traffic for round. I guess I can query the corresponding `OpenMiningRound` to fetch both of this info. But since the  `OpenMiningRound` has been archived, it wont be in ACS, so fetch from update_history will be required.

meiersi-da: Ah I see. I now realize my mistake. I was thinking of the `IssuingMiningRound`. Your request makes sense. Please add the fields to copy the info from the `OpenMiningRound`.

dfordivam: Another related concern is that the test code is still based on `AppRewardCoupon` count. Specifically

```
runNextIssuance
token-standard/splice-token-standard-test/daml/Splice/Testing/Registries/AmuletRegistry.daml

runNextIssuanceD
daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
```

And we even use these APIs as it is in `Splice.Scripts.TestRewardAccountingV2`.
Do you think we should do any modifications to these?

The scala is simply injecting the `totalFeaturedAppRewardCoupons` via offline calcs, so the flow is more or less identical to the existing daml testing even in the case of `RewardVersion_TrafficBasedAppRewards`. And as long as the  `totalFeaturedAppRewardCoupons` is being set properly it would work fine.

dfordivam: We could pass in `Optional Decimal` to `runNextIssuance` and use that alongwith an assert on `rewardConfig`, to keep the logic in sync with the scala.

dfordivam: I did implement this, see PR #4724

meiersi-da: Thanks. yes, good call.

meiersi-da:
all this code has now landed on `main` in its final form

PR #3966, opened 2026-02-16
vagrant: Add more space for cache

giner:
Fresh installation takes 17 GiB of cache image. Increasing it from 20 GiB to 30 GiB.


martinflorian-da (approved):
Looks right to me, thanks!

PR #3975, opened 2026-02-16
Avoid starting flaky sv-2 components in SvOnboardingAddlIntegrationTest

martinflorian-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/7242

[ci]

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvOnboardingAddlIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvOnboardingAddlIntegrationTest.scala
@@ -326,21 +326,9 @@ class SvOnboardingAddlIntegrationTest
       }
 
       clue("Start SV2") {
-        loggerFactory.assertEventuallyLogsSeq(SuppressionRule.Level(Level.ERROR))(
-          startAllSync(
-            sv2ScanBackend,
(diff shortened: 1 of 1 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvOnboardingAddlIntegrationTest.scala
OriolMunoz-da, line 318: should this also be gone?

OriolMunoz-da (approved):
ty

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SvOnboardingAddlIntegrationTest.scala
martinflorian-da: AFAIU no; this is for sv1 where apparently we can be fairly certain that we'll get these logs.

AFAIU the problem is that if we don't get these logs they might happen at some later point in the test... so we need some form of "wait for them to appear".

martinflorian-da: (and just accepting all kinds of errors seems sketchy if we can avoid that)

PR #3987, opened 2026-02-17
Add dependabot.yml for load-tester and pulumi code

### LINKED ISSUE #619: Treat dependabot PRs as [ci]
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#17747). Original author: @moritzkiefer-da ]

Currently a dependabot PR doesn't ctually run CI, seems like we could easily check the PR author or maybe easier the branch name (it always starts with `dependabot/`) and use that to run it.

krzysztofczyz-da:
Add dependabot.yml with assignees.

Fixes #619

#### Cluster Testing
- [x] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [x] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [x] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,8 @@ updates:
       - "martinflorian-da"
     labels:
       - "static"
-    prefix: "[ci] "
+    commit-message:
+      prefix: "[ci] "
   - package-ecosystem: "npm"
     directory: "/load-tester"
     schedule:
@@ -21,7 +22,8 @@ updates:
       - "martinflorian-da"
     labels:
       - "static"
-    prefix: "[ci] "
+    commit-message:
+      prefix: "[ci] "
   - package-ecosystem: "github-actions"
     directory: "/" # For GitHub Actions '/' means .github/workflows/**/*.{yaml,yml}
     # Check for updates to GitHub Actions every week
@@ -33,4 +35,5 @@ updates:
       - "krzysztofczyz-da"
       - "mblaze-da"
       - "martinflorian-da"
-    prefix: "[ci] "
+    commit-message:
+      prefix: "[ci] "


.github/dependabot.yml
```diff
@@ -0,0 +1,33 @@
...
+      - "krzysztofczyz-da"
+      - "mblaze-da"
+      - "martinflorian-da"
+    labels:
+      - "static"
+  - package-ecosystem: "npm"
+    directory: "/load-tester"
+    schedule:
+      interval: "daily"
+    assignees:
```
martinflorian-da, line 19: Can you confirm that this creates issues (and assigns us to them) and not PRs?

If it creates PRs ... that probably won't work (because it probably doesn't know about out `[ci]` / `[static]` labels...) and if it does - we probably want to be added as reviewers and not assignees.

krzysztofczyz-da: > Can you confirm that this creates issues (and assigns us to them) and not PRs?

This creates PRs - dependabot does not create issues AFAICT

> we probably want to be added as reviewers and not assignees

That would mean that we would get added to reviewers on every PR touching those files as dependabot uses [CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) file. Is that okay?

> because it probably doesn't know about out [ci] / [static] labels...)

That piece of config takes care of adding `static` label:
```
    labels:
      - "static"
```

And I've added `[ci] ` as a prefix to commit message 😄 

And also note: this still might not work due to [dependabot restrictions](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-on-actions#restrictions-when-dependabot-triggers-events)

martinflorian-da: Ah missed the label, nice! `[static]` will also be fine if dependabot touches only typescript for now.

> reviewers

Hm gemini lied to me that this also supports an `reviewers:` key. I guess it does not? Assignees seems better then than messing with CODEOWNERS...

martinflorian-da (approved):
Let's give it a try, thanks!

PR #3990, opened 2026-02-17
Support multipart upload of updates to S3

isegall-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnectionForUnitTests.scala
+++ b/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnectionForUnitTests.scala
@@ -25,12 +25,11 @@ class S3BucketConnectionForUnitTests(
       s3Stream =>
         Using.resource(new DataInputStream(s3Stream.asInputStream())) { dataInput =>
           def readAll(): Vector[Array[Byte]] =
-            Vector.unfold(()){ _ =>
+            Vector.unfold(()) { _ =>
               Try {
                 PaddedData.readNext(dataInput)
-              }
-                .toOption
-                .map{d => (d.data.array(), ())}
+              }.toOption
+                .map { d => (d.data.array(), ()) }
             }
 
           def concatenate(chunks: Vector[Array[Byte]]): ByteBuffer = {


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
```diff
@@ -62,6 +69,101 @@ class S3BucketConnection(
...
+          .key(key)
+          .uploadId(id)
+          .partNumber(partNumber)
+          .build()
+        response <- s3Client
+          .uploadPart(partRequest, AsyncRequestBody.fromByteBuffer(content))
+          .asScala
+      } yield {
+        blocking {
+          synchronized {
```
ray-roestenburg-da: I think this be done without `blocking { synchronized`?

Better to never do blocking I/O from within synchronized, which is maybe why this was enforced in the project as duct tape, and I'm just going to say once again that's not a good idea 😄

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
```diff
@@ -62,6 +69,101 @@ class S3BucketConnection(
...
+    val createRequest = CreateMultipartUploadRequest
+      .builder()
+      .bucket(bucketName)
+      .key(key)
+      .build();
+
+    private val uploadId = s3Client.createMultipartUpload(createRequest).asScala.map(_.uploadId())
+    private val parts = new AtomicReference(Seq.empty[Option[CompletedPart]])
+
+    /* Must be called sequentially. It is a fast operation that just initializes the next upload slot */
```
ray-roestenburg-da: what do you mean with sequentially? (is this not thread safe)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
isegall-da: I actually think that `blocking` is wrong here, it's not an I/O operation, so only needs `synchronized`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
isegall-da: not thread safe, and the caller should make sure to call it in the right order for the "next" to make sense, i.e. for the parts to be in the right order in the object. Will make the comment more explicit.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
ray-roestenburg-da: I agree should not have both

isegall-da: As discussed, this is also wrong, and is racing with `prepareUploadNext`, I'll fix it.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
isegall-da: @ray-roestenburg-da I believe I resolved the concurrency issues now, please review again. Thanks

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
```diff
@@ -62,6 +72,94 @@ class S3BucketConnection(
...
+          .put(
+            partNumber,
+            CompletedPart
+              .builder()
+              .partNumber(partNumber)
+              .eTag(response.eTag())
+              .build(),
+          )
+          .fold(())(_ =>
+            throw new RuntimeException(s"Part number $partNumber uploaded more than once")
```
ray-roestenburg-da: nit: rather use Future.failed(new RuntimeException(...)) which requires moving up the parts.put in the for comprehension and wrap in future successful 

res <- parts.put(...)
  .fold(Future.successful(()))( Future.failed(...)) 
} yield res

ray-roestenburg-da: (it does the same thing really, so not a huge deal)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistorySegmentBulkStorage.scala
```diff
@@ -119,45 +133,121 @@ class UpdateHistorySegmentBulkStorage(
...
-        // TODO(#3429): For now, we accumulate the full object in memory, then write it as a whole.
-        //    Consider streaming it to S3 instead. Need to make sure that it then handles crashes correctly,
-        //    i.e. that until we tell S3 that we're done writing, if we stop, then S3 throws away the
-        //    partially written object.
-        for {
-          _ <- s3Connection.writeFullObject(objectKey, ByteBuffer.wrap(zstdObj.toArrayUnsafe()))
-        } yield {
-          s3ObjIdx.addAndGet(1)
-          ()
+      .mapAsync(4) {
```
ray-roestenburg-da: should this be configurable? (4)

isegall-da: yeah, I guess. Not sure what I'm more allergic to, config variables that nobody knows what value to chose for them or arbitrary hard-coded numbers :) 
Will add a TODO for now

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnection.scala
```diff
@@ -62,6 +72,98 @@ class S3BucketConnection(
...
+      } yield {
+        res
+      }
+    }
+
+    def finish(): Future[Unit] = {
+      require(numParts.get() > 0)
+      require(
+        parts.size == numParts.get(),
+        "completeMultiPartUpload may not be called before all parts have finished uploading",
```
stephencompall-DA:
```suggestion
        "finish may not be called before all parts have finished uploading",
```

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnectionForUnitTests.scala
```diff
@@ -0,0 +1,119 @@
...
+          def readAll(acc: Vector[Array[Byte]]): Vector[Array[Byte]] = {
+            val nextObject: Try[PaddedData] = Try {
+              PaddedData.readNext(dataInput)
+            }
+
+            nextObject.toOption match {
+              case Some(data) => readAll(acc :+ data.data.array())
+              case None => acc
+            }
+          }
```
stephencompall-DA:
```suggestion
          def readAll(acc: Vector[Array[Byte]]): Vector[Array[Byte]] =
            Vector.unfold(()){ _ => 
              Try {
                PaddedData.readNext(dataInput)
              }.toOption.map((_, ()))
            }
```

PR #3991, opened 2026-02-17
Use multipart uploads also for ACS snapshots

isegall-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistorySegmentBulkStorage.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistorySegmentBulkStorage.scala
@@ -33,6 +33,19 @@ case class UpdatesSegment(
     toTimestamp: TimestampWithMigrationId,
 )
 
+/** Pekko source for dumping all updates from a segment to S3 objects.
+  * Reads updates from the updateStore, encodes and compresses them
+  * into chunks of size >=config.bulkZstdFrameSize. Each chunk is a frame
+  * in zstd terms (i.e. a complete zstd object). The chunks are written into
+  * s3 objects of size >=config.bulkMaxFileSize (as multi-frame zstd objects, which
+  * are simply a concatenation of zstd objects), using multi-part upload (where
+  * each chunk/frame is a part in the upload).
+  * Whenever an object is fully written, the source emits an Output object
+  * with the segment details, the name of the object just written (useful for monitoring
+  * progress and testing), and a flag of whether this is the last object in this
+  * segment (useful when streaming a sequence of segments, so that we can easily
+  * know when each segment is complete).
+  */
 class UpdateHistorySegmentBulkStorage(
     val config: ScanStorageConfig,
     val updateHistory: UpdateHistory,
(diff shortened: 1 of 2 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3ZstdObjects.scala
```diff
@@ -0,0 +1,154 @@
...
+          chunk.obj.upload(chunk.partNumber, chunk.bytes.asByteBuffer).map(_ => chunk)
+        case chunk => Future.successful(chunk)
+      }
+      .mapAsync(1) {
+        case chunk: ObjectChunk if chunk.isLastChunkInObject =>
+          logger.debug(
+            s"Finished uploading part ${chunk.partNumber}, which is the last one for the object ${chunk.obj.key}, finishing the upload"
+          )
+          chunk.obj
+            .finish()
```
stephencompall-DA, line 109: If we never get here, are the previous uploaded chunks still...uploaded?

isegall-da: no, s3 is write-once, even for multi-part upload. So unless you finish the upload, nothing gets written. But the whole pipeline is built to be retryable, and on any failure, we just retry the whole thing from the last persisted point.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3ZstdObjects.scala
```diff
@@ -0,0 +1,154 @@
...
+          case (state, chunk) if state.s3ObjSize + chunk.bytes.length > config.bulkMaxFileSize =>
+            logger.debug(
+              s"Adding a chunk of ${chunk.bytes.length} bytes. The object size so far has been: ${state.s3ObjSize}, together they cross the threshold of ${config.bulkMaxFileSize}, so this is the last chunk for the object"
+            )
+            (
+              State(
+                s3Connection.newAppendWriteObject(
+                  getObjectKey(state.s3ObjIdx + 1)
+                ),
+                state.s3ObjIdx + 1,
```
ray-roestenburg-da, line 59: Some cleanup: could add a def nextIndex on state, which does this s3ObjIdx + 1

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3ZstdObjects.scala
```diff
@@ -0,0 +1,154 @@
...
+                chunk.isLast,
+                state.obj.prepareUploadNext(),
+              ),
+            )
+          case (state, chunk) =>
+            logger.debug(
+              s"Adding a chunk of ${chunk.bytes.length} bytes. The object size so far has been: ${state.s3ObjSize}, together they are not yet at the threshold of ${config.bulkMaxFileSize}"
+            )
+            (
+              State(
```
ray-roestenburg-da, line 75: This also feels like you could add a def on state that returns this state stateFromChunk or whatever

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3ZstdObjects.scala
isegall-da: tried it now, decided against.
Made the flow harder to follow IMHO by moving this logic somewhere else, with very little upside (it's not that that def is reused anywhere else for exampel)

PR #3993, opened 2026-02-17
[ci] add bulk storage metrics

isegall-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/SingleAcsSnapshotBulkStorage.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/SingleAcsSnapshotBulkStorage.scala
@@ -22,7 +22,6 @@ import io.circe.syntax.*
 import java.nio.charset.StandardCharsets
 import scala.concurrent.duration.FiniteDuration
 import Position.*
-import com.daml.metrics.api.MetricsContext
 import org.apache.pekko.{Done, NotUsed}
 import org.lfdecentralizedtrust.splice.scan.config.ScanStorageConfig
 
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistorySegmentBulkStorage.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistorySegmentBulkStorage.scala
@@ -3,7 +3,6 @@
 
 package org.lfdecentralizedtrust.splice.scan.store.bulk
 
-import com.daml.metrics.api.MetricsContext
 import com.digitalasset.canton.logging.{NamedLoggerFactory, NamedLogging}
 import com.digitalasset.canton.tracing.TraceContext
 import org.apache.pekko.NotUsed


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/SingleAcsSnapshotBulkStorage.scala
```diff
@@ -86,6 +88,11 @@ class SingleAcsSnapshotBulkStorage(
           loggerFactory,
         )
       )
+      .map(o => {
+        historyMetrics.BulkStorage.objectsCount
+          .inc()(MetricsContext("object_type" -> "ACS_snapshots"))
```
ray-roestenburg-da: Minor cleanup, you could put defs on BulkStorageMetrics (or on BulkStorage) that creates the metrics, instead of having that defined here. See HttpClientMetrics, and I think we do that in a couple of other places as well

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistorySegmentBulkStorage.scala
```diff
@@ -135,9 +138,10 @@ class UpdateHistorySegmentBulkStorage(
           loggerFactory,
         )
       )
-      .map((o: S3ZstdObjects.Output) =>
+      .map((o: S3ZstdObjects.Output) => {
+        historyMetrics.BulkStorage.objectsCount.inc()(MetricsContext("object_type" -> "updates"))
```
ray-roestenburg-da: add a def to historyMetrics that does this, and you just call it here

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorageTest.scala
```diff
@@ -90,6 +96,17 @@ class AcsSnapshotBulkStorageTest
           objectKeys.foreach(
             _.key() should startWith("2026-01-02T00:00:00Z-Migration-0-2026-01-03T00:00:00Z/ACS_")
           )
+          val objectCountMetrics = metricsFactory.metrics.counters.get(
+            SpliceMetrics.MetricsPrefix :+ "history" :+ "bulk-storage" :+ "object-count"
```
ray-roestenburg-da, line 100: this you can also define on your Metrics object

isegall-da: I kinda don't want to. I want this test to catch if we unintentionally change the metric name. If the code and the test use the same constant from code, you wouldn't catch that.

ray-roestenburg-da: just have it defined in one place, makes more sense to me in general. Do you mean that the grafana query might not match? this test doesn't really do anything about that right?

isegall-da: The test asserts the name of the metric. So if the name changes (which would in turn break dashboards), the test will catch it. If the test simply takes the name from the same place that the implementation takes it, then the test will never catch regression where the metric name changes.

ray-roestenburg-da: yeah but the dashboard can change without changing the test.. anyway, I'm not too bothered about this

PR #4020, opened 2026-02-18
Added required Helm changes for validator_backend.migrate-validator-party.scan.admin-api.url

pasindutennage-da:
fixes https://github.com/DACH-NY/cn-test-failures/issues/7276

### DIFF
--- a/cluster/helm/splice-validator/templates/validator.yaml
+++ b/cluster/helm/splice-validator/templates/validator.yaml
@@ -223,7 +223,15 @@ spec:
         {{- if  .Values.migrateValidatorParty }}
         - name: ADDITIONAL_CONFIG_MIGRATE_VALIDATOR_PARTY
           value: |
+            {{- if .Values.scanClient}}
+              {{- if eq .Values.scanClient.scanType "trust-single" }}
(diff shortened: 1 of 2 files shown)


cluster/helm/splice-validator/templates/validator.yaml
```diff
@@ -223,7 +223,15 @@ spec:
         {{- if  .Values.migrateValidatorParty }}
         - name: ADDITIONAL_CONFIG_MIGRATE_VALIDATOR_PARTY
           value: |
+            {{- if .Values.scanClient}}
+              {{- if eq .Values.scanClient.scanType "trust-single" }}
+            canton.validator-apps.validator_backend.migrate-validator-party.scan.admin-api.url = {{ .Values.scanClient.scanAddress | quote }}
+              {{- else}}
+            canton.validator-apps.validator_backend.migrate-validator-party.scan.admin-api.url = {{ index .Values.scanClient.seedUrls 0 | quote }}
```
OriolMunoz-da, line 230: for my understanding as to why this is OK, comment says:
```
    // We don't require a BFT read as ACS commitments are a sufficient indicator
    // that something went wrong.
```

PR #4040, opened 2026-02-19
Extend SvAppBackendConfig with a feature flag for permissionedSynchronizer, defaults to false

pasindutennage-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4039

### DIFF
--- a/cluster/deployment/config.yaml
+++ b/cluster/deployment/config.yaml
@@ -160,8 +160,6 @@ sv:
             fillInterval: 60s
 svs:
   default:
-    svApp:
-      permissionedSynchronizer: false  # will be removed once the permissioned synchronizer is the default
     logging:
       appsLogLevel: DEBUG
       appsAsync: false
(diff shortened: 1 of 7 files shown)


cluster/deployment/config.yaml
```diff
@@ -160,6 +160,8 @@ sv:
             fillInterval: 60s
 svs:
   default:
+    svApp:
+      permissionedSynchronizer: false  # will be removed once the permissioned synchronizer is the default
```
martinflorian-da: Any reason to set it here at all, given that this is the default anyway? As you see this will change all resolved configs, and I'm not sure what we gain from having the default re-set also here.

If you just want to make sure that the config gets picked up, consider adding to `mock/config.yaml`?

pasindutennage-da: Thanks, Moving to mock/config.yaml

PR #4044, opened 2026-02-19
add the possibility to configure the number of SVs to upgrade in cncluster partial_upgrade

mblaze-da:
Deno is an alternative runtime for TypeScript that simplifies dependency management in the case of standalone scripts.

I checked that the script also works when called from the `canton-network-internal` repo.

### DIFF
--- a/scripts/partially-upgrade-versions.ts
+++ b/scripts/partially-upgrade-versions.ts
@@ -5,29 +5,58 @@
 
 import { readFileSync } from 'node:fs'
 import process from 'node:process'
+import { command, number, optional, positional, run, string } from 'npm:cmd-ts@0.15.0'
 
-const args = process.argv.slice(2)
-const newVersion = args[0]
-const oldVersionsFile = args[1]
-const numberOfSvsToUpgrade = args.length > 2
-  ? Number.parseInt(args[2])
-  : 1
+function main(
+  newVersion: string,
+  oldVersionsFile: string,
+  numberOfSvsToUpgrade: number,
+) {
+  type ChartVersions = Record<string, Record<string, string>>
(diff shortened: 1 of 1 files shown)


scripts/partially-upgrade-versions.ts
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env -S deno run --allow-all
+
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { readFileSync } from 'node:fs'
+import process from 'node:process'
+
+const args = process.argv.slice(2)
+const newVersion = args[0]
```
martinflorian-da: There is no nice(r) easy way to parse args? Sad.

scripts/partially-upgrade-versions.ts
```diff
@@ -0,0 +1,33 @@
...
+  'splice-participant',
+  'splice-global-domain',
+  'splice-cometbft',
+  'splice-scan',
+  'splice-sv-node',
+  'splice-validator',
+]
+
+for (const sv of svs) for (const chart of svCharts) chartVersions[sv][chart] = newVersion
+for (const chart of ['splice-validator', 'splice-participant', 'splice-splitwell-web-ui'])
```
martinflorian-da: How about replicating the logic for svs here so it's more consistent and easy to understand?

So

`validatorCarts = [ 'splice-validator', ...`

`for (const validator of ['validator1']) for (const chart ...`

scripts/partially-upgrade-versions.ts
mblaze-da: I added some fancier argument parsing just in case other people want to get some quick inspiration.

martinflorian-da: Fancy, thanks!

PR #4047, opened 2026-02-19
add gha resource utilization summary dashboard 

isegall-da:
I started collecting this table manually, then realized there must be a better way:

Screenshot from 2026-02-19 09-35-48

Seems to me like the conclusion is there's not much we need to change right now. So: fixes https://github.com/DACH-NY/canton-network-internal/issues/3433

@martinflorian-da @nicu-da do you agree?

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


nicu-da:
I think it makes sense then, also something to keep in mind when checking resource usage in the future.
I do wonder if it might make sense to lower requests to account for the average usage though, and imo just not have limits. 
What might be worth looking at accounting for bursts is overall node usage as well. Why i created the original issue was because looking at how many resources we were using on one node it seemed like we were wasting lots of resources.
We have fairly large requests, and we usually can fit 1 or max 2 runners on a single node, and it seems that leaves a lot of resources unused during most times.

PR #4084, opened 2026-02-23
Persist external txn hash 

jagathweerasinghe-da:
[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -245,6 +245,15 @@ object SqlIndexInitializationTrigger {
           on round_party_totals (store_id, party, closed_round desc)
         """,
       ),
+    IndexAction
```
moritzkiefer-da, line 248: Have we done some performance tests for this on CILR? How long does it take? Do we need to worry about it overloading the DB while it is running?

moritzkiefer-da: I guess it only has to do a linear scan but it creates an empty index so hopefully not too bad?

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__update_history_for_hash.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE update_history_transactions ADD COLUMN external_transaction_hash bytea;
```
moritzkiefer-da, line 1: can you remind me under which circumstances you can get updates with duplicate tx hashes? If you're unsure yourself probably best to ask someone from Canton.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UpdateHistoryIntegrationTest.scala
```diff
@@ -181,22 +189,100 @@ class UpdateHistoryIntegrationTest
...
       }
 
       eventually() {
         checkUpdateHistoryMetrics(sv1ScanBackend, sv1ScanBackend.participantClient, dsoParty)
       }
     }
-  }
 
+    actAndCheck(
+      "Externally signed txn hash", {
```
moritzkiefer-da, line 210: we already have plenty of tests that create externally signed transactions. Can we add an assertion to one of those instead of adding it to these test? Creating a bunch of transactions takes a non-trivial amount of time so this is a fairly expensive assertion you're adding here.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
jagathweerasinghe-da: It took around 300s on CILR, thats why we moved the index creation to the app.

moritzkiefer-da: thx, that sounds quite reasonable

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__update_history_for_hash.sql
jagathweerasinghe-da: Product people said duplicates possible. But, I'll double check with Canton team.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UpdateHistoryIntegrationTest.scala
jagathweerasinghe-da: Got it. WDYT about adding assertions in ExternalPartySetupProposalIntegrationTest?

moritzkiefer-da: sgtm, or alternatively maybe one of the token standard tests given that we do want to deprecate the functionality used in the ExternalPartySetupProposalIntegrationTest

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__update_history_for_hash.sql
jagathweerasinghe-da: Thibault confirmed the hash is unique. 

Am I correct in assuming we are avoiding UNIQUE constraints in SQL to minimize the performance impact on write operations?

ray-roestenburg-da: In general, if a unique constraint can ensure data integrity, we should define them, and we do define some unique indexes in the repo, not defining them for performance reasons seems the wrong way around and pre-optimization (unless we have clear evidence of performance problems and a special situation that requires not having one). 
In this particular case I would say it's fine not to have it, the uniqueness is guaranteed somewhere else, if we somehow do see duplicates when querying we can error out.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UpdateHistoryIntegrationTest.scala
jagathweerasinghe-da: @moritzkiefer-da Moved the integration test to TokenStandardCliIntegrationTest. BTW, I had to open a new PR as I messed up this PR. Please review this: https://github.com/hyperledger-labs/splice/pull/4109. Sorry about that.

PR #4087, opened 2026-02-23
Change Sv1Initializer to initialize synchronizer in permissioned mode if permissionedSynchronizer = true

pasindutennage-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4082

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
@@ -28,7 +28,6 @@ class PermissionedSynchronizerIntegrationTest
       .withManualStart
 
   "onboard validator in permissioned mode" in { implicit env =>
-
     withClue("onboard SV1 in RestrictedOpen mode") {
       initDsoWithSv1Only()
 


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
```diff
@@ -17,68 +18,62 @@ class PermissionedSynchronizerIntegrationTest
...
+              .propose(
+                decentralizedSynchronizerId,
+                validator.participantClientWithAdminToken.id,
+                permission = ParticipantPermission.Submission,
+              )
+          }
+        }
+        clue("Starting SV" + validator.participantClient.id) {
+          sv.startSync()
+          validator.startSync()
```
pasindutennage-da: In the post integration test log sanity checks (as seen in CI failure), this yields "cannot connect to SV3Scan", simply because it is not yet started.

moritzkiefer-da: The issue is because you try to start sv, validator and scan one after the other with startSync so the validator will fail because scan is not up. You can swap the order to sv -> scan -> validator which works but even better is to start them in parallel as that is also faster. You can use `startAllSync` for that and pass sv, validator and scan of all 3 other SVs in one go.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
```diff
@@ -17,68 +18,62 @@ class PermissionedSynchronizerIntegrationTest
...
-        },
-      )(
-        "Verify parameters are updated to RestrictedOpen",
-        _ => {
-          val currentParams = sv1ValidatorBackend.participantClient.topology.synchronizer_parameters
-            .get_dynamic_synchronizer_parameters(decentralizedSynchronizerId)
-
-          currentParams.onboardingRestriction shouldBe RestrictedOpen
-        },
+    "allow SV1 to authorize and start follower SVs 2-4 sequentially" in { implicit env =>
```
moritzkiefer-da: This test seems to assume that `onboard SV1 in RestrictedOpen mode" ran before. That is generally a bit of an antipattern because then just trying to run this test (e.g. filtering through scalatest's `-z` or in intellij or whatever) will not work.

So either remove this assumption or just merge them in one test. In this case, just merging them makes more sense. If you want to group parts within one test (which also shows up in logs) use `clue`

moritzkiefer-da (approved):
Thanks! lgtm minus the comments on the test

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
pasindutennage-da: Thank you, Addressed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
pasindutennage-da: Merged to a single test, and used withClue

PR #4092, opened 2026-02-23
Support transfer instruction expiry DAML

moritzkiefer-da:
Solves https://github.com/hyperledger-labs/splice/issues/4470


daml/splice-amulet/daml/Splice/ExternalPartyConfigState.daml
```diff
@@ -1,3 +1,6 @@
+-- Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
pasindutennage-da, line 1: pre-commit hook forced me!

daml/dars.lock
```diff
@@ -7,7 +7,7 @@ splice-amulet 0.1.13 6e9fc50fb94e56751b49f09ba2dc84da53a9d7cff08115ebb4f6b7a12d0
 splice-amulet 0.1.14 3ca1343ab26b453d38c8adb70dca5f1ead8440c42b59b68f070786955cbf9ec1
 splice-amulet 0.1.15 67fac2f853bce8dbf0b9817bb5ba7c59f10e8120b7c808696f7010e5f0c8a791
 splice-amulet 0.1.16 c208d7ead1e4e9b610fc2054d0bf00716144ad444011bce0b02dcd6cd0cb8a23
-splice-amulet 0.1.17 387b41b7c40b6015f389e820cd65864caab34e94233df3475f735ec0f3a856d5
+splice-amulet 0.1.17 6fd1f5df98014a7649a3fd7794e18ada00e11add9fedc8b60015b862fb309421
```
pasindutennage-da: precommit hook forced me to sbt damlDarsLockFileUpdate

daml/splice-amulet/daml/Splice/AmuletTransferInstruction.daml
```diff
@@ -77,3 +77,19 @@ abortAmuletTransferInstruction instr extraArgs = do
     senderChangeCids = senderHoldingCids
     output = TransferInstructionResult_Failed
     meta = emptyMetadata
+
+updateAmuletTransferInstruction : AmuletTransferInstruction -> ExtraArgs -> Update TransferInstructionResult
+updateAmuletTransferInstruction instr extraArgs = do
+  now <- getTime
```
moritzkiefer-da: use `assertDeadlineExceeded` instead of explicitly checking against now

daml/splice-amulet/daml/Splice/ExternalPartyConfigState.daml
moritzkiefer-da: I think I fixed it in the base branch so if you rebase you shouldn't need it anymore.

daml/splice-amulet/daml/Splice/AmuletTransferInstruction.daml
```diff
@@ -77,3 +77,19 @@ abortAmuletTransferInstruction instr extraArgs = do
     senderChangeCids = senderHoldingCids
     output = TransferInstructionResult_Failed
     meta = emptyMetadata
+
+updateAmuletTransferInstruction : AmuletTransferInstruction -> ExtraArgs -> Update TransferInstructionResult
+updateAmuletTransferInstruction instr extraArgs = do
```
moritzkiefer-da: missing: check that the extra controllers are empty

daml/splice-dso-governance-test/daml/Splice/Scripts/TestDecentralizedAutomation.daml
```diff
@@ -506,3 +513,52 @@ testTransferPreapprovalExpiry = do
...
+        meta = emptyMetadata
+      extraArgs = ExtraArgs with
+        context = emptyChoiceContext
+        meta = emptyMetadata
+
+  submitWithDisclosures' sender.primaryParty transfer.disclosures $ exerciseCmd transfer.factoryCid transfer.arg
+
+  passTime (hours 2)
+ 
+  [(instrCid, _)] <- query @AmuletTransferInstruction app.dso
```
moritzkiefer-da: missing tests:
- check that you have a locked amulet before and it goes away
- check that you really have no amulets before, otherwise the assertion that you have a non-empty set of amulets is redundant
- check that the expiry choice fails before you advance time through `submitMultiMustFail`
- probably also worth doing: check the actual balances not just that the list of amulets is non-empty

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1114,6 +1124,28 @@ template DsoRules with
     -- Amulet management driven directly by the DSO delegates
     ---------------------------------------------------------
 
+    -- | Cleanup choice for expired Amulet transfer instructions.
+    
+    nonconsuming choice DsoRules_Amulet_ExpireTransferInstruction : DsoRules_Amulet_ExpireTransferInstructionResult
+      with
+        amuletTransferInstructionCid : ContractId AmuletTransferInstruction
+        sv : Optional Party
```
moritzkiefer-da: Remove the optional, this is on some of the existing choices because they existed before the argument added and you can only add optional arguments. For new choices there is no reason to make it an optional.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1114,6 +1124,28 @@ template DsoRules with
     -- Amulet management driven directly by the DSO delegates
     ---------------------------------------------------------
 
+    -- | Cleanup choice for expired Amulet transfer instructions.
+    
+    nonconsuming choice DsoRules_Amulet_ExpireTransferInstruction : DsoRules_Amulet_ExpireTransferInstructionResult
```
moritzkiefer-da: We want to directly make this a batch choice that accepts a list of transfer instructions. We didn't do that for all choices in the past and it came back to bite us so I'd rather do it right from the start.

moritzkiefer-da: also make sure to add the stakeholders as choice observers to reduce the views to 1

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1114,6 +1124,28 @@ template DsoRules with
...
+        _ <- getAndValidateSvParty this sv
+        
+        let transferInstructionCid = toInterfaceContractId @TransferInstruction amuletTransferInstructionCid
+
+        result <- exercise transferInstructionCid TransferInstruction_Update with
+          extraActors = []
+          extraArgs = ExtraArgs with
+            meta = emptyMetadata
+            context = ChoiceContext with
+              values = TextMap.fromList [(expireLockKey, toAnyValue True)]
```
moritzkiefer-da: I don't think you can hardcode this to true. The failing test you'll get with this (which indicates that we also should add for this) should be this:

1. Setup the test as you have right now
2. Pass the time until past the expiry of the transfer instruction
3. Exercise LockedAmulet_OwnerExpireLock on the locked amulet in the transfer instruction.
4. Now try to exercise this. It will fail.

The two options we have for fixing it are:
1. Make this a parameter that gets passed into the choice (related to my batch argument: it needs to be configurable for each transfer instruction separately.
2. Hardcode it to false, that always works. The impact is that the locked amulet does not get unlocked. From a DSO perspective we don't really care, expiry is mainly about ACS size reduction and locked vs unlocked doesn't make a meaningful difference. And not sure it makes a big difference for users either given that you can directly use an expired locked amulet as an input to a token standard transfer.

I think I'm still leaning towards 1. Feels cleaner to me. @meiersi-da wdyt?

moritzkiefer-da (commented):
Good work, thanks a lot! Almost there close but a few more edge cases and some testing that are missing.

daml/splice-amulet/daml/Splice/AmuletTransferInstruction.daml
pasindutennage-da: THanks, Done

daml/splice-amulet/daml/Splice/AmuletTransferInstruction.daml
pasindutennage-da: Thanks, added extraActors

daml/splice-dso-governance-test/daml/Splice/Scripts/TestDecentralizedAutomation.daml
pasindutennage-da: Thanks, All added

daml/splice-dso-governance/daml/Splice/DsoRules.daml
pasindutennage-da: Thanks, Fixed

daml/splice-dso-governance/daml/Splice/DsoRules.daml
pasindutennage-da: THanks, Batched

pasindutennage-da: @moritzkiefer-da 

```
nonconsuming choice DsoRules_Amulet_ExpireTransferInstructions : DsoRules_Amulet_ExpireTransferInstructionsResult
      with
        amuletTransferInstructionCids : [ContractId AmuletTransferInstruction]
        sv : Party
      controller sv
      do
        _ <- getAndValidateSvParty this (Some sv)
        
        individualResults <- forA amuletTransferInstructionCids $ \cid -> do
          let transferInstructionCid = toInterfaceContractId @TransferInstruction cid
          exercise transferInstructionCid TransferInstruction_Update with
            extraActors = []
            extraArgs = ExtraArgs with
              meta = emptyMetadata
              context = ChoiceContext with
                values = TextMap.fromList [(expireLockKey, toAnyValue True)]

        return DsoRules_Amulet_ExpireTransferInstructionsResult with
          results = map (DsoRules_Amulet_ExpireTransferInstructionResult) individualResults
```


if we add all stakeholders as choice observers on a batch command, every sender in the batch will be able to see everyone else's private transfer details and identities because they'd all share visibility of the entire transaction tree.

moritzkiefer-da: > if we add all stakeholders as choice observers on a batch command, every sender in the batch will be able to see everyone else's private transfer details and identities because they'd all share visibility of the entire transaction tree.

Yes that is true, but remember that everything the SVs see is supposed to be public and is disclosed through scan. So you're not leaking any information. There is a scalability impact as nodes need to process more but that's ok here.

pasindutennage-da: Added stakeholders

daml/splice-dso-governance/daml/Splice/DsoRules.daml
pasindutennage-da: > Make this a parameter that gets passed into the choice 

Implemented this

daml/splice-amulet/daml/Splice/ExternalPartyConfigState.daml
pasindutennage-da: Thanks, I merged from the base

daml/splice-amulet/daml/Splice/AmuletRules.daml
```diff
@@ -770,6 +776,24 @@ template AmuletRules
              else pure []
          pure AmuletRules_ConvertFeaturedAppActivityMarkersResult with appRewardCouponCids = concat markerCids
 
+    nonconsuming choice AmuletRules_Amulet_ExpireTransferInstructions : AmuletRules_Amulet_ExpireTransferInstructionsResult
+      with
+        inputs : [(ContractId TransferInstruction, Bool)]
```
moritzkiefer-da: let's define a record for this so we can give a useful name to the Bool

daml/splice-amulet/daml/Splice/AmuletRules.daml
```diff
@@ -770,6 +776,24 @@ template AmuletRules
...
+        observers : [Party] 
+      observer observers
+      controller dso
+      do
+        individualResults <- forA inputs $ \(cid, expireLock) -> do
+          exercise cid TransferInstruction_Update with
```
moritzkiefer-da: missing validating that the admin party on the transfer instruction is the DSO

daml/splice-dso-governance-test/daml/Splice/Scripts/TestDecentralizedAutomation.daml
```diff
@@ -506,3 +516,86 @@ testTransferPreapprovalExpiry = do
...
+  [(dsoRulesCid, _)] <- query @DsoRules app.dso
+  [(amuletRulesCid, _)] <- query @AmuletRules app.dso
+
+  [(instrInterfaceCid, maybeView)] <- queryInterface @TransferInstruction app.dso
+  let instrView = fromSomeNote "Need view" maybeView
+
+  let stakeholders = [instrView.transfer.sender, instrView.transfer.receiver]
+
+  let expireBatchChoice = AmuletRules_Amulet_ExpireTransferInstructions with
+        inputs = [(instrInterfaceCid, True)]
```
moritzkiefer-da: We're still missing the test where the locked amulet is already archived beforehand and this then succeeds with `False`

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1114,6 +1118,20 @@ template DsoRules with
...
+    nonconsuming choice DsoRules_Amulet_ExpireTransferInstructions : DsoRules_Amulet_ExpireTransferInstructionsResult
+      with
+        amuletRulesCid : ContractId AmuletRules
+        transferInsts : AmuletRules_Amulet_ExpireTransferInstructions 
+        sv : Party
+      controller sv
+      do
+        _ <- getAndValidateSvParty this (Some sv)
+        
+        res <- exercise amuletRulesCid transferInsts
```
moritzkiefer-da: missing: validate the dso party on amulet rules

daml/splice-amulet/daml/Splice/AmuletRules.daml
pasindutennage-da: Thanks, Done

daml/splice-amulet/daml/Splice/AmuletRules.daml
pasindutennage-da: Addressed

daml/splice-dso-governance-test/daml/Splice/Scripts/TestDecentralizedAutomation.daml
pasindutennage-da: Added test

daml/splice-dso-governance/daml/Splice/DsoRules.daml
pasindutennage-da: Added

daml/splice-amulet/daml/Splice/AmuletRules.daml
```diff
@@ -770,6 +782,31 @@ template AmuletRules
...
 
+    nonconsuming choice AmuletRules_Amulet_ExpireTransferInstructions : AmuletRules_Amulet_ExpireTransferInstructionsResult
+      with
+        inputs : [AmuletRules_ExpireTransferInstructionInput]
+        observers : [Party] 
+      observer observers
+      controller dso
+      do
+        individualResults <- forA inputs $ \input -> do
+          instr <- fetch input.transferInstructionCid
```
moritzkiefer-da: Can we use `fetchCheckedInterface` instead?

daml/splice-dso-governance-test/daml/Splice/Scripts/TestDecentralizedAutomation.daml
```diff
@@ -506,3 +514,170 @@ testTransferPreapprovalExpiry = do
...
+  submitWithDisclosures' sender.primaryParty transfer.disclosures $ exerciseCmd transfer.factoryCid transfer.arg
+
+  passTime (hours 2)
+
+  [(lockedAmuletCid, _)] <- query @LockedAmulet sender.primaryParty
+  
+  submit sender.primaryParty $ exerciseCmd lockedAmuletCid LockedAmulet_OwnerExpireLockV2
+
+  lockedAmuletsAfterOwnerExpire <- query @LockedAmulet sender.primaryParty
+  length lockedAmuletsAfterOwnerExpire === 0
```
moritzkiefer-da:
```suggestion
  lockedAmuletsAfterOwnerExpire === []
```
Usually length assertions are an antipattern since the errors are quite useless if it fails

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1114,6 +1118,25 @@ template DsoRules with
...
+        amuletRulesCid : ContractId AmuletRules
+        transferInsts : AmuletRules_Amulet_ExpireTransferInstructions 
+        sv : Party
+      controller sv
+      do
+        _ <- getAndValidateSvParty this (Some sv)
+        
+        amuletRules <- fetch amuletRulesCid
+        
+        assertMsg "The AmuletRules dso must match the DsoRules dso"
```
moritzkiefer-da: use `fetchChecked` or probably even better here: Instead of checking here add an `expectedDso` (non-optional) argument to the amulet rules choice and check in there. See for example the expectedDso in `AmuletRules_Transfer`

daml/splice-amulet/daml/Splice/AmuletRules.daml
```diff
@@ -770,6 +782,31 @@ template AmuletRules
...
+          let instrView = view instr
+          
+          assertMsg "The DSO must be the admin of the transfer instruction's instrument"
```
moritzkiefer-da: nit: use our `require` helper instead

daml/splice-amulet/daml/Splice/AmuletRules.daml
pasindutennage-da: Fixed

daml/splice-dso-governance-test/daml/Splice/Scripts/TestDecentralizedAutomation.daml
pasindutennage-da: Fixed

daml/splice-dso-governance/daml/Splice/DsoRules.daml
pasindutennage-da: Used use fetchChecked (mainly for simplicity, but happy to change if needed)

daml/splice-amulet/daml/Splice/AmuletRules.daml
pasindutennage-da: Fixed

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: if you use fetchChecked you can drop the assertMsg you still have here. I think the expectedDso option would still be better though

daml/splice-amulet/daml/Splice/AmuletTransferInstruction.daml
```diff
@@ -77,3 +78,14 @@ abortAmuletTransferInstruction instr extraArgs = do
     senderChangeCids = senderHoldingCids
     output = TransferInstructionResult_Failed
     meta = emptyMetadata
+
+updateAmuletTransferInstruction : AmuletTransferInstruction -> [Party] ->ExtraArgs -> Update TransferInstructionResult
```
moritzkiefer-da:
```suggestion
updateAmuletTransferInstruction : AmuletTransferInstruction -> [Party] -> ExtraArgs -> Update TransferInstructionResult
```

daml/splice-amulet/daml/Splice/AmuletTransferInstruction.daml
```diff
@@ -77,3 +78,14 @@ abortAmuletTransferInstruction instr extraArgs = do
...
+updateAmuletTransferInstruction instr extraActors extraArgs = do
+  assertMsg "Amulet transfer expiry does not support extra actors" (null extraActors)
+  assertDeadlineExceeded "Transfer.executeBefore" instr.transfer.executeBefore
+  let twoStepTransfer = standardTransferToTwoStepTransfer instr.transfer
+  senderHoldingCids <- abortTwoStepTransfer twoStepTransfer instr.lockedAmulet extraArgs
+  pure TransferInstructionResult with
+    senderChangeCids = senderHoldingCids
+    output = TransferInstructionResult_Failed
+    meta = emptyMetadata
```
moritzkiefer-da: missing trailing newline, I recommend configuring your editor to always add those

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1118,6 +1124,18 @@ template DsoRules with
...
+    nonconsuming choice DsoRules_Amulet_ExpireTransferInstructions : DsoRules_Amulet_ExpireTransferInstructionsResult
+      with
+        amuletRulesCid : ContractId AmuletRules
+        transferInsts : AmuletRules_Amulet_ExpireTransferInstructions
+        sv : Party
+      controller sv
+      do
+        _ <- getAndValidateSvParty this (Some sv)
+
+        res <- exercise amuletRulesCid (transferInsts with expectedDso = dso)
```
moritzkiefer-da, line 1136: feels a bit weird to silently overwrite `expectedDso`, let's just add an assertion like
```
require ("expected dso " <> show transferInsts.expectedDso <> " matches actual dso " <> show dso) (transferInst.expectedDso == dso)
```

moritzkiefer-da (commented):
Thanks! Let's also add a line in the release notes something like

```
Implement missing TransferInstruction_Update choice for AmuletTransferInstruction to support expiry.
```

PR #4096, opened 2026-02-24
Add trigger to transfer Lsu traffic

nicu-da:
[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/lsu/LogicalSyncUpgradeTransferTrafficTrigger.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/lsu/LogicalSyncUpgradeTransferTrafficTrigger.scala
@@ -61,7 +61,6 @@ class LogicalSyncUpgradeTransferTrafficTrigger(
   protected def completeTask(task: ScheduledTaskTrigger.ReadyTask[TrafficTransferTask])(implicit
       tc: TraceContext
   ): Future[TaskOutcome] = {
-    logger.info(s"Running LSU traffic state transfer for $task")
     for {
       trafficState <-
         currentSynchronizerNode.sequencerAdminConnection.getLsuTrafficControlState()


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/lsu/LogicalSyncUpgradeTransferTrafficTrigger.scala
```diff
@@ -0,0 +1,117 @@
...
+      Future.successful(Seq.empty)
+    } else {
+      for {
+        physicalSynchronizerId <- currentSynchronizerNode.sequencerAdminConnection
+          .getPhysicalSynchronizerId()
+        announcements <- announcements(now, physicalSynchronizerId)
+        successorInitialized <- isSuccessorInitialized()
+      } yield {
+        announcements
+          .filter(_ => successorInitialized)
```
moritzkiefer-da, line 55: Can we check that the traffic was not already imported to avoid this trigger running repeatedly?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/lsu/LogicalSyncUpgradeTransferTrafficTrigger.scala
```diff
@@ -0,0 +1,117 @@
...
+          .map(result => TrafficTransferTask(result.mapping))
+      }
+    }
+  }
+
+  protected def completeTask(task: ScheduledTaskTrigger.ReadyTask[TrafficTransferTask])(implicit
+      tc: TraceContext
+  ): Future[TaskOutcome] = {
+    logger.info(s"Running LSU traffic state transfer for $task")
```
moritzkiefer-da: don't you get this log already from the trigger infra?

moritzkiefer-da (approved):
Nice thanks!

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/lsu/LogicalSyncUpgradeTransferTrafficTrigger.scala
nicu-da: Not exposed afact, but we have the `transferred` state, so it will run twice only if the app is restarted

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/lsu/LogicalSyncUpgradeTransferTrafficTrigger.scala
nicu-da: fair point, though this should be logged only when we actually do the transfer so the more logging the better there

moritzkiefer-da: > though this should be logged only when we actually do the transfer so the more logging the better there

Not sure I agree with that. Just logging for the same thing twice just seems kinda pointless and people copy this to other triggers without thinking that may run more often. The other log lines make sense since they actually indicate different steps but this one seems to not indicate anything useful over the trigger log.

nicu-da: removed

PR #4106, opened 2026-02-24
add a switch for the logical synchronizer deployment mode

mblaze-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4105

### DIFF
--- a/cluster/pulumi/common/src/config/migrationSchema.ts
+++ b/cluster/pulumi/common/src/config/migrationSchema.ts
@@ -32,6 +32,7 @@ export const MigrationInfoSchema = z
       .gte(0),
     version: CnChartVersionSchema,
     releaseReference: GitReferenceSchema.optional(),
+    enableLogicalSynchronizerDeploymentMode: z.boolean().default(false),
     sequencer: z
       .object({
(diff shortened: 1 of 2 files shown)


cluster/pulumi/common/src/config/migrationSchema.ts
```diff
@@ -32,6 +32,7 @@ export const MigrationInfoSchema = z
       .gte(0),
     version: CnChartVersionSchema,
     releaseReference: GitReferenceSchema.optional(),
+    enableLogicalSynchronizerDeploymentMode: z.boolean().default(false),
```
nicu-da, line 35: Let's maybe add it when we have an actual usage, not sure it's worth merging a flag that does nothing

mblaze-da: I must have misunderstood you earlier. I thought you wanted the configuration early.

nicu-da: Ah sorry, for it to be usable we need the ability to set the config but also not create the participant in the sv-canton stack but in the new stack

PR #4110, opened 2026-02-24
Use ExportPartyAcs/ImportPartyAcs for SV party replication

### LINKED ISSUE #3629: Try out Canton ExportPartyAcs/ImportPartyAcs for SV party replication
We don't expect huge improvements from this on our side but it is the recommended way of doing party replication so makes sense for us to also switch to it. It does have some additional safeguards as well which may come in handy.

Note:
1. This cannot be used for "recover from key". Just keep our existing code for that.
2. This cannot be used for HDMs/network-wide DR. Just keep our existing code for that.
3. There is a related endpoint `ClearOnboardingFlag`. The expectation for this issue is not to switch to that and instead stick with our current mode where we first host the party in observing mode and then switch to confirming which solves the same issue. We should eventually switch but less urgent.

Phrased differently: This is essentially just a replacement of the two old export/import calls for party migration with the two new ones but we don't expect more changes to be required. If more changes are needed, report back to Canton.

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/3629

[cluster test](https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/61226/workflows/c18618d2-2887-4a56-bf4c-5cfb48b2c7b0)


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/sponsor/DsoPartyMigration.scala
```diff
@@ -48,31 +50,43 @@ class DsoPartyMigration(
...
+    loggerFactory,
+  )
 
   def authorizeParticipantForHostingDsoParty(
       participantId: ParticipantId
   )(implicit tc: TraceContext): EitherT[Future, DsoPartyMigrationFailure, ByteString] = {
     logger.info(s"Sponsor SV authorizing DSO party to participant $participantId")
     for {
       dsoRules <- EitherT.liftF(dsoStore.getDsoRules())
+      beforeActivationOffset <- EitherT.liftF(readOnlyConnection.ledgerEnd())
```
OriolMunoz-da: https://docs.digitalasset.com/operate/3.4/howtos/operate/parties/party_replication.html#authorize-new-hosting-for-the-party

> To later find the ledger offset of the topology transaction which authorizes the new hosting arrangement, take the current ledger end offset on the source participant as a starting point

The offset used before is after activation, so it didn't seem to work

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminConnection.scala
```diff
@@ -662,6 +720,7 @@ class ParticipantAdminConnection(
           HostingParticipant(
             newParticipant,
             ParticipantPermission.Observation,
+            onboarding = true,
```
OriolMunoz-da: https://docs.digitalasset.com/operate/3.4/howtos/operate/parties/party_replication.html#authorize-new-hosting-on-the-target-participant

> Please ensure the onboarding flag is set with requiresPartyToBeOnboarded = true.

Following the code, that ends up being the case by setting the flag here

OriolMunoz-da: also:

> For 3.5, there's going to be a change regarding the clearance of the onboarding flag (heads-up) – but the corresponding endpoint will still exist as it is documented now. (edited)

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/sponsor/DsoPartyMigration.scala
moritzkiefer-da: Is the implementation here safe under crashes? Can I not end up in a situation where the offset is now after the topology transaction and it fails because of that?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminConnection.scala
```diff
@@ -616,7 +674,7 @@ class ParticipantAdminConnection(
       // New participants are only given Observation rights. We explicitly promote them to Submission rights later.
       // See SvOnboardingPromoteToSubmitterTrigger.
       val newHostingParticipant =
-        HostingParticipant(newParticipant, ParticipantPermission.Observation)
+        HostingParticipant(newParticipant, ParticipantPermission.Observation, onboarding = true)
```
moritzkiefer-da: When exactly is the onboarding flag cleared?

If we do set it, let's also create an issue to remove the step to add the SV first with observation rights. That exists only to work around the lack of onboarding flag so if we do set that now we shouldn't need it.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/sponsor/DsoPartyMigration.scala
OriolMunoz-da: gah, you're right. What's a safe & stable offset to use here? The offset of the contract in `contracts <- dsoStore.lookupSvOnboardingConfirmedByParty(candidateParty)`?

moritzkiefer-da: That should be safe I hope. i guess the alternative is to actually look for the topology transaction, do timestamp => offset and subtract 1? But I'd also report this to the Canton team. This just feels like pretty questionable UX to me.

yves-da: > i guess the alternative is to actually look for the topology transaction

The ExportPartyAcs endpoint searches internally for the effective topology transaction that "activates" the party on the target participant (where the party is going to be newly hosted).
In principle, you could give an ledger offset taken from before starting the replication process. But the caveat is that the endpoint may take longer to complete, as its topology transaction search also takes longer, traversing through the ledger update stream.

What you need to do is to call that endpoint with a starting ledger offset from where it should search for the topology transaction on the ledger. – The "easiest" is to record the current ledger offset just before the topology transaction becomes effective (that is, all entities have authorized sufficiently the new party hosting).

So, definitely, you should not look up the topology transaction yourself, and then trying to map its validFrom (record time) to a ledger offset.

Unfortunately, a better UX is currently not available/possible. The ledger works with offsets, and all its APIs (external/internal) are "update stream" based. – Sadly, there is not yet a ledger API that would allow to "simply look up a topology transaction" on the ledger (hopefully it will come in the future).

If this flow is not workable:
1. Take offset `o0`,
2. Effective topology transaction for the new party hosting
3. ExportPartyAcs taking `o0` as starting point

then we need to better understand why that not possible.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/offboarding/SvOffboardingPartyToParticipantTrigger.scala
```diff
@@ -47,13 +47,26 @@ class SvOffboardingPartyToParticipantProposalTrigger(
...
+      val (svsToOffboard, svsToKeep) =
+        currentHostingParticipants.partition(e => offboardedParticipants.contains(e.participantId))
+      val participantIdsToOffboard = svsToOffboard.map(_.participantId)
+      if (svsToKeep.forall(_.onboarding)) {
+        // realistically this is only a problem in tests (SvOnboardingViaNonFoundingSvIntegrationTest)
+        logger.info(
+          s"Not offboarding participants $participantIdsToOffboard as otherwise we cannot satisfy the DSO threshold."
+        )
+        Seq.empty
+      } else {
```
OriolMunoz-da: context: https://github.com/hyperledger-labs/splice/actions/runs/22642468244/job/65622178405
```
"message":"processTaskWithRetry failed with an unknown exception, not retrying","logger_name":"o.l.s.s.a.s.o.SvOffboardingPartyToParticipantProposalTrigger:SvOnboardingViaNonFoundingSvIntegrationTest/config=64a1957a/SV=sv2","thread_name":"SvOnboardingViaNonFoundingSvIntegrationTest-64a1957a-env-ec-623","level":"WARN","stack_trace":"java.lang.IllegalArgumentException: Received the non-positive 0 as argument, but we require a positive value here.\n\tat com.digitalasset.canton.config.RequireTypes$PositiveNumeric$.$anonfun$tryCreate$2(RequireTypes.scala:293)\n\tat cats.syntax.EitherOps$.valueOr$extension(either.scala:95)\n\tat com.digitalasset.canton.config.RequireTypes$PositiveNumeric$.tryCreate(RequireTypes.scala:293)\n\tat com.digitalasset.canton.config.RequireTypes$PositiveInt$.tryCreate(RequireTypes.scala:265)\n\tat org.lfdecentralizedtrust.splice.config.Thresholds$.governanceThreshold(Thresholds.scala:70)\n\tat org.lfdecentralizedtrust.splice.config.Thresholds$.partyToParticipantThreshold(Thresholds.scala:37)\n\tat org.lfdecentralizedtrust.splice.environment.ParticipantAdminConnection.$anonfun$ensurePartyToParticipantProposal$5(ParticipantAdminConnection.scala:773)\n\tat org.lfdecentralizedtrust.splice.environment.TopologyAdminConnection.org$lfdecentralizedtrust$splice$environment$TopologyAdminConnection$$proposeNewTopologyTransaction$1(TopologyAdminConnection.scala:725)\n\tat org.lfdecentralizedtrust.splice.environment.TopologyAdminConnection$$anonfun$$nestedInanonfun$ensureTopologyMapping$6$1.applyOrElse(TopologyAdminConnection.scala:771)\n\tat org.lfdecentralizedtrust.splice.environment.TopologyAdminConnection$$anonfun$$nestedInanonfun$ensureTopologyMapping$6$1.applyOrElse(TopologyAdminConnection.scala:765)\n\tat scala.concurrent.impl.Promise$Transformation.run(Promise.scala:490)\n\tat 
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminConnection.scala
```diff
@@ -346,16 +350,7 @@ class ParticipantAdminConnection(
...
-        case Right(offset) => Future.successful(offset)
-        case Left(timestamp) =>
-          offsetByTimestamp(synchronizerId, timestamp, force).map { offset =>
-            logger.debug(
-              s"Resolved timestamp $timestamp to offset $offset for $synchronizerId, force=$force"
-            )
-            offset
-          }
-      }
+      offset <- resolveOffset(timestampOrOffset, synchronizerId, force)
```
OriolMunoz-da, line 374: refactor that doesn't currently matter as this is the only user of this

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminConnection.scala
```diff
@@ -370,6 +365,51 @@ class ParticipantAdminConnection(
...
+          party,
+          synchronizerId,
+          targetParticipantId,
+          beforeActivationOffset,
+          waitForActivationTimeout = None, // i.e., default
+          observer,
+        )
+      )
+      chunks <- observer.resultFuture
+    } yield ByteString.copyFrom(chunks.map(_.chunk).asJava)
```
OriolMunoz-da, line 438: loads everything into memory, but so did the one already in use, so that's not new

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminConnection.scala
```diff
@@ -731,7 +789,8 @@ class ParticipantAdminConnection(
     def promoteParticipantToSubmitter(
```
OriolMunoz-da, line 826: this is effectively just clearing the onboarding flag now, but without using the new endpoint. The "problem" with the new one is the same as with the activation:
```
ParticipantAdminCommands.PartyManagement
                  .ClearPartyOnboardingFlag(
                    party,
                    synchronizerId,
                    beginOffsetExclusive, // THIS OFFSET HERE
                    waitForActivationTimeout = None, // i.e., default
                  )
```

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
```diff
@@ -460,8 +461,8 @@ class HttpSvPublicHandler(
           isCandidateOnboardingConfirmed <- isOnboardingConfirmed(candidateParty)
           dsoRules <- dsoStore.getDsoRules()
           isCandidateSv = SvApp.isSvParty(candidateParty, dsoRules)
-          contracts <- dsoStore.lookupSvOnboardingConfirmedByParty(candidateParty)
-          candidateParticipantId = contracts
+          contracts <- dsoStore.lookupSvOnboardingConfirmedByPartyWithOffset(candidateParty)
```
OriolMunoz-da: this is not the offset of the contract, but the latest ingested offset (in which the contract is still active)
so this is not really crash-safe
I _could_ use `contract.created_at - 1m` and resolve the offset (with the refactored method above)

This wasn't working anyway/already if the SvOnboardingConfirmed had expired, so no changes there

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/offboarding/SvOffboardingPartyToParticipantTrigger.scala
OriolMunoz-da: i.e. we tried to offboard sv-1 but that would leave us with a threshold of 0, which is not allowed (good)
That seems to imply that sv-2 was still onboarding=true

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/onboarding/SvClearOnboardingFlagTrigger.scala
OriolMunoz-da, line 1: should be renamed to ClearOnboardingFlag except it's not even using that call

here we do have the caveat about the SvOnboardingConfirmed, where it _could_ be archived before we manage to clear the flag

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminConnection.scala
moritzkiefer-da: I've briefly looked at the code of the new endpoint. What it does internally with the offset is to just translate it to a timestamp. So the process we would go through is:

1. We have a timestamp
2. We translate it to an offset and pass it to Canton
3. Canton translates it back to a timestamp

That seems to strength my suspicion that:

1. The APIs do seem way harder to use than they need to be and we need to feed this back to Canton.
2. We are probably best off going for the silly but simple translation of timestamp -> offset -> offset - 1.

moritzkiefer-da: I'm still happy to keep the change here to not use that endpoint and leave ethat for a later follow up at some point.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
moritzkiefer-da: Having thought about it a bit more, I'm increasingly convinced the timestamp => offset => offset - 1 translation is the better option.

What we are doing here is effectively using the ledger contract to persist the offset before the migration. Even if that may work for us, I don't think requiring client side persistence (it's on-ledger in our case but that doesn't really matter, conceptually it could just as well be stored in a local db) for party migration feels quite off to me. So in the interest of us also being the guinea pig for those APIs I'd rather go for the option that does not require persistence even if it is a bit silly and feed that back to Canton and see if they can improve their APIs.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/offboarding/SvOffboardingPartyToParticipantTrigger.scala
moritzkiefer-da: I don't understand this. My understanding from your change is that you set the onboarding flag exactly when we set the observation permission before. And the thresholds should also always line up.

So how are we hitting this now but did not hit it before? That seems like before we were able to switch it to confirming state earlier?

OriolMunoz-da: (sorry that it took so long to get back to this)
sv-2 now has submission rights, so it seems to be able to submit the sv-1 offboarding, but threshold-wise any onboarding sv is not counted, so that's why it was 0 and failed

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminConnection.scala
OriolMunoz-da: > I'm still happy to keep the change here to not use that endpoint and leave ethat for a later follow up at some point.

I can keep the old one so as long as they're equivalent, which is not entirely clear to me

> The APIs do seem way harder to use than they need to be and we need to feed this back to Canton.

cc @yves-da (since you're already on this PR), s.a. https://github.com/hyperledger-labs/splice/pull/4110#discussion_r2888524436

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/offboarding/SvOffboardingPartyToParticipantTrigger.scala
moritzkiefer-da: > sv-2 now has submission rights, so it seems to be able to submit the sv-1 offboarding

that doesn't really make sense to me. to offboard sv-1 from the party to participant mapping it needs to be in the decentralized namespace, why do submission rights matters

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
```diff
@@ -194,6 +194,40 @@ abstract class TopologyAdminConnection(
     )
   }
 
+  /** Return the transaction that first added the participant to PartyToParticipant
+    * if the participant is still included in the latest state.
+    */
+  def getDsoPartyToParticipantTransaction(
```
OriolMunoz-da, line 200: moved from apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/DsoPartyHosting.scala

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminConnection.scala
```diff
@@ -391,6 +387,53 @@ class ParticipantAdminConnection(
...
+      targetParticipantId: ParticipantId,
+      activationTime: Instant,
+  )(implicit
+      traceContext: TraceContext
+  ): Future[ByteString] = {
+    val observer = new SeqAccumulatingObserver[ExportPartyAcsResponse]
+
+    for {
+      activationOffset <- resolveOffset(Left(activationTime), synchronizerId, force = true)
+      beforeActivationOffset = activationOffset - 1L
```
moritzkiefer-da, line 423: add a comment on why we commit those crimes

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminConnection.scala
```diff
@@ -746,7 +813,32 @@ class ParticipantAdminConnection(
     )
   }
 
-  def ensureHostingParticipantIsPromotedToSubmitter(
+  def clearOnboardingFlag(party: PartyId, synchronizerId: SynchronizerId, activationTime: Instant)(
```
moritzkiefer-da: Am I stupid? I don't see any usage of this

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/onboarding/SvClearOnboardingFlagTrigger.scala
```diff
@@ -28,17 +28,20 @@ import org.lfdecentralizedtrust.splice.sv.automation.singlesv.SyncConnectionStal
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters.CollectionHasAsScala
 
-/** Trigger that promotes participants from observers to submitters for the DSO party.
+/** Trigger that removes the onboarding flag from participants that host the DSO party.
```
moritzkiefer-da, line 31: and bumps the threshold right?

moritzkiefer-da (approved):
Nice! Thanks a lot for pulling through that painful change

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/onboarding/SvClearOnboardingFlagTrigger.scala
OriolMunoz-da: it's below:
```
  * This trigger then clears said flag and respectively updates the Party to Participant threshold
  * (onboarding participants do not count towards the threshold).
  * ```

PR #4116, opened 2026-02-25
Integrate bulk storage in scan app

isegall-da:

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTimeBasedIntegrationTest.scala
isegall-da, line 1: recommend to view this one with "hide whitespace" in github to see the actual change

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/HasS3Mock.scala
```diff
@@ -0,0 +1,130 @@
...
+  val s3ConfigMock = S3Config(
+    "http://127.0.0.1:9090",
+    "bucket",
+    Region.US_EAST_1.toString,
+    "mock_id",
+    "mock_key",
+  )
+
+  private var container: Option[S3MockContainer] = None
+  override def beforeAll(): Unit = startContainer()
```
isegall-da, line 46: moved the container start to beforeAll() rather than at the beginning of specific relevant tests, so that for the integration test, s3mock will be started before the scan app initializes (otherwise we get errors in the logs where the app tries to write to s3 before s3mock is available).

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTimeBasedIntegrationTest.scala
```diff
@@ -366,6 +385,8 @@ class ScanTimeBasedIntegrationTest
 
     val startTime = getLedgerTime
 
+    val (aliceUserParty, _) = onboardAliceAndBob()
```
rautenrieth-da: Why was this moved? Was there an issue with onboarding at the beginning?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/S3Config.scala
```diff
@@ -0,0 +1,24 @@
...
+final case class S3Config(
+    endpoint: String,
+    bucketName: String,
+    region: String,
+    accessKeyId: String,
+    secretAccessKey: String,
+)
+object S3Config {
+  def hideConfidential(config: S3Config): S3Config = {
+    val hidden = "****"
```
ray-roestenburg-da, line 15: Is this for logs? Would it make sense to use `Pretty`?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTimeBasedIntegrationTest.scala
```diff
@@ -529,5 +550,69 @@ class ScanTimeBasedIntegrationTest
...
+      )(acsObjKey)
+      acsAtMidnightFromScan should contain theSameElementsInOrderAs acsAtMidnightFromS3
+
+      def isInTimeRange(u: definitions.UpdateHistoryItemV2) = {
+        val recordTime = CompactJsonScanHttpEncodings()
+          .httpToLapiUpdate(u)
+          .update
+          .update
+          .recordTime
+        recordTime >= startTime && recordTime <= CantonTimestamp.assertFromInstant(lastMidnight)
```
rautenrieth-da: I see that we need the upper bound, because we only waited for bulk storage to catch up to last midnight above.

But why do we need the lower bound? Shouldn't bulk storage contain all updates since genesis at this point?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -2126,9 +2126,17 @@ class UpdateHistory(
     *         And also for past migrations, whether the SV was present in them or not.
     */
   def isHistoryBackfilled(migrationId: Long)(implicit tc: TraceContext) = {
-    sourceHistory
-      .migrationInfo(migrationId)
-      .map(_.exists(i => i.complete && i.importUpdatesComplete))
+    state
```
rautenrieth-da: I would prefer if we don't do this, and instead have the caller wait for `UpdateHistory.isReady` or `UpdateHistory.waitUntilInitialized` before calling _any_ other method on it.

Otherwise we're conflating "no" and "I don't know" results here. In most cases it should not matter since callers are usually just waiting for a "yes", but there could be cases during initialization where you need a definite "yes" or "no" answer.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/HasS3Mock.scala
rautenrieth-da: Why are those errors? Should the app not retry the write for a while before emitting a warning? We should be able to handle intermittent S3 connection problems in production.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/BulkStorage.scala
```diff
@@ -0,0 +1,66 @@
...
+        currentMigrationId,
+        s3Connection,
+        historyMetrics,
+        loggerFactory,
+      ).getSource().toMat(Sink.ignore)(Keep.left).run(),
+    )
+
+  }
+  override def close(): Unit = {
+    killSwitches.foreach(_.shutdown())
```
rautenrieth-da, line 63: Discussed as part of update/verdict ingestion: triggering a kill switch and waiting for the Future from the Sink's materialized value does not guarantee that all async graph stages have finished.

In SpliceLedgerSubscription, we have code for waiting for database operations triggered by async graph stages to complete before finishing shutdown.

Not sure how relevant it is here.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/S3Config.scala
isegall-da: This is consistent with all other obfuscation that we make for configs, it's just moved here

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTimeBasedIntegrationTest.scala
isegall-da: Why not? The more things happen after "start time", the mire test coverage you have, no? Especially if you filter by >= start time

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTimeBasedIntegrationTest.scala
isegall-da: It definitely should. 
I thought I was getting some noise from past tests in the db, but that actually does not make sense, does it? Will look deeper.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/HasS3Mock.scala
isegall-da: Sorry, was imprecise. They're only warnings, and the retries work well and recover from them, but that still fails the log checker in the test, and adding an ignore for this felt wrong.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
isegall-da: Hmmm not sure I love this method throwing exceptions if uninitialized, but I also don't feel too strong, and can move this check to the caller

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/HasS3Mock.scala
rautenrieth-da: Isn't that still too verbose? We should not be getting warnings during retries, only when a retry gives up. Otherwise we'll just end up ignoring the warning in our giant log explorer filter.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTimeBasedIntegrationTest.scala
isegall-da: Seems like filtering on startTime was indeed unecessary, in which case I no longer care about when the onboarding happens, will move it back where it was.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/HasS3Mock.scala
isegall-da: There's different levels of retries. The highest level continues retrying forever, but logs warnings.
We can adjust thresholds in prod if it gets noisy.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/BulkStorage.scala
isegall-da: Naa, I don't think it matters much. Will just continue after restart from where it got. I'd prefer not to delay shutdown here to optimize that.

Issue #4118, opened 2026-02-25
Scan-app: Add temporal ACS store for doing asOf lookups


dfordivam:
Add a temporal ACS store, called ScanTcsStore, that works similarly to an active contract store, but also stores archived contracts in a separate table together with the record time of their archival.

It would only store FeaturedAppRight,  OpenMiningRound, and AmuletRules contracts, as they are required to be looked-up as of round start times. It is a separate store with its own tables to enable it to have its own indexing scheme and pruning schedule to ensure consistent performance.

PR #4126, opened 2026-02-25
CIP-104: Add temporal ACS store for doing asOf lookups

dfordivam:
Fixes https://github.com/hyperledger-labs/splice/issues/4118

Note this does not implement caching, only the DB based queries

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbScanRewardsReferenceStoreTest.scala
+++ b/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbScanRewardsReferenceStoreTest.scala
@@ -19,7 +19,10 @@ import org.lfdecentralizedtrust.splice.store.{HardLimit, Limit, StoreTestBase}
 import org.lfdecentralizedtrust.splice.util.{ResourceTemplateDecoder, TemplateJsonDecoder}
 import slick.jdbc.JdbcProfile
 
-class DbScanRewardsReferenceStoreTest extends StoreTestBase with SplicePostgresTest with AcsJdbcTypes {
+class DbScanRewardsReferenceStoreTest
+    extends StoreTestBase
+    with SplicePostgresTest
+    with AcsJdbcTypes {
 
   "DbScanRewardsReferenceStore" should {
 


apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__scan_tcs_store.sql
```diff
@@ -0,0 +1,56 @@
...
+
+    -- reestablish foreign key constraint as that one is not copied by the LIKE statement above
+    foreign key (store_id) references store_descriptors(id),
+
+    -- record_time of the transaction that archived the contract, in micros since epoch.
+    archived_at bigint not null
+);
+
+create index acs_store_archived_template_temporal
+    on acs_store_archived_template (store_id, migration_id, template_id_qualified_name, created_at, archived_at);
```
meiersi-da:
```suggestion
    on acs_store_archived_template (store_id, migration_id, template_id_qualified_name, created_at) include (archived_at);
```
probably the better option to have cheaper internal btree nodes and statistics

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__scan_tcs_store.sql
```diff
@@ -0,0 +1,56 @@
...
+    featured_app_right_provider text
+);
+
+-- temporal query support: created_at filtering for point-in-time lookups on the live table
+create index scan_tcs_store_active_temporal
+    on scan_tcs_store_active (store_id, migration_id, template_id_qualified_name, created_at);
+
+-- lookup FeaturedAppRight
+create index scan_tcs_store_active_sid_mid_tid_farp
+    on scan_tcs_store_active (store_id, migration_id, template_id_qualified_name, featured_app_right_provider)
```
meiersi-da:
```suggestion
    on scan_tcs_store_active (store_id, migration_id, featured_app_right_provider)
```

seems enough or not?

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__scan_rewards_reference_store.sql
```diff
@@ -0,0 +1,56 @@
...
+
+create table acs_store_archived_template(
+    like acs_store_template including all,
+
+    -- reestablish foreign key constraint as that one is not copied by the LIKE statement above
+    foreign key (store_id) references store_descriptors(id),
+
+    -- record_time of the transaction that archived the contract, in micros since epoch.
+    archived_at bigint not null
+);
```
meiersi-da, line 12: Do we need this given that we use `like <actual_store_table_active>` below?

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__scan_tcs_store.sql
```diff
@@ -0,0 +1,56 @@
...
+    -- reestablish foreign key constraint as that one is not copied by the LIKE statement above
+    foreign key (store_id) references store_descriptors (id),
+
+    -- record_time of the transaction that archived the contract, in micros since epoch.
+    archived_at bigint not null
+);
+
+-- temporal query support: created_at + archived_at filtering for point-in-time lookups on the archive table
+create index scan_tcs_store_archived_temporal
+    on scan_tcs_store_archived (store_id, migration_id, template_id_qualified_name, created_at, archived_at);
```
meiersi-da:
```suggestion
    on scan_tcs_store_archived (store_id, migration_id, template_id_qualified_name, created_at) include (archived_at);
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -152,6 +155,26 @@ final class DbMultiDomainAcsStore[TXE](
...
+      retryProvider
+        .waitUnlessShutdown(s.offsetChanged.future)
+        .failOnShutdownTo {
+          io.grpc.Status.UNAVAILABLE
+            .withDescription(
+              s"Aborted waitUntilRecordTimeReached($asOf), as RetryProvider(${retryProvider.loggerFactory.properties}) is shutting down in store $acsStoreDescriptor"
+            )
+            .asRuntimeException()
+        }
+        .flatMap(_ => waitUntilRecordTimeReached(asOf))
```
meiersi-da: @nicu-da : I remember you fixing some memory leak problems wrt this kind of recursive loop on futures. Is this construction here safe?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1333,11 +1425,13 @@ final class DbMultiDomainAcsStore[TXE](
                   .mapValues(trees =>
                     CantonTimestamp.assertFromInstant(trees.last.tree.getRecordTime)
                   )
+                val maxRecordTime = synchronizerIdToRecordTime.values.maxOption
```
meiersi-da: This looks wrong. Record time from different synchronizers must not be interchanged. Given that we are adjusting the multi-domain store, we need to use a vector clock for record time tracking; i.e., use a `Map[SynchronizerId, CantonTimestamp]` both at the tracking level, and for waiting. 

The wait in the `ScanVerdictIngestion` then needs to use the synchronizer-id of the global synchronizer to wait on.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
meiersi-da, line 1418: It probably is there, but I can't find it: can you point me to where you are updating the last ingested record times for the `OffsetCheckpoint`?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/TemporalAcsQueries.scala
```diff
@@ -0,0 +1,110 @@
...
+  AcsStoreId,
+  SelectFromAcsTableResult,
+  SelectFromAcsTableWithStateResult,
+}
+import slick.dbio.Effect
+import slick.jdbc.canton.ActionBasedSQLInterpolation.Implicits.actionBasedSQLInterpolationCanton
+import slick.jdbc.canton.SQLActionBuilder
+import slick.sql.SqlStreamingAction
+
+trait TemporalAcsQueries extends AcsQueries {
```
meiersi-da: Side note: @adetokunbo -- renaming `DbAppActiviyRecordStore` to `AppActivityRecordQueries` would seem to be more clear, as it really is mostly a helper class to build the queries akin to the `TemporalAcsQueries` here.

Food for thought. Not key to change, but might help with code clarity.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbTemporalAcsStoreTest.scala
```diff
@@ -0,0 +1,236 @@
...
+      IngestionConfig(),
+      defaultLimit = HardLimit.tryCreate(Limit.DefaultMaxPageSize),
+      acsArchiveConfigOpt = acsArchiveConfigOpt,
+    )
+  }
+
+  protected def c(i: Int): Contract[AppRewardCoupon.ContractId, AppRewardCoupon] =
+    appRewardCoupon(i, dsoParty, contractId = validContractId(i))
+
+  "DbTemporalAcsStore" should {
```
meiersi-da: missing test for `waitUntilRecordTimeReached`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanTemporalAcsStore.scala
```diff
@@ -0,0 +1,105 @@
...
+import org.lfdecentralizedtrust.splice.migration.DomainMigrationInfo
+import org.lfdecentralizedtrust.splice.scan.store.{ScanStore, ScanTemporalAcsStore}
+import org.lfdecentralizedtrust.splice.store.{Limit, TemporalAcsStore}
+import org.lfdecentralizedtrust.splice.store.MultiDomainAcsStore.ContractCompanion
+import org.lfdecentralizedtrust.splice.store.db.{DbAppStore, DbMultiDomainAcsStore, StoreDescriptor}
+import org.lfdecentralizedtrust.splice.util.{ContractWithState, TemplateJsonDecoder}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class DbScanTemporalAcsStore(
```
meiersi-da: Consider using the `Tcs` as an abbreviation for `TemporalContractStore` analogous to `Acs` for `ActiveContractStore`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
```diff
@@ -0,0 +1,105 @@
...
+      companion: C,
+      asOf: CantonTimestamp,
+      limit: Limit,
+  )(implicit
+      companionClass: ContractCompanion[C, TCid, T],
+      traceContext: TraceContext,
+  ): Future[Seq[ContractWithState[TCid, T]]] =
+    multiDomainAcsStore.listContractsAsOf(companion, asOf, limit)
+
+  def lookupFeaturedAppRightsAsOf(
```
meiersi-da, line 85: we're missing the one for looking them up by party. Ideally as a batch call.

meiersi-da: and in a later PR with caching.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanTemporalAcsStore.scala
```diff
@@ -0,0 +1,59 @@
...
+package org.lfdecentralizedtrust.splice.scan.store
+
+import com.digitalasset.daml.lf.data.Time.Timestamp
+import com.digitalasset.canton.topology.PartyId
+import org.lfdecentralizedtrust.splice.codegen.java.splice
+import org.lfdecentralizedtrust.splice.scan.store.db.ScanTemporalAcsTables.ScanTemporalAcsStoreRowData
+import org.lfdecentralizedtrust.splice.store.{AppStore, MultiDomainAcsStore}
+import org.lfdecentralizedtrust.splice.store.db.AcsInterfaceViewRowData
+
+trait ScanTemporalAcsStore extends AppStore {
```
meiersi-da: Consider repeating the comment wrt purpose here from design doc. Goal is to inform future readers as to the tight scoping of this. Perhaps a name that relates to that would help `ScanRewardsReferenceDataStore` might be an option.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
nicu-da: After the fix in https://github.com/DACH-NY/canton-network-node/pull/9212 it should be safe. Modern futures are also pretty safe in chaining this way as they will flatten the recursion loop but it can still cause issues. Main issue if the `waitUntilRecordTimeReached` is called from a "hotspot" in the ingestion pipeline with a asOf which is relatively far in the future, thus we will have large number of `waitUntilRecordTimeReached` running in loops for each ingestion from the looks of it. 
Ideally we should reverse and just add a promise that gets fulfilled when the ingestion reaches the asOf timestamp, removing the constant looping.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__scan_rewards_reference_store.sql
dfordivam: This is used only for the unit test of `DbTemporalAcsStoreTest`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
dfordivam: Hmmm. did we need [Party] based lookup in our flow?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: At the moment the temporal store APIs have no external endpoints, and a single automation doing lookups, hence I went with the simpler approach. I doubt we will ever be exposing this to an external endpoint. The lookup record_time also would typically not be far in future. So I will let this be as it is for now.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__scan_tcs_store.sql
dfordivam: I will investigate

meiersi-da: also make the index partial to exclude `NULL`s. That ensures a small index size.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__scan_rewards_reference_store.sql
meiersi-da: Consider renaming it to `test_` instead of `template_` then.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
meiersi-da: > The lookup record_time also would typically not be far in future. So I will let this be as it is for now.

This is likely true. It is though one of these things that's going to be expensive to rediscover as part of scalability work. Let's create at least a tech-debt item the Scalability milestone to switch to a promise-based implementation.

meiersi-da: The situation where it may not be true are catch-up situations. The annoying thing is that these are exceptional, but operationally critical circumstances. So keeping them in mind during development is valuable.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
meiersi-da: We'll need the featured app status for all confirmers in the verdict of a confirmation request. My intuition was that organizing the code such that we pre-resolve all of them in a single call is not too onerous and has better performance properties.

meiersi-da: Happy to postpone that though as a performance optimization, so we get to a working implementation more quickly.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__scan_tcs_store.sql
meiersi-da: actually I just realized that we will typically be looking up recent timestamps, which means changing the index to efficiently skip over past events is valuable.

Concretely
```suggestion
    on acs_store_archived_template (store_id, migration_id, template_id_qualified_name, archived_at) include (created_at);
```

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__scan_rewards_reference_store.sql
dfordivam: Done

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__scan_tcs_store.sql
dfordivam: Done

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__scan_tcs_store.sql
dfordivam: I got rid of this as discussed yesterday we won't be doing lookup on party IDs

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V057__scan_tcs_store.sql
dfordivam: Done, ie using archived_at in index

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: Ok, I fixed this to work similar to how offset based wakeup of Futures is currently being done.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: Fixed

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: Actually I missed it. Added it now, along with a test.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbTemporalAcsStoreTest.scala
dfordivam: Added

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanTemporalAcsStore.scala
dfordivam: Done

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
dfordivam: As discussed yesterday on call, we plan on doing a single AsOf lookup of all parties per round.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanTemporalAcsStore.scala
dfordivam: Added

PR #4146, opened 2026-02-25
Revert "Revert "create a new Pulumi project to enable separation of the participant for LSU (#4129)""

mblaze-da:
Reverts hyperledger-labs/splice#4139


martinflorian-da (approved):
thanks but needs `[static]`

PR #4153, opened 2026-02-25
[ci] switch s3mock to s3proxy

isegall-da:
s3mock in-process does not work well, and the s3mock container (which is what we've been using so far) implies a docker dependency, which I wanted to get rid of to avoid introducing that dependency into the scan integration test. This seems to work fine too.

Argh, but then it started flaking..... `Unable to execute HTTP request: Connection refused: /127.0.0.1:9090` (in an integration test that uses this, [here](https://github.com/hyperledger-labs/splice/actions/runs/22424328284/job/64929092755))

Might fall back to s3mock again after all... :thinking: 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -65,7 +65,7 @@ object Dependencies {
   lazy val aws_s3 = "software.amazon.awssdk" % "s3" % "2.30.22"
 
   lazy val s3proxy = "org.gaul" % "s3proxy" % "3.0.0" % Test
-  
+
   lazy val jclouds_fs = "org.apache.jclouds.api" % "filesystem" % "2.5.0" % Test
 
 }
--- a/test-full-class-names-local-net-based.log
+++ b/test-full-class-names-local-net-based.log
@@ -1,3 +1,1 @@
 org.lfdecentralizedtrust.splice.integration.tests.LocalNetFrontendIntegrationTest
-org.lfdecentralizedtrust.splice.scan.store.bulk.AcsSnapshotBulkStorageTest
-org.lfdecentralizedtrust.splice.scan.store.bulk.UpdateHistoryBulkStorageTest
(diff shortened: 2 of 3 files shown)


project/Dependencies.scala
```diff
@@ -62,7 +62,10 @@ object Dependencies {
 
   lazy val zstd = "com.github.luben" % "zstd-jni" % "1.5.7-6"
 
-  lazy val aws_s3 = "software.amazon.awssdk" % "s3" % CantonDependencies.aws_version
+  lazy val aws_s3 = "software.amazon.awssdk" % "s3" % "2.30.22"
```
rautenrieth-da, line 65: [Canton main defines](https://github.com/DACH-NY/canton/blob/main/project/Dependencies.scala#L41C3-L41C34) `aws_version = "2.41.1"`. Can we just bump our own `aws_version` instead of overriding it here?

ray-roestenburg-da:
What makes s3mock testcontainers not work, just curious? for integration tests testcontainers seems like a fine option to me..

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/HasS3Mock.scala
```diff
@@ -7,54 +7,81 @@ import io.grpc.netty.shaded.io.netty.buffer.{ByteBufInputStream, Unpooled}
...
+  val s3Config = S3Config(
+    "http://localhost:9090",
+    "bucket",
+    Region.US_EAST_1.toString,
+    "mock_id",
+    "mock_key",
+  )
+
+  // Note: we used Adobe s3mock before. It worked well in a container, but that adds a docker dependency,
+  // and it doesn't work well in-process (restarts are a pain).
```
ray-roestenburg-da, line 40: Can't follow completely, what do you want to run in-process here, which restarts are a pain? Don't know all the details of course but it just seems to make sense to do this in testcontainers, is it something about running it in CI that makes it more difficult or something like that?

isegall-da: I was trying to avoid a dependency on docker because unit tests run on runners that don't have docker. But I'm no longer convinced I can make this thing reliable enough, so might need to instead make our CI setup even more complex and add an option to have docker for some integration tests.

isegall-da: > because unit tests run on runners that don't have docker

Sorry, meant integration tests

ray-roestenburg-da: 😢 for more complexity but makes sense

PR #4155, opened 2026-02-26
Sketch reward reassignment API

meiersi-da:
[no ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/daml/splice-amulet/daml/Splice/Amulet/RewardAccountingV2.daml
+++ b/daml/splice-amulet/daml/Splice/Amulet/RewardAccountingV2.daml
@@ -28,10 +28,6 @@ overall:
 - separate the minting API and its implementation into its own PR
 
 
-api:
-- introduce api package: splice-api-reward-assignment-v1
-  - take a note wrt integration with HoldingV2 interface introduced by token standard v2
-
 tests:
 - minting delegation of RewardCouponV2
 


daml/splice-api-reward-minting-v1/daml/Splice/Api/RewardMintingV1.daml
```diff
@@ -8,26 +8,74 @@ import Splice.Api.Token.MetadataV1
...
+      newBeneficiaries : [RewardBeneficiary]
+        -- ^ The new beneficiaries to assign part of the minting amount of the
+        -- coupon. Left-over change is assigned to the original beneficiary.
+        --
+        -- There MUST be at most `maxNumNewBeneficiaries` new beneficiaries, as specified in the coupon view.
+      extraArgs : ExtraArgs
+        -- ^ Extra arguments for extensibility. Set to empty, unless needed for specific implementations.
+    observer map (.beneficiary) newBeneficiaries
+    controller (view this).beneficiary
+    do rewardCoupon_assignBeneficiariesImpl this self arg
```
meiersi-da: @moritzkiefer-da : I've created the API we discussed yesterday. I realize though that:

1. the API may well lead to some app providers creating lots of small reward coupons ==> creating high load on DSO ACS size and DSO processing
2. the API may lead to parties using reward reassignment as an alternative form of CC transfers ==> not sure about the impact
3. the API does not obviate the need for implementing self-service reward minting for external parties

So I see three options for the first delivery of traffic-based app rewards:

1. Only implement the RewardCoupon interface, and require reward minting to be done using `AmuletRules_Transfer` for parties with short prepare-submission delays, and `MintingDelegation` otherwise.
2. Only implement the `RewardMintingFactory` interface, and require apps that need to redistribute rewards to use that.
3. Implement both options for maximum flexibility.

I'm still leaning towards Option 1 as it has the lowest implementation effort for both Splice and app providers. The DSO ACS size problem should not become much worse than for CC, as the short, DSO-controlled expiry on our coupons serves the same purpose as holding fees, and our efforts to make traffic cost observable will lead to people starting to observe the cost of creating many coupons.

The alternative form of transfer also seems OK, as the original source of the coupon's amount is always visible, and app providers can guard against minting coupons from sources they don't like.

wdyt?

moritzkiefer-da: > the API may well lead to some app providers creating lots of small reward coupons ==> creating high load on DSO ACS size and DSO processing

You can't really prevent that can you? Even if you don't provide the API you'll get some people that just try to do the same by minting to one party and then splitting (if they're ok with that from a legal perspective). The protection against that should just be that traffic costs make this unattractive. I guess the one difference may be that we don't have merging for rewards. But given that they expire relatively quickly that seems ok.

> the API may lead to parties using reward reassignment as an alternative form of CC transfers ==> not sure about the impact

Hm yeah, so I guess the question is: What do you actually gain from this API as opposed to having multiple outputs on your minting transaction (which at least afaiu is not an issue from an accounting/legal perspective). I can come up with these points:

1. You don't have the issue with unavailable counterparties as they're not confirmers. Ofc you still have the vetting issues. IIRC when we discussed with Bernhard he didn't seem worried about that being a limitation at all so that's a relatively weak argument.
2. It works for external parties. That's obviously valuable. But this on its own doesn't let you do reward minting so in combination with your suggestion 1 it falls apart again.
3. You don't have to setup authorization from your receiver to be able to do the mint.

So it seems like 3 is really the main argument initially. I think I'm leaning towards this API (in combination with your option 1) being simple enough and it maybe makes it easier to do full minting for external parties as a later iteration that it's worth doing.

meiersi-da: > So it seems like 3 is really the main argument initially. I think I'm leaning towards this API (in combination with your option 1) being simple enough and it maybe makes it easier to do full minting for external parties as a later iteration that it's worth doing.

Thanks for adding your analysis. Argument 3 really seems to carry the main weight, together with saving the effort of building (and app users integration with) yet another off-ledger API.

So, I'll go with Option 1 then.

PR #4162, opened 2026-02-26
add a separate participant deployment to the sv project to enable testing LSU

mblaze-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4136

I'm aware that these changes assume that there exists a service account that is currently not present but I'm planning to rectify that in a separate PR for the sake of small PRs.

### DIFF
--- a/cluster/pulumi/sv/README.md
+++ b/cluster/pulumi/sv/README.md
@@ -1,5 +1,5 @@
 # sv (Pulumi project)
 
-Deploys the logical part of the synchronizer and SV apps for an SV.
+Deploys the participant, i.e. the Canton components that don't change on an LSU.
 
 Deployment typically managed via an [operator](../operator). For manual deployment s.a. `cncluster apply`.


cluster/pulumi/sv/src/installNode.ts
```diff
@@ -1,6 +1,55 @@
...
+} from '@lfdecentralizedtrust/splice-pulumi-common-sv';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-export function installNode(sv: string, auth0Client: Auth0Client): void {}
+import { installParticipant } from './participant';
+
+export function installNode(sv: string, auth0Client: Auth0Client): void {
+  const staticConfig = findStaticConfigOrFail(sv);
+  const config = configForSv(staticConfig.nodeName);
+  // I don't get this comment VVV
```
moritzkiefer-da, line 22: the canton network and the sv-canton stack both rely on the sv namespaces. So you need to agree on which one is allowed to delete the namespace or you end up with the sv-canton stack also taking down the sv app because it deletes the whole namespace. As the comment suggests currently the canton-network stack is the one that will delete the namespace

cluster/pulumi/common-sv/src/synchronizer/participant.ts
```diff
@@ -0,0 +1,138 @@
...
+    'participant-pg',
+    'participant-pg',
+    version,
+    participant?.cloudSql ?? spliceConfig.pulumiProjectConfig.cloudSql,
+    true,
+    { disableProtection }
+  );
+}
+
+function installParticipantChart(
```
moritzkiefer-da, line 76: How soon do you expect to be able to get rid of the sv-canton stack files? I'm trying to figure out how much we need to worry about the duplication and whether it's not better to add a utility library to share that code betwene sv and sv-canton.

moritzkiefer-da (changes requested):
Thanks for the pr, started a discussion in slack to understand what will actually go in the sv stack because the readme and this PR don't line up. Requesting changes until I understand why this is in line with the plan.

cluster/pulumi/common-sv/src/synchronizer/participant.ts
mblaze-da: If I correctly recall the timeline that Nicu described it might coexist for a few months.

moritzkiefer-da: then I'd lean towards factoring it out

cluster/pulumi/sv/src/installNode.ts
mblaze-da: Yes, I understand that's why we use `retainOnDelete`, but I don't understand why do we `new k8s.core.v1.Namespace` instead of `k8s.core.v1.Namespace.get`. I assume the k8s provider is smart enough to reuse the namespace if it already exists.

moritzkiefer-da: Will Namespace.get not fail if the sv-canton stack comes up first? We don't control the order for which one starts first.

mblaze-da: I keep forgetting that we run all these stacks at once. Makes sense.

cluster/pulumi/sv/README.md
```diff
@@ -1,5 +1,5 @@
 # sv (Pulumi project)
 
-Deploys the physical part of the synchronizer and SV apps for an SV.
+Deploys the logical part of the synchronizer and SV apps for an SV.
```
moritzkiefer-da:
```suggestion
Deploys the participant, i.e. the Canton components that don't change on an LSU.
```

Probably clearer until we actually move stuff like sv, validtaro and scan

PR #4163, opened 2026-02-26
Release notes for 0.5.13

rautenrieth-da:
- Manually added missing notes from https://github.com/DACH-NY/canton/blob/release-line-3.4/UNRELEASED.md
- Checked the git log for splice, did not spot any missing entry

### DIFF
--- a/docs/src/release_notes.rst
+++ b/docs/src/release_notes.rst
@@ -27,6 +27,11 @@
         <sv_participant_pruning>` for instructions on how to enable
         it.
 
+    - Scan txlog script
+
+      - scan_txlog.py script has been deprecated. It will not be maintained moving forward, and will be removed completely in a future release.
+        A user interested in understanding how to parse transactions from history is reffered to the :ref:`Reading and parsing transaction history involving Token Standard contracts <token_standard_usage_reading_tx_history>` section.
+
     - Canton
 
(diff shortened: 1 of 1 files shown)


docs/src/release_notes.rst
```diff
@@ -18,6 +18,101 @@
...
+.. release-notes:: 0.5.13
+
+    - SV Participant
+
+      - Participant pruning for super validators is now supported and
+        recommended. Follow the :ref:`documentation
+        <sv_participant_pruning>` for instructions on how to enable
+        it.
+
+    - Canton
```
martinflorian-da, line 35: nit: shall we move this to one end of the release notes, such as "always first item"?

Interspersing Splice release notes and Canton release notes seems a bit confusing to me.

martinflorian-da: given that this is quite verbose and, I'd say, overall less relevant to most people than the Splice release notes... how about we always put the Canton block last?

moritzkiefer-da: yeah I also prefer last

ray-roestenburg-da: yeah

rautenrieth-da: Moved to the end

PR #4170, opened 2026-02-27
Bump gha resources to avoid over-committing splice cluster

julientinguely-da:
fixes https://github.com/DACH-NY/cn-test-failures/issues/7390

see https://grafana.splice.network.canton.global/d/ae9lqwimiigw0d/resource-utilization?orgId=1&from=now-12h&to=now&timezone=UTC&var-test_suite=$__all

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/deployment/mock/config.yaml
+++ b/cluster/deployment/mock/config.yaml
@@ -323,11 +323,11 @@ gha:
         resources: {
           requests: {
             cpu: '4',
-            memory: '8Gi',
+            memory: '6Gi',
           },
           limits: {
-            cpu: '6',
-            memory: '10Gi',
+            cpu: '4',
+            memory: '6Gi',
           },
         },
       },
@@ -338,11 +338,11 @@ gha:
         resources: {
           requests: {
             cpu: '6',
-            memory: '16Gi',
+            memory: '18Gi',
           },
           limits: {
-            cpu: '8',
-            memory: '20Gi',
+            cpu: '6',
+            memory: '18Gi',
           },
         },
       },
@@ -353,11 +353,11 @@ gha:
         resources: {
           requests: {
             cpu: '8',
-            memory: '24Gi',
+            memory: '26Gi',
           },
           limits: {
-            cpu: '10',
-            memory: '30Gi',
+            cpu: '8',
+            memory: '26Gi',
           },
         },
       },
@@ -367,12 +367,12 @@ gha:
         docker: true,
         resources: {
           requests: {
-            cpu: '10',
+            cpu: '8',
             memory: '36Gi',
           },
           limits: {
-            cpu: '12',
-            memory: '42Gi',
+            cpu: '8',
+            memory: '36Gi',
           },
         },
       },
@@ -382,12 +382,12 @@ gha:
         docker: false,
         resources: {
           requests: {
-            cpu: '12',
-            memory: '42Gi',
+            cpu: '10',
+            memory: '48Gi',
           },
           limits: {
-            cpu: '14',
-            memory: '52Gi',
+            cpu: '10',
+            memory: '48Gi',
           },
         },
       },


cluster/pulumi/gha/src/runners.ts
```diff
@@ -55,7 +55,7 @@ const runnerSpecs = [
     resources: {
       requests: {
         cpu: '4',
-        memory: '10Gi',
+        memory: '8Gi',
```
moritzkiefer-da: can we make this stuff configurable through config instead of hardcoding it? That way changing it becomes much easier than having to jump betwene splice and internal

julientinguely-da: good idea

julientinguely-da: Now we have to be careful and first define the runnerSpecs in internal before to merge this PR. right?

julientinguely-da: https://github.com/DACH-NY/canton-network-internal/pull/3879

cluster/deployment/mock/config.yaml
```diff
@@ -294,6 +294,104 @@ cluster:
       nodeType: c4-standard-16
       minNodes: 0
       maxNodes: 1
+gha:
```
moritzkiefer-da, line 297: why not put this in the default config.yaml? Then we can overwrite in internal but it still works if we don't overwrite

julientinguely-da: because it seems kind of weird to have half of the config in internal/main and the other half in the default. I usually just paste it in mock to have a reference, but here it doesn't update the mock expected

julientinguely-da: Oh thanks for the refs

> Maybe for now just merge only the configurability but keep the values as before.

I'll just wait Monday to see what @nicu says on that. To me the grafana shows that the resources are tight

julientinguely-da: ok yes, we could indeed lower the requests a bit more while keeping the limits :+1:

moritzkiefer-da: > because it seems kind of weird to have half of the config in internal/main and the other half in the default. I usually just paste it in mock to have a reference, but here it doesn't update the mock expected


fair

nicu-da: @julientinguely-da also check the discussion in https://github.com/hyperledger-labs/splice/pull/4047 around changing any resource requests. I think it makes sense but maybe check with @isegall-da as well

isegall-da: I'm a bit worried that requests != limits means we are not dedicating resources and things might start failing in non-deterministic and hard to investigate ways

julientinguely-da: I've made the changes to make them equal and bump their value: https://github.com/DACH-NY/canton-network-internal/pull/3879

cluster/pulumi/gha/src/config.ts
```diff
@@ -13,6 +14,19 @@ const GhaConfigSchema = z.object({
     // this is a https://github.com/actions/actions-runner-controller version
     runnerScaleSetVersion: z.string(),
     workPvcSize: z.string().default('20Gi'),
+    runnerSpecs: z
+      .array(
+        z.object({
+          name: z.enum(['tiny', 'x-small', 'small', 'medium', 'large', 'x-large']),
+          k8s: z.boolean(),
+          docker: z.boolean(),
```
isegall-da, line 22: See my comment on the other PR re just supporting all sizes in both

julientinguely-da: Merging for now, it seems like a bigger change, happy to have a follow-up PR for that.

PR #4176, opened 2026-02-27
add a switch for the logical synchronizer deployment mode

mblaze-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4105


cluster/pulumi/sv/src/installNode.ts
```diff
@@ -27,19 +28,23 @@ export function installNode(sv: string, auth0Client: Auth0Client): void {
...
-      version: config.versionOverride ?? activeVersion,
-      disableProtection: staticConfig.nodeName === svRunbookConfig.nodeName,
-      participantAdminUserNameFrom: ledgerApiUserSecretSource,
-      imagePullServiceAccountName: serviceAccountName,
-    },
-    { dependsOn: [...imagePullDeps, ledgerApiUserSecret] }
-  );
+  const isLsuDeployment =
+    DecentralizedSynchronizerUpgradeConfig.active.enableLogicalSynchronizerDeploymentMode;
+  if (isLsuDeployment) {
```
moritzkiefer-da, line 33: I wonder if adding those flags to the sv stack makes sense. I was expecting we just don't install the sv stack when we don't need this?

mblaze-da: I plan on adding the sv project to all the `cncluster` commands. I'm not sure how easy it is to opt out of a particular project or stacks when doing `cncluster apply`.

PR #4181, opened 2026-02-27
Generate DarResources.scala

moritzkiefer-da:
[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -371,6 +371,7 @@ To edit the files in a particular Daml project, for example, `/apps/wallet/daml`
    If you change any of them, then you can propagate these changes across the .dars as follows:
    1. redo Step 2
     3. use Ctrl-Shift-P "Developer: Reload Window" in VS code to restart the `daml studio` language server with the updated package dependencies.
+7. When updating a DAR, make sure to run `sbt updateDarResuorces` before running integration tests. CI enforces that this is always done.
 
 *Tip:* if `damlBuild` fails with weird errors, then that might be due to stale `damlBuild` outputs.
 Try forcing a clean rebuild by cleaning via SBT, e.g., `apps-common/clean` and similar for the dependent project.
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/DarResources.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/DarResources.scala
@@ -1,7 +1,7 @@
 // Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-// DO NOT EDIT! This file is generated automatically by DarResourcesGenerator.scala
+// DO NOT EDIT! This file is generated automatically by DarResourcesGenerator.scala via `sbt updateDarResources`
 
 package org.lfdecentralizedtrust.splice.environment
 


apps/dar-resources-generator/src/main/scala/org/lfdecentralizedtrust/splice/darutils/DarResourcesGenerator.scala
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package org.lfdecentralizedtrust.splice.darutils
+
+import better.files.File
+import com.digitalasset.canton.discard.Implicits.DiscardOps
+
+object DarResourcesGenerator {
+  def render(): String = {
```
moritzkiefer-da, line 10: There are a bunch of libraries to produce actual scala ASTs instead of doing string concatenation. I honestly have a hard time justifying any of them here. This seems much simpler 🤷

.github/workflows/build.static_tests.yml
```diff
@@ -144,12 +144,12 @@ jobs:
...
         uses: ./splice-shared-gha/.github/actions/sbt/execute_sbt_command
         with:
-          cmd: "Test/compile lint updateTestConfigForParallelRuns"
+          cmd: "Test/compile lint updateTestConfigForParallelRuns updateDarResources"
 
       - name: Verify no changes in SBT test files
         uses: ./splice-shared-gha/.github/actions/nix/run_bash_command_in_nix
         with:
-          cmd: git add . && git diff --staged --exit-code test*.log
+          cmd: git add . && git diff --staged --exit-code test*.log '**/DarResources.scala'
```
moritzkiefer-da, line 152: did confirm that this does indeed fail CI as expected

.github/workflows/build.static_tests.yml
```diff
@@ -144,12 +144,12 @@ jobs:
       - name: SBT-based static checks
         uses: ./splice-shared-gha/.github/actions/sbt/execute_sbt_command
         with:
-          cmd: "Test/compile lint updateTestConfigForParallelRuns"
+          cmd: "Test/compile lint updateTestConfigForParallelRuns updateDarResources"
```
moritzkiefer-da, line 147: 1. I did consider instead making this a proper SBT code generator that automatically runs when you compile. I decided against it because it seems like we run clean more often than we change Daml models so not having to rely on SBT caching for this seems better.
2. It doesn't seem to be that slow, around 20s maybe (and we could make it faster) so putting it here in the CI job seemed fine.

apps/dar-resources-generator/src/main/scala/org/lfdecentralizedtrust/splice/darutils/DarResources.scala
```diff
@@ -0,0 +1,486 @@
...
+import com.digitalasset.daml.lf.archive.Dar
+import com.digitalasset.daml.lf.language.Ast
+import com.digitalasset.daml.lf.language.Ast.Package
+
+import java.io.{File, FileInputStream, InputStream}
+import java.util.zip.ZipInputStream
+import java.nio.file.Path
+import scala.util.Using
+
+object DarResources {
```
moritzkiefer-da, line 19: This is a 1:1 copy of the existing DAR resources. I think the next step here would be to auto-generate this as well from the dars we have in the dars directory. That seems to naturally fit into https://github.com/hyperledger-labs/splice/issues/3667

build.sbt
```diff
@@ -1974,6 +1975,40 @@ lazy val `apps-metrics-docs` =
       Headers.ApacheDAHeaderSettings
     )
 
+lazy val `apps-dar-resources-generator` =
```
isegall-da, line 1978: this is not caching anything, hence we do not want to put it as a "dependsOn" for all our apps, right? That would reduce mistakes, but IIUC, would slow down every single compilation, regardless of whether there was a daml change or not?

That's my main concern with this PR, that it would make it easier to make confusing mistakes. Having said that, not too often do people make daml changes, and then in the same PR app&integration test changes that would be affected, so I think I can live with this extra complexity.

.github/workflows/build.static_tests.yml
```diff
@@ -144,12 +144,12 @@ jobs:
...
 
       - name: Verify no changes in SBT test files
```
isegall-da, line 149:
```suggestion
      - name: Verify no changes in files auto-generated via SBT
```

isegall-da: Should we add a single sbt target `codegen` or whatever, such that any failure of this test can be fixed by running that without remembering too much? (I know I personally always need to look for that magic `updateTestConfigForParallelRuns` word because I never remember it)

moritzkiefer-da: sbt updateDarResources sound pretty close to that already?

build.sbt
moritzkiefer-da: see this comment here https://github.com/hyperledger-labs/splice/pull/4181#discussion_r2873560765

moritzkiefer-da: I think the best fix to guard against that is probably to make it much faster and then put it in a precommit hook.

apps/dar-resources-generator/src/main/scala/org/lfdecentralizedtrust/splice/darutils/DarResourcesGenerator.scala
isegall-da: Given that the DarResources file hasn't changed (in structure) in a long time, I tend to agree

apps/dar-resources-generator/src/main/scala/org/lfdecentralizedtrust/splice/darutils/DarResourcesGenerator.scala
```diff
@@ -0,0 +1,127 @@
...
+    val lines = Seq(
+      "// DO NOT EDIT! This file is generated automatically by DarResourcesGenerator.scala",
```
isegall-da:
```suggestion
      "// DO NOT EDIT! This file is generated automatically by DarResourcesGenerator.scala via `sbt updateDarResources`",
```

(again, to make it easier to find the thing you need to run in order to update this)

.github/workflows/build.static_tests.yml
isegall-da: I meant one that includes both updateDarResources and updateTestConfigForParallelRuns, but I'm also fine if you think I'm over-thinking this

moritzkiefer-da: alsoe added it to the readme

moritzkiefer-da: > I meant one that includes both updateDarResources and updateTestConfigForParallelRuns, but I'm also fine if you think I'm over-thinking this

I'd lean towards not adding that. These catch all commands tend to slow down people needlessly. I think you're better off understanding when you need them here given that the rules are not that complex.

isegall-da: fair

PR #4185, opened 2026-03-01

canton-network-da:


martinflorian-da (approved):
newer is always better

FYI @julientinguely-da ; on of the next "Bump GHA runner" PRs on internal will consume this.

PR #4194, opened 2026-03-02
Document workaround for OwnerToKeyMapping signature issue during validator recovery

martinflorian-da:
- **Document workaround for OwnerToKeyMapping signature issue during validator recovery**

Add troubleshooting entry for a rare failure where ImportTopologySnapshot
rejects an OwnerToKeyMapping due to insufficient signatures. This can
happen when the topology snapshot was created under an older Canton version.
The workaround involves manually proposing the corrected mapping via
Canton console before starting the validator app.

Fixes https://github.com/DACH-NY/canton-network-internal/issues/2705 (see comments there)

- **Split troubleshooting section into subsections for readability**


docs/src/validator_operator/validator_disaster_recovery.rst
```diff
@@ -274,6 +281,41 @@ To address a failed :term:`ACS` import, you can usually:
    You can now take down the node to which you originally tried to restore and try the restore procedure again with your adjusted dump on a fresh node with a different participant ID prefix
    (i.e., a different ``newParticipantIdentifier`` / ``<new_participant_id>`` depending on your deployment model).
 
+Troubleshooting rejected topology snapshots
+""""""""""""""""""""""""""""""""""""""""""""
+
+In rare cases, the re-onboarding process may fail at the ``ImportTopologySnapshot`` step
+because an ``OwnerToKeyMapping`` for the old participant ID has an insufficient number of
+signatures in the topology snapshot. This can happen when the topology snapshot was created
```
moritzkiefer-da: let's add the version where this was the case. That is gonna exclude the majority of validators and avoid sending them down the wrong track.

martinflorian-da: good idea!

PR #4195, opened 2026-03-02
init sv stacks for scratchnets and start integrating automation

mblaze-da:


cluster/pulumi/pulumiDown.ts
```diff
@@ -18,6 +19,8 @@ async function runStacksDown() {
   operations.push(downOperation(mainStack, abortController));
   const cantonDown = startDownOperationsForCantonStacks(abortController);
   operations = operations.concat(cantonDown);
+  const svsDown = startDownOperationsForSvStacks(abortController);
```
mblaze-da, line 22: I'm a bit worried about adding this because after a splice bump we start running this for internal clusters which will initialize new stacks but their configuration won't be versioned. I'm not sure what will happen on subsequent downs in that case.

nicu-da: It's going to fail unfortunately until we check-in the stack files. 
What's missing though is a global config that controls if the sv stacks are used. If we have that we won't have any issue right? 
And then we can enable that by default for the scratch clusters and I would also enable for cimain

mblaze-da: I added
https://github.com/hyperledger-labs/splice/blob/4360f1b1da006d620db0b1700b7fa602ca78f9a2/cluster/pulumi/sv/src/installNode.ts#L33
but I guess it doesn't solve the issue here.

mblaze-da: Do I understand correctly that you're suggesting using this config in the Pulumi automation scripts to disable running `sv` altogether?

nicu-da: Yes, run the automation for the sv stacks only if that config is set to true, that should cover us  for now and make rolling this out easier.

mblaze-da: @nicu-da I adjusted the implementation accordingly.

cluster/pulumi/sv/src/installNode.ts
```diff
@@ -28,23 +27,19 @@ export function installNode(sv: string, auth0Client: Auth0Client): void {
   const auth0Config = auth0Client.getCfg();
   const ledgerApiUserSecret = installLedgerApiUserSecret(auth0Client, xns, 'sv', 'sv');
   const ledgerApiUserSecretSource = auth0UserNameEnvVarSource('sv', true);
-  const isLsuDeployment =
-    DecentralizedSynchronizerUpgradeConfig.active.enableLogicalSynchronizerDeploymentMode;
-  if (isLsuDeployment) {
```
nicu-da, line 33: Not sure I fully understand this change

nicu-da, line 33: Not sure I fully understand this change

mblaze-da: In this commit I added a condition in the Pulumi automation scripts which controls whether operations on the `sv` project get applied. It is based on the `enableLogicalSynchronizerDeploymentMode` configuration property. After this addition, hiding resource definition in the `sv` project behind the same flag became obsolete because when the whole stack is disabled, then there is no point in disabling particular resources.

nicu-da: Ah fair point, we don't have support for it in the operator yet so this is safe. Thanks for the explanation

nicu-da (approved):

PR #4109, opened 2026-02-24
Persist ext txn hashes

jagathweerasinghe-da:
[ci]

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #4199, opened 2026-03-02
Add ext txn hash to update api

jagathweerasinghe-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
@@ -1622,6 +1622,7 @@ class UpdateHistory(
       )
     }
     val events: Seq[Event] = (createEvents ++ exerciseEvents).sortBy(_.getNodeId)
+
     UpdateHistoryResponse(
       update = TransactionTreeUpdate(
         new Transaction(
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodings.scala
@@ -741,7 +741,7 @@ object ScanHttpEncodings {
       case (_, event) => sys.error(s"Unexpected event type: $event")
     }
 
-    // Only include the external transaction hash based on the hash inclusion policy.
+    // Include the external transaction hash based on the hash inclusion policy.
     val externalTransactionHash: ByteString =
       hashInclusionPolicy match {
         case ExternalHashInclusionPolicy.AlwaysInclude =>


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1566,7 +1567,7 @@ class UpdateHistory(
           /*synchronizerId = */ updateRow.synchronizerId,
           /*traceContext = */ TraceContextOuterClass.TraceContext.getDefaultInstance,
           /*recordTime = */ updateRow.recordTime.toInstant,
-          /*externalTransactionHash = */ ByteString.EMPTY, // TODO(#3408): Revisit when ingesting to DB
+          /*externalTransactionHash = */ ByteString.EMPTY,
```
jagathweerasinghe-da, line 1573: TBD: Not sure whether we need to add hashes here for "getImportUpdates"?

OriolMunoz-da: From the openAPI docs in this PR:

> For transaction externally signed, contains the external transaction hash signed by the external party.

The import updates are never externally signed, so it's safe to set it to ByteString.EMPTY afaik

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodingsTest.scala
```diff
@@ -286,6 +292,8 @@ class ScanHttpEncodingsTest extends StoreTestBase with TestEssentials with Match
           rightChildId1 -> mkCreate(rightChildId1),
           rightChildId2 -> mkCreate(rightChildId2),
         ),
+        externalTransactionHash =
+          Option("4d68f590e4a298d9617ebe07b98c6ecbe04b7f3d7a5327f0e0ad4719638302b7"),
```
OriolMunoz-da:
```suggestion
          Some("4d68f590e4a298d9617ebe07b98c6ecbe04b7f3d7a5327f0e0ad4719638302b7"),
```

OriolMunoz-da (commented):
PR itself LGTM,though there is one concern I have wrt backfilling (which is already the case in https://github.com/hyperledger-labs/splice/pull/4109 afaict):

`external_transaction_hash` is now introduced for `/v0/backfilling/updates-before`, i.e., we will backfill the `external_transaction_hash`.

There's now the situation where e.g.:
- sv-1 and sv-2 upgrade to the version that has this code
- sv-3 and sv-4 are slower to upgrade so they don't for a few hours
- an externally signed transaction comes => sv-1 and sv-2 set the hash, sv-3 and sv-4 don't

How should:
- A client doing BFT reads from Scan handle that?
- A newly onboarded / recovered SV doing backfilling?

The problem being that for these 4 SVs (f=1), there's not really consensus as to whether the transaction does or does not have a transaction hash, so the client will randomly get one or the other.

(the answer might very well be that I'm missing some chunk of code for this)

cc @ray-roestenburg-da

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/UpdateHistoryTestUtil.scala
```diff
@@ -258,6 +255,33 @@ trait UpdateHistoryTestUtil extends TestCommon {
     succeed
   }
 
+  def compareHistoryViaLosslessScanApiWithExtTxnHashes(
```
rautenrieth-da, line 257: This does two independent checks:
- Compare responses from the scan api with responses from the store
- Compare tx hashes from the scan api responses with a given sequence of strings

How about we split these? I.e., remove `compareHistoryViaLosslessScanApi()` from the body and  `scanBackend` from the parameter list.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/UpdateHistoryTestUtil.scala
```diff
@@ -258,6 +255,33 @@ trait UpdateHistoryTestUtil extends TestCommon {
...
+    val recordedExtTxnHashes = historyThroughApi.flatMap { update =>
+      update.update.update match {
+        case TransactionTreeUpdate(tx) =>
+          val hash = tx.getExternalTransactionHash
+          if (hash.isEmpty) None else Some(HexString.toHexString(hash))
+        case _ => None
+      }
+    }
+    recordedExtTxnHashes should not be empty
+    recordedExtTxnHashes should contain allElementsOf extTxnHashes
```
rautenrieth-da, line 278: Either change `extTxnHashes` to a `Set` (if the input hashes are not ordered), or change `allElementsOf` to `theSameElementsInOrderAs` (if the input hashes are ordered)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -2589,6 +2592,7 @@ object UpdateHistory {
       workflowId: Option[String],
       commandId: Option[String],
       eventId: String,
+      externalTransactionHash: Array[Byte],
```
rautenrieth-da: This column is nullable in the schema, can we make it `Option[Array[Byte]]` here, and let users handle missing values?

ray-roestenburg-da: It will be an empty array if it is missing. Otherwise you have None, empty, and non-empty that you would have to deal with.

rautenrieth-da:
> the SVs will agree that it's either empty, or set (never disagree on the actual value if it's defined), meaning that there would be always f+1 agreeing on None or Some(x)

That's only the case for the updates endpoint. I was talking about bulk storage, where we write a large number of updates into one file, and then take the hash of the file contents. Unless all updates within the file are bitwise identical, the hash of the whole file will differ.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
jagathweerasinghe-da: Added a comment to explicitly mention this.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanHttpEncodingsTest.scala
jagathweerasinghe-da: done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/UpdateHistoryTestUtil.scala
jagathweerasinghe-da: Make sense, comparing hashes from API responses is enough. Done.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/UpdateHistoryTestUtil.scala
jagathweerasinghe-da: done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/UpdateHistoryTestUtil.scala
```diff
@@ -142,12 +170,55 @@ trait UpdateHistoryTestUtil extends TestCommon {
...
+      case (actual, recorded) =>
+        actual shouldBe recorded
+    }
+    succeed
+  }
+
+  private def compareExternalTxnHashes(
+      actualUpdates: Seq[UpdateHistoryResponse],
+      recordedUpdates: Seq[UpdateHistoryResponse],
+      extTxnHashes: Set[String],
```
rautenrieth-da: Why is this a Set and not a Seq? Don't we know the order in which the external transactions are created in the test?

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__update_history_for_hash.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE update_history_transactions ADD COLUMN external_transaction_hash bytea;
```
rautenrieth-da: You'll have to rename this file to `V061_...`, I just merged a PR that adds migration 60.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsJdbcTypes.scala
```diff
@@ -366,6 +366,24 @@ trait AcsJdbcTypes {
...
+      extTxnHash: ByteString,
+      maxLength: Int = 1024,
+  ): Option[Array[Byte]] = {
+    Option(extTxnHash)
+      .filterNot(_.isEmpty)
+      .map { nonNullExtTxnHash =>
+        val trimmedExtTxnHash =
+          if (nonNullExtTxnHash.size() > maxLength) nonNullExtTxnHash.substring(0, maxLength)
+          else nonNullExtTxnHash
+        trimmedExtTxnHash.toByteArray
```
rautenrieth-da: In the unexpected case that the transaction hash is too long, this will store trimmed data.

When inserting strings that should represent canton ids, we use `lengthLimited()` to make sure the strings are shorter than 2066 characters, and crash the ingestion if they are not.

AFAIK, the transaction hash is chosen by canton and is coming from the ledger API. We should therefore have the same amount of trust that it has the correct length as we have in the length of template ids. I would therefore prefer if we also crashed ingestion rather than storing incorrect data.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1614,6 +1623,19 @@ class UpdateHistory(
       )
     }
     val events: Seq[Event] = (createEvents ++ exerciseEvents).sortBy(_.getNodeId)
+    // Only include the external transaction hash for transactions recorded on or after the threshold timestamp.
```
rautenrieth-da: So far, UpdateHistory _mostly_ served data as it is stored in the database. We already apply various transformations to this data before we return it to external users. In particular, in `ScanHttpEncodings.makeConsistentAcrossSvs` we rewrite offsets and event ids in order to make the responses BFT safe. 

What we're doing here is another kind of transformation for making responses BFT safe, so I think for consistency it should also go in `ScanHttpEncodings.makeConsistentAcrossSvs`.

Alternatively, move the existing offset and event ids rewriting to `UpdateHistory` as well, but that sounds like more refactoring work.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -1783,6 +1795,7 @@ class HttpScanHandler(
             txWithMigration,
             encoding = encoding,
             version = if (consistentResponses) ScanHttpEncodings.V1 else ScanHttpEncodings.V0,
+            hashInclusionPolicy = ExternalHashInclusionPolicy.AlwaysInclude,
```
rautenrieth-da: Can you please remind me, why did we decide to always include the external hash for pointwise lookups? This will make the `getUpdateById` method not BFT safe, and it will return data inconsistent with the update stream.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -2431,6 +2430,7 @@ object UpdateHistory {
   }
 
   sealed trait BackfillingRequirement
+
```
rautenrieth-da: nit: avoid useless whitespace changes

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
jagathweerasinghe-da: My assumption is vX/updates/{update_id} and vX/updates/hashes/{hash} should always include; 
My assumption might be wrong; Its a valid point, once we expose it we don't how the users are going to use it. 

Does this mean BFT safety must be applied for all API endpoints, including updates by hash (vX/updates/hashes/{hash})?  If so, we have to communicate this to the product team.

rautenrieth-da: IMO all endpoints should be BFT safe by default, as users may not trust any single scan and do a BFT read for every scan request. We could exempt individual responses fields from BFT safety, but that does not work for clients that use the whole opaque response when checking response equality (like we do in the validator scan proxy).

For this reason I would keep at least `vX/updates/{update_id}` BFT safe. 

For `vX/updates/hashes/{hash}` I'm less sure, I guess here it's OK if we always include the hash.

jagathweerasinghe-da: As the product team has no specification for vX/updates/{update_id} endpoint, this approach should be fine. They are concerned about having the vX/updates/hashes/{hash} endpoint. 

So, we can go ahead with:

**vX/updates: BFT Safe
vX/updates/{update_id} : BFT Safe
vX/updates/hashes/{hash}: Not BFT Safe**

So, taking this approach, WDYT about mentioning that vX/updates/hashes/{hash} is not BFT safe in the API documentation?

ray-roestenburg-da: The main request has been to be able to lookup a transaction by transaction id, so I think the get by update id API does not have to include the hash. If we need that for some reason, we can add a request parameter so people can decide they want that returned in the response as well, and we can explain that in certain cases they might get different responses from SVs though only for some period, iiuc.

rautenrieth-da: > So, we can go ahead with

Sounds good to me, just make sure to document everything in detail in the API spec.

From a technical point of view, "vX/updates/hashes/{hash}: Not BFT Safe" would mean that some scans return 404 while all others return 200 with the exact same payload, which is less annoying that different scans returning different payloads. You have to deal with 404 responses anyway because some scans might not have caught up yet.

ray-roestenburg-da: @rautenrieth-da Is per hash not safe _after a specific cut over point_  or not safe at all? The latter would surprise me.

jagathweerasinghe-da: vX/updates/hashes/{hash} is not safe only for the hashes created before the threshold date.

ray-roestenburg-da: Please document it as such, and I would assume at some point we'll know the actual threshold date and can add that. Does that also mean we can include the has in by update id if we know the threshold in the system right?

jagathweerasinghe-da: We have already set a conservative threshold date of 2030/01/01, which we can change later as needed. 
The APIs are going to be like below:

vX/updates: BFT Safe. Hashes created **after the threshold** date will be included in the response.
vX/updates/{update_id} : BFT Safe. Hashes created **after the threshold** date will be included in the response.
vX/updates/hashes/{hash}: Not BFT Safe. Supported irrespective of the created date. 

Does that answer your question.

rautenrieth-da: > not safe only for the hashes created before the threshold date

It's even slightly better, the exact behavior for `vX/updates/hashes/{hash}` should be:

There are record times R0 and R1 where:
- For transactions commited before R0, all scans will return 404
- For transactions committed between R0 and R1, some scans will return 404 and some scans will return a transaction
- For transactions committed after R1, all scans will return a transaction
- If two scans return a non-404 response, then the response payloads will be identical
- The difference between R0 and R1 is expected to be in the order of a few days
- R1 is expected to be weeks before the cutoff date after which transaction hashes will start to appear in the `v2/updates` stream

"not BFT safe for the hashes created before a threshold date" sounds good as a short description.

> I would assume at some point we'll know the actual threshold date

Yes, but it depends on the network and changes after network resets.

ray-roestenburg-da: @jagathweerasinghe-da I would document, `vX/updates/hashes/{hash}` is BFT safe for hashes that were computed for transactions that were created after the threshold date. It's not, by definition, 'not-BFT safe'.

jagathweerasinghe-da: > > not safe only for the hashes created before the threshold date
> 
> It's even slightly better, the exact behavior for `vX/updates/hashes/{hash}` should be:
> 
> There are record times R0 and R1 where:
> 
> * For transactions commited before R0, all scans will return 404
> * For transactions committed between R0 and R1, some scans will return 404 and some scans will return a transaction
> * For transactions committed after R1, all scans will return a transaction
> * If two scans return a non-404 response, then the response payloads will be identical
> * The difference between R0 and R1 is expected to be in the order of a few days
> * R1 is expected to be weeks before the cutoff date after which transaction hashes will start to appear in the `v2/updates` stream
> 
> "not BFT safe for the hashes created before a threshold date" sounds good as a short description.
> 
> > I would assume at some point we'll know the actual threshold date
> 
> Yes, but it depends on the network and changes after network resets.

Great, we will add this in the next PR, which brings the vX/updates/hashes/{hash} endpoint.

PR #4207, opened 2026-03-03
Add GCP log alert for leaked secrets in production logs

martinflorian-da:
Adds a new GCP Cloud Monitoring alert that fires when secrets (JWTs, Bearer tokens, key=value credentials) appear in cluster logs. This complements the existing integration test check in check-logs.sh with a cluster alert.

Fixes https://github.com/DACH-NY/canton-network-internal/issues/2647

### DIFF
--- a/cluster/deployment/scratchneta/config.resolved.yaml
+++ b/cluster/deployment/scratchneta/config.resolved.yaml
@@ -64,7 +64,7 @@ monitoring:
         overMinutes: 5
         quantile: 0.95
     enableNoDataAlerts: false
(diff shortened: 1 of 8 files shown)


cluster/deployment/scratchneta/config.resolved.yaml
```diff
@@ -64,6 +64,7 @@ monitoring:
         overMinutes: 5
         quantile: 0.95
     enableNoDataAlerts: false
+    loggedSecretsFilter: "(jsonPayload.message=~\"(secret|token|private-key|password)=[^, ]+\" OR jsonPayload.message=~\"eyJhbGc[A-Za-z0-9_-]{2,}\\.[A-Za-z0-9_-]{2,}\\.[A-Za-z0-9_-]{2,}\" OR jsonPayload.message=~\"Bearer\\s+eyJ[A-Za-z0-9_-]{2,}\")\n-jsonPayload.message=~\"(secret|token|private-key|password)=(\\\"\\\\*\\\\*\\\\*\\\\*\\\"|hidden)\"\n"
```
moritzkiefer-da: so I just need to make sure I log it as privateKey 🙈

moritzkiefer-da: more serious suggestion: is a case insenitive key= better?

moritzkiefer-da (approved):
thx, assuming you checked this is clean on the exsiting clusters

cluster/deployment/scratchneta/config.resolved.yaml
martinflorian-da: Hmm so just `key` could give a lot of false positives...

Let me just add `privateKey` by itself here (and also to the integration test log checker, for consistency) and try to make the whole matching here case-insensitive.

PR #4216, opened 2026-03-03
add support for logical synchronizer deployment to the operator

mblaze-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4208

### DIFF
--- a/cluster/pulumi/deployment/src/stacks/splice.ts
+++ b/cluster/pulumi/deployment/src/stacks/splice.ts
@@ -115,23 +115,22 @@ function installSvStacks(
   gcpSecret: k8s.core.v1.Secret
 ): void {
   if (
-    !deploymentConf.projectsToDeploy.has('sv') ||
-    !DecentralizedSynchronizerUpgradeConfig.active.enableLogicalSynchronizerDeploymentMode
+    deploymentConf.projectsToDeploy.has('sv') &&
+    DecentralizedSynchronizerUpgradeConfig.active.enableLogicalSynchronizerDeploymentMode
   ) {
-    return;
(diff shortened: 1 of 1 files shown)


cluster/pulumi/deployment/src/stacks/splice.ts
```diff
@@ -97,4 +105,33 @@ export function installSpliceStacks(
...
+  envRefs: EnvRefs,
+  namespace: string,
+  gcpSecret: k8s.core.v1.Secret
+): void {
+  if (
+    !deploymentConf.projectsToDeploy.has('sv') ||
+    !DecentralizedSynchronizerUpgradeConfig.active.enableLogicalSynchronizerDeploymentMode
+  ) {
+    return;
+  }
```
nicu-da: Can we use if else without negating the conditions? it is easier to read if we have a block that checks if the project is deployed and the lsu mode is enabled compared to two negated or conditions. 
Also having an if/else return is easier to read as well and less probe to mistakes in the future compared to returning early from an if block

mblaze-da: I lean towards the opinion that nesting is a greater hindrance to readability but it's not a strong conviction so I changed the implementation as you've suggested.

nicu-da: I would strongly recommend avoiding returning early as a control of flow. It leads to great probability of errors in the future when removing the if that contains the return but the code afterwards is not affected by that thus becoming a probable cause of bugs.

I do agree that nesting is not a good sing but I don't consider an if/else case as nesting

PR #4229, opened 2026-03-04
Bump `splice-shared-gha`

martinflorian-da:
To get https://github.com/hyperledger-labs/splice-shared-gha/pull/6

I don't think we have autobumps for this. We might want to add them though.

[static]


moritzkiefer-da (approved):
autobumps sound great, do you wanna create tech debt for it? If only we knew someone that was doing their tech debt week right now 🧌

PR #4245, opened 2026-03-04
do not attempt to delete unused DBs that are protected from deletion

mblaze-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/7406

IMO there is no point in attempting to delete DBs that have protection on. Besides, failing the whole process could prevent us from successfully deleting other unused DBs.

### DIFF
--- a/cluster/pulumi/purge_unused_dbs.ts
+++ b/cluster/pulumi/purge_unused_dbs.ts
@@ -78,9 +78,13 @@ async function deleteDb(db: cloudsql.protos.google.cloud.sql.v1.IDatabaseInstanc
     instance: db.name,
     project: db.project,
   };
-  console.log(`Deleting ${db.name}...`);
-  await gcpSqlClient.delete(request);
-  console.log(`Done deleting ${db.name}`);
+  if (db.settings?.deletionProtectionEnabled?.value ?? false) {
+    console.warn(`Skipping ${db.name} deletion due to active deletion protection.`);
+  } else {
+    console.log(`Deleting ${db.name}...`);
+    await gcpSqlClient.delete(request);
+    console.log(`Done deleting ${db.name}`);
+  }
 }
 
 async function runPurgeUnusedDbs() {


cluster/pulumi/purge_unused_dbs.ts
```diff
@@ -78,9 +78,13 @@ async function deleteDb(db: cloudsql.protos.google.cloud.sql.v1.IDatabaseInstanc
     instance: db.name,
     project: db.project,
   };
-  console.log(`Deleting ${db.name}...`);
-  await gcpSqlClient.delete(request);
-  console.log(`Done deleting ${db.name}`);
+  if (db.settings?.deletionProtectionEnabled?.value ?? false) {
+    console.warn(`Skipping ${db.name} deletion due to active deletion protection.`);
```
julientinguely-da, line 82: Thanks, sounds right to me. Just wanna be sure, @nicu-da is it fine not to error in the case we have a custom protected db? I'm just a bit worried we lose track of staled dbs by just having a warning.

nicu-da: But why are we trying to delete dbs with deletion protection? In what scenario did this happen?

julientinguely-da: @jagathweerasinghe-da is doing some test with some test dbs. He added protection himself I think so that it doesn't get deleted.

nicu-da: I would much rather we have some label we can add to the db so that it's ignored by this script.
This produces the issue that if you have deletion protection by mistake then we will let that db survive and cost money

julientinguely-da: good point, `label=test` sounds like the right solution :+1:

nicu-da: we get the dbs witht his filter:

```
  const filter = `settings.userLabels.cluster:${CLUSTER_BASENAME}`;
```

so just removing that label should be enough for this script to not delete it

moritzkiefer-da: > @jagathweerasinghe-da is doing some test with some test dbs. He added protection himself I think so that it doesn't get deleted.

isn't the fix for that just to remove the cluster label so that our purge DB logics ignores it? Why do we need an ignore label?

julientinguely-da: hmm, yes you're right :see_no_evil: 
I removed the cluster labels for, this should fix the issue @mblaze-da : 
- https://console.cloud.google.com/sql/instances/sv-1-cn-apps-pg-c8f4841-wee-clone-new-index-test/overview?project=da-cn-ci-2
- https://console.cloud.google.com/sql/instances/sv-1-participant-7-pg-50b6da8-clone-wee-hash-test/overview?project=da-cn-ci-2

mblaze-da: Then I'm closing this PR and the corresponding issue.

PR #4246, opened 2026-03-04
Add get update by hash api

jagathweerasinghe-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/docs/src/release_notes_upcoming.rst
+++ b/docs/src/release_notes_upcoming.rst
@@ -9,7 +9,6 @@
 
 .. release-notes:: Upcoming
 
-
     - Deployment
 
         - SV apps now support a ``copyVotesFrom`` setting that automatically mirrors governance votes


apps/scan/src/main/openapi/scan.yaml
```diff
@@ -409,6 +409,43 @@ paths:
         "500":
           $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/500"
 
+  /v2/updates/hashes/{hash}:
```
jagathweerasinghe-da: TBD: hash or transaction_hash or ext_txn_hash or ?

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -444,6 +444,43 @@ paths:
...
+      x-jvm-package: scan
+      operationId: "getUpdateByHash"
+      description: |
+        Returns the update associated with the given hash.This endpoint is BFT safe only for the updates recorded after a threshold time (t1).
+        For record times of t0 and t1 where:
+          - t0: The time when the Scan backend begin indexing external transaction hashes
+          - t1: The time at which transaction hashes are included in the /v1/updates and /v2/updates streams
+          - For transactions recorded before t0, 404 will be returned
+          - For transactions recorded between t0 and t1, either 404 or the update will be returned
+          - For transactions recorded after t1, the update will be returned
```
jagathweerasinghe-da: Once we agree on this, Ill add the release note.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1405,6 +1406,58 @@ class UpdateHistory(
...
+        participant_offset,
+        domain_id,
+        migration_id,
+        effective_at,
+        root_event_ids,
+        workflow_id,
+        command_id,
+        external_transaction_hash
+      from  update_history_transactions
+      where external_transaction_hash = $safeExtTxnHash
```
rautenrieth-da, line 1434: Looks reasonable, we have an async created index `updt_hist_tran_hi_eth` for this.

apps/scan/src/main/openapi/scan.yaml
rautenrieth-da: Don't have a strong opinion, I'd use `/hash/{hash}`, as I'm not aware of any other kind of hash used for transactions.

If you do need a longer name, use `/v2/updates/external-transaction-hash/{hash}` and not `/v2/updates/ext_txn_hash/{hash}`.

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -444,6 +444,43 @@ paths:
...
+          in: "query"
+          schema:
+            $ref: "#/components/schemas/DamlValueEncoding"
+      responses:
+        "200":
+          description: ok
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/UpdateHistoryItemV3"
```
rautenrieth-da: Why do we need a V3 version? And why is `/v2/updates/{update_id}` and `/v2/updates/hash/{hash}` returning different data types?

If it's just about making the external_transaction_hash non-optional, then I would not bother creating a new type and just describe it in the description.

apps/scan/src/main/openapi/scan.yaml
rautenrieth-da:
```suggestion
        Returns the update associated with the given hash of the prepared transaction.
        
        This endpoint is not always BFT safe.
        For transactions committed before a scan instance started indexing hashes, the instance will return a 404 error.
        For transactions committed around the time different scans started indexing hashes, some scan instances might return a 404 error while others return the matching update.
        
       This is in contrast to the `v2/updates` and `v2/updates/{update_id}` endpoints, which are guaranteed to be always BFT safe.
```

I'd keep it more general, something like this. We might also want to improve the description of the external transaction hash field, to note that the field might be missing for old transactions.

apps/scan/src/main/openapi/scan.yaml
jagathweerasinghe-da: Yes, the purpose of new schema type is to introduce the non-optional field. 

If we use a schema with optional field for a schema needing non-optional field, that creates some issues:

1. Degraded type-safety in the codebase, we can make a mistake by not including a required field in the response
2. API doc rendering usually/could drop optional fields from rendering, which makes it confusing for the API users (I saw some inconsistency, it could depend on the HTML renderer)

The cleanest approach for this case is to use allOf and override the required field like below:
```
    UpdateHistoryTransactionV3:
      allOf:
        - $ref: "#/components/schemas/UpdateHistoryTransactionV2"
        - type: object
          required:
            - external_transaction_hash
```
Even though OpenAPI specification allows this, guardrail does not override an optional field. 

What we can do here is to define a base transaction and use it in both UpdateHistoryTransactionV2 and UpdateHistoryTransactionV3 like below:
WDYT?

```
    UpdateHistoryTransactionV2:
      allOf:
        - $ref: "#/components/schemas/BaseUpdateHistoryTransaction"
        - type: object
          properties:
            external_transaction_hash:
              description: |
                For transaction externally signed, contains the external transaction hash
                signed by the external party. Can be used to correlate an external submission with a committed transaction.
              type: string
    UpdateHistoryTransactionV3:
      allOf:
        - $ref: "#/components/schemas/BaseUpdateHistoryTransaction"
        - type: object
          required:
            - external_transaction_hash
          properties:
            external_transaction_hash:
              description: |
                For transaction externally signed, contains the external transaction hash
                signed by the external party. Can be used to correlate an external submission with a committed transaction.
              type: string
```

rautenrieth-da: Yes, we're trading type safety for reduced code duplication (not only in our code base, but also for anyone who uses some client codegen), as UpdateHistoryTransactionV2 and UpdateHistoryTransactionV3 produce unrelated types in Scala, even though they are almost identical. In this particular case I think it's more convenient if you can reuse the type.

If you do want to stick with a separate type to enforce the non-optional field (which is also fine by me), please call it `UpdateHistoryTransactionV2WithHash` or similar, the V2 suffix means the type is used for the `/v2/updates` endpoint.

jagathweerasinghe-da: I would like to move forward with the separate type for type-safety and clarity. As you suggested, renamed it to ```UpdateHistoryTransactionV2WithHash```.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -1032,6 +1035,30 @@ class HttpScanHandler(
         )
     }
 
+  private def toUpdateV3(update: UpdateHistoryItem): UpdateHistoryItemV2WithHash =
```
rautenrieth-da:
```suggestion
  private def toUpdateV2WithHash(update: UpdateHistoryItem): UpdateHistoryItemV2WithHash =
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -1920,6 +1947,55 @@ class HttpScanHandler(
     }
   }
 
+  def getUpdateByHash(
+      hash: String,
+      encoding: DamlValueEncoding,
+      consistentResponses: Boolean,
```
rautenrieth-da: Do we need this parameter? It will always be `true`.

docs/src/release_notes_upcoming.rst
```diff
@@ -7,4 +7,11 @@
 
 .. NOTE: add your upcoming release notes below this line. They are included in the `release_notes.rst`.
 
+- Scan
+
+   - Added a new ``GET /v2/updates/hash/{hash}`` endpoint that returns the update associated with a given external transaction hash of a prepared transaction.
+     This endpoint is not always BFT safe: for transactions committed before a scan instance started indexing hashes, the instance will return a 404 error.
+     For transactions committed around the time different scans started indexing hashes, some scan instances might return a 404 while others return the matching update.
+     This is in contrast to the ``GET /v2/updates`` and ``GET /v2/updates/{update_id}`` endpoints, which are guaranteed to be always BFT safe.
+
```
rautenrieth-da:
```suggestion
     This endpoint is not always BFT safe, see the Scan OpenAPI documentation for details.
```
I'd keep it short, we have referred to the spec before in release notes.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardCliIntegrationTest.scala
```diff
@@ -260,6 +260,19 @@ class TokenStandardCliIntegrationTest
           )
         }
       }
+
+      clue(
+        "SV1 scan API (/vX/updates/hashes/{hash})  should return the update with external txn hash"
```
ray-roestenburg-da:
```suggestion
        "SV1 scan API (/vX/updates/hashes/{hash}) should return the update with external txn hash"
```

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -2509,11 +2554,28 @@ components:
...
-            signed by the external party. Can be used to correlate an external submission with a committed transaction.
-          type: string
+    UpdateHistoryTransactionV2:
+      allOf:
+        - $ref: "#/components/schemas/BaseUpdateHistoryTransaction"
+        - type: object
+          properties:
+            external_transaction_hash:
+              description: |
+                For transaction externally signed, contains the external transaction hash
```
```suggestion
                For an externally signed transaction, contains the external transaction hash
```

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -2509,11 +2554,28 @@ components:
...
+    UpdateHistoryTransactionV2WithHash:
+      allOf:
+        - $ref: "#/components/schemas/BaseUpdateHistoryTransaction"
+        - type: object
+          required:
+            - external_transaction_hash
+          properties:
+            external_transaction_hash:
+              description: |
+                For transaction externally signed, contains the external transaction hash
```
```suggestion
                For an externally signed transaction, contains the external transaction hash
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
ray-roestenburg-da: Agree, there is no v3

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
ray-roestenburg-da: +1

ray-roestenburg-da: No need to add hash to v0 or v1 responses (also confusing that there is no V2, even if V2 and V1 would be the same, I know, not done in this PR, though it reads weird to see getUpdateV2 call consistentResponses = true, and then in getUpdateById: version = if (consistentResponses) ScanHttpEncodings.V1 else ScanHttpEncodings.V0 Maybe you can clean this up in this PR or in a separate one?

ray-roestenburg-da:
Let's make sure this lands in the same release as https://github.com/hyperledger-labs/splice/pull/4796 so we don't end up with two versions, one responding possibly with null, and one omitting nulls.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
jagathweerasinghe-da: > Do we need this parameter? It will always be `true`.

Removed

jagathweerasinghe-da: > getUpdateV2

lets do the refactoring related getUpdateV2, in a later PR.

PR #4247, opened 2026-03-04
Handle sequencer connections per physical synchronizer

nicu-da:
Add scan support for sequencer serial id and also to return the current active serial according to the SV participant. We keep backwards compatiblity as if there are no sequencer that defined a serial then we use the migration id.

[ci]

fixes  https://github.com/hyperledger-labs/splice/issues/4240

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DecentralizedSynchronizerMigrationIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DecentralizedSynchronizerMigrationIntegrationTest.scala
@@ -1113,7 +1113,8 @@ class DecentralizedSynchronizerMigrationIntegrationTest
                     inside(sv1ScanLocalBackend.listDsoSequencers()) {
                       case Seq(DomainSequencers(synchronizerId, sequencers)) =>
                         synchronizerId shouldBe decentralizedSynchronizerId
-                        sequencers.map { sequencer =>
+                        val migrationSeqeuncers = sequencers.filter(_.serial.isEmpty)
+                        migrationSeqeuncers.map { sequencer =>
                           (sequencer.migrationId, sequencer.url)
                         }.toSet shouldBe Set(
                           (1L, getPublicSequencerUrl(sv1LocalBackend)),


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -633,32 +635,77 @@ class HttpScanHandler(
...
+        nodeName: String,
+        synchronizerConfig: SynchronizerNodeConfig,
+    ) = {
+      synchronizerConfig.physicalSynchronizers.toScala.toList.flatMap(_.asScala.flatMap {
+        case (serial, nodeConfig) =>
+          nodeConfig.sequencer
+            .flatMap(sequencerConfig =>
+              sequencerConfig.availableAfter.map { availableAfter =>
+                definitions.DsoSequencer(
+                  -1,
```
moritzkiefer-da: does -1 mean physical synchronizer? Maybe define it as a consant somewhere with a clear name and description.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/domainmigration/DomainMigrationInitializer.scala
```diff
@@ -416,7 +416,8 @@ class DomainMigrationInitializer(
                   .initializeFromGenesisState(
                     genesisState,
                     localSynchronizerNodes.current.staticSynchronizerParameters(
-                      NonNegativeInt.zero
+                      // workaround to still keep the HDM test working
+                      NonNegativeInt.one
```
moritzkiefer-da, line 420: I don't understand this. why do we need this?

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/domain/DomainConnector.scala
```diff
@@ -87,7 +87,19 @@ class DomainConnector(
...
     config.domains.global.url match {
       case None =>
-        waitForSequencerConnectionsFromScan(clock)
+        participantAdminConnection
+          .lookupPhysicalSynchronizerId(config.domains.global.alias)
+          .flatMap {
+            case Some(serial) =>
+              waitForSequencerConnectionsFromScan(clock, serial.serial)
+            case None =>
+              logger.info("No registerd physicial synchronizer id, using active id from scan")
```
moritzkiefer-da:
```suggestion
              logger.info("No registered physicial synchronizer id, using active id from scan")
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
nicu-da: Yes, migrationId is marked as required and it's part of the public api so we can't easily make it optional with possibly breaking parsers. 
Will use a nice constant.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/domainmigration/DomainMigrationInitializer.scala
nicu-da: For the HDM we basically go from serial 0 to serial 0, so with the new way we track sequencers by the serial some could actually be from migration 0 and some from migration 1 for the same serial 0. So we would break the HDM test as we would mix sequencers. 
This way we do the migration from migration id 0 and serial 0 to migration id 1 and serial 1

PR #4252, opened 2026-03-04
Use LedgerIngestionService (now named RetryingService) for Bulk Storage

isegall-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
@@ -11,16 +11,13 @@ import org.apache.pekko.Done
 import org.apache.pekko.actor.ActorSystem
 import org.apache.pekko.stream.KillSwitches
 import org.apache.pekko.stream.scaladsl.{Keep, Sink, Source}
-import org.lfdecentralizedtrust.splice.automation.{
-  InfiniteServiceWithShutdown,
-  RetryableInfiniteService,
-}
+import org.lfdecentralizedtrust.splice.automation.{ServiceWithShutdown, RetryingService}
 import org.lfdecentralizedtrust.splice.config.AutomationConfig
 import org.lfdecentralizedtrust.splice.environment.RetryProvider
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class PekkoRetryableInfiniteService[S](
+class PekkoRetryingService[S](
     source: Source[S, ?],
     // Note that the killSwitch for termination will be placed between the source and the sink.
     sink: Sink[Any, Future[Done]],
@@ -30,13 +27,13 @@ class PekkoRetryableInfiniteService[S](
     override protected val retryProvider: RetryProvider,
     override protected val loggerFactory: NamedLoggerFactory,
 )(implicit as: ActorSystem, ec: ExecutionContext, tracer: Tracer)
-    extends RetryableInfiniteService(automationConfig, backoffClock, description)
+    extends RetryingService(automationConfig, backoffClock, description)
     with NamedLogging {
 
-  // TODO: don't we always just want to do that in the constructor of RetryableInfiniteService?
-  startIngestion()
+  // kick-off processing
+  start()
 
-  private class PekkoInfiniteServiceWithShutdown extends InfiniteServiceWithShutdown {
+  private class PekkoServiceWithShutdown extends ServiceWithShutdown {
 
     private val (killSwitch, done) =
       source.viaMat(KillSwitches.single)(Keep.right).toMat(sink)(Keep.both).run()
@@ -53,6 +50,6 @@ class PekkoRetryableInfiniteService[S](
 
   override protected def instantiateService()(implicit
       traceContext: TraceContext
-  ): Future[InfiniteServiceWithShutdown] = Future.successful(new PekkoInfiniteServiceWithShutdown())
+  ): Future[ServiceWithShutdown] = Future.successful(new PekkoServiceWithShutdown())
 
 }
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/RetryingService.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/RetryingService.scala
@@ -18,7 +18,7 @@ import io.opentelemetry.api.trace.Tracer
 import java.util.concurrent.atomic.AtomicReference
 import scala.concurrent.{ExecutionContext, Future}
 
-abstract class InfiniteServiceWithShutdown {
+abstract class ServiceWithShutdown {
   def initiateShutdown()(implicit tc: TraceContext): Unit
   def completed: Future[Done]
   def isActive: Boolean
@@ -27,7 +27,7 @@ abstract class InfiniteServiceWithShutdown {
 /** Abstract class to share the retry and shutdown logic between different infinite services,
   * e.g. ones for ingesting ledger data using a subscription to a Ledger API stream.
   */
-abstract class RetryableInfiniteService(
+abstract class RetryingService(
     config: AutomationConfig,
     backoffClock: Clock,
     description: String,
@@ -43,28 +43,28 @@ abstract class RetryableInfiniteService(
   /** Allocate a new instance of the service. */
   protected def instantiateService()(implicit
       traceContext: TraceContext
-  ): Future[InfiniteServiceWithShutdown]
+  ): Future[ServiceWithShutdown]
 
-  // Note that we are tracking the current subscription outside the retry loop instead of just
-  // calling 'runOnShutdown' on every newly acquired subscription, as that would leak memory.
+  // Note that we are tracking the current instance of the service outside the retry loop instead of just
+  // calling 'runOnShutdown' on every newly acquired instantiation, as that would leak memory.
   private val currentServiceInstance =
-    new AtomicReference[Option[InfiniteServiceWithShutdown]](None)
+    new AtomicReference[Option[ServiceWithShutdown]](None)
   private val serviceTerminatedF = new AtomicReference[Future[Done]](Future.successful(Done))
 
   retryProvider.runOnOrAfterClose_(new RunOnClosing {
     override def name: String = s"terminate service $description"
-    // this is not perfectly precise, but SpliceLedgerSubscription.initiateShutdown is idempotent
+    // this is not perfectly precise, but `initiateShutdown`s should be idempotent
     override def done: Boolean = false
     override def run()(implicit tc: TraceContext): Unit = currentServiceInstance
       .get()
-      .foreach(subscription => {
+      .foreach(instance => {
         logger
           .debug(s"Terminating service $description, as we are shutting down.")(TraceContext.empty)
-        subscription.initiateShutdown()
+        instance.initiateShutdown()
       })
   })(TraceContext.empty)
 
-  protected def startIngestion(): Unit = {
+  protected def start(): Unit = {
     withNewTrace(description)(implicit traceContext =>
       _ => {
         logger.debug(s"Starting service $description")
@@ -82,10 +82,10 @@ abstract class RetryableInfiniteService(
                 // 2. If we get very infrequent errors (e.g. stale stream authorization on user addition), no error is logged an we get fast retries
                 //    instead of sleeping for the polling interval just because a certain number of users got allocated.
                 RetryFor.LongRunningAutomation,
-                "infinite_service",
+                "service",
                 description, {
                   instantiateService().flatMap(service => {
-                    // Smuggle the current subscription out of the body here, so that we can use
+                    // Smuggle the current instance out of the body here, so that we can use
                     // runOnShutdown outside to signal the termination via a call to .initiateShutdown().
                     currentServiceInstance.set(Some(service))
                     // The creation of the new service races with the call to close the content of `currentServiceInstance`, which is issued
@@ -98,7 +98,7 @@ abstract class RetryableInfiniteService(
                       service.initiateShutdown()
                     }
                     // The actual return value of the future being retried is the future inside the SpliceLedgerConnection,
-                    // which signals when the subscription terminated.
+                    // which signals when the instance terminated.
                     service.completed.map(_ =>
                       if (retryProvider.isClosing)
                         Done // This is the normal path that we hit when we are shutting down.
@@ -107,7 +107,7 @@ abstract class RetryableInfiniteService(
                         // We consider it transient error that we want to retry on in the hope of hitting a live server.
                         throw Status.ABORTED
                           .withDescription(
-                            "Unexpected closing of subscription, likely due to server shutdown."
+                            "Unexpected closing of a service instance, likely due to server shutdown."
                           )
                           .asRuntimeException()
                       }
@@ -143,7 +143,7 @@ abstract class RetryableInfiniteService(
   }
 
   final override def isHealthy: Boolean =
-    // Healthy if there's an active subscription
+    // Healthy if there's an active instance
     currentServiceInstance.get().exists(_.isActive)
 
   final override def closeAsync(): Seq[AsyncOrSyncCloseable] = {
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
@@ -18,13 +18,7 @@ import com.digitalasset.canton.admin.api.client.data.parties.PartyDetails
 import com.digitalasset.canton.config.NonNegativeFiniteDuration
 import com.digitalasset.canton.data.CantonTimestamp
 import com.digitalasset.canton.ledger.error.LedgerApiErrors
-import com.digitalasset.canton.lifecycle.{
-  AsyncCloseable,
-  AsyncOrSyncCloseable,
-  FlagCloseableAsync,
-  FutureUnlessShutdown,
-  SyncCloseable,
-}
+import com.digitalasset.canton.lifecycle.FutureUnlessShutdown
 import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
 import com.digitalasset.canton.logging.{NamedLoggerFactory, NamedLogging}
 import com.digitalasset.canton.protocol.LocalRejectError.ConsistencyRejections.InactiveContracts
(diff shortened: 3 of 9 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
```diff
@@ -665,14 +666,19 @@ class BaseLedgerConnection(
...
+  *  (hence the name).
+  */
 class SpliceLedgerSubscription[S](
     source: Source[S, NotUsed],
     map: S => Future[Unit],
     override protected[this] val retryProvider: RetryProvider,
     override protected val loggerFactory: NamedLoggerFactory,
 )(implicit ec: ExecutionContext, mat: Materializer)
-    extends RetryProvider.Has
+    extends InfiniteServiceWithShutdown
```
moritzkiefer-da: This seems subtly different from PekkoInfiniteServiceWithShutdown and I don't understand why.

In the end this function also has a sink. So why can we not put all the generic code in PekkoInfiniteServiceWithShutdown and then make this extend PekkoInfiniteServiceWithShutdown passing the Flow[S].mapAsync(…).toMap(Sink.ignore) part as the sink?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorageTest.scala
```diff
@@ -151,49 +154,52 @@ class AcsSnapshotBulkStorageTest
...
+        retryProvider,
+      )
 
-        _ = clue("Initially, a single snapshot is dumped") {
-          probe.request(2)
-          probe.expectNext(2.minutes) shouldBe TimestampWithMigrationId(ts1, 0)
-          probe.expectNoMessage(10.seconds)
+      Using.resources(svc, retryProvider) { (_, _) =>
+        clue("Initially, a single snapshot is dumped") {
+          eventually(4.minutes) {
```
moritzkiefer-da, line 180: why do we now need a 4 minute eventually in here?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/RetryableInfiniteService.scala
```diff
@@ -22,12 +18,19 @@ import io.opentelemetry.api.trace.Tracer
 import java.util.concurrent.atomic.AtomicReference
 import scala.concurrent.{ExecutionContext, Future}
 
-/** Abstract class to share the retry and shutdown logic between different services for ingesting ledger data using
-  * a subscription to a Ledger API stream.
+abstract class InfiniteServiceWithShutdown {
```
ray-roestenburg-da: i think I would just drop the `Infinite`. it's a service with shutdown and a service implies that it will run until it is shutdown.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/RetryableInfiniteService.scala
```diff
@@ -22,12 +18,19 @@ import io.opentelemetry.api.trace.Tracer
...
+  def initiateShutdown()(implicit tc: TraceContext): Unit
+  def completed: Future[Done]
+  def isActive: Boolean
+}
+
+/** Abstract class to share the retry and shutdown logic between different infinite services,
+  * e.g. ones for ingesting ledger data using a subscription to a Ledger API stream.
   */
-abstract class LedgerIngestionService(
+abstract class RetryableInfiniteService(
```
ray-roestenburg-da:
```suggestion
abstract class RetryableService(
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/RetryableInfiniteService.scala
```diff
@@ -78,28 +82,28 @@ abstract class LedgerIngestionService(
                 // 2. If we get very infrequent errors (e.g. stale stream authorization on user addition), no error is logged an we get fast retries
                 //    instead of sleeping for the polling interval just because a certain number of users got allocated.
                 RetryFor.LongRunningAutomation,
-                "ledger_ingestion",
-                "ledger ingestion subscription", {
-                  newLedgerSubscription().flatMap(subscription => {
+                "infinite_service",
```
ray-roestenburg-da:
```suggestion
                "service",
```
and everywhere you use infinite

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryableInfiniteService.scala
```diff
@@ -0,0 +1,58 @@
...
+    backoffClock: Clock,
+    description: String,
+    override protected val retryProvider: RetryProvider,
+    override protected val loggerFactory: NamedLoggerFactory,
+)(implicit as: ActorSystem, ec: ExecutionContext, tracer: Tracer)
+    extends RetryableInfiniteService(automationConfig, backoffClock, description)
+    with NamedLogging {
+
+  // TODO: don't we always just want to do that in the constructor of RetryableInfiniteService?
+  startIngestion()
```
ray-roestenburg-da: I would prefer if we have a separate `start` method for this, and then of course also have a `stop` method (or call it shutdown if you want).

ray-roestenburg-da: You should be able to construct a service and not necessarily immediately start it

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
```diff
@@ -0,0 +1,58 @@
...
+    extends RetryableInfiniteService(automationConfig, backoffClock, description)
+    with NamedLogging {
+
+  // TODO: don't we always just want to do that in the constructor of RetryableInfiniteService?
+  startIngestion()
+
+  private class PekkoInfiniteServiceWithShutdown extends InfiniteServiceWithShutdown {
+
+    private val (killSwitch, done) =
+      source.viaMat(KillSwitches.single)(Keep.right).toMat(sink)(Keep.both).run()
```
ray-roestenburg-da, line 39: would also prefer an explicit `start` method that does this

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/RetryableInfiniteService.scala
ray-roestenburg-da: I think it makes sense for this to be more symmetrical, and also have a `start()` method that indicates that it has started by returning a `Future[Done]`. Right now it is implied that the service starts, but you don't know when it is started, or if it fails (say asynchronously and the exception is thrown from another thread, so the caller of constructor does not know).

Also, initiateShutdown and completed could be simplified into `shutdown(): Future[Done]`

ray-roestenburg-da: Also this can be a trait, no need for abstract class

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/RetryingService.scala
```diff
@@ -37,33 +40,34 @@ abstract class LedgerIngestionService(
...
       .foreach(subscription => {
         logger
-          .debug(s"Terminating ledger ingestion loop, as we are shutting down.")(TraceContext.empty)
+          .debug(s"Terminating service $description, as we are shutting down.")(TraceContext.empty)
         subscription.initiateShutdown()
       })
   })(TraceContext.empty)
 
   protected def startIngestion(): Unit = {
-    withNewTrace("ledger ingestion loop")(implicit traceContext =>
```
ray-roestenburg-da, line 64: it's still called startIngestion? or am I reading the diff wrong?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/RetryableInfiniteService.scala
ray-roestenburg-da: This in itself could also extend InfiniteServiceWithShutdown so you can shut it down, start it etc in the same way (and internally delegate to the service that you want to wrap, which you can pass as a constructor argument). So you get something like `RetryableInfiniteService(service)` though I would suggest to rename that maybe to `ResilientService(service)` or `AutoRestart(service)` or something like that, in that way other services don't even need to extend this, you just wrap it around when you construct it. (this does not have to be an abstract class at all)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/RetryableInfiniteService.scala
```diff
@@ -37,33 +40,34 @@ abstract class LedgerIngestionService(
...
     // this is not perfectly precise, but SpliceLedgerSubscription.initiateShutdown is idempotent
     override def done: Boolean = false
-    override def run()(implicit tc: TraceContext): Unit = currentSubscription
+    override def run()(implicit tc: TraceContext): Unit = currentServiceInstance
       .get()
       .foreach(subscription => {
         logger
-          .debug(s"Terminating ledger ingestion loop, as we are shutting down.")(TraceContext.empty)
+          .debug(s"Terminating service $description, as we are shutting down.")(TraceContext.empty)
         subscription.initiateShutdown()
```
ray-roestenburg-da: rename `subscription`? there is also many comments still using that name

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
```diff
@@ -665,14 +666,19 @@ class BaseLedgerConnection(
   }
 }
 
-/** Subscription for reading the ledger */
+/** An infinite service that reads from a Pekko source, and runs an async function on it.
+  *  Optimized for closing the source quickly, so should be used in cases where the source may be closing in parallel
+  *  to the service itself, thus might cause shutdown noise. Currently, this is true for ledger ingestion services
+  *  (hence the name).
+  */
 class SpliceLedgerSubscription[S](
```
ray-roestenburg-da: Please move this class outside of SpliceLedgerConnection, fine to be a separate file. it has really not much to do with the SpliceLedgerConnection class.

ray-roestenburg-da: (I would also suggest renaming it for what it does, it's really quite abstract, like a SourceMapper or whatever, it does nothing specific with a ledger)

ray-roestenburg-da: Even the log messages don't make sense if this class is used anywhere else:
"Ignoring failure to handle transaction, as we are shutting down"
"terminating ledger api stream"
"ledger api stream terminated"

It is a `Source[S, NotUsed]` ..

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
```diff
@@ -0,0 +1,58 @@
...
+  InfiniteServiceWithShutdown,
+  RetryableInfiniteService,
+}
+import org.lfdecentralizedtrust.splice.config.AutomationConfig
+import org.lfdecentralizedtrust.splice.environment.RetryProvider
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class PekkoRetryableInfiniteService[S](
+    source: Source[S, ?],
```
ray-roestenburg-da, line 21: Is fine to be `NotUsed` as mat cause you can always keep another Mat in a bigger flow, or just to have another generic type `Mat` if you really wanted to keep track of it.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
```diff
@@ -0,0 +1,58 @@
...
+    // Note that the killSwitch for termination will be placed between the source and the sink.
+    sink: Sink[Any, Future[Done]],
```
ray-roestenburg-da, line 23: It's easy enough to compose source and sink somewhere else

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryableInfiniteService.scala
```diff
@@ -0,0 +1,58 @@
...
+    automationConfig: AutomationConfig,
+    backoffClock: Clock,
+    description: String,
+    override protected val retryProvider: RetryProvider,
+    override protected val loggerFactory: NamedLoggerFactory,
+)(implicit as: ActorSystem, ec: ExecutionContext, tracer: Tracer)
+    extends RetryableInfiniteService(automationConfig, backoffClock, description)
+    with NamedLogging {
+
+  // TODO: don't we always just want to do that in the constructor of RetryableInfiniteService?
```
ray-roestenburg-da: I would definitely have a separate method for this, not during constructor time.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
ray-roestenburg-da, line 39: so you could have passed in a function `() => (KillSwitch, Future[Done]`  and call that

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
isegall-da: Now that I understand this better, I need to improve the comment that I added.
What this class does mostly is make sure that the tail of the flow (everything after the killswitch) fully completes before reporting completed. Naively putting that Flow[S].mapAsync(…).toMap(Sink.ignore) part as the sink could very easily destroy that property if we're not careful.
So while I agree that it would be definitely more elegant to refactor this into a "ClosingSink" that provides these guarantees that it's a sink that will never report completed too early, that would be a slightly bigger refactoring, and I don't want to fall into the rabbit hole of breaking this code that's been working fine for a long time.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
isegall-da: I could, but why is that better?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
isegall-da: Sorry, I don't understand what you're suggesting and why.

ray-roestenburg-da: usually you construct a flow in many ways, and then this thing can run _any_ graph, not just `source.viaMat(KillSwitches.single)(Keep.right).toMat(sink)(Keep.both).run()`, so in that way it would be more flexible to use

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
ray-roestenburg-da: what if you want to add some operation to the flow, but only for a certain case, you just construct it differently, and construct it with`PekkoRetryableInfiniteService(f)` and you never have to change PekkoRetryableInfiniteService when you change that flow / graph, as long as it ends up with a killswitch and a completion signal

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
isegall-da: Sure, but I'd do that if and when we have a need for more complex graphs. Right now, just moving that logic to the caller adds nothing and IMHO making this interface less intuitive, as the caller will need to make sure they put a killSwitch in their flow while the caller does not need the killSwitch themselves.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
isegall-da: Let's solve that problem if and when we have it, not make things more complex right now for a "what if" scenario

ray-roestenburg-da: sure, not a big issue

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
ray-roestenburg-da: sure no big deal

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
isegall-da: "is fine" in what sense? That would be more restrictive, and forces the caller to have a `Source[S, NotUsed]` (which is actually not the case right now), for no reason.

ray-roestenburg-da: Ok maybe too restrictive, never mind. NotUsed would be clear to show it really is not used but not too important, since you don't know the type with ? that is also implied.

isegall-da: yup, ? says exactly that. `NotUsed` here says "I force _the caller_ to not use it", which is more restrictive.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryableInfiniteService.scala
isegall-da: While I agree with this directionally, I tried doing that now and it propagates to a larger change that make other things inconsistent, as for all other services (e.g. automation trigger) we do assume that they start on construction. So for now I'm just keeping the call to start() here.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorageTest.scala
isegall-da: The retries were broken (or rather, the error injection was. It was injecting errors on a method that was no longer used, `writeFullObject`), so now that that's fixed, things are slower because we actually inject errors and restart the whole workflow a couple of times.

PR #4255, opened 2026-03-04
Fix validator scan seed error from failed scan service

pasindutennage-da:
Fixes https://github.com/DACH-NY/canton-network-internal/issues/3804

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
@@ -1332,19 +1332,26 @@ abstract class TopologyAdminConnection(
     )
 
   def initId(id: NodeIdentity)(implicit traceContext: TraceContext): Future[Unit] = {
-    runCmd(
-      TopologyAdminCommands.Init
-        .InitId(id.uid.identifier.toProtoPrimitive, id.uid.namespace.toProtoPrimitive, Seq.empty)
-    ).map(_ => ()).recover {
-      // We catch the FAILED_PRECONDITION error specifically when it mentions being already initialized
-      case e: io.grpc.StatusRuntimeException
-          if e.getStatus.getCode == io.grpc.Status.Code.FAILED_PRECONDITION &&
-            e.getMessage.contains("Already initialised") =>
-        logger.info(
-          "initId called on an already initialized participant; treating as success (likely resumed from backup)."
-        )
-        ()
-    }
+    retryProvider.ensureThatB(
+      RetryFor.WaitingOnInitDependency,
+      "init_id",
+      show"Node is initialized with ID $id",
+      getIdOption().map {
+        case GetIdResult(true, Some(uid)) if uid == id.uid => true
+        case GetIdResult(true, Some(uid)) =>
+          throw Status.FAILED_PRECONDITION
+            .withDescription(
+              s"Node is already initialized with a different ID: $uid (expected ${id.uid})"
+            )
+            .asRuntimeException()
+        case _ => false
+      },
+      runCmd(
+        TopologyAdminCommands.Init
+          .InitId(id.uid.identifier.toProtoPrimitive, id.uid.namespace.toProtoPrimitive, Seq.empty)
+      ).map(_ => ()),
+      logger,
+    )
   }
 
   def identity()(implicit


apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
```diff
@@ -141,8 +139,6 @@ class ValidatorApp(
   override def preInitializeBeforeLedgerConnection()(implicit
       traceContext: TraceContext
   ): Future[Unit] = for {
-    // TODO(tech-debt) consider removing early version check once we switch to a non-dev Canton protocol version
```
martinflorian-da, line 144: So this looks fine because we're on PV 34, which is non-dev.

But what if we bump the PV to 35 and someone didn't upgrade? Then it should just fail on the Canton side without data corruption or bigger problems?

martinflorian-da: Actually, I don't understand what this version check was doing in the first place. Removing seems fine....

martinflorian-da (approved):
Thanks!

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
martinflorian-da: Hm hm reading the code some more... if we set `failOnVersionMismatch = true` it actually does crash here if the version is not right. Not sure if we're ever going to use that though in the current form.

martinflorian-da: Ignore me, I had missed the conversation at https://github.com/DACH-NY/canton-network-internal/issues/3804#issuecomment-4005925053 :shrug:

pasindutennage-da: Thanks

project/ignore-patterns/canton_network_test_log.ignore.txt
```diff
@@ -117,6 +117,7 @@ Failed to connect to scan of Digital-Asset-Eng-3.*AppUpgradeIntegrationTest
 Failed to connect to scan of .*DisasterRecoveryIntegrationTest
 Failed to connect to scan of .*AppUpgradeIntegrationTest
 Failed to connect to scan of .*ConfigurationProvidedBftScanConnectionIntegrationTest
+Failed to connect to scan of .*BftScanConnectionIntegrationTest
```
pasindutennage-da: when aliceValidator tries to connect to Sv1Scan, which is down, we get this.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
```diff
@@ -299,8 +295,16 @@ class ValidatorApp(
...
                 else
                   appInitStep("Ensuring decentralized synchronizer registered") {
-                    domainConnector
-                      .ensureDecentralizedSynchronizerRegisteredAndConnectedWithCurrentConfig(clock)
+                    retryProvider.retry(
+                      RetryFor.WaitingOnInitDependency,
+                      "ensure_synchronizer_registered",
+                      "registering decentralized synchronizer",
+                      domainConnector
+                        .ensureDecentralizedSynchronizerRegisteredAndConnectedWithCurrentConfig(
```
martinflorian-da: Doesn't that already have retries internally?

project/ignore-patterns/canton_network_test_log.ignore.txt
martinflorian-da: Why not suppress these logs in the test? Much more targeted if it works; we only ignoring them exactly where we expect them to happen.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
pasindutennage-da: Actually yes, it has

```
private def waitForSequencerConnectionsFromScan(clock: Clock)(implicit
      tc: TraceContext
  ): Future[Map[SynchronizerAlias, SequencerConnections]] = {
    retryProvider.getValueWithRetries(
      // Short retries since usually a failure here is just a misconfiguration error.
      // The only case where this can happen is during a domain migration and even then
      // it is fairly unlikely outside of tests for validators to come up fast enough that
      // scan has not yet updated.
      RetryFor.WaitingOnInitDependency, // because the scan connections might still be in bootstrap phase
      "scan_sequencer_connections",
      "non-empty sequencer connections from scan",
      getSequencerConnectionsFromScan(Right(clock))
 
```

pasindutennage-da: Will remove redundant one

project/ignore-patterns/canton_network_test_log.ignore.txt
pasindutennage-da: Fixed

moritzkiefer-da (approved):
thanks!

moritzkiefer-da:
consider also including it in the release notes

moritzkiefer-da:
oops sorry didn't intend to actually trigger one, I'll just ignore that one and let you trigger one

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/TopologyAdminConnection.scala
```diff
@@ -1335,7 +1335,16 @@ abstract class TopologyAdminConnection(
     runCmd(
       TopologyAdminCommands.Init
         .InitId(id.uid.identifier.toProtoPrimitive, id.uid.namespace.toProtoPrimitive, Seq.empty)
-    )
+    ).map(_ => ()).recover {
```
pasindutennage-da: @moritzkiefer-da @martinflorian-da 

### Failing DockerComposeValidatorFrontendIntegrationTest (3 times)

```
Seq(
      "build-tools/splice-compose.sh",
      "restore_node",
      validatorBackup.toString,
      participantBackup.toString,
      "0",
    ) !

    val identities = backupsDir.resolve("identities.json")

    withComposeValidator("after restoring from backup") {
```

### Logs

16:22:20.720 validator_backend  /GetId to participant:5002: sending request

16:22:20.906 participant Resuming with existing uid da-composeValidator-13

16:22:20.943  participant /GetId by /172.19.0.5:59356: received a message

16:22:20.962 validator_backend  /GetId to participant:5002: received a message: Node has no identity, generating a new one.

16:22:21.368 validator_backend  /InitId to participant:5002: sending request 

16:22:21.398Z  participant /InitId by /172.19.0.5:59356: failed with FAILED_PRECONDITION/Already initialised Init node id  

16:22:21.410  validator_backend  Initialize app failed io.grpc.StatusRuntimeException: FAILED_PRECONDITION: Already initialised Init node id      

### Conclusion

The test failure was caused by a race condition. The Validator prematurely assumed the participant was uninitialized while it was still resuming identity state. Consequently, it attempted a redundant initId call, which Canton rejected with a FAILED_PRECONDITION: Already initialised error, crashing the app.

moritzkiefer-da: Can we instead use something ilke
`ensureThat(
  getIdOption == expectedId,
initId
)
`

PR #4268, opened 2026-03-05
[canton-3.5] Fix some DABFT tests

moritzkiefer-da:
Factoring out the stuff that is just a pretty clear fix so we can already merge it.

[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/resources/bft-sequencer-global-domain-overrides.conf
+++ b/apps/app/src/test/resources/bft-sequencer-global-domain-overrides.conf
@@ -3,18 +3,22 @@ canton {
       globalSequencerSv1.sequencer = ${_sequencer_bft_template}
       globalSequencerSv1.sequencer.config.initial-network.server-endpoint.port = 5110
       globalSequencerSv1.sequencer.config.initial-network.server-endpoint.external-port = 5110
+      globalSequencerSv1.sequencer.config.consensus-empty-block-creation-timeout = 250ms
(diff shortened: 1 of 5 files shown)


apps/app/src/test/resources/bft-sequencer-global-domain-overrides.conf
```diff
@@ -3,18 +3,22 @@ canton {
       globalSequencerSv1.sequencer = ${_sequencer_bft_template}
       globalSequencerSv1.sequencer.config.initial-network.server-endpoint.port = 5110
       globalSequencerSv1.sequencer.config.initial-network.server-endpoint.external-port = 5110
+      globalSequencerSv1.sequencer.config.consensus-empty-block-creation-timeout = 250ms
```
nicu-da, line 6: out of curiosity why do we need this?

moritzkiefer-da: ensures time advances faster which can help speed up some tests when you need that to happen.
I initially thought you may also need it to speed things up when you have a submission against one node because you need to wait for the other nodes to complete their work which can involve sending an empty block if they don't have a submission. There is some logic that should actually make it produce an empty block faster in those cases so in theory it shouldn't matter but that logic seems at least partially buggy (although that was when I configured different timeouts on different nodes).

I'm honestly not sure if we did end up relying on it in the end but Canton sets it in all their integration tests and I couldn't be bothered to debug it without that in the end.

PR #4275, opened 2026-03-05
Release Notes for 0.5.14

OriolMunoz-da:
see latest screenshot below

### DIFF
--- a/docs/src/release_notes_upcoming.rst
+++ b/docs/src/release_notes_upcoming.rst
@@ -7,4 +7,4 @@
 
 .. NOTE: add your upcoming release notes below this line. They are included in the `release_notes.rst`.
 
-.. release-notes:: Upcoming
+.. .. release-notes:: Upcoming


docs/src/release_notes.rst
```diff
@@ -18,6 +18,43 @@
 
 .. _release_notes:
 
+.. release-notes:: 0.5.14
```
moritzkiefer-da, line 21: did you copy the release notes from Canton? I thought there were more changes?

OriolMunoz-da: discussed with @martinflorian-da , I omitted some changes that didn't seem relevant to Splice from https://github.com/DACH-NY/canton/blob/f37c36b228480d4bcb7b354a3ac11a58dc7d5d2a/UNRELEASED.md?plain=1#L7
Would you copy all as-is?

moritzkiefer-da: Definitely happy to edit it but the edit here feels too aggressive. E.g. the entire section under `JSON Ledger API OpenAPI/AsyncAPI Specification Updates` does seem relevant.

martinflorian-da: These are already in 0.5.13, no?

moritzkiefer-da: ah I see got tripped up by the dates in Canton not aligning with our releases, alright probably fine then

OriolMunoz-da: GAH, added and deleted.

docs/src/release_notes.rst
```diff
@@ -18,6 +18,55 @@
...
+
+      - *BREAKING* The
+
+          - ``/v2/updates`` HTTP POST and websocket GET endpoints
+          - ``/v2/updates/flats`` HTTP POST and websocket GET endpoints
+
+        were incorrectly retuning LedgerEffects events (i.e., ``CreatedEvent`` and ``ExercisedEvent``). They are now corrected to return
+        AcsDelta (flat) events (i.e., ``CreatedEvent`` and ``ArchivedEvent``).
+
+      - Corrected the OpenAPI and AsyncAPI specification files to properly reflect field requirements as defined in the Ledger API `.proto` files.
```
moritzkiefer-da: sorry for sending you down the wrong track, I think Martin is correct that those are already in 0.5.13 and should be removed again. The `updateFormat` one does seem new though

docs/src/release_notes.rst
martinflorian-da: I got confused at least 3 times.

I think we're only really missing this now but on careful reflection maybe let's add it after all?

```
#### Specification bugfix
`GetUpdatesRequest.updateFormat` now appears marked as required in the OpenAPI/AsyncAPI specs even though it was previously optional.
This is a specification fix to align with the actual behavior of the API, which requires this field.
If your client code was previously omitting this field, you will need to update it to include a valid value.
```

Sorry for the back and forth @OriolMunoz-da ... also I suggest we continue the practice of having a dedicated `Canton` bullet will all upstream release notes under there.

Specifically, let's distinguish between "changes to participant/mediator/sequencer things in Splice only" and "whatever upstream did".

OriolMunoz-da: done

docs/src/release_notes_upcoming.rst
```diff
@@ -8,18 +8,3 @@
 .. NOTE: add your upcoming release notes below this line. They are included in the `release_notes.rst`.
 
 .. release-notes:: Upcoming
```
martinflorian-da: Don't you want to comment that out?

martinflorian-da (approved):
Thank you!

PR #4279, opened 2026-03-05
Reconnect on synchronizer connection changes

moritzkiefer-da:
Kept the code to disable it given that we'll need it again in 3.5. Does mean until then it's untested but still seems easier to revive than ripping it out completely now only to then add it back.

[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/project/ignore-patterns/canton-standalone-participants-before-disaster-lost-all-sequencers-most-participants.ignore.txt
+++ b/project/ignore-patterns/canton-standalone-participants-before-disaster-lost-all-sequencers-most-participants.ignore.txt
@@ -9,3 +9,4 @@ Failed to acknowledge clean timestamp.*closed stream
 Now retrying operation 'request current time'
 
 'sequencer-client' is now in state Failed
+Connection has failed validation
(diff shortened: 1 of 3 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceConfig.scala
```diff
@@ -70,7 +70,9 @@ final case class CircuitBreakersConfig(
 
 final case class EnabledFeaturesConfig(
     enableNewAcsExport: Boolean = true,
-    newSequencerConnectionPool: Boolean = true,
+    // For now, we always need to do this as Canton does not update the configuration until a reconnect.
+    // On 3.5 we should be able to set it to false.
+    reconnectOnSynchronizerConfigurationChange: Boolean = true,
```
nicu-da, line 75: This is going to break things for the validators that disabled it explicitly. Should we add some backwards compat to the config reader? Or add release notes to force them to just drop the flag and add a note about the new flag?

moritzkiefer-da: > This is going to break things for the validators that disabled it explicitly. Should we add some backwards compat to the config reader?

Did we even tell anyone about that flag? I thought when we told people to disable the old pools we just mentioned the participant flag and skipped over this?

moritzkiefer-da: let me add it to the release notes though I guess, cheap enough.

nicu-da: Ah ok, I thought we mentioned both of them but it makes sense then. :shipit:

PR #4281, opened 2026-03-05
Stop vetting packages older than minimumInitialization

julientinguely-da:
fixes https://github.com/hyperledger-labs/splice/issues/4119

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BootstrapPackageConfigDarUploadIntegrationTest.scala
```diff
@@ -148,15 +148,32 @@ class BootstrapPackageConfigDarUploadIntegrationTest
       packageResource: PackageResource,
       requiredVersion: String,
   ): Unit = {
-    withClue(
-      s"dars for package ${packageResource.latest.metadata.name} should be up to $requiredVersion"
+    val dars =
```
moritzkiefer-da, line 190: I don't think this test is complete. There are two parts to check:

1. When you initialize a new node it does not vet the packages below minimumInitialization. This is what you are testing here.
2. A node that has already vetted packages below minimumInitialization (all existing nodes) unvet older versions. I don't think you are testing this.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/DarResources.scala
```diff
@@ -6611,4 +6611,27 @@ object DarResources {
 
   def lookupAllPackageVersions(name: PackageName): Seq[DarResource] =
     packageResources.view.flatMap(_.all).toSeq.filter(_.metadata.name == name)
+
+  def getRequiredPackageVersions(
```
moritzkiefer-da: I don't think your change works for actually unvetting things for validators that have already vetted the old versions. What we need to do is get the package ids for the old versions and explicitly remove them.

julientinguely-da: I made the changes to `PackageVettingTrigger` to unvet existing unsupported packages and added a test for both SV and validator.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnsupportedPackageVettingIntegrationTest.scala
```diff
@@ -0,0 +1,80 @@
...
+
+class UnsupportedPackageVettingIntegrationTest extends IntegrationTest {
+
+  override def environmentDefinition: SpliceEnvironmentDefinition =
+    EnvironmentDefinition
+      .simpleTopology1Sv(this.getClass.getSimpleName)
+
+  "Unsupported vetted packages are automatically removed by the package vetting trigger for SV and validator" in {
+    implicit env =>
+      val unsupportedDars = Seq(
```
moritzkiefer-da: Can we upload old splitwell versions that depend on amulet?

I suspect there is likely an issue because Canton has some checks that prevent unvetting a package that still has depnedencies on it. There is a force flag for it but trying to understand the impact of it. See thread in slack.

julientinguely-da: I've taken a softer approach by not removing these at all: https://github.com/hyperledger-labs/splice/blob/95d1db14b5e5e878d5ee8b02f53957b3c64bb012/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminDarsConnection.scala#L202 
But yeah, not transitive..

moritzkiefer-da: thx will review more carefully after lunch. The request for the test with splitwell stands either way though.

julientinguely-da: I added two unsupported splitwell dars and the tests succeeded

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PackageVettingTrigger.scala
```diff
@@ -54,6 +56,23 @@ abstract class PackageVettingTrigger(
...
+      vettedPackages <- participantAdminConnection.listVettedPackages(
+        participantId,
+        domainId,
+        AuthorizedState,
+      )
+      unsupportedPackages = DarResourcesUtil.filterUnsupportedPackageVersions(
+        vettedPackages.flatMap(_.mapping.packages).map(_.packageId)
+      )
+      _ = if (unsupportedPackages.nonEmpty) {
+        vetting.unvetPackages(
```
moritzkiefer-da, line 76: topology transactions are expensive, we shouldn't make one call to vetPackages and one to unvetPackages but instead make a single one that tries to add new packages and remove old ones at the same time.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminDarsConnection.scala
```diff
@@ -137,31 +137,99 @@ trait ParticipantAdminDarsConnection {
...
     )
-  private def updateVettingStateForDars(
+
+  private def removeDarsFromVettingState(
+      dars: Seq[DarResource],
+      currentVetting: VettedPackages,
+  ): VettedPackages = {
+    val packageIdsToRemove: Seq[String] = dars.map(_.packageId)
+    // filter out all dependencies that may exist with the resulting packages before removing a package
+    val packageIdsToKeep = DarResourcesUtil
```
moritzkiefer-da: I don't think this check works. Let's say we have a package X that depends on a splice-amulet version but is not in the splice repo. DarResourceUtils.getDarResources will give you nothing here since our code only knows our packages. So `packageIdsToKeep` will be empty. So then `safePackageIdsToRemove` will include the splice amulet version. But unvetting will fail as you can't unvet it without also unvetting X or using a force flag.


As discussed in slack and in https://github.com/DACH-NY/canton/issues/29834#issuecomment-4030515946 ideally we would just always be able to use the force flag but that doesn't work well with the way Canton currently handles vetting. I'm hoping we can fix this in PV 3.5 so I also wouldn't want to invest into workarounds until then (the best one would probably to query `PackageService.GetPackageReferences`).

What do you think of this:

The problem applies only to validators. Super validators should not have third-party packages vetted (we probably want to double check this on dev/test/mainnet). For security we only need unvetting of SVs as that is enough to block use of the old version for anything meaningful.

So how about we change this PR to only unvet for SVs for now but not validators. You can probably use the `svValidator` flag in the validator config to distinguish. That way we can get the security benefits of unvetting now and then hopefully get the real fix for unvetting around dependencies in Canton PV 3.5 and then also enable it for validators.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PackageVettingTrigger.scala
julientinguely-da: Not sure about that. Adding packages via the trigger uses the vote requests `effective from` logic. My idea was to separate these so that unsupported packages get removed directly - same for the following issues where we want to update packages from the config map that I will add instantly.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminDarsConnection.scala
julientinguely-da: I think it makes sense. I will make the change. Let me know if I can help in the upstream issue somehow.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BootstrapPackageConfigDarUploadIntegrationTest.scala
```diff
@@ -148,15 +148,32 @@ class BootstrapPackageConfigDarUploadIntegrationTest
...
     ) {
-      val dars =
-        uploadedDars.filter { case (name, _) =>
-          name == packageResource.latest.metadata.name
-        }
       dars should not be empty withClue s"dars for ${packageResource.latest.metadata.name}"
       dars.map(_._2).max shouldBe PackageVersion.assertFromString(requiredVersion)
     }
+    clue(
+      s"versions for package ${packageResource.latest.metadata.name} should strictly contain the required ones"
```
moritzkiefer-da, line 201: Can we also check that the serial is only 2 and that each serial only added packages? In particular I want to test that we don't end up with something silly like this (for new validators):

1. validator app first vets all versions
2. then we unvet some of them again

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnsupportedPackageVettingIntegrationTest.scala
```diff
@@ -0,0 +1,96 @@
...
+        DarResources.dsoGovernance_0_1_0,
+        DarResources.walletPayments_0_1_0,
+        DarResources.amuletNameService_0_1_0,
+        DarResources.amulet_0_1_0,
+        DarResources.splitwell_0_1_0,
```
moritzkiefer-da: I think for SVs we don't want to upload splitwell. What we should do is check on CILR, dev, test mainnet that the SVs do not have anything weird vetted atm. Can you do that please?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnsupportedPackageVettingIntegrationTest.scala
```diff
@@ -0,0 +1,96 @@
...
+            unsupportedDars.map(UploadablePackage.fromResource),
+            RetryFor.Automation,
+          )
+          .futureValue
+        participantAdminConnection
+          .vetDars(synchronizerId, unsupportedDars, None, None)
+          .futureValue
+      },
+    )(
+      s"the unsupported packages are vetted on $name",
```
moritzkiefer-da, line 76: I don't understand what you're testing here. For SVs you set unsupportedDars = removedDars

Then you first check that they are vetted here, why is "unsupported" vetted? And then you check that the unsupported packages are removed based on the clue but the actual comment checks that removedDars is gone not unsupportedDars.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PackageVettingTrigger.scala
moritzkiefer-da: fair, it's rare enough that we need to unvet to not need to worry about this I think.

docs/src/release_notes.rst
```diff
@@ -18,6 +18,24 @@
 
 .. _release_notes:
 
+.. release-notes:: 0.5.15
+
+  - SV and Validator app
```
```suggestion
  - SV
```

docs/src/release_notes.rst
```diff
@@ -18,6 +18,24 @@
...
+
+    - Going forward unsupported splice DARs will no longer be uploaded or vetted.
```
```suggestion
    - Going forward unusable splice DARs will be automatically unvetted by the super validators. This will be used for DARs that can already not be used, e.g., because a downgrade of AmuletRules to that version is not possible so it does not force more aggressive upgrades for validators or app devs.
```

docs/src/release_notes.rst
```diff
@@ -18,6 +18,24 @@
...
+      The minimum supported versions are:
```
moritzkiefer-da: Let's use the versions from https://github.com/hyperledger-labs/splice/pull/3529

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PackageVettingTrigger.scala
```diff
@@ -16,6 +17,7 @@ abstract class PackageVettingTrigger(
     packages: Set[PackageIdResolver.Package],
     maxVettingDelay: NonNegativeFiniteDuration,
     latestPackagesOnly: Boolean,
+    svValidator: Boolean,
```
moritzkiefer-da: Let's give that a more descriptive name like "enableUnvetting" here in the generic class. svValidator is fine as a name in the validator specific code

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminDarsConnection.scala
```diff
@@ -137,31 +137,100 @@ trait ParticipantAdminDarsConnection {
...
+
+  private def removeDarsFromVettingState(
+      dars: Seq[DarResource],
+      currentVetting: VettedPackages,
+  ): VettedPackages = {
+    val packageIdsToRemove: Seq[String] = dars.map(_.packageId)
+    // filter out all dependencies that may exist with the resulting packages before removing a package
+    // See https://github.com/DACH-NY/canton/issues/29834: make it work for validators
+    val packageIdsToKeep = DarResourcesUtil
+      .getDarResources(currentVetting.packages.map(_.packageId))
```
moritzkiefer-da: Now that we only enable it for SVs do we need any of that logic here? They shouldn't have vetted third-party packages so I would have expected you can just remove it without looking at this.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/DarResourcesUtil.scala
```diff
@@ -0,0 +1,72 @@
...
+        .getOrElse(
+          throw new NoSuchElementException(s"Could not find PackageResource ${pkg.metadata.name}.")
+        )
+        .minimumInitialization
+        .metadata
+        .version > pkg.metadata.version
+    )
+  }
+
+  private def supportedPackageVersions: Seq[DarResource] =
```
moritzkiefer-da: any reason this is not a `val`?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnsupportedPackageVettingIntegrationTest.scala
julientinguely-da: By calling directly `vetDars()` I bypass the `vetPackages` logic that make the version checks, which is good, that way I can test that `PackageVettingTrigger` unvets unsupported packages (instead of hacking a way to bootstrap a node with unsupported packages).

Now, I added `removedDars` as parameter to test that sv validators unvet the unsupported packages but not the validators. Does it make sense?

moritzkiefer-da: Ok I think I see what you're after, maybe we can make it clearer with some renaming:

Call `unsupportedDars` `unsupportedDarsToVet` and
`removedDars` `darsUnvettedByAutomation`

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminDarsConnection.scala
julientinguely-da: I would keep it in case; it doesn't cost much?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BootstrapPackageConfigDarUploadIntegrationTest.scala
julientinguely-da: Not sure it's expected. In the test for sv1 I'm getting the following Vetted Packages transaction, which makes serial 3 the latest, and that when running from main or my branch:
```
2026-03-16T12:55:54.024Z [⋮] DEBUG - o.l.s.a.a.c.ApiClientRequestLogger:BootstrapPackageConfigDarUploadIntegrationTest/config=355e9b4c/SV=sv1 (cb0efe0c79cf8ee1c0a5da0323d17e6b-com.digitalasset.canton.topology.admin.v30.TopologyManagerReadService/ListVettedPackages--a4841bd87a907dfe) - Request (no caller tid) com.digitalasset.canton.topology.admin.v30.TopologyManagerReadService/ListVettedPackages to 127.0.0.1:27102: sending request ListVettedPackagesRequest(
  BaseQuery(
    StoreId(Synchronizer(Synchronizer(Id(global-domain::1220aa3a68850129f5f35304ff535562252c805839a835201f0d378ec760aa8b2362)))),
    false,
    TOPOLOGY_CHANGE_OP_UNSPECIFIED,
    HeadState(Empty(UnknownFieldSet(Map()))),
    "",
    None
  ),
```

moritzkiefer-da: Ah right, it's:

1. Canton initializes the node and vets the initial package.
2. SV app vets dars.
3. Validator app vets dars.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminDarsConnection.scala
moritzkiefer-da: Less code is always better. Easier to understand, less things that can be buggy. So strongly lean towards removing it.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PackageVettingTrigger.scala
```diff
@@ -81,10 +81,12 @@ abstract class PackageVettingTrigger(
   private def runIfInputChanged(
       input: Seq[String]
   )(run: => Future[Unit])(implicit tc: TraceContext) = {
-    val previoslyRunInput = previouslyRunInputRef.get()
-    if (previoslyRunInput != input.toSet) {
```
julientinguely-da, line 64: previouslyRunInput is always empty at start making the trigger run

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BootstrapPackageConfigDarUploadIntegrationTest.scala
julientinguely-da: all good then it works as exectted :+1:

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminDarsConnection.scala
julientinguely-da: alright, removing this logic

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/BootstrapPackageConfigDarUploadIntegrationTest.scala
```diff
@@ -141,22 +144,50 @@ class BootstrapPackageConfigDarUploadIntegrationTest
...
     }
+    clue(s"Vetted package transactions have the expected serial $serial") {
+      participantAdminConnection
+        .listVettedPackages(
+          participantAdminConnection.getParticipantId().futureValue,
+          domainId,
+          AuthorizedState,
+        )
+        .futureValue
+        .map(_.base.serial.value) should contain only serial
```
moritzkiefer-da:
```suggestion
        .loneElement.base.serial.value shouldBe serial
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnsupportedPackageVettingIntegrationTest.scala
moritzkiefer-da: That checks our own participant and it checks uploaded packages not vetted. We should query the actual vetted packages for all SVs

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnsupportedPackageVettingIntegrationTest.scala
```diff
@@ -0,0 +1,96 @@
...
+      .simpleTopology1Sv(this.getClass.getSimpleName)
+
+  "Unsupported vetted packages are automatically removed by the package vetting trigger for SV and validator" in {
+    implicit env =>
+      val unsupportedDarsToVet = Seq(
+        DarResources.dsoGovernance_0_1_0,
+        DarResources.walletPayments_0_1_0,
+        DarResources.amuletNameService_0_1_0,
+        DarResources.amulet_0_1_0,
+        DarResources.splitwell_0_1_0,
```
moritzkiefer-da: why are we still uploading splitwell for SVs?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PackageVettingTrigger.scala
```diff
@@ -54,16 +57,36 @@ abstract class PackageVettingTrigger(
...
+      }
     } yield false
   }
 
   private def runIfInputChanged(
       input: Seq[String]
   )(run: => Future[Unit])(implicit tc: TraceContext) = {
-    val previoslyRunInput = previouslyRunInputRef.get()
-    if (previoslyRunInput != input.toSet) {
+    val previouslyRunInput = previouslyRunInputRef.get()
```
moritzkiefer-da, line 88: i don't understand this change, isn't running once on startup exactly what we want?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminDarsConnection.scala
```diff
@@ -137,31 +137,92 @@ trait ParticipantAdminDarsConnection {
...
   }
 
+  def unvetDars(
+      synchronizerId: SynchronizerId,
+      dars: Seq[DarResource],
+      maxVettingDelay: Option[(Clock, NonNegativeFiniteDuration)],
+  )(implicit
+      tc: TraceContext
+  ): Future[Unit] = {
+    retryProvider.retry(
```
moritzkiefer-da: I don't understand the point of the outer retry.  ensureTopologyMapping already takes care of any retries no?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PackageVettingTrigger.scala
julientinguely-da: Ah yes you're right

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminDarsConnection.scala
julientinguely-da: I followed the same pattern we had with vetDars. Let me change that

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnsupportedPackageVettingIntegrationTest.scala
julientinguely-da: removing now

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/DarResourcesUtil.scala
```diff
@@ -29,16 +29,22 @@ object DarResourcesUtil {
   def getRequiredPackageVersions(
       name: PackageName,
       upToRequiredVersion: PackageVersion,
+      enableUnsupportedDarsUnvetting: Boolean,
```
julientinguely-da, line 32: fyi @moritzkiefer-da added feature flag here

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnsupportedPackageVettingIntegrationTest.scala
julientinguely-da: Ok, doing the pattern as discussed:
```
val sdaf = svs.map(e=>ParticipantId.tryFromProtoPrimitive(s"PAR::$e")) 

val participantId = participant.id 

val allSvsPackages = participant.topology.vetted_packages.list().filter(e => sdaf.contains(e.item.participantId)).flatMap(_.item.packages.map(_.packageId)).distinct 

val sv1Packages = participant.topology.vetted_packages.list().filter(e =>e.item.participantId==participantId).flatMap(_.item.packages.map(_.packageId)).distinct 

allSvsPackages.diff(sv1Packages)
```

julientinguely-da: so no weird package on the way for all prod cluster and cilr

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/UnsupportedPackageVettingIntegrationTest.scala
```diff
@@ -0,0 +1,94 @@
...
+          .futureValue
+        participantAdminConnection
+          .vetDars(synchronizerId, unsupportedDarsToVet, None, None)
+          .futureValue
+      },
+    )(
+      s"the unsupported packages are vetted on $name",
+      _ =>
+        participantAdminConnection
+          .listVettedPackages(participantId, synchronizerId, AuthorizedState)
```
martinflorian-da, line 79: @julientinguely-da Is there some reason I'm missing why this is not inherently flaky?

AFAICT you are relying on timing here which sounds quite risky. Probably you want to pause automation and then resume it again for the second check?

julientinguely-da: You're probably right, stopping the trigger might be better here.

I'm working on similar testing for my followup PR. Don't bother, I'll have a deeper look anyway afterwards.

PR #4283, opened 2026-03-05
fix issue with root notification  channel specifying matchers

nicu-da:
[static]

tested by fixing cimain by removing the field and also hacking the mock cluster to generate the expected file and everything looks good 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/infra/src/observability.ts
+++ b/cluster/pulumi/infra/src/observability.ts
@@ -950,7 +950,7 @@ function grafanaAlertNotificationPolicies() {
   const notificationPolicies = [];
   const defaultPolicy = yaml.load(
     readGrafanaAlertingFile('notification_policies/default_slack.yaml')
-  );
+  ) as Record<string, unknown>;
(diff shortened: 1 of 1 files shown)


cluster/pulumi/infra/src/observability.ts
```diff
@@ -975,6 +975,8 @@ function grafanaAlertNotificationPolicies() {
       ],
     });
   } else {
+    // we can't have a single policy with matchers, the root policy must be the default if nothing else matches
+    delete defaultPolicy.object_matchers;
```
moritzkiefer-da: can we not set it in the first place instead of first setting it and then deleting it here?

nicu-da: d'oh we actually ignore it for the other branch indeed. let me fix this PR and will let the dumb version in 0.5.14

nicu-da: actually we can't do that as we use the same config when we have mutliple policies.

PR #4304, opened 2026-03-06
Apply vetting split to expired amulet triggers

moritzkiefer-da:
part of #4251

[ci]

I can't quite convince myself that setting up integration tests for this only to then have some log assertions or something like that is worth it but happy to be convinced it is.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/BatchedMultiDomainExpiredContractTrigger.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/BatchedMultiDomainExpiredContractTrigger.scala
@@ -58,14 +58,14 @@ abstract class BatchedMultiDomainExpiredContractTrigger[
         PackageIdResolver.Package.SpliceAmulet,
         c => stakeholders(c.payload),
         expiredContracts,
+        batchSize,
       )
       .map {
         _.toSeq.flatMap {
-          case (Some(version), contracts) =>
-            contracts.grouped(batchSize).map(Batch(pkg, version, _))
+          case (Some(version), contractBatches) => contractBatches.map(Batch(pkg, version, _))
           case (None, contracts) =>
             logger.warn(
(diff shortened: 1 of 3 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/BatchedMultiDomainExpiredContractTrigger.scala
```diff
@@ -41,7 +48,28 @@ abstract class BatchedMultiDomainExpiredContractTrigger[
...
+      .map {
+        _.toSeq.flatMap {
+          case (Some(version), contracts) =>
+            contracts.grouped(batchSize).map(Batch(pkg, version, _))
+          case (None, contracts) =>
+            logger.warn(
+              show"No vetted $pkg version for ${contracts.map { _.contractId.contractId }}"
+            )
+            Seq.empty
+        }
```
meiersi-da, line 71: Seems like code that might well live in `vettingLookupService.splitBatch`. Any reason to not just pass in the batchSize and have it do the whole splitting?

meiersi-da (commented):
> I can't quite convince myself that setting up integration tests for this only to then have some log assertions or something like that is worth it but happy to be convinced it is.

I guess there are two kinds of bugs we'd like to avoid:
1. splitting into too many batches
2. not splitting among versions

The former seems fine to discover in prod and then fix later; the latter is tested by the test you added in the PR introducing the infrastructure. Furthermore, if you can code share the `batchSize`-batch splitting the former test also covers the new triggers you changed; and it seems fine to not add a test.

So yeah, seems OK to take the risk.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/BatchedMultiDomainExpiredContractTrigger.scala
moritzkiefer-da: moved

PR #4309, opened 2026-03-06
replace implicit base config with explicit !include in splice

mblaze-da:
Fixes https://github.com/DACH-NY/canton-network-internal/issues/3973

These changes are the first step in simplifying the configuration loading in Pulumi. The aim is to remove imlicit loading of `cluster/deployment/config.yaml` and `../config.yaml`, replacing it with explicit `!include`. This will also make the merging logic consistent with the merging semantics of `!include` and allow us to use `!append` on lists included from the `cluster/configs/shared/base.yaml` config.


cluster/configs/shared/scratchnet.yaml
```diff
@@ -1,3 +1,4 @@
+!include(base.yaml)
```
martinflorian-da, line 1: So this is effectively a NOP during the transition, correct?

Did you verify this works as expected? No YAML errors because of duplicate keys and the like?

mblaze-da: Yes, during transition it does nothing. The static checks did the verification for me.

mblaze-da: Actually, there is one thing it does. We can already start using `!append` to append to lists from `base.yaml` which wasn't possible with the implicit merge.

martinflorian-da: > The static checks did the verification for me.

Note that we don't have expected files for scratchnets... But I guess we'll see soon enough if it doesn't work. Maybe FYI @rautenrieth-da @jagathweerasinghe-da so they know what to revert :stuck_out_tongue:

mblaze-da: But we have resolved config files. They remain unchanged and their update suceeded so I'm optimistic.

PR #4317, opened 2026-03-06
[ci] Prevent javaValue deserialization when updates are requested in protobuf-json daml value encoding.

ray-roestenburg-da:
Related to #4308 
Removed the RawJson type since it is confusing in the openAPI spec, and doesn't improve performance, circe marshalling is required in how guardrails + pekko + circe is used.

The protobuf-json damlValueEncoding side of the code paths gets more complicated, with raw data passed along everywhere, and dummy records in the events that can get confusing, so not sure if this PR is very useful right now, especially since we don't know yet how many people opt in for the protobuf-json format which is harder to parse.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
@@ -7,6 +7,7 @@ import cats.data.{NonEmptyList, OptionT}
 import cats.syntax.semigroup.*
 import com.daml.ledger.api.v2.TraceContextOuterClass
 import com.daml.ledger.javaapi.data.codegen.{ContractId, DamlRecord}
+import com.daml.ledger.javaapi.{data as javaApi}
 import com.daml.ledger.javaapi.data.{CreatedEvent, Event, ExercisedEvent, Identifier, Transaction}
 import com.daml.metrics.api.MetricsContext
 import com.google.protobuf.ByteString
@@ -1090,6 +1091,7 @@ class UpdateHistory(
       filters: NonEmptyList[SQLActionBuilder],
       orderBy: SQLActionBuilder,
       limit: Limit,
+      valueDeserializer: String => Option[javaApi.Value] = UpdateHistory.deserializeValue,
   )(implicit tc: TraceContext): Future[Seq[TreeUpdateWithMigrationId]] = {
     def makeSubQuery(afterFilter: SQLActionBuilder): SQLActionBuilderChain = {
       sql"""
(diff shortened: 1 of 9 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/plugins/EventHistorySanityCheckPlugin.scala
```diff
@@ -79,7 +79,7 @@ class EventHistorySanityCheckPlugin(
                 case TreeEvent.members.ExercisedEvent(exercised)
                     if exercised.choice == "DsoRules_AddSv" =>
                   DsoRules_AddSv
-                    .fromJson(exercised.choiceArgument.noSpaces)
+                    .fromJson(exercised.choiceArgument.value)
```
ray-roestenburg-da: This is now a `RawJson`

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1121,13 +1123,16 @@ class UpdateHistory(
...
+        )
         TreeUpdateWithMigrationId(
-          decodeTransaction(
-            row,
-            creates.getOrElse(row.rowId, Seq.empty),
-            exercises.getOrElse(row.rowId, Seq.empty),
-          ),
+          response,
           row.migrationId,
+          rawProtobufJsonEventData = rawData,
```
ray-roestenburg-da, line 1135: Keeping rawdata around so it can be passed directly to circe encoder, without using value deserializer.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1602,20 +1634,28 @@ class UpdateHistory(
         ).toJava,
         /*contractId = */ row.contractId,
         /*choice = */ row.choice,
-        /*choiceArgument = */ ProtobufCodec.deserializeValue(row.argument),
+        /*choiceArgument = */ valueDeserializer(row.argument).getOrElse(UpdateHistory.emptyRecord),
```
ray-roestenburg-da, line 1637: this is not great (emptyRecord) because some consumers of some methods, like getUpdatesBefore, that does not use the encoding step to http, will return update history response with these placeholders.

moritzkiefer-da:
@ray-roestenburg-da definitely makes sense to not worry about the protobuf-json encoding, I think noone other than backfilling uses that.

Can you summarize why we can't optimize away some of the intermediate conversions for the json output?

moritzkiefer-da:
or consider using transcode. Not actually sure which library that uses atm but I'm hoping it's circe.

PR #4322, opened 2026-03-09
Add TcsStore and ScanRewardsReferenceStore 

dfordivam:
Fixes: https://github.com/hyperledger-labs/splice/issues/4118

This was partially reviewed in https://github.com/hyperledger-labs/splice/pull/4126

But due to deletion of its base I had to open a new PR.
Note this does not implement caching, only the DB based queries

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
```diff
@@ -0,0 +1,64 @@
...
+      limit: Limit,
+  )(implicit
+      companionClass: ContractCompanion[C, TCid, T],
+      traceContext: TraceContext,
+  ): Future[Seq[ContractWithState[TCid, T]]]
+
+  def listContractsInAsOfRange[C, TCid <: ContractId[?], T](
+      companion: C,
+      minAsOf: CantonTimestamp,
+      maxAsOf: CantonTimestamp,
```
meiersi-da:
```suggestion
      lowerBoundIncl: CantonTimestamp,
      upperBoundIncl: CantonTimestamp,
```

make inequalities obvious...

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
```diff
@@ -0,0 +1,64 @@
...
+      companion: C,
+      asOf: CantonTimestamp,
+      synchronizerId: SynchronizerId,
+      limit: Limit,
+  )(implicit
+      companionClass: ContractCompanion[C, TCid, T],
+      traceContext: TraceContext,
+  ): Future[Seq[ContractWithState[TCid, T]]]
+
+  def listContractsInAsOfRange[C, TCid <: ContractId[?], T](
```
meiersi-da:
```suggestion
  def listContractsActiveWithin[C, TCid <: ContractId[?], T](
```

meiersi-da: clarify in comment that these are all contracts whose activeness interval intersects with the given range

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  /** A contract with its temporal bounds from the TCS range query. */
+  case class TemporalContractWithState[TCid, T](
+      contractWithState: ContractWithState[TCid, T],
+      createdAt: CantonTimestamp,
+      archivedAt: Option[CantonTimestamp],
+  )
+
+  /** Pure function: filter contracts alive at a specific timestamp. */
+  def contractsAsOf[TCid, T](
```
meiersi-da:
```suggestion
  def contractsActiveAsOf[TCid, T](
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
```diff
@@ -0,0 +1,64 @@
...
+import com.digitalasset.canton.data.CantonTimestamp
+import com.digitalasset.canton.topology.SynchronizerId
+import com.digitalasset.canton.tracing.TraceContext
+import org.lfdecentralizedtrust.splice.store.MultiDomainAcsStore.ContractCompanion
+import org.lfdecentralizedtrust.splice.util.ContractWithState
+
+import scala.concurrent.Future
+
+trait TcsStore {
+
```
meiersi-da, line 16: why no function to wait until a record time was ingested?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
```diff
@@ -0,0 +1,133 @@
...
+import org.lfdecentralizedtrust.splice.scan.store.{ScanStore, ScanRewardsReferenceStore}
+import org.lfdecentralizedtrust.splice.store.{Limit, TcsStore}
+import org.lfdecentralizedtrust.splice.store.MultiDomainAcsStore.ContractCompanion
+import org.lfdecentralizedtrust.splice.store.db.{DbAppStore, DbMultiDomainAcsStore, StoreDescriptor}
+import org.lfdecentralizedtrust.splice.util.{ContractWithState, TemplateJsonDecoder}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class DbScanRewardsReferenceStore(
+    override val key: ScanStore.Key,
```
meiersi-da: Is the sharing of the key type intentional?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
```diff
@@ -0,0 +1,133 @@
...
+      ingestionConfig = ingestionConfig,
+      acsArchiveConfigOpt = Some(
+        DbMultiDomainAcsStore.AcsArchiveConfig.withIndexColumns(
+          ScanRewardsReferenceTables.archiveTableName,
+          ScanRewardsReferenceTables.ScanRewardsReferenceStoreRowData.hasIndexColumns.indexColumnNames,
+        )
+      ),
+    )
+    with ScanRewardsReferenceStore
+    with TcsStore {
```
meiersi-da:
```suggestion
    {
```

consider whether we do want to expose the generic methods. I could see that we are better off keeping an explicit control over the specific reads we expose to ensure we fully understand their usage and read access patterns.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
```diff
@@ -0,0 +1,62 @@
...
+
+import com.digitalasset.daml.lf.data.Time.Timestamp
+import com.digitalasset.canton.topology.PartyId
+import org.lfdecentralizedtrust.splice.codegen.java.splice
+import org.lfdecentralizedtrust.splice.scan.store.db.ScanRewardsReferenceTables.ScanRewardsReferenceStoreRowData
+import org.lfdecentralizedtrust.splice.store.{AppStore, MultiDomainAcsStore}
+import org.lfdecentralizedtrust.splice.store.db.AcsInterfaceViewRowData
+
+/** This is a temporal contract store (TcsStore) to provide efficient asOf round
+  * lookups of FeaturedAppRight, OpenMiningRound, and AmuletRules contracts
```
meiersi-da:
```suggestion
  * lookups of FeaturedAppRight and OpenMiningRound contracts
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -186,6 +216,119 @@ final class DbMultiDomainAcsStore[TXE](
...
+      id: ContractId[?],
+      asOf: CantonTimestamp,
+      synchronizerId: SynchronizerId,
+  )(implicit
+      companionClass: ContractCompanion[C, TCid, T],
+      traceContext: TraceContext,
+  ): Future[Option[ContractWithState[TCid, T]]] = {
+    val archiveConfig = requireArchiveConfig("lookupContractByIdAsOf")
+    waitUntilAcsIngested {
+      waitUntilRecordTimeReached(synchronizerId, asOf).flatMap { _ =>
```
meiersi-da: Why do we need to wait for the acs to be ingested? Isn't that implied by reaching a record time?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -186,6 +216,119 @@ final class DbMultiDomainAcsStore[TXE](
...
+        for {
+          result <- storage.query(
+            selectFromTcsTableWithStateAsOf(
+              acsTableName,
+              archiveConfig.archiveTableName,
+              acsStoreId,
+              domainMigrationId,
+              companion,
+              asOf,
+              orderLimit = sql"""order by event_number limit ${sqlLimit(limit)}""",
```
meiersi-da: We need to check the query plan for this. I suspect that we'll want to order by `(archived_at, event_number)` for the archived ones, and `(created_at, event_number)` for non-archived ones. This also hints at wanting to include the `event_number` in their indices.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/TcsQueries.scala
```diff
@@ -0,0 +1,170 @@
...
+       from #$archiveTableName acs
+       where acs.store_id = $storeId
+         and acs.migration_id = $migrationId
+         and acs.package_name = ${packageQualifiedName.packageName}
+         and acs.template_id_qualified_name = ${packageQualifiedName.qualifiedName}
+         and acs.created_at <= $asOf
+         and acs.archived_at > $asOf
+         """ ++ additionalWhere ++ sql"""
+       )
+       """ ++ orderLimit).toActionBuilder
```
meiersi-da: as pointed out above. I suspect we want different orders for the two segments; and we want to orders to align with the main index used to retrieve the results.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/TcsQueries.scala
```diff
@@ -0,0 +1,170 @@
...
+       select #$columns, acs.archived_at
+       from #$archiveTableName acs
+       where acs.store_id = $storeId
+         and acs.migration_id = $migrationId
+         and acs.package_name = ${packageQualifiedName.packageName}
+         and acs.template_id_qualified_name = ${packageQualifiedName.qualifiedName}
+         and acs.created_at <= $maxAsOf
+         and acs.archived_at > $minAsOf
+       )
+       """ ++ orderLimit).toActionBuilder.as[TcsQueries.SelectFromTcsTableRangeResult]
```
meiersi-da: same concern wrt table orders applies

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
```diff
@@ -0,0 +1,62 @@
...
+    import MultiDomainAcsStore.mkFilter
+    val dso = key.dsoParty.toProtoPrimitive
+
+    MultiDomainAcsStore.SimpleContractFilter(
+      key.dsoParty,
+      Map(
+        mkFilter(splice.round.OpenMiningRound.COMPANION)(co => co.payload.dso == dso) { contract =>
+          ScanRewardsReferenceStoreRowData(
+            contract = contract,
+            contractExpiresAt = Some(Timestamp.assertFromInstant(contract.payload.targetClosesAt)),
```
meiersi-da: Why do we need expiry in this store? Note that the round selection is driven off of the actual `archivedAt` and not the target. Otherwise, extended downtimes of the SV round automation can lead to activity without an associated round. We prefer to over-subscribe a round and thus smoothen the loss.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__scan_rewards_reference_store.sql
```diff
@@ -0,0 +1,52 @@
...
+    -- reestablish foreign key constraint as that one is not copied by the LIKE statement above
+    foreign key (store_id) references store_descriptors(id),
+
+    -- record_time of the transaction that archived the contract, in micros since epoch.
+    archived_at bigint not null
+);
+
+create index acs_store_archived_test_temporal
+    on acs_store_archived_test (store_id, migration_id, template_id_qualified_name, archived_at) include (created_at);
+
```
meiersi-da:
```suggestion

-- Store for reference data for the off-ledger reward computations.
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
dfordivam: This was merely getting ingested because we have this coloumn in our table, like the Scan Acs Store, but we are not using it. I guess not ingesting it would save on db space, so I will remove it.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: Not necessary, I will remove

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
dfordivam: Done

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
dfordivam: Changed API and added comment

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
dfordivam: The lookup APIs wait internally, so do we see a need for a separate API to wait?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
dfordivam: I have removed this

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
dfordivam: Not really, I looked around and did not see similar pattern, so I added separate Key type.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
dfordivam: Ok, removed these, we could expose only APIs we have tested

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
dfordivam: Fixed

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: Removed.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: I have added `event_number` to indices. And added pagination support in lookup APIs.

Also did an investigation on the query plan, and will share concrete results soon.

But I would like to better understand why we would want to order by "(archived_at, event_number) for the archived ones, and (created_at, event_number) for non-archived ones." Why don't we simply use the `event_number` ordering for these lookups too?
cc @rautenrieth-da

meiersi-da: This comes down to the question of what query exeuction plans postgres has available to do a LIMIT on an ordered result. AFAIK, there is only one option that works well with large result sets: deliver the results directly from a premade index.

See these considerations wrt paging for intuition on the problem: https://use-the-index-luke.com/sql/partial-results/fetch-next-page

The alternative that we might have available is to only specify a LIMIT and not use any order. The result is non-deterministic, but that can be OK. That strategy allows more free index usage for the query execution plans.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
meiersi-da: ah I see. no all good then.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: yeah, (archived_at, event_number) for the archived ones, and (created_at, event_number) for non-archived ones, gives much better results. Similar in perf to no order.

I also discovered an issue in index creation, `like scan_rewards_reference_store_active including all,` also creates the `created_at` index in `scan_rewards_reference_store_archived` which messed up its index selection.

Now I am considering the design of pagination wrt the cursor

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__scan_rewards_reference_store.sql
```diff
@@ -0,0 +1,54 @@
+-- Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+create table acs_store_archived_test(
```
rautenrieth-da: What's the purpose of this table? I only see it being used in the test. It feels strange to have a table that is used in a unit test but not in production.

FWIW, our `acs_store_template` table is used both for unit tests and as a template for other tables.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -2210,6 +2426,8 @@ object DbMultiDomainAcsStore {
       acsSize: Int,
       offsetChanged: Promise[Unit],
       offsetIngestionsToSignal: SortedMap[Long, Promise[Unit]],
+      lastIngestedRecordTimes: Map[SynchronizerId, CantonTimestamp],
```
rautenrieth-da, line 2278: Please update the scaladoc of this class with the new fields.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -1849,26 +2008,52 @@ final class DbMultiDomainAcsStore[TXE](
...
-          ) ++ sql" returning contract_id").toActionBuilder.as[String].map { deletedCids =>
+        DBIO.sequence(deletes.grouped(ingestionConfig.maxDeletesPerStatement).map { deletes =>
+          val performDeleteSql = acsArchiveConfigOpt match {
+            case Some(DbMultiDomainAcsStore.AcsArchiveConfig(archiveTableName, baseColumns)) =>
+              val valuesPairs = deletes.map { d =>
+                val cid = lengthLimited(d.evt.getContractId)
+                val archivedAt = CantonTimestamp.assertFromInstant(d.recordTime).toMicros
+                sql"($cid, $archivedAt)"
+              }
+              val valuesClause = sqlCommaSeparated(valuesPairs)
```
rautenrieth-da, line 1900: This looks like we could use array parameters instead of comma separated values, similar to how we do in `inClause`. I would leave it like this for now though and worry about performance finetuning later.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbTcsStoreTest.scala
```diff
@@ -0,0 +1,615 @@
...
+                ),
+              )
+            )
+          )
+        )
+        _ <- waitD2Future
+      } yield succeed
+    }
+
+    "waitUntilRecordTimeReached complete when record time is reached via transaction ingestion" in {
```
rautenrieth-da, line 290: Please also add a test for the edge case where we `waitUntilRecordTimeReached` on a synchronizer that has not ingested any contracts yet.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
```diff
@@ -0,0 +1,69 @@
...
+  def lookupContractByIdAsOf[C, TCid <: ContractId[?], T](
+      companion: C
+  )(id: ContractId[?], asOf: CantonTimestamp, synchronizerId: SynchronizerId)(implicit
+      companionClass: ContractCompanion[C, TCid, T],
+      traceContext: TraceContext,
+  ): Future[Option[ContractWithState[TCid, T]]]
+
+  def listContractsAsOf[C, TCid <: ContractId[?], T](
```
meiersi-da:
```suggestion
  /** List ALL contracts of the given template that are active as of the given record time. Use this with great caution as the result set is unbounded in size. */
  def listAllContractsAsOf[C, TCid <: ContractId[?], T](
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
```diff
@@ -0,0 +1,69 @@
...
+      companion: C,
+      asOf: CantonTimestamp,
+      synchronizerId: SynchronizerId,
+      after: Option[Long],
+      limit: Limit,
+      sortOrder: SortOrder,
```
meiersi-da:
```suggestion
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
```diff
@@ -0,0 +1,69 @@
...
+  )(implicit
+      companionClass: ContractCompanion[C, TCid, T],
+      traceContext: TraceContext,
+  ): Future[ResultsPage[ContractWithState[TCid, T]]]
+
+  /** Returns all contracts whose activeness interval intersects with [lowerBoundIncl, upperBoundIncl]. */
+  def listContractsActiveWithin[C, TCid <: ContractId[?], T](
```
meiersi-da:
```suggestion
  def listAllContractsActiveWithin[C, TCid <: ContractId[?], T](
```

meiersi-da: same comment wrt unbounded result set

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
```diff
@@ -0,0 +1,69 @@
...
+      companion: C,
+      lowerBoundIncl: CantonTimestamp,
+      upperBoundIncl: CantonTimestamp,
+      synchronizerId: SynchronizerId,
+      after: Option[Long],
+      limit: Limit,
+      sortOrder: SortOrder,
```
meiersi-da:
```suggestion
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
dfordivam: Fixed

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbTcsStoreTest.scala
dfordivam: Added test

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: Done

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__scan_rewards_reference_store.sql
dfordivam: Yes, this was only being used in test, so I have removed this from migration and instead added in test's setup to create it dynamically during `cleanDb`. There was also the option to reuse the `scan_rewards_reference_store_archived`, but that didn't feel right to use for tests in common code.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: Order and limit have been removed entirely as discussed yesterday.

dfordivam:
@rautenrieth-da , @meiersi-da I have addressed the review comments. PTAL.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
```diff
@@ -0,0 +1,105 @@
...
+      tc: TraceContext
+  ): Future[
+    Seq[TcsStore.TemporalContractWithState[FeaturedAppRight.ContractId, FeaturedAppRight]]
+  ] =
+    multiDomainAcsStore.listAllContractsActiveWithin(
+      FeaturedAppRight.COMPANION,
+      lowerBoundIncl,
+      upperBoundIncl,
+      synchronizerId,
+    )
```
meiersi-da: Where do you expect to need this function?

dfordivam: This could have been used for fetching for multiple rounds, but we won't need to do that in current approach, so I will remove it. Its trivial to add back if necessary.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -186,6 +216,107 @@ final class DbMultiDomainAcsStore[TXE](
       .value
   }
 
+  private def requireArchiveConfig(methodName: String): DbMultiDomainAcsStore.AcsArchiveConfig =
+    acsArchiveConfigOpt.getOrElse(
+      throw new IllegalStateException(
```
meiersi-da:
```suggestion
      throw new UnsupportedOperationException(
```

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V061__scan_rewards_reference_store.sql
```diff
@@ -0,0 +1,40 @@
+-- Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
meiersi-da, line 1: @OriolMunoz-da : review welcome

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -44,6 +44,7 @@ import com.digitalasset.canton.util.ShowUtil.showPretty
 
 import java.util.concurrent.atomic.AtomicReference
 import scala.collection.immutable.{Seq, SortedMap, VectorMap}
+import scala.jdk.CollectionConverters.*
```
meiersi-da, line 47: @OriolMunoz-da : review welcome

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/TcsQueries.scala
```diff
@@ -0,0 +1,164 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package org.lfdecentralizedtrust.splice.store.db
```
meiersi-da, line 4: @OriolMunoz-da : review welcome

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package org.lfdecentralizedtrust.splice.store
```
meiersi-da, line 4: @OriolMunoz-da : review welcome

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/TcsQueries.scala
```diff
@@ -0,0 +1,164 @@
...
+       )
+       UNION ALL
+       (
+       select #$columns
+       from #$archiveTableName acs
+       where acs.store_id = $storeId
+         and acs.migration_id = $migrationId
+         and acs.package_name = ${packageQualifiedName.packageName}
+         and acs.template_id_qualified_name = ${packageQualifiedName.qualifiedName}
+         and acs.created_at <= $asOf
```
OriolMunoz-da, line 67: if we're filtering by created_at, why is it not added to the index? it can still be added in last position to support what the comment says:
`-- Since we would be typically doing lookups for recently archived contracts, index on archived_at will help skip past events more efficiently`, but I'm probably missing something

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__scan_rewards_reference_store.sql
```diff
@@ -0,0 +1,40 @@
...
+);
+
+-- temporal query support: created_at filtering for point-in-time lookups on the live table
+create index scan_rewards_reference_store_active_temporal
+    on scan_rewards_reference_store_active (store_id, migration_id, template_id_qualified_name, created_at);
+
+-- temporal query support: created_at + archived_at filtering for point-in-time lookups on the archive table
+-- Since we would be typically doing lookups for recently archived contracts, index on archived_at will help skip past events more efficiently
+create index scan_rewards_reference_store_archived_temporal
+    on scan_rewards_reference_store_archived (store_id, migration_id, template_id_qualified_name, archived_at) include (created_at);
```
OriolMunoz-da: is the exclusion of package_name from the indexes deliberate? it's being used in the union all query

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__scan_rewards_reference_store.sql
```diff
@@ -0,0 +1,40 @@
+-- Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Store for reference data for the off-ledger reward computations.
+create table scan_rewards_reference_store_active
+(
+    like acs_store_template including all,
```
OriolMunoz-da: note that this includes all indexes defined in the template table:

Image

those are useful for our other stores, but I'm not sure that they are for you here, so you might want to not include them to avoid the performance impact of maintaining extra indexes

you might want also to validate that the primary keys make sense

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  def lookupContractByIdAsOf[C, TCid <: ContractId[?], T](
+      companion: C
+  )(id: ContractId[?], asOf: CantonTimestamp, synchronizerId: SynchronizerId)(implicit
+      companionClass: ContractCompanion[C, TCid, T],
+      traceContext: TraceContext,
+  ): Future[Option[ContractWithState[TCid, T]]]
+
+  /** List ALL contracts of the given template that are active as of the given
+    * record time. Use this with great caution as the result set is unbounded in size.
```
OriolMunoz-da, line 25: > Use this with great caution as the result set is unbounded in size.

We definitely cannot accept this, we've already had instances of returning too many rows.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
```diff
@@ -0,0 +1,69 @@
...
+      companion: C,
+      asOf: CantonTimestamp,
+      synchronizerId: SynchronizerId,
+  )(implicit
+      companionClass: ContractCompanion[C, TCid, T],
+      traceContext: TraceContext,
+  ): Future[Seq[ContractWithState[TCid, T]]]
+
+  /** Returns all contracts whose activeness interval intersects with
+    * [lowerBoundIncl, upperBoundIncl]. Use this with great caution as the
```
OriolMunoz-da, line 37: likewise

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  /** A contract with its temporal bounds from the TCS range query. */
+  case class TemporalContractWithState[TCid, T](
+      contractWithState: ContractWithState[TCid, T],
+      createdAt: CantonTimestamp,
+      archivedAt: Option[CantonTimestamp],
+  )
+
+  /** Pure function: filter contracts alive at a specific timestamp. */
+  def contractsActiveAsOf[TCid, T](
```
OriolMunoz-da, line 61: this is test-only code, let's move it there

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -2256,6 +2472,28 @@ object DbMultiDomainAcsStore {
...
     }
 
+    def signalRecordTimesReached(
+        recordTimes: Map[SynchronizerId, CantonTimestamp]
+    ): Unit = {
+      recordTimes.foreach { case (syncId, reachedTime) =>
+        recordTimeWaiters.get(syncId).foreach { waiters =>
+          waiters.foreach { case (waitTime, promise) =>
+            if (waitTime <= reachedTime) {
+              promise.success(())
```
OriolMunoz-da, line 2353: I can't spot where we remove the promises from the Map to avoid filling it up with sucessful promises, can you add a comment here? (or restructure the code so it happens here, if possible)

Also document why calling `success` and not `trySuccess` is okay - this would blow up if called twice

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -2233,15 +2431,33 @@ object DbMultiDomainAcsStore {
...
+          acc.updated(syncId, acc.get(syncId).fold(rt)(_.max(rt)))
+      }
+      val remainingWaiters = recordTimeWaiters.flatMap { case (syncId, waiters) =>
+        updatedRecordTimes.get(syncId) match {
+          case Some(reachedTime) =>
+            val remaining = waiters.filter { case (waitTime, _) => waitTime > reachedTime }
+            if (remaining.isEmpty) None else Some(syncId -> remaining)
+          case None => Some(syncId -> waiters)
+        }
+      }
```
OriolMunoz-da, line 2322: these two could use comments as to what exactly they're doing

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -2194,6 +2357,37 @@ final class DbMultiDomainAcsStore[TXE](
...
+    def withIndexColumns(
+        archiveTableName: String,
+        indexColumns: Seq[String],
+    ): AcsArchiveConfig =
+      AcsArchiveConfig(
+        archiveTableName,
+        if (indexColumns.isEmpty) defaultBaseColumns
+        else defaultBaseColumns + indexColumns.mkString(", ", ", ", ""),
+      )
+  }
```
OriolMunoz-da: let's move this into a separate file, as this one is already bloated enough as-is

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -186,6 +216,107 @@ final class DbMultiDomainAcsStore[TXE](
...
+  override def lookupContractByIdAsOf[C, TCid <: ContractId[?], T](companion: C)(
+      id: ContractId[?],
+      asOf: CantonTimestamp,
+      synchronizerId: SynchronizerId,
+  )(implicit
+      companionClass: ContractCompanion[C, TCid, T],
+      traceContext: TraceContext,
+  ): Future[Option[ContractWithState[TCid, T]]] = {
+    val archiveConfig = requireArchiveConfig("lookupContractByIdAsOf")
+    waitUntilRecordTimeReached(synchronizerId, asOf).flatMap { _ =>
```
OriolMunoz-da: design question: why do we wait until a record time is ingested instead of just returning what's available (maybe with some field indicating the last ingested record time)?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -2256,6 +2472,28 @@ object DbMultiDomainAcsStore {
...
+        }
+      }
+    }
+
+    def signalWaiters(
+        newOffset: Long,
+        recordTimes: Map[SynchronizerId, CantonTimestamp],
+    ): Unit = {
+      signalOffsetChanged(newOffset)
+      signalRecordTimesReached(recordTimes)
```
OriolMunoz-da, line 2365: note that this is called when an event is _ingested_, not when it is _seen_

Ingested: event's stuff got inserted/deleted into the store
Seen: event might have gotten ignored or ingested

See the two related metrics:
- metrics.updateLastSeenMetrics(batch.last)
- metrics.getLastIngestedRecordTimeMsForSynchronizer(synchronizer).updateValue(recordTime.toEpochMilli)

this means that if your store ignores most events, it will take a while for the signal for a record time to be emitted. I cannot tell if this matters to you, or if I'm missing something

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -96,14 +97,17 @@ final class DbMultiDomainAcsStore[TXE](
...
 )(implicit
     ec: ExecutionContext,
     templateJsonDecoder: TemplateJsonDecoder,
     closeContext: CloseContext,
 ) extends MultiDomainAcsStore
     with AcsTables
     with AcsQueries
+    with TcsQueries
     with TxLogQueries[TXE]
+    with TcsStore
```
OriolMunoz-da: > are you fine adding the TcsStore capability to the code that you and the other "store experts" are maintaining?

If at all possible, I would rather have the queries outside of this class. Ingestion is likely too much work / not worth it / not possible

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: This matches the existing pattern of changing the state first (ie `offsetIngestionsToSignal`), and then doing the actual signal (`signalOffsetChanged`). I am wary of doing the restructure as you suggest, so I will simply add a comment.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: It seems that the actual filtering for ingestion happens in `ingestTransactionTrees`, but this is being called at the point where `metrics.updateLastSeenMetrics(batch.last)` gets updated, which would include even the filtered out events. So we are in fact considering all seen events.
 Note this is called at the same place existing `offsetIngestionsToSignal` signalling was happening.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: We do need to wait for ingestion to complete, otherwise we will have to process the batch of verdicts in even smaller batches and it would complicate the ingestion. Waiting is a better/simpler approach.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: Ok, I have moved these out, it has made the code a bit clearer.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__scan_rewards_reference_store.sql
dfordivam: Thanks for pointing this out, I have fixed this to not include the indexes.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__scan_rewards_reference_store.sql
dfordivam: Nope, it was a mistake, Thanks for pointing out!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: I got rid of this, now the exception happens when the `DbTcsStore` is initialized, I guess using `IllegalStateException` at that point in fine?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: Ok, moved

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
dfordivam: Added comments.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/TcsQueries.scala
dfordivam: The reasoning behind this is that we would be mostly interested in contracts which got archived recently, and their count would be quite small, a few dozen at max. So it is fine to not have `created_at` in index.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
dfordivam: The reason we read all the rows in a single lookup is because we do need to fetch all the contracts matching the filter. and doing multiple lookups would unnecessarily complicate both the lookup and sql limit+order.
Now we do have an upper bound on how much we expect to read, which is a max of 100k rows, (specifically 100k FeaturedAppRight contracts).
So how about adding 100k as a hard limit in the SQL, and the API will throw exception if the number of rows hits this limit.
cc @meiersi-da

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
dfordivam: We plan to use it in app code in subsequent PR

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TcsStore.scala
meiersi-da: @dfordivam : yeah, that's a good idea. Let's make it a config parameter, so that it is easy to up the limit in production.

meiersi-da: Actually, after another chat with @OriolMunoz-da I managed to convince us both that the parameter is not the right option. The only thing that one can do when hitting the parameter is to bump the limit and continue. That sounds worse than reacting to an increase in memory usage, which also will not happen suddenly.

So let's scrap this idea.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/TcsQueries.scala
meiersi-da: and using only `INCLUDE (created_at)` means that the index does not need to store these values in internal pages. They will only be stored on leaf pages, which shrinks index size.

PR #4337, opened 2026-03-09
DO NOT MERGE: ts2 - split AmuletAllocation and AmuletAllocationV2 and polish V2

meiersi-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/token-standard/splice-token-standard-test-v2/daml/Splice/Tests/TestAmuletTokenDvP_V1V2Mixed.daml
+++ b/token-standard/splice-token-standard-test-v2/daml/Splice/Tests/TestAmuletTokenDvP_V1V2Mixed.daml
@@ -1,16 +1,12 @@
 -- Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 -- SPDX-License-Identifier: Apache-2.0
 
--- | Daml script tests showing that the token standard can be used to execute
--- DvP settlements of Amulet tokens; and how to do so.
+-- | Daml script tests showing how to batch settle allocations created using
+-- both the V1 and V2 allocation factories in a single settlement using the V2
+-- settlement factory.
 --
--- See this test and the 'Splice.Testing.TradingApp' module for an example of
--- how to integrate with the allocation APIs of the token standard to execute
--- DvP settlements.
---
--- Note also that the delivery part of a DvP settelement can be both another
--- token implementing the standard, as well as the creation of on-ledger state
--- specific to your registry; e.g., a license contract.
+-- Uses Amulet as the test token, but can be extened to use multiple tokens from
+-- different registries as well.
 module Splice.Tests.TestAmuletTokenDvP_V1V2Mixed where
 
 import Daml.Script
@@ -22,6 +18,7 @@ import DA.Optional (isSome)
 import DA.TextMap as TextMap hiding (filter)
 import DA.Map as Map hiding (filter)
 import DA.Time
+import qualified DA.Traversable
 
 import Splice.Api.Token.MetadataV1 as Api.Token.MetadataV1
 import Splice.Api.Token.HoldingV2
@@ -37,13 +34,14 @@ import Splice.Amulet.TokenApiUtils (burnedMetaKey)
 
 import Splice.Testing.Apps.TradingAppV2 hiding (require)
 import Splice.Testing.UtilsV2
-import Splice.Testing.Registries.AmuletRegistryV2 qualified as AmuletRegistry
-import Splice.Testing.TokenStandard.RegistryApiV2 qualified as RegistryApi
-import Splice.Testing.TokenStandard.WalletClientV2 qualified as WalletClient
+import Splice.Testing.Registries.AmuletRegistryV2 qualified as AmuletRegistryV2
+import Splice.Testing.TokenStandard.RegistryApiV2 qualified as RegistryApiV2
+import Splice.Testing.TokenStandard.WalletClientV2 qualified as WalletClientV2
 
 import Splice.Testing.Registries.AmuletRegistry qualified as AmuletRegistryV1
 import Splice.Testing.TokenStandard.RegistryApi qualified as RegistryApiV1
 import Splice.Testing.TokenStandard.WalletClient qualified as WalletClientV1
+
 
 data AllocatedOTCTrade = AllocatedOTCTrade
   with
@@ -53,7 +51,7 @@ data AllocatedOTCTrade = AllocatedOTCTrade
     providerBeneficiary1 : Party
     providerBeneficiary2 : Party
     providerBeneficiaries : [(Party, Decimal)]
-    registry : AmuletRegistry.AmuletRegistry
+    registry : AmuletRegistryV2.AmuletRegistry
     otcTradeCid : ContractId OTCTrade
     otcTrade : OTCTrade
     requestCids : [ContractId OTCTradeAllocationRequest]
@@ -63,7 +61,7 @@ data AllocatedOTCTrade = AllocatedOTCTrade
 
 setupOtcTrade : Script AllocatedOTCTrade
 setupOtcTrade  = do
-  registry <- AmuletRegistry.initialize AmuletRegistry.defaultAmuletRegistryConfig
+  registry <- AmuletRegistryV2.initialize AmuletRegistryV2.defaultAmuletRegistryConfig
 
   let amuletId = registry.instrumentId
   let registryV1 = AmuletRegistryV1.AmuletRegistry with
@@ -82,11 +80,11 @@ setupOtcTrade  = do
         , (providerBeneficiary1, 0.1)
         , (providerBeneficiary2, 0.1)
         ]
-  AmuletRegistry.featureApp registry provider
+  AmuletRegistryV2.featureApp registry provider
 
   -- get some funds for alice and bob
-  AmuletRegistry.tapLockedAndUnlockedFunds registry alice 1000.0
-  AmuletRegistry.tapLockedAndUnlockedFunds registry bob 1000.0
+  AmuletRegistryV2.tapLockedAndUnlockedFunds registry alice 1000.0
+  AmuletRegistryV2.tapLockedAndUnlockedFunds registry bob 1000.0
 
   let
     mkTransfer id sender receiver amount =
@@ -97,13 +95,21 @@ setupOtcTrade  = do
         amount
         instrumentId = amuletId
         meta = emptyMetadata
+
   -- Note: here we use Amulet as the single asset, as that is all we have available in this repository.
   -- Given that we are using only the DvP interface, that should though be a fine stand-in for
   -- a DvP across two assets from different registries.
   let aliceLeg = mkTransfer "leg0_a" alice bob 100.0
   let feeLeg = mkTransfer "fee" bob provider 1.0
 
-  -- venue creates a trade
+  -- venue creates a trade where
+  -- - alice -> bob for 100.0 Amulet
+  -- - bob -> provider for 1.0 Amulet
+  --
+  -- and alice allocates using V1, while Bob and Venue allocate using V2.
+  --
+  -- FIXME: increase test coverage for the other combinations of V1 and V2 allocations, allocation requests, and settlement factories
+  --
   now <- getTime
   let settleAt = now `addRelTime` hours 1
       deadline = now `addRelTime` hours 2
@@ -135,11 +141,10 @@ setupOtcTrade  = do
   aliceAlloc.transferLeg === (downcast aliceLeg)
 
   -- alice accepts allocation request directly via her wallet
-  inputHoldingCids <- WalletClient.listHoldingCids alice amuletId
+  inputHoldingCids <- WalletClientV2.listHoldingCids alice amuletId
   now <- getTime
 
   -- alice accepts allocation request directly via her wallet
-  -- TODO(DACH-NY/canton-network-node#18633): test here and for all steps that the expected ledger time bounds are present on the submissions
   enrichedChoice <- RegistryApiV1.getAllocationFactory registryV1 V1.AllocationFactory_Allocate with
     expectedAdmin = registry.dso
     allocation = aliceAlloc
@@ -153,26 +158,13 @@ setupOtcTrade  = do
   -- check metadata
   expectNoBurn result.meta
 
-  -- check lock context
-  {-
-  [(_, lockedHolding)] <- WalletClient.listLockedHoldings alice registry.instrumentId
-  let expectedLock = Lock with
-        expiresAt = None
-        expiresAfter = None
-        holders = [registry.dso]
-        context = Some "allocation for transfer leg 'leg0_a' to Bob"]
-  ((fromSome lockedHolding.lock) with expiresAt = None) === expectedLock
-  assertMsg "locks expire should be in the future"
-    (fromSome (fromSome lockedHolding.lock).expiresAt > now)
-  -}
-
   -- Bob sees the allocation request in his wallet as well
-  [bobAlloc] <- WalletClient.listAllocationRequestsV2 bob
+  [bobAlloc] <- WalletClientV2.listAllocationRequestsV2 bob
   bobAlloc.transferLegs === [aliceLeg, feeLeg]
-  WalletClient.acceptAllocationRequestV2 (AmuletRegistry.amuletToMultiRegistry registry) (basicAccount bob) bob bobAlloc
+  WalletClientV2.acceptAllocationRequestV2 (AmuletRegistryV2.amuletToMultiRegistry registry) (basicAccount bob) bob bobAlloc
 
   -- Lastly, the venue creates an allocation for the fee leg.
-  enrichedChoice <- RegistryApi.getAllocationFactory registry AllocationFactory_Allocate with
+  enrichedChoice <- RegistryApiV2.getAllocationFactory registry AllocationFactory_Allocate with
     expectedAdmin = registry.dso
     allocation = AllocationSpecification with
       authorizer = basicAccount provider
@@ -200,8 +192,6 @@ setupOtcTrade  = do
     feeLeg
 
 
--- FIXME: test multiple executor
-
 -- | Test that a DvP settlement of an OTC trade works when using Amulet via the token standard.
 testDvP_mixed : Script ()
 testDvP_mixed = script do
@@ -221,7 +211,7 @@ testDvP_mixed = script do
             ((allocAdmin, _authorizer), (allocCid, _allocView)) <- Map.toList allocationsMap
             guard (allocAdmin == admin)
             pure allocCid
-      context <- RegistryApi.getSettlementFactory_SettleBatch registry SettlementFactory_SettleBatch with
+      context <- RegistryApiV2.getSettlementFactory_SettleBatch registry SettlementFactory_SettleBatch with
         expectedAdmin = registry.dso
         settlement = mkOtcTradeSettlementInfo otcTradeCid otcTrade
         transferLegs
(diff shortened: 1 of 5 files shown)


token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2.daml
```diff
@@ -150,6 +183,311 @@ instance BackwardCompatible HoldingV1.HoldingView HoldingV2.HoldingView where
...
+      -- execute the allocation using the V1 API
+      result <- exercise factoryCid $ AllocationInstructionV1.AllocationFactory_Allocate with
+        expectedAdmin
+        allocation = allocationV1
+        requestedAt = requestedAt
+        inputHoldingCids = map downcast inputHoldingCids
+        extraArgs
+      pure $ Some (upcast result)
+
+tryDowncast_v2_v1_AllocationSpecification : AllocationV2.AllocationSpecification -> Optional AllocationV1.AllocationSpecification
```
meiersi-da: This is not really compatible with how the extra authorizers are determined and checked for v1 allocations. Let's remove it.

token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2.daml
```diff
@@ -635,3 +974,14 @@ transferFactory_v2_publicFetchImpl this self arg =
   upcast <$> TransferInstructionV1.transferFactory_publicFetchImpl this (downcast self) (downcast arg)
 
 
+-- FIXME: consider whether to inline or use from `splice-util`
```
meiersi-da: Keep inline to reduce version churn from splice-util bumping.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -62,117 +74,336 @@ data TransferLeg = TransferLeg with
...
+    -- FIXME: consider whether we can replace this ^^ in favor of `extraSettlementAuthorizers`
+
+    canWithdraw : [Party]
+      -- ^ The parties that can unilaterally withdraw this allocation.
+      --
+      -- Provided for wallets to determine whether their user can call the `Allocation_Withdraw` choice.
+      --
+      -- Registries that require joint authorization for a withdrawal to complete SHOULD
+      -- set this to the empty list, and provide a registry specific UI to
+      -- initiate and complete the withdrawal.
```
meiersi-da: Note: I'm not sure about this construction, it seems a bit too ad-hoc.

I'm though also wary about the generic construction where the possible actors for every choice are encoded in the view, as that requires a type of the form `Map ActionId [[Party]]`, which can become large for rich action graphs.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -62,117 +74,340 @@ data TransferLeg = TransferLeg with
...
+      -- to authorize the whole batch settlement.
+
+      -- V1 settlement is only available if there is a single transfer leg. It
+      -- guarantees that the owners of its sender and receiver account
+      -- are also required to authorize the whole batch settlement in addition
+      -- to the executor.
+      --
+      -- Additional versions may defined in future extensions of the token standard.
+      --
+    -- FIXME: consider whether we can replace this ^^ in favor of `extraSettlementAuthorizers`
```
meiersi-da, line 121: I will give this a try. Left it as-is for now to get a first version into a reviewable state.

daml/splice-amulet/daml/Splice/AmuletAllocationV2.daml
```diff
@@ -0,0 +1,166 @@
...
+        canWithdraw = [[allocation.authorizer.owner]]
+        meta = emptyMetadata
+
+      -- no extra observers required, as the choice bodies do not introduce further informees
+      allocation_settleExtraObservers _arg = []
+      allocation_cancelExtraObservers _arg = []
+      allocation_withdrawExtraObservers _arg = []
+
+      allocation_withdrawImpl self arg@(V2.Allocation_Withdraw{..}) = do
+        archive self
```
meiersi-da, line 71: TODO: simplify the check using the `canWithdraw` information

token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2/Allocations.daml
```diff
@@ -0,0 +1,558 @@
...
+  -- * Allocation implementation
+  allocationV2_defaultSettleImplUsingV1,
+  allocationV2_defaultWithdrawImplUsingV1,
+  allocationV2_defaultCancelImplUsingV1,
+
+  -- * SettlementFactory implementation
+  settlementFactoryV2_defaultSettleBatchImpl,
+
+  validateAndPrepareAllocations,
+  netAllocationCreditAmount,
```
meiersi-da, line 18: @bame-da: The default choices in here are worth reviewing jointly with their usage in `AmuletAllocation` and `AmuletAllocationV2`.

token-standard/splice-token-standard-test-v2/daml/Splice/Tests/TestAmuletTokenDvP_V1V2Mixed.daml
```diff
@@ -0,0 +1,370 @@
...
+    registry
+    providerBeneficiaries
+    otcTradeCid
+    otcTrade
+    requestCids
+    amuletId
+    feeLeg
+
+
+-- | Test that a DvP settlement of an OTC trade works when using Amulet via the token standard.
```
meiersi-da, line 195: @bame-da: mixed version settlement worth reviewing together with the code used in the `TradingAppV2` to manage the actors.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -1,18 +1,30 @@
 -- Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 -- SPDX-License-Identifier: Apache-2.0
```
meiersi-da, line 2: @bame-da: worth reviewing in full. Apart from the FIXME's the API here is polished.

I expect that we want to get rid of the `settlementVersion` though; and I expect that we want to make a more consistent decision wrt actors vs. controller overrides.

token-standard/splice-api-token-allocation-request-v2/daml/Splice/Api/Token/AllocationRequestV2.daml
```diff
@@ -6,8 +6,6 @@
 -- for the purpose of settling a DvP or a payment as part of an app's workflow.
 module Splice.Api.Token.AllocationRequestV2 where
```
meiersi-da, line 7: @bame-da: ready for review, but nothing surprising in here.

token-standard/splice-api-token-allocation-instruction-v2/daml/Splice/Api/Token/AllocationInstructionV2.daml
```diff
@@ -1,7 +1,7 @@
 -- Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 -- SPDX-License-Identifier: Apache-2.0
 
--- | Interfaces to enable wallets to instruct the registry to create allocations.
```
meiersi-da, line 4: @bame-da: this module is still a bit WIP. The main construction site is how to deal best with the actors pattern and different authorization policies by different registries. I don't think it's going to be terribly difficult. Just didn't have time to build a convincing solution proposal.

I suggest we first align on `AllocationV2` + mixed version settlement + allocation request, before we finish this one.

token-standard/splice-token-standard-test-v2/daml/Splice/Testing/TokenStandard/WalletClientV2.daml
```diff
@@ -91,43 +104,97 @@ checkHoldingBounds p holdingCid (lb, ub) = do
...
+          transferLegs
   pure amuletAllocs
+
+listAllocationRequestsV2 : Party -> Script [V2.AllocationRequestView]
+listAllocationRequestsV2 p = do
+  reqs <- queryInterface @V2.AllocationRequest p
+  return [ reqView | (_, Some reqView) <- reqs ]
+
+
+-- | Accept the allocation request by authorizing the transfer of all its requested legs.
```
meiersi-da, line 155: @bame-da: this is how I expect wallets to accept allocation requests. Worth reviewing to check alignment.

I haven't yet built the analoguous function for V1 requests. I expect it will work by creating one v1 allocation for every single leg.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -62,117 +74,343 @@ data TransferLeg = TransferLeg with
...
+      -- V1 settlement is only available if there is a single transfer leg. It
+      -- guarantees that the owners of its sender and receiver account
+      -- are also required to authorize the whole batch settlement in addition
+      -- to the executor.
+      --
+      -- Additional versions may defined in future extensions of the token standard.
+      --
+    -- FIXME: consider whether we can replace this ^^ in favor of `extraSettlementAuthorizers`
+
+    canWithdraw : [[Party]]
```
bame-da, line 123: I think this may be overkill. I'd use the `actors` pattern, say that implementatios SHOULD let the authorizer account owner withdraw (though they MAY decide just to create withdrawal request to the account admin) and leave it at that.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -62,117 +74,343 @@ data TransferLeg = TransferLeg with
...
+    -- the body of this choice.
     with
+      actors : [Party]
+        -- ^ Parties authorizing the withdrawal.
+        --
+        -- Implementations MUST check these parties to avoid unauthorized withdrawal.
+        --
+        -- Implementations are free to decide on the required authorization for withdrawal.
+        -- They MUST however allow the parties listed in the
+        -- `AllocationView.canWithdraw` to call this choice.
```
bame-da, line 238: For V1 allocations, the authorizer.owner should be allowed to call this.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -62,117 +74,343 @@ data TransferLeg = TransferLeg with
...
-  deriving (Show, Eq)
+      -- ^ Additional metadata specific to the settlement, used for extensibility.
+  deriving (Eq, Show)
 
 -- | The result of the `Allocation_Cancel` choice.
 data Allocation_CancelResult = Allocation_CancelResult
   with
-    senderHoldingCids : [ContractId Holding]
-      -- ^ The holdings that were released back to the sender.
+    authorizerHoldingCids : [ContractId Holding]
```
bame-da, line 283: I'm not keen on renaming `sender` to `authorizer`. In some cases it improves clarity, in others like here it doesn't. Since it's not a clear improvement I'd favour fewer changes.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -62,117 +74,343 @@ data TransferLeg = TransferLeg with
...
+
+  nonconsuming choice SettlementFactory_SettleBatch : SettlementFactory_SettleBatchResult
+    -- ^ Settle a batch of allocations for instruments with the same instrument admin.
+    --
+    -- The choice is structured in this form for efficiency and privacy. It
+    -- enables the instrument admin to only perform net debits and credits for
+    -- each account across all transfers being settled; and restrict visibility of
+    -- each credit or debit to executor, admin, and affected account only.
+    with
+      expectedAdmin : Party
```
bame-da, line 329: I've never really understood or liked this field. Why not just check the signatory off-chain before calling?

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
```diff
@@ -62,117 +74,343 @@ data TransferLeg = TransferLeg with
...
+        --
+        -- Implementations MUST check this value to avoid unauthorized settlement execution.
+        --
+        -- This choice's default controllers are the `settlement.executor` and the
+        -- `settlementV1Authorizers`. Implementations SHOULD allow executing the
+        -- choice with these default controllers to maximize compatibility with apps.
+      extraArgs : ExtraArgs
+        -- ^ Additional choice arguments.
+    observer settlementFactory_settleBatchExtraObservers this arg
+    controller fromOptional (defaultSettlementFactory_SettleBatchControllers arg) controllersOverride
```
bame-da, line 375: I cannot quite make sense of the settlementV1Authorizers and controllerOverride pattern here.

If `settlementV1Authorizers` are not controllers, how can you use their authority?

And why would you set the controllersOverride here rather than on SettlementInfo? I think it would be clearer to force apps to advertise the authorization upfront, not determine it only at the time of settlement.

token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2/Allocations.daml
```diff
@@ -0,0 +1,558 @@
...
+allocationV2_defaultSettleImplUsingV1 this self arg@(AllocationV2.Allocation_Settle{..}) = do
+  -- no archive of self, as the V1 choice is consuming
+  onlyAllowDefaultControllers' "V2.Allocation_Settle" this arg
+  let v1Cid = coerceInterfaceContractId self : ContractId AllocationV1.Allocation
+  result <- exercise v1Cid AllocationV1.Allocation_ExecuteTransfer with extraArgs
+  -- Archive the receiver holdings,
+  -- which is required, as the receiver will get their credit as part of
+  -- settling their allocation. The receiver parses this transaction using the
+  -- V2 choices and will thus understand it.
+  forA_ result.receiverHoldingCids $ \receiverHoldingCid ->
```
bame-da, line 80: Wouldn't it be nicer if we passed this into the receiver allocation_settle choice as a previous credit? Ie can we avoid the create-archive-create chain?

token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2/Allocations.daml
```diff
@@ -0,0 +1,558 @@
...
+  let admin = (view factory).admin
+
+  requireExpectedAdminMatch admin expectedAdmin
+  onlyAllowDefaultControllers "SettlementFactory_SettleBatch" factory arg
+
+  -- validate and prepare the allocations
+  prep <- validateAndPrepareAllocations admin arg
+
+  -- settle all v1 allocations
+  newHoldingCidssV1 <- forA prep.allocationsSettledUsingV1 $ \((allocationCid, allocationView), leg) -> do
```
bame-da, line 171: Shouldn't you be using `allocationV2_defaultSettleImplUsingV1` here? As written, if you have a V1 from sender and V2 from receiver, the receiver gets two credits, don't they?

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
meiersi-da: > I've never really understood or liked this field. Why not just check the signatory off-chain before calling?

Yeah, it's not a pretty field at all

Note checking the signatory-off chain is neither easily possible (Ledger API does not yet offer decoding disclosed contracts) nor is it an inadequate safe-guard, as it is very easy for clients to forget to do.

That said I believe we can remove it in V2 because every factory choice does specify the expected admin already as part of the other arguments of the choice. So if you agree, then I'd suggest we add a note in the CIP text, and I remove the expected admin.

token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2/Allocations.daml
meiersi-da: Only the v1 allocations that are not matched with v2 allocations are settled this way.

bame-da: I feel like this way of calling the V1 choice as part of V2 settlement is too dangerous. It's too easy to miss the needed tidy-up of the receiver outputs to justify the value of tx parsing for the sender.

Maybe here's a middle ground: Add a metadata field `batchSettleMode` to the call of the V1 ExecuteTransfer choice. If set, you don't create receiver outputs.
Since this is only ever to the _disadvantage_ of the receiver, there's no risk this being set outside of a batchSettlement. And it's much easier to remember setting this than switching between the direct choice call and the `allocationV2_defaultSettleImplUsingV1`. Lastly, it avoids create-archive-create chains.

bame-da: In fact if you do this, you no longer need to do the switching inside the batchSettle choice at all. You can make the switch inside the Allocation_Settle choice. It can decide whether it's a V1 allocation and if so, call the V1 with the metadata internally. Now the V2 settle choice because simple which also removes

```
-- The logic for doing this correctly is nontrivial, which is why registries
-- SHOULD use this default implementation where possible.
```

bame-da: > Only the v1 allocations that are not matched with v2 allocations are settled this way.
This surprises me. I thought the point of this whole dance was to make sure that the V1 sender gets to see a V1 choice call. If you settle a matched V1 allocation through a V2 call, then do you still get the event?

token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2/Allocations.daml
```diff
@@ -0,0 +1,558 @@
...
+        leg <- allocationView.allocation.transferLegs
+        pure (authorizer, leg)
+
+  let hasMatchingV2Allocation (_, v1Leg) = (v1Leg.receiver, v1Leg) `Set.member` v2Legs
+  let (matchedV1Allocations, unmatchedV1Allocations) = partition hasMatchingV2Allocation v1Allocations
+  let allocationsSettledUsingV1 = unmatchedV1Allocations
+
+  -- We settle V1 allocations that are matched with a V2 allocation using the V2 API to
+  -- avoid having to archive the holding for the receiver that would be created as part of its
+  -- V1 settlement. This seems too finnicky. Wallets will have to use their fallback tx history
```
bame-da, line 234: I agree: too finnicky.
I'd maybe just partition into matched/unmatched.
Call V2 settlement on every allocation with a metadata field that says whether it's matched/unmatched.
V1 allocations call the V1 choice internally preserving that metadata field.
If matched, they don't output receiver holdings.

token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2/Allocations.daml
meiersi-da: Not sure this is worth optimizing for is it? The `Archive` won't create an extra view neither will the `create` on the receipt allocation settlement, which may anyways be required because there are other transfers.

The complexity increase for the logic in `Allocation_Settle` is though significant, as it now needs to take extra params and net that extra input. Likewise parsing the choice with these extra inputs seems more complex than just parsing each `V2.Allocation_Settle` as a debit or credit for the authorizer.

token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2/Allocations.daml
meiersi-da: > If you settle a matched V1 allocation through a V2 call, then do you still get the event?

Yes you do, as the V2 call does call the `V1.Allocation_ExecuteTransfer` choice.

> In fact if you do this, you no longer need to do the switching inside the batchSettle choice at all. You can make the switch inside the Allocation_Settle choice. It can decide whether it's a V1 allocation and if so, call the V1 with the metadata internally. Now the V2 settle choice because simple which also removes

I'm working on a version where there's no version dependent logic, and instead the required features to settle v1 allocations are available in general. It looks like this approach removes much of the complexity. I'd like to first give that a shot before doing the metadata dance.

I also suspect that the metadata dance is not available for all registries, as allowing everybody to burn their asset holdings may not be possible.

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
meiersi-da: undone

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
meiersi-da: removed

token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2/Allocations.daml
meiersi-da: kept, as per our slack discussion

token-standard/splice-api-token-utils-v2/daml/Splice/Api/Token/UtilsV2/Allocations.daml
meiersi-da: Switched away from explicit v1 compat, but did not use the metadata-based switched for the following reason:
> I would like to do the V1/V2 switch, but how can we justify that it is OK to burn one's asset holdings without admin approval? That is what becomes possible with that metadata switch. To avoid that we'd need flexible actors, which we don't have on V1

token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml
meiersi-da: moved to actors pattern uniformly

PR #4363, opened 2026-03-10
Add checksums to bulk storage uploads

isegall-da:
Serves two purposes:
integrity checks (the S3 API validates the uploaded data automatically)
persists a checksum for the whole object, for BFT guarantees, i.e. easy comparison between Scan instances

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnectionForTests.scala
+++ b/apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnectionForTests.scala
@@ -3,7 +3,11 @@ package org.lfdecentralizedtrust.splice.store
 import com.digitalasset.canton.logging.NamedLoggerFactory
 import org.lfdecentralizedtrust.splice.config.S3Config
 import software.amazon.awssdk.core.async.AsyncResponseTransformer
-import software.amazon.awssdk.services.s3.model.{GetObjectRequest, GetObjectResponse, GetObjectTaggingRequest}
+import software.amazon.awssdk.services.s3.model.{
+  GetObjectRequest,
+  GetObjectResponse,
+  GetObjectTaggingRequest,
+}
 
 import java.io.{ByteArrayInputStream, IOException}
 import java.nio.ByteBuffer


apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3BucketConnectionForUnitTests.scala
```diff
@@ -50,12 +49,6 @@ class S3BucketConnectionForUnitTests(
     }
   }
 
-  override def writeFullObject(key: String, content: ByteBuffer)(implicit
```
isegall-da, line 53: no longer used anywhere

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
```diff
@@ -4,56 +4,62 @@
...
+        .tagSet()
+        .asScala
+        .find(_.key() == "splice-checksum")
+        .map(_.value())
+        .getOrElse(throw new RuntimeException("Missing checksum tag"))
+    } yield {
+      val bytes = new Array[Byte](data.remaining)
+      data.duplicate.get(bytes)
+      val bis = new ByteArrayInputStream(bytes)
+      // We compare the computed & stored checksum to one we independtly compute via the system's `sha256sum` executable for sanity
```
ray-roestenburg-da:
```suggestion
      // We compare the computed & stored checksum to one we independently compute via the system's `sha256sum` executable for sanity
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
```diff
@@ -4,56 +4,62 @@
...
+      val expectedChecksum =
+        ("sha256sum" #| "awk '{print $1}'" #| "xxd -r -p" #| "base64" #< bis).!!.trim
```
ray-roestenburg-da: This feels really weird to do here, (to call out to a process), can you use a sha256 Java lib instead?

ray-roestenburg-da: You can just do something line `MessageDigest.getInstance("SHA-256").digest(...)` and then `Base64.getEncoder.encodeToString(...)` ? I really don't like that this code now requires the callout to sys process as well as these shell tools to be available, not to mention overhead and all kinds of other problems.

ray-roestenburg-da: Maybe an option to move to a test, if you are uncertain about the code..

isegall-da: Thanks for calling out my laziness. This is actually test code (as documented above the method) in a prod class, which I agree is terrible and should be moved out of here.

> MessageDigest.getInstance("SHA-256").digest(...)

Of course, but that's what we do on the write path. I wanted to cross-check it against a different source. Just rerunning the same thing on read doesn't give you much guarantees that you're doing something standard. Just for testing of course, which goes back to me thanking you for forcing me to move this code out of the prod class :)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
isegall-da: Glad to see that your spelling skills are just as bad as mine :D

ray-roestenburg-da: LOL!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
isegall-da: Done. Mind reviewing again please?

PR #4388, opened 2026-03-12
Persist ext txn hashes and expose with threshold date

jagathweerasinghe-da:
[74a8c8d](https://github.com/hyperledger-labs/splice/pull/4388/commits/74a8c8d4658f2ce73e6eda8c36b28b6b8db7d978) is from the reverted PR. 

Delta is the threshold date and changes to the tests.

### DIFF
--- a/apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V062__update_history_for_hash.sql
+++ b/apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V062__update_history_for_hash.sql
@@ -1,4 +1,4 @@
 ALTER TABLE update_history_transactions
-    ADD COLUMN external_transaction_hash bytea
+    ADD COLUMN external_transaction_hash bytea;
 
 -- To avoid long migration waits, the index is created asynchronously by the application (SqlIndexInitializationTrigger)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -1614,6 +1620,14 @@ class UpdateHistory(
       )
     }
     val events: Seq[Event] = (createEvents ++ exerciseEvents).sortBy(_.getNodeId)
+    // Only include the external transaction hash for transactions recorded on or after the threshold timestamp.
+    val externalTransactionHash: ByteString =
+      externalTransactionHashThresholdTimestamp match {
+        case Some(threshold) if updateRow.recordTime >= threshold =>
```
jagathweerasinghe-da: I assume record_time is the correct timestamp to use.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
```diff
@@ -80,6 +80,9 @@ case class ScanAppBackendConfig(
     acsStoreDescriptorUserVersion: Option[Long] = None,
     txLogStoreDescriptorUserVersion: Option[Long] = None,
     bulkStorage: BulkStorageConfig = BulkStorageConfig(),
+    // The thresholdDate from which external transaction hashes are included in the updates from internal ScanAPIs.
+    // TODO(#4249): use on-ledger synchronization for switching record times
+    externalTransactionHashThresholdDate: Option[String] = Some("2026-07-01T00:00:00Z"),
```
jagathweerasinghe-da: TBD: date

OriolMunoz-da: I'd just pick 2030 so we don't have to worry about it until we have the actual ledger-based solution (because priorities change so it's better to be more conservative here)

OriolMunoz-da: don't use Option[String], use `Option[Instant]` so you don't have to parse it later. There should already be a ConfigReader/Writer for Instant, so it should just work

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/ScanApp.scala
```diff
@@ -215,6 +216,8 @@ class ScanApp(
         enableissue12777Workaround = true,
         enableImportUpdateBackfill = config.updateHistoryBackfillImportUpdatesEnabled,
         nodeMetrics.dbScanStore.history,
+        externalTransactionHashThresholdTimestamp = config.externalTransactionHashThresholdDate
+          .map(s => CantonTimestamp.assertFromInstant(java.time.Instant.parse(s))),
```
OriolMunoz-da: and once you change the config, you won't need this anymore

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
OriolMunoz-da: yes

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/UpdateHistoryTest.scala
```diff
@@ -957,5 +959,168 @@ class UpdateHistoryTest extends UpdateHistoryTestBase {
...
+
+        val store = mkStore(
+          storeName = "null_hash_store",
+          externalTransactionHashThresholdTimestamp =
+            Some(CantonTimestamp.assertFromInstant(externalTxnHashThresholdDate)),
+        )
+
+        // Simulate receiving null from java APIs and writing it to Scan DB
+        @SuppressWarnings(Array("org.wartremover.warts.Null"))
+        val javaNullHash: ByteString = null
```
OriolMunoz-da: does this actually happen in the API?
It's defined as NonNull:

`@NonNull private final ByteString externalTransactionHash;`

so no need to test for that

(this is a Java convention, but that is respected in the javaapi)

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/UpdateHistoryTest.scala
```diff
@@ -957,5 +959,168 @@ class UpdateHistoryTest extends UpdateHistoryTestBase {
...
+        }
+      }
+
+      "return stored external transaction hash when not empty" in {
+        val externalTxnHashThresholdDate = Instant.parse("2026-06-30T23:59:59Z")
+        // recordDate must be after the externalTxnHashThresholdDate for the hash to be included
+        val recordDate = Instant.parse("2026-07-01T00:00:00Z")
+
+        val store = mkStore(
+          storeName = "store",
```
OriolMunoz-da: just use the default here and in the test above

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsJdbcTypes.scala
```diff
@@ -366,6 +366,24 @@ trait AcsJdbcTypes {
     */
   protected def lengthLimited(s: String): String2066 = String2066.tryCreate(s)
 
+  /** Transaction hash is SHA-256 (32 bytes), but we apply a relaxed future-proof limit of 1024 bytes just in case.
+    *  The hash is not guaranteed to be present. For consistency with historical data,
+    *  we represent a missing hash as NULL in the database, and not as an empty byte array (\x).
+    */
+  protected def sanitizedExtTxnHash(
```
OriolMunoz-da: this file already defines a SetParameter[ByteString], so you can just define one for SetParameter[Option[ByteString]] analogous to it - no need to complicate it with length checks

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -245,6 +245,15 @@ object SqlIndexInitializationTrigger {
           on round_party_totals (store_id, party, closed_round desc)
         """,
       ),
+    IndexAction
+      .Create(
+        indexName = "updt_hist_tran_hi_eth",
+        createAction = sqlu"""
+          create index concurrently if not exists updt_hist_tran_hi_eth
+          on update_history_transactions (history_id, external_transaction_hash)
+          where external_transaction_hash is not null
```
OriolMunoz-da, line 254: in your other comment:

> For consistency with historical data, we represent a missing hash as NULL in the database, and not as an empty byte array (\x).

it's not only about consistency, but also to prevent bloating the index with empty hashes that are not null

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__update_history_for_hash.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE update_history_transactions ADD COLUMN external_transaction_hash bytea;
```
OriolMunoz-da: add a check constraint that it's non-empty. By which I mean:
- null => allowed
- [] => NOT allowed
- [...] => allowed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/UpdateHistoryTestUtil.scala
```diff
@@ -142,12 +169,59 @@ trait UpdateHistoryTestUtil extends TestCommon {
...
+    (actual.update, recorded.update) match {
+      case (
+            TransactionTreeUpdate(actualTx),
+            TransactionTreeUpdate(recordedTx),
+          ) =>
+        def toByteArray(hash: ByteString): Array[Byte] =
+          Option(hash).map(_.toByteArray).getOrElse(Array.emptyByteArray)
+
+        toByteArray(actualTx.getExternalTransactionHash) shouldBe
+          toByteArray(recordedTx.getExternalTransactionHash)
```
OriolMunoz-da: pretty sure ByteString equality works as-is and you shouldn't need this

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
jagathweerasinghe-da: done

jagathweerasinghe-da: done

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsJdbcTypes.scala
jagathweerasinghe-da: done

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V060__update_history_for_hash.sql
jagathweerasinghe-da: done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/UpdateHistoryTestUtil.scala
jagathweerasinghe-da: yes, that works. changed.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/UpdateHistoryTest.scala
jagathweerasinghe-da: @NonNull is just an indicator, we could get Null;

OriolMunoz-da: yes - what I mean is that the participant *will* respect it (meaning: will never give you null there), so you don't need to check for it

jagathweerasinghe-da: Right, protobuf initialize it to ByteString.EMPTY
 
 ```
 public static final int EXTERNAL_TRANSACTION_HASH_FIELD_NUMBER = 10;
 private com.google.protobuf.ByteString externalTransactionHash_ = com.google.protobuf.ByteString.EMPTY;
 ```
We can use hasExternalTransactionHash() to check whether the hash is explicit value, but since the default value is ByteString.EMPTY, we don't even need to check that. 

dropped the null test.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -588,9 +588,10 @@ class UpdateHistory(
       .map(lengthLimited)
     val safeWorkflowId = lengthLimited(tree.getWorkflowId)
     val safeCommandId = lengthLimited(tree.getCommandId)
-    val safeExternalTransactionHash = sanitizedExtTxnHash(tree.getExternalTransactionHash)
+    val externalTransactionHash = Option(tree.getExternalTransactionHash)
```
ray-roestenburg-da: should this limit the length that the getExternalTransactionHash is allowed to be, even if it is a generous length limit, to prevent say an attacker to upload a very large byte array? (maybe I'm overthinking it)
Note that the workflowId and commandId are lengthLimited, it seems strange that we go through all that trouble and then just accept a byte array of any size

jagathweerasinghe-da: I was thinking the same while adding sanitizedExtTxnHash(). Since we’re pulling hashes from the Ledger API rather than end-users, there’s an inherent level of trust there. However, adding sanity checks on the scan side provides a defensive layer against unexpected behavior or bugs on the Ledger side. The flip side, of course, is that it adds a bit of complexity to the implementation.

WDYT @OriolMunoz-da ?

ray-roestenburg-da: There seems to be a LengthLimitedByteString in canton util, with ByteString190, 256, 4096, 6144 I think, so it should not be too hard to do

ray-roestenburg-da: Of course you need to know the max size of the hash

OriolMunoz-da: > should this limit the length that the getExternalTransactionHash is allowed to be, even if it is a generous length limit, to prevent say an attacker to upload a very large byte array?

they can do that? I thought it's a SHA that comes from the participant. Our apps already trust it to not do insane/malformed stuff.

rautenrieth-da: Commented on an outdated commit: for other fields, we crash if we the participant returns unexpectedly long values (via `lengthLimited`).

jagathweerasinghe-da: Yes, figured out we throw IllegalArgumentException. Makes sense to make it consistent. Added a lenthlimit check for hash.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V062__update_history_for_hash.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE update_history_transactions
+    ADD COLUMN external_transaction_hash bytea
+        CONSTRAINT external_transaction_hash_not_empty CHECK (length(external_transaction_hash) > 0);
```
rautenrieth-da: AFAICT, this constraint will not be used for query planning, and will add a negligible overhead when inserting rows. We usually don't include such constraints, but I don't see anything wrong with keeping it here.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -587,17 +587,21 @@ class UpdateHistory(
       .map(lengthLimited)
     val safeWorkflowId = lengthLimited(tree.getWorkflowId)
     val safeCommandId = lengthLimited(tree.getCommandId)
+    val safeExternalTransactionHash =
+      lengthLimitedExtTxnHash(tree.getExternalTransactionHash)
+
+    import storage.DbStorageConverters.setParameterOptionalByteArray
```
rautenrieth-da, line 594: Why do we need this implicit? `safeExternalTransactionHash` is of type `ByteString`, the implicit is for `Option[Array[Byte]]`.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsJdbcTypes.scala
```diff
@@ -366,6 +378,19 @@ trait AcsJdbcTypes {
     */
   protected def lengthLimited(s: String): String2066 = String2066.tryCreate(s)
 
+  private def lengthLimitedByteString(bs: ByteString, maxLength: Int): ByteString = {
```
rautenrieth-da, line 369: I assume this should work on `Array[Byte]` and not `ByteString` (which is a protobuf thing)?

I would also encode the decision to store empty arrays as NULL here (or in a similar helper), i.e., have this method return something like an `Option[Array[Byte]]`. 

Having the transformation in an implicit like `nullableByteStringSetParameter` is more dangerous as different callers might want to store empty byte arrays as such in the database, and it's not immediately visible what implicits are used in your code.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
jagathweerasinghe-da: Right, that implementation is not easily readable; Made it clear with sanitizedExtTxnHash() returning Option[Array[Byte]].

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/AcsJdbcTypes.scala
jagathweerasinghe-da: Agreed, dropped nullableByteStringSetParameter().

jagathweerasinghe-da: Now, sanitizedExtTxnHash() does the sanitization explicitly.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V062__update_history_for_hash.sql
jagathweerasinghe-da: Makes sense; I checked the repo and we don't usually add these constraints. so dropped it for consistency; our app code make sure we convert empty bytestrings to SQL null;

PR #4392, opened 2026-03-12
make participant_console command work with an LSU participant

mblaze-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4391

As the participant in the logical synchronizer deployment scheme does not have a migration ID in its name, the `participant_console` subcommand needed a way to specify that when looking for the participant pod it should not expect a migration ID in its name.

### DIFF
--- a/build-tools/cncluster
+++ b/build-tools/cncluster
@@ -436,9 +436,9 @@ function subcmd_logs() {
 function k8s_labels_for_app() {
     local global_domain_specific_apps=("global-domain-cometbft" "global-domain-sequencer" "global-domain-mediator")
     local domain_specific_apps="@(participant|multi-participant-[0-9][0-9])"
-    local logical_synchronizer_deployment_mode=$(
-        "${SPLICE_ROOT}/cluster/scripts/get-resolved-config.sh"
-        | yq ".synchronizerMigration[] | select(.id == ${domain}).enableLogicalSynchronizerDeploymentMode"
+    local logical_synchronizer_deployment_mode
+    logical_synchronizer_deployment_mode=$( \
+        get_resolved_config | yq ".synchronizerMigration[] | select(.id == ${domain}).enableLogicalSynchronizerDeploymentMode" \
     )
 
     # shellcheck disable=SC2076


build-tools/cncluster
```diff
@@ -443,7 +443,7 @@ function k8s_labels_for_app() {
        # If the app is in the list, append the domain to the app
        app_container=${app#"global-domain-"}
        app="global-domain-${domain}-${app#"global-domain-"}"
-    elif [[ "$app" == $domain_specific_apps ]]; then
+    elif [[ "$app" == $domain_specific_apps && "$domain" != "-" ]]; then
```
nicu-da: can we just make the `domain` optional? if not set instead of -

mblaze-da: The domain already is optional but the default is `0`. If I change the default I'll be breaking backwards compatibility and peoples expectations.

nicu-da: fair point. any reason to not just read the config from the resolved file though? and check if LSU is enabled for that migration id

mblaze-da: We already have references in our docs to using this with the default migration ID. If you think that such change would be fine I can do it and follow up with a PR to change the following doc:
https://github.com/DACH-NY/canton-network-internal/blob/6721d5b366eb2f2f949c354eeeecd74a10384180/cluster/README.md?plain=1#L1920-L1921

mblaze-da: > fair point. any reason to not just read the config from the resolved file though? and check if LSU is enabled for that migration id

That's what I meant by
> I also thought of a solution which looked at the config to decide whether the participant pod has the migration ID in its name or not but it was a bit more complicated.

The `synchronizerMigration` is a mapping not a list which results in a bit more complicated `yq` expression to filter by migration ID. I was hesitant about doing it considering the added complexity but I guess it would be the most user friendly solution.

nicu-da: I would just do that instead. find the migration config for that id and check if it's enabled.

build-tools/cncluster
```diff
@@ -436,14 +436,18 @@ function subcmd_logs() {
 function k8s_labels_for_app() {
     local global_domain_specific_apps=("global-domain-cometbft" "global-domain-sequencer" "global-domain-mediator")
     local domain_specific_apps="@(participant|multi-participant-[0-9][0-9])"
+    local logical_synchronizer_deployment_mode=$(
+        "${SPLICE_ROOT}/cluster/scripts/get-resolved-config.sh"
```
nicu-da: check the existing get_resolved_config

PR #4400, opened 2026-03-12
proxy s3 downloads via scan

isegall-da:
I tried hard to get guardrail to accept `application/octet-stream` and do the right thing. Managed to convince it to generate server-side (with just a little help in form of a small manually-written implicit), but completely failed client-side. No matter what I tried, I couldn't get it to not treat it as json and try to unmarshal it. So I ended up telling it not to generate the client, and I took the broken code that it generated, and manually modified it to code that works.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/ScanStreamClient.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/ScanStreamClient.scala
@@ -1,3 +1,6 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
 package org.lfdecentralizedtrust.splice.scan.admin.api.client
 
 import org.lfdecentralizedtrust.splice.http.v0.Implicits.*


apps/scan/src/main/openapi/scan-stream-server.yaml
isegall-da, line 1: In a separate file because AFAICT, that's the only granularity in which one can tell guardrail to generate only server-side and not client-side.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ResponseEntityGuardrailSupport.scala
isegall-da, line 1: Yes, it forced me to write a marshaller that does nothing...

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/ScanStreamClient.scala
isegall-da, line 1: This is basically guardrail codegen'ed client, my change is the ok() method, where guardrail insisted on trying to unmarshal the response as json.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTimeBasedIntegrationTest.scala
```diff
@@ -600,12 +603,14 @@ class ScanTimeBasedIntegrationTest
 
       val updateObjs = s3Objs.filter(_.key().contains("/updates"))
       val updatesFromS3 = updateObjs
-        .flatMap(
-          readUncompressAndDecode(
-            bucketConnection,
+        .flatMap { obj =>
+          val out = new ByteArrayOutputStream()
+          sv1ScanBackend.bulkStorageDownload(obj.key(), out).futureValue
```
isegall-da, line 625: this is the main change in the test that demonstrates that this whole mess actually works, where we read the object via the scan proxy instead of directly from s3.

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/ScanAppReference.scala
```diff
@@ -748,6 +752,18 @@ abstract class ScanAppReference(
       )
     }
   }
+
+  @Help.Summary("Download a bulk storage object")
+  def bulkStorageDownload(objectKey: String, output: OutputStream)(implicit
+      ec: ExecutionContext,
+      as: ActorSystem,
+  ): Future[Long] =
```
ray-roestenburg-da, line 760: Is it not easier for testing if you return the binary data (as a Source) directly? now this only returns the size in bytes I guess?

ray-roestenburg-da: Oh wait the output is by reference in the parameter to bulkStorageDownload.. ok I guess that works

ray-roestenburg-da: You could also use a FileIO Sink from connectors and write it to a file, or if it fits in mem, materialize into a Seq[ByteString], then to Array[Byte] if you really wanted to, I think as a user of this I would prefer a Source I can do stuff with for testing

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/commands/HttpScanAppClient.scala
```diff
@@ -2473,4 +2484,27 @@ object HttpScanAppClient {
         .leftMap(_.toString)
     }
   }
+
+  case class BulkStorageDownload(
+      objectKey: String
+  ) extends ScanStreamBaseCommand[
+        BulkStorageDownloadResponse,
+        Source[org.apache.pekko.util.ByteString, Any],
```
ray-roestenburg-da, line 2492: The Any is a bit of a smell

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/commands/HttpScanAppClient.scala
```diff
@@ -2473,4 +2484,27 @@ object HttpScanAppClient {
...
+    ): EitherT[Future, Either[Throwable, HttpResponse], BulkStorageDownloadResponse] =
+      client.bulkStorageDownload(objectKey)
+
+    override protected def handleOk()(implicit
+        decoder: TemplateJsonDecoder
+    ): PartialFunction[BulkStorageDownloadResponse, Either[
+      String,
+      Source[org.apache.pekko.util.ByteString, Any],
+    ]] = { case BulkStorageDownloadResponse.OK(response) =>
+      Right(response.dataBytes)
```
ray-roestenburg-da, line 2507: You need to discardEntityBytes if the source is not fully read, which is in the error cases, I don't see that happening here, please have a look at https://pekko.apache.org/docs/pekko-http/current/implications-of-streaming-http-entity.html (maybe you should do this in the base or in the ScanStreamClient

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/ScanStreamClient.scala
```diff
@@ -0,0 +1,132 @@
...
+      HttpEntity.Empty,
+      HttpProtocols.`HTTP/1.1`,
+    ).flatMap(req =>
+      EitherT(
+        httpClient(req)
+          .flatMap(resp =>
+            resp.status match {
+              case StatusCodes.OK => ok(resp.entity)
+              case StatusCodes.NotFound => notFound(resp)
+              case _ => FastFuture.successful(Left(Right(resp)))
```
ray-roestenburg-da, line 112: here in the 'other' case you need to call resp.discardEntityBytes

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/ScanStreamClient.scala
```diff
@@ -0,0 +1,132 @@
...
+          )
+          .recover({ case e: Throwable => Left(Left(e)) })
+      )
+    )
+  }
+}
+sealed abstract class BulkStorageDownloadResponse {
+  def fold[A](
+      handleOK: org.apache.pekko.http.scaladsl.model.ResponseEntity => A,
+      handleNotFound: _root_.org.lfdecentralizedtrust.splice.http.v0.definitions.ErrorResponse => A,
```
ray-roestenburg-da: Since this is now not generated code anymore you could cleanup the imports

ray-roestenburg-da (changes requested):
Main thing is the discardEntityBytes

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/ScanAppReference.scala
isegall-da: As much as I'm enjoying Pekko, I still think that it introduces mental complexity over a simple OutputBuffer, so until there's an actual need to use Pekko here, I'd rather not force it.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/commands/HttpScanAppClient.scala
isegall-da: Hmmm... I guess the Guardrail maintainers never read this article, this is purely codegen code that I didn't touch...

isegall-da: I'm not sure I agree though. In this case, this just returns the entire HttpResponse as it is, and passes the responsibility of handling it (e.g. deciding whether to process or discard the entity bytes) up to the caller. That seems fine to me. Especially given that we use this client only for testing code.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/ScanStreamClient.scala
isegall-da: Yeah, I tried to minimize how much I touch the generated code, but I agree that a bit of cleanup won't hurt. Will do

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/commands/HttpScanAppClient.scala
ray-roestenburg-da: You need to discardBytes if you do not read the response fully, or if no one can read the response fully, I maybe missed something, but thought that all of this eventually ends up as NotFound or OK to the user, and errors (other cases) are handled somewhat transparently, if that's the case, in that error handling, you need to be sure to discard, because no one is going to get to those responses?

ray-roestenburg-da: I think this unhandled error stuff is handled somewhere in retryProvider in `case ex: BaseAppConnection.UnexpectedHttpResponse =>` following runHttpCmd in BaseAppConnection which calls the submitRequest.. but there the entity bytes is never discarded. (and actually the guardrails code does do discardEntityBytes in other cases, for instance  (random example)

```
  def updateAmuletPriceVote(body: _root_.org.lfdecentralizedtrust.splice.http.v0.definitions.UpdateAmuletPriceVoteRequest, headers: List[HttpHeader] = Nil): EitherT[Future, Either[Throwable, HttpResponse], UpdateAmuletPriceVoteResponse] = {
    val allHeaders = headers ++ scala.collection.immutable.Seq[Option[HttpHeader]]().flatten
    makeRequest(HttpMethods.PUT, host + basePath + "/v0/admin/sv/amulet-price/vote", allHeaders, body, HttpProtocols.`HTTP/1.1`).flatMap(req => EitherT(httpClient(req).flatMap(resp => resp.status match {
      case StatusCodes.OK =>
        resp.discardEntityBytes().future.map(_ => Right(UpdateAmuletPriceVoteResponse.OK))
```

isegall-da: But in that example, it discards the entity for the case where it handles the response locally (specifically, for the StatusCodes.OK case). Also there, for the "else" case, it just returns `Left(Right(resp))`, and leaves it to the caller to either read or discard the entity:
```
      case _ =>
        FastFuture.successful(Left(Right(resp)))
```

Or maybe I misunderstood you, and your comment was not about the `case _` case?

isegall-da: In other words, in what case am I not reading the response, nor allowing someone else to do that?

In the OK and NotFound case, I am reading the response fully. In the `_` case, I'm returning the full response to the caller, so they can read it fully (or discard).

ray-roestenburg-da: you always need to read all of the entity or discard it (which is like reading it but doing nothing with it) so that example shows when guardrails only returns on OK object further down it discards the entity because no one else gets access to it (the OK object doesn't have it so the caller gets bo access to it)

isegall-da: I understand that example, what I don't get is where you think that is missing in my client.
As above: In the OK and NotFound case, I am reading the response fully. In the `_` case, I'm returning the full response to the caller, so they can read it fully (or discard). 
And this is very aligned with all other generated clients.

isegall-da: So the "else" case (the `case _`) basically says "I don't know this response, I don't know what to do with it, so here's the full response, do whatever you want with it". I don't see how it's correct to discard anything in this case.
And again, that's what all our client code does for this case.

ray-roestenburg-da: The whole pipeline is a bit complicated. I thought about discardEntityBytes cause it's a common problem with streaming, and I don't see it handled in the not Ok and not NotFound case, please check. 
So I got Claude to double-check:

```
There was one entity leak — in the ScanStreamClient.bulkStorageDownload catch-all case _ branch (line 109). When an unexpected HTTP status code (not 200 or 404) is received, the HttpResponse was returned as Left(Right(resp)) without discarding the entity bytes. In Pekko HTTP, unconsumed response entities hold onto the connection from the pool, eventually leading to connection starvation.
```

```
Now I can see the full chain:
ScanStreamClient.bulkStorageDownload returns EitherT[Future, Either[Throwable, HttpResponse], BulkStorageDownloadResponse]
The Left(Right(resp)) means: Left side of the EitherT, containing Right(resp) (i.e., Either[Throwable, HttpResponse] with the Right case being the HttpResponse).
In HttpCtlRunner.run (line 45-50), the .leftMap handles this. resp is Either[Throwable, HttpResponse], and it matches case err => err.toString() — which is Right(resp).toString().
So: No, it is not discarded. The leftMap just calls .toString() on the Right(HttpResponse(...)), which converts it to a string error message. The HttpResponse entity (the streaming body) is never consumed or discarded. It just gets garbage collected with an unconsumed stream, which in Pekko HTTP means the underlying connection is not returned to the pool until it times out.
So the discardBytes() call I added is correct and necessary — it must happen at the ScanStreamClient level before the response is passed back, because nothing upstream takes care of it.
```

ray-roestenburg-da: It might even be that we don't handle this correctly in other cases.. and we have seen some strange issues with requests timing out recently, so maybe there is a leak in here somewhere

ray-roestenburg-da: I would not assume that it is right everywhere else

PR #4404, opened 2026-03-12
Fix automation for `optValidatorFaucetCap=0` world

martinflorian-da:
The idea is that this works even if we don't have the Daml fix in https://github.com/hyperledger-labs/splice/pull/4377

Once above PR is merged, and assuming we still have concerns that we won't be able to get the new DAR versions in before cap=0 becomes effective, I'll follow-up with a sedond PR to make the new test force an older DAR version for testing. I have a weird idea for continuing to test the "newest DAR + newest automation" case without duplicating the test; but that's for the next PR.

I also wonder whether we should set `optValidatorFaucetCap=0` on CILR sooner rather than later.

Part of #3461


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -108,6 +108,7 @@ object SvOnboardingConfig {
       initialRound: Long = 0L,
       developmentFundPercentage: Option[BigDecimal] = None,
       developmentFundManager: Option[PartyId] = None,
+      optValidatorFaucetCap: Option[BigDecimal] = None,
```
martinflorian-da, line 111: Yes we're adding a config option just for a test. The only viable alternative I'm seeing is to make the test slower by having a voting step in there to change this value.

moritzkiefer-da: that seems fine, I think really what we want is just an option to specifyf all of AmuletConfig initially to avoid having to add one option for each stupid field here. But that's definitely out of scope of this PR

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorFaucetCapZeroTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,95 @@
...
+
+        val openRounds = sv1ScanBackend
+          .getOpenAndIssuingMiningRounds()
+          ._1
+          .filter(_.payload.opensAt.isBefore(getLedgerTime.toInstant))
+        openRounds should not be empty
+
+        val targetRound = openRounds.minBy(_.payload.round.number)
+
+        aliceValidatorBackend.participantClientWithAdminToken.ledger_api_extensions.commands
```
moritzkiefer-da: use actAndCheck

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorFaucetCapZeroTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,95 @@
...
+    extends IntegrationTestWithIsolatedEnvironment
+    with WalletTestUtil
+    with TimeTestUtil {
+
+  override def environmentDefinition: SpliceEnvironmentDefinition =
+    EnvironmentDefinition
+      .simpleTopology1SvWithSimTime(this.getClass.getSimpleName)
+      .addConfigTransform((_, config) => ConfigTransforms.withValidatorFaucetCap(0)(config))
+
+  "system works with optValidatorFaucetCap=0 and handles stale liveness records" in {
```
moritzkiefer-da, line 29: can we check that the `lastActiveAt` update and therefore also the version number update still works? Maybe the existing test for that is also already sufficient, leaving up to your judgement

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/automation/ReceiveFaucetCouponTrigger.scala
```diff
@@ -74,6 +74,16 @@ class ReceiveFaucetCouponTrigger(
           )
           .minByOption(_.contract.payload.opensAt)
       )
+      // Skip if validator faucet cap is zero — no rewards to distribute
+      _ <- OptionT.fromOption[Future] {
+        val cap = firstOpenNotClaimed.payload.issuanceConfig.optValidatorFaucetCap.toScala
+        if (cap.exists(_.compareTo(java.math.BigDecimal.ZERO) <= 0)) {
+          logger.info(
```
moritzkiefer-da: I wonder if we should just pause the trigger after seeing this once so it doesn't log it every 10min. Probably not worth it and we just remove it at some point?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorFaucetCapZeroTimeBasedIntegrationTest.scala
martinflorian-da: Yeah I was thinking about this as well, but it makes the test a bit more complex. Maybe better to add, let me try it and see how it looks in the end.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/automation/ReceiveFaucetCouponTrigger.scala
martinflorian-da: > I wonder if we should just pause the trigger after seeing this once so it doesn't log it every 10min.

Ha, that didn't even occur to me... I was debating whether it should be debug, but it does seem more than that.

IDK let me try it? If it's easy why not, if it's too annoying/adds too much extra complexity IMO validators can live with an extra log line every 10min... Maybe it's even a feature, because people sometimes only read the last 10min of log anyway :see_no_evil:

martinflorian-da: So we now print the info log only once. It's a tiny bit hacky but surely we'll just kill this trigger soon, so I hope I'll be forgiven.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ValidatorFaucetCapZeroTimeBasedIntegrationTest.scala
martinflorian-da: added, don't know why I thought this will add too much bloat

PR #4416, opened 2026-03-13
Stop dropping http entities on unknown responses

isegall-da:
The contract with the guardrail-generated client is that if it receives an unknown response, it passes that response as-is to the caller, and it's the caller's responsibility to handle it. That means either consuming it or discarding it, but in certain cases we do neither.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/AppConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/AppConnection.scala
@@ -43,10 +43,12 @@ import org.apache.pekko.http.scaladsl.model.{
   Uri,
 }
 import org.apache.pekko.stream.Materializer
+import org.apache.pekko.util.ByteString
 
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ExecutionContext, ExecutionContextExecutor, Future}
 import scala.util.{Failure, Success}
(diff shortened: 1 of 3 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/AppConnection.scala
```diff
@@ -68,21 +76,44 @@ abstract class BaseAppConnection(
...
   }
 }
 
 object BaseAppConnection {
-  final class UnexpectedHttpResponse(val response: HttpResponse)
-      extends Throwable(s"Unexpected Http Response: $response")
+  final class UnexpectedHttpJsonResponse(val statusCode: StatusCode, val entity: ResponseEntity)
+      extends Throwable(s"Unexpected Http Response: $statusCode: $entity")
+  final class UnexpectedHttpNonJsonResponse(val statusCode: StatusCode)
+      extends Throwable(s"Unexpected Http Response: $statusCode")
```
nicu-da, line 141: can we also handle text/plain,text/html response types at least to include the response entity? it would be useful in debugging. maybe consume only up to x bytes from the entity before discarding if possible?

isegall-da: Good idea, added.

PR #4420, opened 2026-03-16
Traffic-based App Rewards: Compute aggregateActivityTotals, add ComputeAppRewardTrigger, expose new endpoints

### LINKED ISSUE #4380: Scan app: Traffic-based App Rewards: compute aggregateActivityTotals; add access via a new endpoint

adetokunbo:
Fixes #4380

@meiersi-da: I replaced `NoOpActivityComputation` with `FakeAppActivityComputation` in 41b32f1.  The `Fake` will be replaced on this feature branch once its rebased on the nearly-done TCS changes  

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanAppRewardsComputationTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanAppRewardsComputationTimeBasedIntegrationTest.scala
@@ -53,7 +53,7 @@ class ScanAppRewardsComputationTimeBasedIntegrationTest
     }
 
     "return activity totals after aggregation trigger runs" in { implicit _env =>
-      // TODO(#4381): Update the time management here
+      // TODO(#4384): Update the time management here
       //
       // Advance enough ticks so that at least three rounds are fully closed and aggregated.
       // (advanceRoundsToNextRoundOpening advances by 1 tick)
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
@@ -25,7 +25,7 @@ import scala.concurrent.{ExecutionContext, Future}
   *   2. Compute reward totals (CC minting allowances with threshold filtering)
   *   3. Build the Merkle tree of batched reward hashes
   *
-  * TODO(#4381): use ScanRewardsReferenceStore for synchronization when computeAndStoreRewards requires it
+  * TODO(#4384): use ScanRewardsReferenceStore for synchronization when computeAndStoreRewards requires it
   */
 class RewardComputationTrigger(
     appRewardsStore: ScanAppRewardsStore,


apps/scan/src/main/openapi/scan.yaml
```diff
@@ -1617,6 +1617,49 @@ paths:
         "500":
           $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/500"
 
+  /v0/reward-accounting-process/rounds/earliest-available:
+    get:
+      tags: [internal, scan]
+      x-jvm-package: scan
+      operationId: "getRewardAccountingEarliestAvailableRound"
+      description: |
+        Return the earliest round for which CIP-0104 reward accounting activity
```
meiersi-da:
```suggestion
        SV node internal API that returns the earliest round for which CIP-0104 reward accounting activity
```

add disclaimer to all `/reward-accounting-process/` endpoints

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -3914,3 +3957,50 @@ components:
...
+        - total_app_activity_weight
+      properties:
+        app_provider_party:
+          type: string
+        app_provider_party_seq_num:
+          type: integer
+          format: int32
+        total_app_activity_weight:
+          type: integer
+          format: int64
```
meiersi-da: Not sure we need to serve this data. 

The reward minting process won't need it. I'm inclined to not add it, as w/o pagination it represents a scalability bottleneck; and we can always later expose it from a proper endpoint with pagination, if we really do need to do so. 

I'm not sure we need, as the raw data is well available.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -122,6 +123,7 @@ class HttpScanHandler(
     sequencerAdminConnection: SequencerAdminConnection,
     protected val storeWithIngestion: AppStoreWithIngestion[ScanStore],
     updateHistory: UpdateHistory,
+    appRewardsStore: DbScanAppRewardsStore,
```
meiersi-da:
```suggestion
    appRewardsStore: ScanAppRewardsStore,
```

This is a case where I believe we'd be well-served to factor out the trait defining the public functions to help us avoid accidental spaghetti-dependencies. The implementation is pretty large, and having a compact summary of what's really required from the implementation is helpful to understand it better.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2321,6 +2323,71 @@ class HttpScanHandler(
...
+                    roundNumber = roundTotal.roundNumber,
+                    totalAppActivityWeight = roundTotal.totalRoundAppActivityWeight,
+                    activePartiesCount = roundTotal.activeAppProviderPartiesCount,
+                    partyTotals = partyTotals.map { pt =>
+                      definitions.RewardAccountingPartyActivityTotal(
+                        appProviderParty = pt.appProviderParty,
+                        appProviderPartySeqNum = pt.appProviderPartySeqNum,
+                        totalAppActivityWeight = pt.totalAppActivityWeight,
+                      )
+                    }.toVector,
```
meiersi-da: Yeah, we really want to avoid retrieving the party totals when retrieving round totals. We need the round totals to compute the issuance per featured app CC burn, and we need to be able to retrieve these totals with constant memory.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -449,6 +449,140 @@ class DbScanAppRewardsStore(
...
+  }
+
+  /** Aggregate per-party and per-round activity totals for the given round from
+    * `app_activity_record_store`, restricted to the supplied featured app parties.
+    * Both inserts use ON CONFLICT DO NOTHING for idempotency.
+    */
+  def aggregateActivityTotals(
+      historyId: Long,
+      roundNumber: Long,
+      featuredAppParties: Seq[String],
```
meiersi-da: This should not be required. Only featured app parties get activity records. So just summing up the records in the DB is enough. No extra filtering required.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,76 @@
...
+  * feature branch.
+  *
+  * For now, when an ACS-based implementation is used.
+  */
+trait FeaturedAppRightProvider {
+  def getFeaturedParties(roundNumber: Long)(implicit
+      tc: TraceContext
+  ): Future[Seq[String]]
+}
+
```
meiersi-da:
```suggestion

```

not required

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,76 @@
...
+  * triggering
+  */
+class RewardComputationTrigger(
+    store: ScanStore,
+    appRewardsStore: DbScanAppRewardsStore,
+    updateHistory: UpdateHistory,
+    featuredAppRightProvider: FeaturedAppRightProvider,
+    override protected val context: TriggerContext,
+)(implicit val ec: ExecutionContext, val tracer: Tracer)
+    extends PollingTrigger {
```
meiersi-da: Plain polling triggers are the most difficult to implement.

Please moving to a `PollingParallelTaskTrigger` and make it retrieve the rounds to compute rewards for as tasks.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def performWorkIfAvailable()(implicit traceContext: TraceContext): Future[Boolean] = {
+    if (!updateHistory.isReady) {
+      Future.successful(false)
+    } else {
+      val historyId = updateHistory.historyId
```
meiersi-da: That looks like having a `updateHistory.getHistoryId : Future[HistoryId]` would nicely serve this pattern.

Also: can you check what `updateHistory.isReady` blocks on? If this blocks on backfilling having completed, then we have a problem, as that can take days.

We want round computation to work as soon as possible after onboarding an SV.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,76 @@
...
+      for {
+        lastClosedO <- store.lookupRoundOfLatestData()
```
meiersi-da: I'm wary that this is slightly off. I think we are better off driving the trigger only from data in the `AppRewardsReferenceStore` and the `AppRewardsStore`.

So what I'd expect is that we drive the trigger based off of the `OpenMiningRound` contracts that have been observed to be archived in the `AppRewardsReferenceStore`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -449,6 +449,140 @@ class DbScanAppRewardsStore(
...
+           )
+           insert into #${Tables.appActivityPartyTotals}
+             (history_id,
+              round_number,
+              total_app_activity_weight,
+              app_provider_party_seq_num,
+              app_provider_party)
+           select $historyId, $roundNumber, total_weight, seq_num, party
+           from numbered
+           on conflict do nothing""").asUpdate
```
meiersi-da: let's throw an error on conflict. It should not happen, and if it does, we want to see it in the logs. Furthermore, when it happens then we'll have our trigger infrastructure retry.

meiersi-da: here and everywhere else

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -449,6 +449,140 @@ class DbScanAppRewardsStore(
...
+  private def insertEmptyRoundTotal(
+      historyId: Long,
+      roundNumber: Long,
+  ) = {
+    sqlu"""insert into #${Tables.appActivityRoundTotals}
+             (history_id,
+             round_number,
+             total_round_app_activity_weight,
+             active_app_provider_parties_count)
+           values ($historyId, $roundNumber, 0, 0)
```
meiersi-da: What do we use this for? If we use it as a marker, then this sounds risky, as there can be rounds without any rewards.

adetokunbo: Fixed.  It became unnecessary once the unnecessary filtering by featured apps was removed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: `updateHistory.isReady` blocks on `ingestionSink.initialize()`, which assigns the `historyId`, but which does not itself do back-filling.  There are separate triggers responsible for back-filling different entities.

Changed this to use `updateHistory.waitUntilInitialized`, which returns a `Future[Unit]` once `isReady` is complete

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed: along with the removal of FeaturedAppProvider

apps/scan/src/main/openapi/scan.yaml
adetokunbo: Removed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
adetokunbo: Factored out.  

The new interface exposes two functions:
`getNextRoundWithoutRootHash` and `computeRewards`.

`computeRewards` currently just forwards to 'aggregateActivityTotals', 
it will do more as more PRs ared add to the feature branch, likely with signature changes

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
adetokunbo: Removed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Switched to use PollingParallelTaskTrigger

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Removed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: I updated this to reflect what we discussed at the last meeting.  the synchronization logic is incomplete; it still uses ScanStore to obtain a round instead of obtaining multiple rounds from ScanRewardsReferenceStore; there is TODO to correct this. which will be addressed after #4322 is merged into this branch

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanAppRewardsStore.scala
```diff
@@ -0,0 +1,29 @@
...
+
+/** Store interface for the CIP-0104 reward computation pipeline.
+  * Decouples RewardComputationTrigger from the DB implementation.
+  */
+trait ScanAppRewardsStore {
+
+  /** Finds the next round requiring reward computation — one that has
+    * ingested activity records but has not yet produced a root hash.
+    */
+  def getNextRoundWithoutRootHash(historyId: Long, lastClosedRound: Long)(implicit
```
meiersi-da:
```suggestion
  def getNextRoundWithoutRootHash(lastClosedRound: Long)(implicit
```

historyId is an implementation detail. In one instance of the `ScanApp` you never need to use multiple history-ids.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanAppRewardsStore.scala
```diff
@@ -0,0 +1,29 @@
...
+      tc: TraceContext
+  ): Future[Option[Long]]
+
+  /** Runs the full reward computation pipeline for a single round:
+    * aggregation, CC conversion, and Merkle tree hashing.
+    * Raises on duplicate key if already computed.
+    */
+  def computeRewards(historyId: Long, roundNumber: Long)(implicit
```
meiersi-da:
```suggestion
  def computeRewards(roundNumber: Long)(implicit
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,87 @@
...
+import org.lfdecentralizedtrust.splice.scan.store.{AppActivityStore, ScanAppRewardsStore, ScanStore}
+import org.lfdecentralizedtrust.splice.store.UpdateHistory
+import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
+import com.digitalasset.canton.tracing.TraceContext
+import io.opentelemetry.api.trace.Tracer
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/** Trigger that drives the CIP-0104 reward computation pipeline through three
+  * idempotent computation steps:
```
meiersi-da: I wonder whether we shouldn't just do all of them in a single slightly longer-running transaction, and thus get a very simple idempotency argument. wdyt?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,87 @@
...
+    override val tracer: Tracer,
+    mat: Materializer,
+) extends PollingParallelTaskExecutionTrigger[RewardComputationTrigger.Task] {
+
+  override def retrieveTasks()(implicit
+      tc: TraceContext
+  ): Future[Seq[RewardComputationTrigger.Task]] = {
+    for {
+      _ <- updateHistory.waitUntilInitialized
+      lastClosedO <- store.lookupRoundOfLatestData()
```
meiersi-da:
```suggestion
      // TODO(#4118): replace this approximation with proper retrieval from the ScanRewardsReferenceStore
      lastClosedO <- store.lookupRoundOfLatestData()
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,87 @@
...
+      earliestCompleteO <- appActivityStore.earliestRoundWithCompleteAppActivity()
+      tasks <- (lastClosedO, earliestCompleteO) match {
+        case (Some((lastClosed, _)), Some(earliestComplete)) =>
+          appRewardsStore
+            .getNextRoundWithoutRootHash(updateHistory.historyId, lastClosed)
```
meiersi-da: This feels off as a function. It seems to do too much. 

Why not query the latest round for which the computation completed; and also query the latest round with complete data. Then the tasks available are all the round from `max(earliestComplete, latestRoundWithRootHash + 1)` to `min(lastClosed, latestComplete)` (with some adjustments for the optionals).

meiersi-da: You can then just pick the first 4 round numbers from this interval.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,87 @@
...
+        case _ => Future.successful(Seq.empty)
+      }
+    } yield tasks
+  }
+
+  override protected def completeTask(
+      task: RewardComputationTrigger.Task
+  )(implicit tc: TraceContext): Future[TaskOutcome] =
+    appRewardsStore
+      .computeRewards(updateHistory.historyId, task.roundNumber)
```
meiersi-da:
```suggestion
      .computeRewards(task.roundNumber)
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
```diff
@@ -65,21 +67,59 @@ class DbAppActivityRecordStore(
       )
   }
 
+  /** Check whether app activity ingestion is complete for a round.
+    * Returns true if activity records exist for the given round AND for
+    * any later round (proving ingestion has moved past round N, since
+    * verdicts are ingested in record_time order).
+    */
```
meiersi-da: That's not sufficient. There could be a prefix that is missing. We need to see activity records for the prior round.

meiersi-da: Actually this this hints at an alternative query:
```
lookupBoundsOfRoundsWithCompleteActivity(): Future[Option[(Int, Int)]]
```
and just return the `min(round) + 1` and `max(round) -1`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -449,6 +451,123 @@ class DbScanAppRewardsStore(
     )
   }
 
+  // -- Aggregation ------------------------------------------------------------
+
+  /** Find the earliest closed round that has activity records but no root hash.
+    * Returns None if all rounds with activity already have root hashes,
+    * or if there are no activity records at all.
+    */
```
meiersi-da: As stated in another comment: let's focus this function to just deliver the latest round for which reward computation was done:
```
lookupLatestRoundWithRewardComputation(): Future[Option[Long]]
```

also note the convention that we have in the codebase: `lookup` prefix for functions returning `Option` and `get` prefix for functions throwing an exception on not found.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -449,6 +451,123 @@ class DbScanAppRewardsStore(
...
+            where not exists (
+              select 1 from #${Tables.appRewardRootHashes} rh
+              where rh.history_id = $historyId and rh.round_number = aar.round_number
+            )
+      """.as[Option[Long]].headOption.map(_.flatten),
+      "appRewards.getNextRoundWithoutRootHash",
+    )
+  }
+
+  def getEarliestActivityRound(historyId: Long)(implicit
```
meiersi-da:
```suggestion
  def getEarliestRoundWithActivityTotals(historyId: Long)(implicit
```

let's make the names telling given that there are quite a few variants of queries releated to activity records that we use

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -449,6 +451,123 @@ class DbScanAppRewardsStore(
...
+              round_number,
+              total_round_app_activity_weight,
+              active_app_provider_parties_count)
+           select $historyId,
+                  $roundNumber,
+                  coalesce(sum(total_app_activity_weight), 0),
+                  count(*)
+           from #${Tables.appActivityPartyTotals}
+           where
+             history_id = $historyId and round_number = $roundNumber"""
```
meiersi-da: I'd have expected that we combine this with the `INSERT` of app_party_activity_total using `RETURNING`. This way we do not have to make another pass over the inserted data.

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -1617,6 +1617,49 @@ paths:
...
           $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/500"
 
+  /v0/reward-accounting-process/rounds/earliest-available:
+    get:
+      tags: [internal, scan]
+      x-jvm-package: scan
+      operationId: "getRewardAccountingEarliestAvailableRound"
+      description: |
+        SV node internal API that returns the earliest round for which CIP-0104 reward accounting activity
+        totals have been computed.
```
meiersi-da:
```suggestion
        totals are available.
```

once we do pruning, they will have been computed, but are no longer available.

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -1617,6 +1617,49 @@ paths:
...
+        "404":
+          $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/404"
+
+  /v0/reward-accounting-process/rounds/{round_number}/activity-totals:
+    get:
+      tags: [internal, scan]
+      x-jvm-package: scan
+      operationId: "getRewardAccountingActivityTotals"
+      description: |
+        Return the CIP-0104 per-round and per-party activity totals for the
```
meiersi-da:
```suggestion
        Return the CIP-0104 per-round activity totals for the
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
meiersi-da: Missing: apply this suggestion

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2321,6 +2323,62 @@ class HttpScanHandler(
...
+    implicit val tc = extracted
+    withSpan(s"$workflowId.getRewardAccountingActivityTotals") { _ => _ =>
+      for {
+        roundTotalO <- appRewardsStore.getAppActivityRoundTotalByRound(
+          updateHistory.historyId,
+          roundNumber,
+        )
+        result <- roundTotalO match {
+          case None =>
+            Future.successful(
```
meiersi-da:
```suggestion
        yield (roundTotalO match {
          case None =>
```

no need to use the futures for a pure computation

meiersi-da: or just use `.map` as above

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanAppRewardsComputationTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,90 @@
...
+      // Run the reward computation trigger
+      sv1ScanBackend.automation
+        .trigger[RewardComputationTrigger]
+        .runOnce()
+        .futureValue
+
+      val earliest = clue("Verify earliest available round is returned") {
+        val e = sv1ScanBackend.getRewardAccountingEarliestAvailableRound()
+        e should not be empty
+        e.value
```
meiersi-da, line 78: Given that you are starting fresh this should be Round 1 or not?

Round 0 will be the one used to prove that activity records are complete for Round 1.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanAppRewardsComputationTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,90 @@
...
+      }
+      clue("activity-totals returns None for round 0") {
+        sv1ScanBackend.getRewardAccountingActivityTotals(0L) shouldBe None
+      }
+    }
+
+    "return activity totals after aggregation trigger runs" in { implicit _env =>
+      // Advance enough rounds so that at least one round is fully closed and aggregated.
+      // Round 0 requires multiple ticks to close.
+      for (_ <- 1 to 7) {
```
meiersi-da, line 60: Why do you need 7 rounds? 

I'd expect that closing rounds 0 and 1 should be sufficient; i.e, you need to advance rounds two times.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanAppRewardsComputationTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,90 @@
...
+
+      clue("Verify activity totals for the aggregated round") {
+        val totals = sv1ScanBackend.getRewardAccountingActivityTotals(earliest)
+        totals should not be empty
+        totals.value.roundNumber shouldBe earliest
+      }
+
+      val (lastRound, _) = sv1ScanBackend.getRoundOfLatestData()
+      clue("Verify 404 for non-existent round") {
+        sv1ScanBackend.getRewardAccountingActivityTotals(lastRound + 1000L) shouldBe None
```
meiersi-da: Can you be more precise: shouldn't we have the `earliest + 1` does not have data?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
meiersi-da: AFAIS, that's what you plan to do. The comment though reads as if these would be separate steps and each of which is made idempotent on its own.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
adetokunbo: Fixed, used .map

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Fixed the comment

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanAppRewardsStore.scala
adetokunbo: Corrected

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanAppRewardsStore.scala
adetokunbo: Nice - correcting this means that DbScanAppRewardsStore gets passed an UpdateHistory in the constructor, which has initialize() called already in ScanApp, so the trigger no longer needs to reference it at all.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanAppRewardsComputationTimeBasedIntegrationTest.scala
adetokunbo: Added the assertion

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanAppRewardsComputationTimeBasedIntegrationTest.scala
adetokunbo: Changed the comment, that function increments by ticks not rounds.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanAppRewardsComputationTimeBasedIntegrationTest.scala
adetokunbo: Done

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Fixed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Changed to lookupLatestRoundWithRewardComputation with the given signature

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Done

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AppActivityStore.scala
```diff
@@ -0,0 +1,32 @@
...
+    * proving ingestion was running continuously through it.
+    * Returns None if fewer than two consecutive rounds have been ingested.
+    */
+  def earliestRoundWithCompleteAppActivity()(implicit
+      tc: TraceContext
+  ): Future[Option[Long]]
+
+  /** Find the latest round with complete app activity.
+    * A round is complete if the prior round also has activity records,
+    * proving ingestion was running continuously through it.
```
meiersi-da: copy-pasta

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AppActivityStore.scala
```diff
@@ -0,0 +1,32 @@
...
+
+/** Store interface for app activity record queries.
+  * Decouples callers from the DB implementation.
+  */
+trait AppActivityStore {
+
+  /** Find the earliest round with complete app activity.
+    * A round is complete if the prior round also has activity records,
+    * proving ingestion was running continuously through it.
+    * Returns None if fewer than two consecutive rounds have been ingested.
```
meiersi-da: This actually leaks the implementation.
```suggestion
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AppActivityStore.scala
```diff
@@ -0,0 +1,32 @@
...
+import com.digitalasset.canton.tracing.TraceContext
+
+import scala.concurrent.Future
+
+/** Store interface for app activity record queries.
+  * Decouples callers from the DB implementation.
+  */
+trait AppActivityStore {
+
+  /** Find the earliest round with complete app activity.
```
meiersi-da:
```suggestion
  /** Find the earliest round for which all app activity records have been ingested.
```

meiersi-da: apply below as well

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanAppRewardsStore.scala
```diff
@@ -0,0 +1,29 @@
...
+  /** Returns the latest round for which reward computation has completed
+    * (i.e. a root hash exists). None if no rounds have been computed.
+    */
+  def lookupLatestRoundWithRewardComputation()(implicit
+      tc: TraceContext
+  ): Future[Option[Long]]
+
+  /** Runs the full reward computation pipeline for a single round:
+    * aggregation, CC conversion, and Merkle tree hashing.
+    * Raises on duplicate key if already computed.
```
meiersi-da:
```suggestion
    * MUST only be called on rounds for which all app activity records have
    * been ingested and for which the reward information has not yet been computed.
    
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
```diff
@@ -65,21 +67,57 @@ class DbAppActivityRecordStore(
...
+  def latestRoundWithCompleteAppActivity()(implicit
+      tc: TraceContext
+  ): Future[Option[Long]] = {
+    runQuerySingle(
+      sql"""select max(aar.round_number)
+            from #${Tables.appActivityRecords} aar
+            where exists(
+              select 1 from #${Tables.appActivityRecords}
+              where round_number = aar.round_number - 1
```
meiersi-da: Check the query plan for this. This might be expensive to do that way.

If you do want to check the existence, then first query for the max and do a join against that single value, instead of potentially forcing an expensive join.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -449,6 +459,115 @@ class DbScanAppRewardsStore(
...
+  }
+
+  /** Unnest per-verdict activity arrays and aggregate weights by party. */
+  private def unnestAndAggregate(roundNumber: Long) =
+    sql"""unnested as (
+            select party, weight
+            from app_activity_record_store,
+                 lateral unnest(app_provider_parties, app_activity_weights)
+                 as party_and_weight(party, weight)
+            where round_number = $roundNumber
```
meiersi-da: you are missing the `where history_id = $historyId` constraint, which means you'd accidentally aggregate data from other store instances.

Please add a test that shows that this is missing.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,86 @@
...
+      earliestCompleteO <- appActivityStore.earliestRoundWithCompleteAppActivity()
+      latestCompleteO <- appActivityStore.latestRoundWithCompleteAppActivity()
+      latestComputedO <- appRewardsStore.lookupLatestRoundWithRewardComputation()
+    } yield {
+      (lastClosedO, earliestCompleteO, latestCompleteO) match {
+        case (Some((lastClosed, _)), Some(earliestComplete), Some(latestComplete)) =>
+          val start = math.max(earliestComplete, latestComputedO.fold(0L)(_ + 1))
+          val end = math.min(lastClosed, latestComplete)
+          (start to end)
+            .take(RewardComputationTrigger.MaxParallelRounds)
```
meiersi-da: change to return at most the next round that has not yet been computed.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,86 @@
...
+      task: RewardComputationTrigger.Task
+  )(implicit tc: TraceContext): Future[TaskOutcome] =
+    appRewardsStore
+      .computeRewards(task.roundNumber)
+      .map(_ => TaskSuccess(s"Computed rewards for round ${task.roundNumber}"))
+
+  override protected def isStaleTask(
+      task: RewardComputationTrigger.Task
+  )(implicit tc: TraceContext): Future[Boolean] =
+    Future.successful(false)
```
meiersi-da: It is stale if the latest completed round is larger than the task round. Should not happen, but seems like the better definition.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  override def retrieveTasks()(implicit
+      tc: TraceContext
+  ): Future[Seq[RewardComputationTrigger.Task]] = {
+    for {
+      // TODO(#4118): replace this approximation with proper retrieval from the ScanRewardsReferenceStore
+      lastClosedO <- store.lookupRoundOfLatestData()
+      earliestCompleteO <- appActivityStore.earliestRoundWithCompleteAppActivity()
+      latestCompleteO <- appActivityStore.latestRoundWithCompleteAppActivity()
+      latestComputedO <- appRewardsStore.lookupLatestRoundWithRewardComputation()
```
meiersi-da: This actually turns out to be wrong when the trigger executes tasks in parallel and the backend gets restarted: round 2 might complete before round 1, and thus round 1 will never be completed.

So what I'd suggest to do for now is:
1. return at most one task and thus ensure that they get solved one after the other
2. create a TODO for an issue to compute tasks in parallel, which will require introducing a watermark table to remember the round number up to which all rewards have been computed. We might not need that. It will depend on how fast catchup works for the round trigger.

FYI: @rautenrieth-da -- catchup speed of reward computation is something to add to the test plan

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,86 @@
...
+
+/** Trigger that drives the CIP-0104 reward computation pipeline via
+  * ScanAppRewardsStore.computeRewards, which will eventually run three
+  * idempotent steps in one transaction:
+  *   1. Aggregate activity totals from app activity records
+  *   2. Compute reward totals (CC minting allowances with threshold filtering)
+  *   3. Build the Merkle tree of batched reward hashes
+  *
+  * TODO(#4384): Add CalculateRewardsV2 / ProcessRewardsV2 contract that gates
+  * triggering
```
meiersi-da: I'm no longer sure we want to do this. It complicates synchronization and makes reasoning about what data is available more difficult.

I'd suggest we first attempt to build a solution that has all totals from the earliest round it started computing totals.

```suggestion
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
meiersi-da: something along the lines of
```sql
select max_round
from
  (select max(round) - 1 
   from app_activity_records
   where history_id = $historyId
  )
where exists (
  select 1 
  from app_activity_records
  where history_id = $history_id and round = max_round
)
```

also: please add tests that surface the problem wrt missing check of history_id, and check all your queries wrt the missing comparison of `history_id`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -449,6 +459,115 @@ class DbScanAppRewardsStore(
...
+
+  /** Runs the full reward computation pipeline for a single round.
+    *
+    * TODO(#4384): Will be extended to run CC conversion (stage 2) and
+    * Merkle tree hashing (stage 3) in a single transaction.
+    */
+  def computeRewards(
+      roundNumber: Long
+  )(implicit tc: TraceContext): Future[Unit] =
+    aggregateActivityTotals(roundNumber)
```
meiersi-da, line 490: prudent engineering: add safety check that all activity records have been ingested for the round in question. Better safe than sorry.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2321,6 +2323,52 @@ class HttpScanHandler(
...
   }
+
+  def getRewardAccountingEarliestAvailableRound(
+      respond: ScanResource.GetRewardAccountingEarliestAvailableRoundResponse.type
+  )()(extracted: TraceContext): Future[
+    ScanResource.GetRewardAccountingEarliestAvailableRoundResponse
+  ] = {
+    implicit val tc = extracted
+    withSpan(s"$workflowId.getRewardAccountingEarliestAvailableRound") { _ => _ =>
+      appRewardsStore.getEarliestRoundWithActivityTotals().map {
```
meiersi-da: AFAIR, the goal of the HTTP function is to enable other SVs to determine whether the earliest round for which this SV will eventually have the activity totals  available. So we should return the earliest round for which there is a complete set of activity records and not the earliest one for which activity totals have been computed.

meiersi-da: I suspect we actually have no need to retrieve the earliest round with activity totals, so we can remove that function from the `ScanAppRewardsStore`.

adetokunbo: Fixed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed: added an assertion; verified with a store test

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Done.  Opened #4570

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Implemented this.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
adetokunbo: Fixed; added, verifying with store tests that would have otherwise failed due to the missing historyId

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AppActivityStore.scala
adetokunbo: Removed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed, along with a verifying test

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Fixed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanAppRewardsComputationTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,90 @@
...
+      clue("earliest-available returns None") {
+        sv1ScanBackend.getRewardAccountingEarliestAvailableRound() shouldBe None
+      }
+      clue("activity-totals returns None for round 0") {
+        sv1ScanBackend.getRewardAccountingActivityTotals(0L) shouldBe None
+      }
+    }
+
+    "return activity totals after aggregation trigger runs" in { implicit _env =>
+      // TODO(#4118): Update the time management here
```
meiersi-da: This issue is closed: https://github.com/hyperledger-labs/splice/issues/4118

Seems like the wrong reference.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2367,6 +2371,61 @@ class HttpScanHandler(
...
+  )(roundNumber: Long)(extracted: TraceContext): Future[
+    ScanResource.GetRewardAccountingActivityTotalsResponse
+  ] = {
+    implicit val tc = extracted
+    withSpan(s"$workflowId.getRewardAccountingActivityTotals") { _ => _ =>
+      appRewardsStore.getAppActivityRoundTotalByRound(roundNumber).map {
+        case None =>
+          ScanResource.GetRewardAccountingActivityTotalsResponse.NotFound(
+            ErrorResponse(
+              s"Activity totals not yet computed for round $roundNumber"
```
meiersi-da:
```suggestion
              s"Activity totals not (yet) computed for round $roundNumber"
```
it might be never

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,77 @@
...
+import org.lfdecentralizedtrust.splice.scan.store.{AppActivityStore, ScanAppRewardsStore}
+import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
+import com.digitalasset.canton.tracing.TraceContext
+import io.opentelemetry.api.trace.Tracer
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/** Trigger that drives the CIP-0104 reward computation pipeline via
+  * ScanAppRewardsStore.computeRewards, which will eventually run three
+  * idempotent steps in one transaction:
```
meiersi-da:
```suggestion
  * computation steps in one transaction:
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AppActivityStore.scala
```diff
@@ -0,0 +1,27 @@
...
+trait AppActivityStore {
+
+  /** Find the earliest round for which all app activity records have been ingested.
+    */
+  def earliestRoundWithCompleteAppActivity()(implicit
+      tc: TraceContext
+  ): Future[Option[Long]]
+
+  /** Find the latest round for which all app activity records have been ingested.
+    * Returns None if fewer than two consecutive rounds have been ingested.
```
meiersi-da:
```suggestion
```
Let's not leak implementation details on the API. We don't want callers to rely on it, and break when the implementation changes.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanAppRewardsStore.scala
```diff
@@ -0,0 +1,30 @@
...
+  def lookupLatestRoundWithRewardComputation()(implicit
+      tc: TraceContext
+  ): Future[Option[Long]]
+
+  /** Runs the full reward computation pipeline for a single round:
+    * aggregation, CC conversion, and Merkle tree hashing.
+    * MUST only be called on rounds for which all app activity records have
+    * been ingested and for which the reward information has not yet been computed.
+    */
+  def computeRewards(roundNumber: Long)(implicit
```
meiersi-da:
```suggestion
  def computeAndStoreRewards(roundNumber: Long)(implicit
```

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -3987,3 +4032,28 @@ components:
...
+        earliest_round:
+          type: integer
+          format: int64
+
+    GetRewardAccountingActivityTotalsResponse:
+      type: object
+      required:
+        - round_number
+        - total_app_activity_weight
+        - active_parties_count
```
meiersi-da, line 4049: @rautenrieth-da :  wdyt about also exposing the number of activity records in the round? Seems like very useful debugging info.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -449,6 +460,131 @@ class DbScanAppRewardsStore(
...
```
meiersi-da, line 490: @rautenrieth-da : wdyt about adding counters for:
- number of parties with activity
- number of activity records summarized
- number of parties with rewards
- number of batches created

tagged with synchronizer-id and migration-id.

It seems that having dashboards for these would seem valuable to observe the proper working of reward computation. However, I'm lacking the ops experience to judge their usefulness.

@adetokunbo : If we do so, then I'd still do that in a separate PR. Just create an issue from this comment.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -449,6 +460,131 @@ class DbScanAppRewardsStore(
...
+        throw Status.FAILED_PRECONDITION
+          .withDescription(
+            s"Incomplete app activity for round $roundNumber: " +
+              s"round ${roundNumber - 1} exists=$hasPrev, round ${roundNumber + 1} exists=$hasNext"
+          )
+          .asRuntimeException()
+    } yield ()
+
+  /** Unnest per-verdict activity arrays and aggregate weights by party. */
+  private def unnestAndAggregate(historyId: Long, roundNumber: Long) =
```
meiersi-da, line 517: @rautenrieth-da : could I ask you to review these DB queries from a maintainers perspective? They look fine to me.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanAppRewardsComputationTimeBasedIntegrationTest.scala
adetokunbo: Switch to #4381

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanAppRewardsStore.scala
adetokunbo: Fixed

apps/scan/src/main/openapi/scan.yaml
rautenrieth-da: Good idea. I had a couple of cases where I needed to get debug info from a scan of another SV through some update history endpoint, so having debug info in scan endpoints is valuable even if we have metrics for the same information.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -0,0 +1,77 @@
...
+      latestCompleteO <- appActivityStore.latestRoundWithCompleteAppActivity()
+      latestComputedO <- appRewardsStore.lookupLatestRoundWithRewardComputation()
+    } yield {
+      (earliestCompleteO, latestCompleteO) match {
+        case (Some(earliestComplete), Some(latestComplete)) =>
+          val start = math.max(earliestComplete, latestComputedO.fold(0L)(_ + 1))
+          // TODO(#4570): Support parallel execution
+          Seq(start).filter(_ <= latestComplete).map(RewardComputationTrigger.Task(_))
+        case _ => Seq.empty
+      }
```
rautenrieth-da: I don't see any test coverage for this logic. I don't have a strong opinion on whether we need it, but if we do then consider moving this logic to a static method on the companion object and then adding unit tests for it. I had good experience with this approach recently in `AcsSnapshotTrigger`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
```diff
@@ -65,21 +69,81 @@ class DbAppActivityRecordStore(
       )
   }
 
+  private val verdictTable = "scan_verdict_store"
```
rautenrieth-da: Move to the `Tables` object?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -198,9 +203,10 @@ class DbScanAppRewardsStore(
     }
   }
 
-  def getAppActivityPartyTotalsByRound(historyId: Long, roundNumber: Long)(implicit
+  def getAppActivityPartyTotalsByRound(roundNumber: Long)(implicit
       tc: TraceContext
   ): Future[Seq[DbScanAppRewardsStore.AppActivityPartyTotalT]] = {
+    val historyId = updateHistory.historyId
```
rautenrieth-da: FWIW, in other stores we have a `private def historyId = updateHistory.historyId` at class level.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
rautenrieth-da: Metrics and dashboard definitely sounds like a good idea. Happy to help with creating the dashboard.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -449,6 +460,133 @@ class DbScanAppRewardsStore(
...
+      "appRewards.aggregateActivityTotals",
+    )
+  }
+
+  /** Safety check: assert that activity records exist for both roundNumber - 1
+    * and roundNumber + 1, which is the completeness condition for roundNumber.
+    * A round with no activity of its own is valid (no featured app traffic),
+    * but surrounding rounds must have data to prove ingestion has covered it.
+    */
+  private def assertCompleteActivity(roundNumber: Long) =
```
rautenrieth-da: This feels like it should be a method on `DbAppActivityRecordStore`, next to `earliestRoundWithCompleteAppActivity`. Also, is it intentional that this method only looks at `app_activity_record_store` while  the `.*RoundWithCompleteAppActivity` methods also join with the verdicts table?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
rautenrieth-da: This is similar in structure to the old query for generating ACS snapshots (a sequence of dependent CTEs, some of which insert data). The code looks fine.

I would like to run an `EXPLAIN ANALYZE` on this query to confirm that it's not doing anything stupid, or at least look at the query timing with real-world data. I don't see anything suspicious, but I also don't trust myself and I want to avoid realizing too late that we need a new index after we already have a large amount of data.

Measuring the query performance could be a bit difficult as IIRC CILR doesn't have featured app activity - we should include it in the test plan.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
```diff
@@ -65,21 +69,77 @@ class DbAppActivityRecordStore(
...
+
+  /** Find the earliest round with complete app activity.
+    * A round is complete if the prior round also has activity records,
+    * proving ingestion was running continuously through it.
+    * Returns None if fewer than two consecutive rounds have been ingested.
+    */
+  def earliestRoundWithCompleteAppActivity()(implicit
```
rautenrieth-da, line 85: I feel like I'm missing something here:
 
1. Why do we need the `where exists` part? If we see an activity record from round N, we know that activity records from round N+1 must (eventually) exist, no?
2. The `min_round + 1` seems only safe if all activity records for round N are ingested before the first record for round N+1. Is that the case? I vaguely remember that we wanted to have the option to assign activity records randomly to one of the two open overlapping mining rounds.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
adetokunbo: Fixed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Moved.
Not joining with the verdicts table was an oversight, which is now fixed

The move means `DbScanAppRewardStore` now depends on `DbAppActivityRecordStore` and is injected in constructor.  This meant a few other constructors had to change, which was useful, the resulting changes are all beneficial

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Good idea, thank you.  The logic is now extracted into the companion object and checked with a unit test, as suggested.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Added https://github.com/hyperledger-labs/splice/issues/4634

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Please see this [gist](https://gist.github.com/adetokunbo/d78c67508ce66b6a8ddb3e12329ff97a) which contains the EXPLAIN ANALYZE output from a modified store test

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
adetokunbo: IUC
1.  app activity recording for round __N__ might not yet be complete *now*
  - Ongoing ingestion may include verdicts with app activity for __N__.  
  - Once `where exists` succeeds, ingestion has started to fill __N+1__, so we can be sure __N__ is complete.  
2.  the random assignment to either open round occurs at the daml layer
  - however ledger api ingestion order for app activity should still be sequential
  - i.e, activity in round __N__ will be inserted before activity in round __N + 1__, though they might have occurred at the same time before being assigned to different rounds

adetokunbo:
@rautenrieth-da I've addressed all your comments, thanks for the review

There was also a bug that got surfaced: `lastRoundWithCompleteActivity` returned `max_round` instead of `max_round - 1` which was inconsistent with `assertCompleteActivity`.  That's fixed now.

> wdyt about also exposing the number of activity records in the round? Seems like very useful debugging info.

> > Good idea. I had a couple of cases where I needed to get debug info from a scan of another SV through some update history endpoint, so having debug info in scan endpoints is valuable even if we have metrics for the same information.

@meiersi-da  should that be added to this PR, or in one of the follow-ups ?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
rautenrieth-da: Thanks for the comment! My misunderstanding was that I thought that activity records for rounds N and N+1 can be interleaved. If that's not the case then the code makes sense. Consider describing this assumption in some comment, to help readers unfamiliar with the design.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
rautenrieth-da: Nice! That looks like it's fast enough. The query plans tend to depend on actual data, but yours was done with 10k activity records, which looks like a reasonable approximation of the expected scale.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
adetokunbo: Done: just updated the scaladoc with these assumptions, PTAL

rautenrieth-da: Looks good, thanks!

apps/scan/src/main/openapi/scan.yaml
adetokunbo: Opened #4645 to track this and added to #4384

PR #4423, opened 2026-03-16
Fix: FAILED_PRECONDITION: Node is already initialized with a different ID

pasindutennage-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4415

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
@@ -130,7 +130,7 @@ class NodeInitializer(
             Future.successful(result)
           } else {
             // No ID set yet according to GetId. Check the status endpoint to determine whether
-            // the node is genuinely waiting for an ID or still bootstrapping.
+            // the node is genuinely waiting for an ID. If not we assume that it is still bootstrapping.
             connection.getStatus.flatMap {
               case NodeStatus.NotInitialized(_, Some(WaitingForId)) =>
                 Future.successful(result)


docs/src/release_notes_upcoming.rst
```diff
@@ -23,6 +23,17 @@
          ``/v0/admin/external-party/transfer-preapproval/submit-send``
          are deprecated and will be removed in a future version. Use the token standard APIs for initiating transfers instead.
 
+       - Fix a bug that caused the Validator App to fail during restarts when the Scan Apps defined
+         in ``scanClient.seedUrls`` were unavailable. This fix ensures the Validator App uses its
+         persisted scan connections from previous runs, removing the dependency on seedUrls
+         scan availability for successful reboots.
+
+       - Fix a bug that caused the Validator App to fail with a ``FAILED_PRECONDITION`` error
```
martinflorian-da: On the fence about mentioning this; the bug never materialized in a release.

Leaning towards "let's not mention in release notes" but happy to leave the final decision to you.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
```diff
@@ -109,6 +110,7 @@ class NodeInitializer(
       nodeIdentity: UniqueIdentifier => Member & NodeIdentity,
   )(implicit tc: TraceContext, ec: ExecutionContext): Future[Unit] = {
     logger.info(s"Making sure canton node has an identity")
+    val bootDeadline = 20.seconds.fromNow
```
martinflorian-da: I thought about this some more... how about going for 10 seconds instead? Avoid slowing down some of our integration tests (to the point that thing might time out there?)

AFAIU 10s would still count as "conservative".

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
```diff
@@ -118,8 +120,20 @@ class NodeInitializer(
       nodeId <- retryProvider.retry(
         RetryFor.WaitingOnInitDependency,
         "node_id",
-        s"${connection.serviceName} answers the getId request",
-        connection.getIdOption(),
+        s"${connection.serviceName} answers the getId request with an ID or confirms uninitialized state",
+        connection.getIdOption().flatMap { result =>
+          if (result.uniqueIdentifier.isDefined || result.initialized) {
+            Future.successful(result)
+          } else if (bootDeadline.hasTimeLeft()) {
```
martinflorian-da: This is needs a comment explaining the why.

And the comment should probably be a TODO(#XXXX) comment, pointing to an issue where we asked Canton to improve the situation upstream... I'd like this to be a workaround that is at least in theory temporary :)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
pasindutennage-da: Thanks, Fixed

docs/src/release_notes_upcoming.rst
pasindutennage-da: Thanks, lets not have it

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
martinflorian-da: Ah on a second look - don't we want to change our implementation of `getIdOption` instead? Seems closer to the root of the problem, no?

pasindutennage-da: I added the status check to disambiguate between the 2 states: no id and still bootstraping.

We can of course merge this logic inside the getIdOption. But then getIdOption is more like a check for both status and getIdOption. Also getIdOption is used in 3 other instances (which I have not looked fully), so not sure we should change it there.

IMO, it should be the highler level  NodeInitializer.scala which should handle this check.

But, the cleaner solution would be to have `.isBootstrapped` method, from canton, so I would suggest to live with the current status check for now, and once we have .isBootstrapped, lets switch to a more cleaner code inside NodeINitializer.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
```diff
@@ -115,11 +115,36 @@ class NodeInitializer(
...
         "node_id",
-        s"${connection.serviceName} answers the getId request",
-        connection.getIdOption(),
+        s"${connection.serviceName} answers the getId request with an ID or confirms uninitialized state",
+        connection.getIdOption().flatMap { result =>
+          if (result.uniqueIdentifier.isDefined || result.initialized) {
+            Future.successful(result)
+          } else {
+            // No ID set yet according to GetId. Check the status endpoint to determine whether
+            // the node is genuinely waiting for an ID or still bootstrapping.
```
martinflorian-da:
```suggestion
            // the node is genuinely waiting for an ID. If not we assume that it is still bootstrapping.
```

nit

PR #4439, opened 2026-03-16
fix handling completion in ZstdGroupedWeight

isegall-da:
The added test is confirmed to explode without the fix.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
@@ -81,14 +81,16 @@ case class GroupedWeightS3ObjectFlow(
     new GraphStageLogic(shape) with InHandler with OutHandler {
       // The usage of callbacks makes this thread safe, so we use vars here and not Atomic's
       @SuppressWarnings(Array("org.wartremover.warts.Var"))
-      @volatile
       private var state = State.initial()
       @SuppressWarnings(Array("org.wartremover.warts.Var"))
-      @volatile
       private var upstreamFinished = false
 
       private def objectDone =
         upstreamFinished || state.currentObjectSize >= maxObjectSize
+
+      private val failCallback = getAsyncCallback[Throwable] { ex =>
+        failStage(ex)
+      }
 
       private val uploadCallback = getAsyncCallback[Unit] { _ =>
         state = state.completePart()
@@ -96,7 +98,7 @@ case class GroupedWeightS3ObjectFlow(
           s"Part upload completed, waiting for ${state.numPendingPartUploads} more"
         )
         if (state.numPendingPartUploads == maxParallelPartUploads - 1) {
-          if (!objectDone) {
+          if (!objectDone && !isClosed(in)) {
             logger.debug(
               "More parallel upload capacity freed up, and we're not done yet, pull more input"
             )
@@ -156,7 +158,7 @@ case class GroupedWeightS3ObjectFlow(
           )
           curState.currentObject.upload(partNumber, elem.bytes.asByteBuffer).onComplete {
             case Success(_) => uploadCallback.invoke(())
-            case Failure(ex) => failStage(ex)
+            case Failure(ex) => failCallback.invoke(ex)
           }
           if (!objectDone) {
             logger.debug(
@@ -182,12 +184,32 @@ case class GroupedWeightS3ObjectFlow(
       private def finishCurrentObject(): Unit =
         state.currentObject.finish().onComplete {
           case Success(_) => finishCallback.invoke(())
-          case Failure(ex) => failStage(ex)
+          case Failure(ex) => failCallback.invoke(ex)
         }
 
       override def onUpstreamFinish(): Unit = {
         logger.debug("upstream finished")
-        // Note that we're not finishing the stage here, we're doing that above when the last object finishes uploading
+        upstreamFinished = true
+        if (state.numPendingPartUploads == 0) {
+          if (state.currentObjectSize > 0) {
+            // All uploads completed and the current object has data — finish it.
+            logger.debug(
+              s"No pending uploads, finishing current object ${state.currentObject.key}"
+            )
+            finishCurrentObject()
+          } else {
+            // No data was written to the current object (e.g., stream was empty
+            // or upstream closed right after the previous object was finalized).
+            logger.debug("No pending uploads and no data in current object, completing stage.")
+            completeStage()
+          }
+        } else {
+          // Uploads are still in flight — the uploadCallback will trigger
+          // finishCurrentObject() when the last one completes.
+          logger.debug(
+            s"Waiting for ${state.numPendingPartUploads} pending uploads to complete before finishing object ${state.currentObject.key}"
+          )
+        }
       }
 
       setHandlers(in, out, this)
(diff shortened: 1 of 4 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
```diff
@@ -181,11 +181,7 @@ case class GroupedWeightS3ObjectFlow(
         }
 
       override def onUpstreamFinish(): Unit = {
-        if (!upstreamFinished) {
```
isegall-da, line 184: The same false assumption on my part that the finish signal is queued after all pushed objects, but that's not true, it propagates down immediately (which feels weird to me, but c'est la vie...)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
```diff
@@ -123,7 +123,7 @@ case class GroupedWeightS3ObjectFlow(
             isLastObject = upstreamFinished,
           ),
         )
-        if (upstreamFinished) {
+        if (upstreamFinished || isClosed(in)) {
           logger.debug("This was the last input in the stream, completing.")
```
ray-roestenburg-da, line 129: maybe trace?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
```diff
@@ -135,6 +135,9 @@ case class GroupedWeightS3ObjectFlow(
       override def onPush(): Unit = {
         val elem = grab(in)
         val curState = state
+        logger.debug(
+          s"Received new element of size ${elem.bytes.length} (isLast=${elem.isLast}) from upstream, current object ${curState.currentObject.key} has size ${curState.currentObjectSize} and ${curState.numPendingPartUploads} pending uploads"
```
ray-roestenburg-da, line 141: maybe also trace

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
```diff
@@ -181,11 +186,7 @@ case class GroupedWeightS3ObjectFlow(
         }
 
       override def onUpstreamFinish(): Unit = {
-        if (!upstreamFinished) {
-          throw new RuntimeException(
-            "upstream finished unexpectedly, without omitting a final object first"
-          )
-        }
+        logger.debug("upstream finished")
```
ray-roestenburg-da, line 191: maybe this was during test debugging, could also be trace level (not a big deal)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
ray-roestenburg-da: onUpStreamFinished does happen after all objects are pushed, but maybe you're looking at your internal state buffer?

ray-roestenburg-da: I would suggest looking at https://github.com/apache/pekko/blob/0d017f1fce7394d9bb836c4b0b1ab9757b481351/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala#L1737 or https://github.com/apache/pekko/blob/0d017f1fce7394d9bb836c4b0b1ab9757b481351/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala#L798 for examples. (I'll look a little deeper in the code as well)

ray-roestenburg-da: can also look at built-in MapAsync https://github.com/apache/pekko/blob/0d017f1fce7394d9bb836c4b0b1ab9757b481351/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala#L1284

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
```diff
@@ -181,11 +186,7 @@ case class GroupedWeightS3ObjectFlow(
         }
 
       override def onUpstreamFinish(): Unit = {
```
ray-roestenburg-da, line 190: If you override this you must call completeStage (the default does this) at some point. In this case only when the buffer is empty (otherwise the default completes the stage immediately). It's OK to implement as you've done now, but maybe a bit clearer to do what MapAsync does, https://github.com/apache/pekko/blob/0d017f1fce7394d9bb836c4b0b1ab9757b481351/stream/src/main/scala/org/apache/pekko/stream/impl/fusing/Ops.scala#L1345, only completing when the buffer is empty (or to add a comment,"doing nothing, the graphstage will drain itself and completeStage is called when it is drained")

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
```diff
@@ -181,11 +186,7 @@ case class GroupedWeightS3ObjectFlow(
...
-        if (!upstreamFinished) {
-          throw new RuntimeException(
-            "upstream finished unexpectedly, without omitting a final object first"
-          )
-        }
+        logger.debug("upstream finished")
         // Note that we're not finishing the stage here, we're doing that above when the last object finishes uploading
```
ray-roestenburg-da: Ah here is your comment, sorry I missed that :-) all good.
```suggestion
        // Note that we're not completing the stage here, we're doing that above when the last object finishes uploading
```

ray-roestenburg-da:
I approved too soon, sorry, I think there are a couple of improvements that still can be made with handling upstream finish and buffering (the internal state), by looking at the internal pekko operators that do very similar work.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
ray-roestenburg-da: I think I would model the stage you are working on as much as possible on MapAsync, since it basically does what you need: async calls, buffering, and draining the right way, completing the stage the right way, etc.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
ray-roestenburg-da: These volatiles are not necessary, only need the suppresswarning

ray-roestenburg-da: can't call failStage from Future onComplete on another thread, should use callback like in success case,

```
private val failCallback = getAsyncCallback[Throwable] { ex =>
  failStage(ex)
}
```
and then `failCallback.invoke(ex)` so everything is safe

isegall-da:
Superseded by #4471

PR #4205, opened 2026-03-03
Add logic for computing app-activity records using traffic-summaries

dfordivam:
Fixes: https://github.com/hyperledger-labs/splice/issues/4204

The base of this branch is still under review (https://github.com/hyperledger-labs/splice/pull/4037), so I will rebase this one over the feature branch when it gets merged.

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #4443, opened 2026-03-17
ScanApp: AppActivityRecord ingestion and serve on /v0/events

dfordivam:
Fixes: https://github.com/hyperledger-labs/splice/issues/4204 and https://github.com/hyperledger-labs/splice/issues/4442

This contains the commits from already reviewed PR https://github.com/hyperledger-labs/splice/pull/4205

- The `/v0/events` endpoint addition commits are provided by @adetokunbo 
- Removes `ScanEventHistoryExtraDataIntegrationTest`, as its functionality is now being tested in the new `TrafficBasedRewardsTimeBasedIntegrationTest`

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
```diff
@@ -63,6 +63,7 @@ case class ScanAppBackendConfig(
     mediatorVerdictIngestion: MediatorVerdictIngestionConfig = MediatorVerdictIngestionConfig(),
     sequencerTrafficIngestion: SequencerTrafficIngestionConfig = SequencerTrafficIngestionConfig(),
     serveTrafficSummaries: Boolean = false,
+    serveAppActivityRecords: Boolean = false,
```
dfordivam: This was added originally by @adetokunbo , but it seems that reusing `serveTrafficSummaries` could be ok?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/RewardsReferenceDataProvider.scala
```diff
@@ -0,0 +1,65 @@
...
+  def lookupActiveOpenMiningRounds(
+      recordTimes: Seq[CantonTimestamp]
+  )(implicit tc: TraceContext): Future[Map[CantonTimestamp, (Long, CantonTimestamp)]]
+
+  def lookupFeaturedAppPartiesAsOf(
+      asOf: CantonTimestamp
+  )(implicit tc: TraceContext): Future[Set[String]]
+}
+
+// store-backed implementation, cache will be implemented later for FAP lookup at this level.
```
dfordivam: This is the point where cache would be implemented, for `lookupFeaturedAppPartiesAsOf`

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,361 @@
...
+        activity.records.map(_.party).toSet shouldBe expectedProviders.map(_.toProtoPrimitive)
+      }
+      withClue(s"$cluePrefix each app activity weight should be positive") {
+        activity.records.foreach { r =>
+          r.weight should be > 0L
+        }
+      }
+      val weightSum = activity.records.map(_.weight).sum
+      withClue(s"$cluePrefix sum of weights should not exceed total traffic cost") {
+        weightSum should be <= totalTrafficCost
```
meiersi-da, line 247: I guess it's not equal due to rounding errors or not? It should probably be within `[totalTrafficCost - numFeaturedAppParties, totalTrafficCost]` or not?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -850,27 +851,49 @@ class HttpScanHandler(
...
+          Future.successful(
+            Left(definitions.ErrorResponse(s"Event with id $updateId not found"))
+          )
         case Some((verdictWithViewsO, updateO)) =>
-          val encodedUpdateV2 = updateO
-            .map(ScanHttpEncodings.encodeUpdate(_, encoding, ScanHttpEncodings.V1))
-            .map(toUpdateV2)
-          val verdictEncoded = verdictWithViewsO.map { case (v, views) =>
-            ScanHttpEncodings.encodeVerdict(v, views)
+          val verdictRowIdO = verdictWithViewsO.map { case (v, _) => v.rowId }
```
meiersi-da, line 869: It feels like this should ultimately be a single call to the verdict store to retrieve the verdict with activity records once we get rid of the feature flags and always turn on serving them. wdyt?

If yes, then let's create a cleanup ticket for Day 2 in the traffic-based app rewards milestone, so we don't forget about it.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictStoreIngestion.scala
```diff
@@ -169,11 +169,16 @@ class ScanVerdictStoreIngestion(
           summaryByTime.get(recordTime).map(_ -> v)
         }
 
-        // Compute app activity records (pure, before DB transaction).
+        // Compute app activity records (before DB transaction).
         // Records have verdictRowId = 0; the store resolves actual row_ids during insertion.
-        computedActivities = appActivityComputation.computeActivities(summariesWithVerdicts)
-        pendingAppActivity = computedActivities.flatMap { case (summary, _, recordO) =>
-          recordO.map(summary.sequencingTime -> _)
+        pendingAppActivity <- appActivityComputationO match {
```
meiersi-da: just for my understanding: in what sense is this app activity "pending"?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
meiersi-da: If you want to use only one, then it's probably better to only use `serveAppActivityRecordsAndTraffic` so that the name is self-explanatory. Yeah, it's not pretty, but this way logs and configs are more telling.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/RewardsReferenceDataProvider.scala
```diff
@@ -0,0 +1,65 @@
...
+class StoreBackedRewardsReferenceDataProvider(
+    rewardsStore: DbScanRewardsReferenceStore
+)(implicit ec: ExecutionContext)
+    extends RewardsReferenceDataProvider {
```
meiersi-da: Why not reanme `RewardsReferenceDataProvider` to `ScanRewardsReferenceStore` and implement it directly in `DbScanRewardsReferenceStore`? Not clear what the indirection here buys you.

That's the pattern we typically use. You can then also add a `CachingScanRewardsReferenceStore` as a separate class that adds caching to any `ScanRewardsReferenceStore`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictStoreIngestion.scala
dfordivam: I guess this is trying to convey that don't insert this directly in DB, fix the verdictRowId, and then insert. May be a better naming could be used, or don't bother with indicating that here as the ingestion logic takes care of fixing it.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/AppActivityComputation.scala
```diff
@@ -20,26 +35,120 @@ trait AppActivityComputation {
...
+                    summary,
+                    verdict,
+                    computeForSingleVerdict(summary, verdict, roundNumber, providers),
+                  )
+                }
+              case None =>
+                logger.error(
+                  s"Open mining round info not provided by lookupActiveOpenMiningRounds for ${summary.sequencingTime}"
+                )
+                Future.successful((summary, verdict, None))
```
meiersi-da: Can you explain:
- how are we ensuring that this error is not triggered during normal operations?
- why is it right to return `Future.successful` here instead of throwing a `RuntimeException`?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
dfordivam: Fixed, now confirming that it is withing this range

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/AppActivityComputation.scala
dfordivam: I have changed this logic; now we get a `None` value for a `record_time`  which was before the tcs store ingestion started, and we simply ignore the activity computation for those verdicts.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/RewardsReferenceDataProvider.scala
dfordivam: Ok, moved the APIs to `DbScanRewardsReferenceStore`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictStoreIngestion.scala
meiersi-da:
```suggestion
        activityRecords <- appActivityComputationO match {
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/AppActivityComputation.scala
```diff
@@ -3,11 +3,27 @@
...
+
+/** Computation of app activity records, ie per featured app traffic weight for a single verdict
+  * See CIP-104 for details.
+  */
+object AppActivityComputation {
+  val MaxTrafficCostBytes: Long = 100L * 1024 * 1024 // 100 MB
+}
+
+class AppActivityComputation(
+    dataProvider: ScanRewardsReferenceStore,
```
meiersi-da:
```suggestion
    rewardsReferenceStore: ScanRewardsReferenceStore,
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/AppActivityComputation.scala
```diff
@@ -20,26 +36,117 @@ trait AppActivityComputation {
...
     )
-  ]
-}
+  ]] = {
+    val tagged = summariesWithVerdicts.map { case (summary, verdict) =>
+      val isEligible =
+        if (verdict.verdict != v30.VerdictResult.VERDICT_RESULT_ACCEPTED) false
+        else if (summary.totalTrafficCost <= AppActivityComputation.MaxTrafficCostBytes) true
+        else {
+          logger.warn(
```
meiersi-da, line 53:
```suggestion
          // Note we skip the computation to avoid integer overflows. 
          // This should never happen as Canton does not process 100MB transactions.
          logger.warn(
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/AppActivityComputation.scala
```diff
@@ -20,26 +36,117 @@ trait AppActivityComputation {
...
+              case Some((roundNumber, roundOpensAt)) =>
+                dataProvider.lookupFeaturedAppPartiesAsOf(roundOpensAt).map { providers =>
+                  (
+                    summary,
+                    verdict,
+                    computeForSingleVerdict(summary, verdict, roundNumber, providers),
+                  )
+                }
+              case None =>
+                Future.successful((summary, verdict, None))
```
meiersi-da, line 90:
```suggestion
                // Skip activity record computation as we don't have the necessary round data ingested.
                // This can happen for freshly onboarded SVs, but is not expected to happen once the first activity record has been computed.
                Future.successful((summary, verdict, None))
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/AppActivityComputation.scala
```diff
@@ -20,26 +36,117 @@ trait AppActivityComputation {
...
+    else {
+      val aggregatedWeights = computeAggregatedWeights(
+        envelopesWithFeaturedAppConfirmers,
+        summary.totalTrafficCost,
+        totalFeaturedAppEnvelopesTraffic,
+      )
+
+      Some(
+        DbAppActivityRecordStore.AppActivityRecordT(
+          verdictRowId = 0L,
```
meiersi-da:
```suggestion
          verdictRowId = -123456789,
```
defensive proramming: use a value that has a very low likelihood of satisfying the foreign key constraint, and is easy to grep

meiersi-da: and define that as `AppActivityRecordT.DUMMY_VERDICT_ROW_ID`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
```diff
@@ -20,6 +24,24 @@ trait ScanRewardsReferenceStore extends AppStore {
 
   def key: ScanRewardsReferenceStore.Key
 
+  /** For a batch of record times, resolve the oldest open mining round at each time.
+    * Returns map from record_time to (roundNumber, roundOpensAt).
+    * This will block till the round info could be obtained for record_times
```
meiersi-da:
```suggestion
    * This will wait till the round info could be obtained for record_times
```

It does not block the thread executing this function, but makes the resulting future wait. Blocking the thread is typically a problem, which is why I want to stay away from the word to avoid confusing devs.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
```diff
@@ -20,6 +24,24 @@ trait ScanRewardsReferenceStore extends AppStore {
...
+    * which are yet to be ingested.
+    *
+    * On the other hand if round info could not be obtained for a particular record_time
+    * then the Map will not contain the entry for that.
+    * This could happen when no contracts ingestion has happened in the store,
+    * or if the record_time is before the ingestion start.
```
meiersi-da:
```suggestion
    * This could happen when the record_time is before the ingestion start.
```

It looks to me that this is the only case or not?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/ScanApp.scala
```diff
@@ -337,7 +338,24 @@ class ScanApp(
         appInitConnection,
         loggerFactory,
       )
-      appActivityComputation = NoOpAppActivityComputation
+      rewardsReferenceStoreO =
+        if (config.sequencerTrafficIngestion.enabled) {
```
meiersi-da:
```suggestion
        if (config.appActivityRecordAndTrafficIngestion.enabled) {
```
as before: make flags have descriptive names

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbTcsStore.scala
```diff
@@ -135,4 +136,35 @@ class DbTcsStore(
...
+      case some @ Some(_) => Future.successful(some)
+      case None =>
+        val storeId = acsStore.acsStoreId
+        val migrationId = acsStore.domainMigrationId
+        storage
+          .query(
+            sql"""SELECT MIN(archived_at) FROM #$archiveTableName
+                  WHERE store_id = $storeId
+                    AND migration_id = $migrationId
+               """.as[Option[Long]].head,
```
rautenrieth-da: Does
```suggestion
               """.as[Option[CantonTimestamp]].head,
```
work? We are using `.as[CantonTimestamp]` in other places.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/AppActivityComputation.scala
rautenrieth-da: How about including the suggested comment in the log message? That way operators have more information ready at hand should this very unlikely case ever happen.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
dfordivam: Actually this could happen if archived table has no records, ie the ingestion has just started. So approx during the first 10min in our case. I guess its ok to return None and not wait in this scenario?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
rautenrieth-da: FYI, we also have `should beWithin` and `should beAround` helpers in `SpliceTests` (pick whatever you find more readable, it won't affect the log output).

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/AppActivityComputation.scala
```diff
@@ -20,26 +36,122 @@ trait AppActivityComputation {
...
+                rewardsReferenceStore.lookupFeaturedAppPartiesAsOf(roundOpensAt).map { providers =>
+                  (
+                    summary,
+                    verdict,
+                    computeForSingleVerdict(summary, verdict, roundNumber, providers),
+                  )
+                }
+              case None =>
+                // Skip activity record computation as we don't have the necessary round data ingested.
+                // This can happen for freshly onboarded SVs, but is not
```
rautenrieth-da, line 85: How about including an debug level log here? It might be spammy for a while for new SVs but also might provide useful debug info for integration tests and unexpected situations.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/AppActivityComputation.scala
```diff
@@ -20,26 +36,122 @@ trait AppActivityComputation {
...
+    */
+  private def computeAggregatedWeights(
+      envelopesWithFeaturedAppConfirmers: Seq[(DbScanVerdictStore.EnvelopeT, Set[String])],
+      totalConfirmationRequestTraffic: Long,
+      totalFeaturedAppEnvelopesTraffic: Long,
+  ): SortedMap[String, Long] = {
+    val perPartyNumerator = envelopesWithFeaturedAppConfirmers
+      .foldLeft(SortedMap.empty[String, Long]) { case (acc, (envelope, featuredAppConfirmers)) =>
+        val perEnvelopeNumerator: Long =
+          (envelope.trafficCost * totalConfirmationRequestTraffic) / featuredAppConfirmers.size.toLong
```
rautenrieth-da, line 154: Can we add a comment why it's ok to use integer division?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
```diff
@@ -69,6 +69,45 @@ class DbScanRewardsReferenceStore(
     descriptor => SynchronizerId.tryFromString(descriptor.key("synchronizerId")),
   )
 
+  override def lookupActiveOpenMiningRounds(
```
rautenrieth-da, line 72: This is sufficiently complex that I think it would be good to have a store test for this (fine to do it later).

dfordivam: Added test

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/AppActivityComputation.scala
dfordivam: Added

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/AppActivityComputation.scala
dfordivam: Added

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbTcsStore.scala
dfordivam: Right, simplified this

PR #4462, opened 2026-03-17
[ci] Make UpdateHistory and EventHistory page size limits configurable

martinflorian-da:
Add `updateHistoryMaxPageSize` config field to `ScanAppBackendConfig`
(defaults to 1000, preserving current behavior). Thread it through
`ScanApp` to `HttpScanHandler` and use it in `PageLimit.tryCreate()`
calls for `getUpdateHistory` and `getEventHistory` endpoints.

Operators can now override the max page size via HOCON config:
  canton.scan-apps.scan-app.update-history-max-page-size = 500

Fixes: #4401


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
moritzkiefer-da, line 729: afaict you are missing getUpdateHistoryV1 and getUpdateHistoryV2

martinflorian-da: argh indeed 🤦

martinflorian-da: Wait doesn't that just call this exact method? What am I missing?

moritzkiefer-da: ah indeed, we have too many methods called `getUpdateHistory`

PR #4471, opened 2026-03-18
revisit s3&zstd flows

isegall-da:
Dropping the idea of labeling the last element using a boolean in `ByteStringWithTermination` allows to simplify everything, and also not fall into the trap that https://github.com/hyperledger-labs/splice/pull/4439 does (this PR will replace #4439)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
@@ -108,10 +108,10 @@ class AcsSnapshotBulkStorage(
         .filter { ts =>
           val ret = storageConfig.shouldDumpSnapshotToBulkStorage(ts.timestamp)
           if (ret) {
-            logger.debug(s"Dumping snapshot at timestamp $ts to bulk storage")
+            logger.debug(s"Dumping snapshot at timestamp ${ts.timestamp} to bulk storage")
           } else {
             logger.info(
-              s"Skipping snapshot at timestamp $ts for bulk storage, not required per the configured period of ${storageConfig.bulkAcsSnapshotPeriodHours}"
+              s"Skipping snapshot at timestamp ${ts.timestamp} for bulk storage, not required per the configured period of ${storageConfig.bulkAcsSnapshotPeriodHours}"
             )
           }
           ret


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
ray-roestenburg-da: Nitpick / minor issue, maybe better to do 
```
filter { twid => 
```
and then on 109 
```
val ret = storageConfig.shouldDumpSnapshotToBulkStorage(twid.ts)      
```
when anyone reads this right now they might think you wanted to do collect and ignore anything else and used filter instead which could match error. But actually the source is always of type `TimestampWithMigrationId` so might as well just access the property

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
```diff
@@ -83,12 +83,8 @@ case class GroupedWeightS3ObjectFlow(
       @SuppressWarnings(Array("org.wartremover.warts.Var"))
       @volatile
```
ray-roestenburg-da: You can remove `@volatile`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
```diff
@@ -116,56 +112,42 @@ case class GroupedWeightS3ObjectFlow(
...
-        push(
-          out,
-          Output(
-            objectKey = state.currentObject.key,
-            isLastObject = upstreamFinished,
-          ),
-        )
-        if (upstreamFinished) {
-          logger.debug("This was the last input in the stream, completing.")
+        push(out, state.currentObject.key)
```
ray-roestenburg-da, line 114: Check isAvailable(out) before push

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
```diff
@@ -116,56 +112,42 @@ case class GroupedWeightS3ObjectFlow(
...
-            )
-            if (state.numPendingPartUploads < maxParallelPartUploads) {
-              pull(in)
-            }
-          } else {
-            logger.debug(
-              s"New object size for ${curState.currentObject.key} is ${curState.currentObjectSize + elem.bytes.length}, done with this object"
-            )
+          if (state.numPendingPartUploads < maxParallelPartUploads) {
+            pull(in)
```
ray-roestenburg-da, line 149: guard with !hasBeenPulled(in) and use tryPull(in) (which guards with !isClosed internally)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
```diff
@@ -98,18 +93,22 @@ case class ZstdGroupedWeight(
...
       private def reset(): Unit = {
         zstd.get().close()
-        zstd.set(new ZSTD(3))
+        zstd.set(new ZSTD(compressionLevel))
         state.set(State.empty())
       }
 
+      private def finishAndPush(): Unit = {
+        state.set(state.get().append(zstd.get().zstdFinish()))
+        push(out, state.get().bytes)
```
ray-roestenburg-da: guard with isAvailable(out)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
ray-roestenburg-da: I think you can drop these atomics inside the GraphStageLogic

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
isegall-da: I disagree.
This now follows the model of the built-in GroupedWeighted. We are triggered by a pull(), so are guaranteed that when we get to this point out is already available. If it's not, that's a bug, and we want this to fail.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
isegall-da: Same as above. This should not be closed at this point, and if it is, I want it to crash.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: No strong opinion, but we use case's just for unmarshalling objects all over the place. Don't see any specific reason why here it would be especially confusing.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
ray-roestenburg-da: `GroupWeighted[T]` is not using an async callback though.. so here it executes on the right thread, but I don't think you can be sure that out is available, it can happen long after the onPull, and I don't see why you would want it to crash in that case.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
ray-roestenburg-da: I think that's incorrect and you should guard it. Why do you think it should not be closed, what guarantees this?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
isegall-da: We need strong guarantees of consistency. If we generate an object and it does not make it downstream, then something is wrong, and we cannot just dump that element and move on.

The flow is: downstream pulls an element => we pull one or two elements from upstream to produce that output => we push an element. If downstream disappeared between step 1&3 we definitely want to fail.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
ray-roestenburg-da: in this case, where you do `filter { case Foo(, , )=> ` you provide a partial function where a total function is expected, (collect would be better which expects a partial function and would never fail). Now you lift that in a total function which can result in a match error normally, which you would look out for in this case. But not a big issue cause as said here it's safe, so leaving up to you.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
isegall-da: Ah, this one's actually even easier: this is in the else of `is (objectDone)`, where `objectDone` is `.. || isClosed(in)`. So we already are guarding against it.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: Hmmm... Actually, wouldn't `map` be the cleanest here?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/ZstdGroupedWeight.scala
isegall-da: Same response as above: out should be available, and if it's not, something bad has happened, and we need to fail to maintain consistency guarantees.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
ray-roestenburg-da: all good for isClosed, though looking at internal stages the !hasBeenPulled(in) is also important, but leave it up to you

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
ray-roestenburg-da: yes, correct

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
ray-roestenburg-da: push when not available will get you an `new IllegalArgumentException(s"Cannot push port ($this) twice, or before it being pulled")`, which can be avoided, and it's not necessarily because of a failure in S3 pushing things but just concurrency issues, and I think because of asynccallback you could get push, push. But I will leave it up to you, maybe have a look at MapAsync that is close to what you are doing and has an async callback and does all the stuff you need to do.

isegall-da: >  which can be avoided

And do what instead? If we got to this point, we have uploaded an object to S3, and need to report that downstream. If the out port is closed, we cannot do that, so what do you propose we do instead of the exception? We can throw a slightly nicer exception, but just not pushing the element because there's nowhere to push to is just incorrect IMHO.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: Ah, no, I don't want a map, I do want a filter. Indeed it's not a partial function, but I want to filter in only those timestamps for which we need to dump, so `filter` is the right thing (or `collect`, but that's just less specific than `filter`). I'll just do what you proposed first, and drop the `case`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
isegall-da: This is inside `onPush` (not in async callback or something). AFAIU, checking `hasBeenPulled` here is pointless, because it will always be `false`, no?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
ray-roestenburg-da: no, buffer and push when available, eventually a  pull should happen again

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
ray-roestenburg-da: Yes, that's correct, overthinking the async nature of this stuff.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/GroupedWeightS3ObjectFlow.scala
isegall-da: We should always get here only after a pull has happened, and a push has not yet (*). So the only risk is that downstream closed the connection in between, not that it's not yet ready.

(*) the reasoning for that is quite simple:
After pushing, we do not pull again, until another pull from downstream happens. So since new elements are created only as a result of an upstream element, no two pushes can happen without a pull in between.

ray-roestenburg-da: Ok, I'll go along with it, don't want to stop the PR, it's already a lot better and if we encounter issues /edge cases we can revisit again

isegall-da: Thanks. I do plan on rerunning the CILR dump from genesis at least once more, which is what uncovered many issues so far.

PR #4493, opened 2026-03-18
Work around for validator-participant race condition, when restarting

pasindutennage-da:
Fix https://github.com/DACH-NY/cn-test-failures/issues/7567

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
@@ -28,6 +28,7 @@ import org.lfdecentralizedtrust.splice.environment.{
 import org.lfdecentralizedtrust.splice.identities.NodeIdentitiesDump
 import org.lfdecentralizedtrust.splice.util.PrettyInstances.prettyString
 
+import java.util.concurrent.atomic.AtomicReference
 import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration.*
 
@@ -110,7 +111,7 @@ class NodeInitializer(
       nodeIdentity: UniqueIdentifier => Member & NodeIdentity,
   )(implicit tc: TraceContext, ec: ExecutionContext): Future[Unit] = {
     logger.info(s"Making sure canton node has an identity")
-    val deadline = 5.seconds.fromNow
+    val emptyIdDeadline = new AtomicReference[Option[Deadline]](None)
     for {
       // If the node was started concurrently with the app, it might not immediately be responding, so we're
       // retrying the getId() call.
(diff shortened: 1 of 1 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
```diff
@@ -109,39 +110,38 @@ class NodeInitializer(
...
-                    .withDescription(
-                      s"Node is still bootstrapping (status: $status). Retrying getId..."
-                    )
-                    .asRuntimeException()
-                )
+            // currently we don't have a mechanism to distinguish between the following 2 states.
+            //    a. a participant that has completed bootstrapping and waiting for the validator to set a new Id
+            //    b. a participant that is still bootstrapping, and not yet set its Id from the db
+            // as a (temporary) work around, if participant has no id set, we retry getId for a maximum of 5s duration
+            // we chose 5s, because, according to logs, 5s is enough for the participant to set it's id, if it already has one
```
martinflorian-da: Shall we crate an issue to track the upstream limitation and add a `TODO(#X)` line in here to make it more clear that we don't want this around forever?

martinflorian-da:
```suggestion
            // we chose 5s, because, according to logs, 5s is enough for the participant to set its id, if it already has one
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
```diff
@@ -109,39 +110,38 @@ class NodeInitializer(
       nodeIdentity: UniqueIdentifier => Member & NodeIdentity,
   )(implicit tc: TraceContext, ec: ExecutionContext): Future[Unit] = {
     logger.info(s"Making sure canton node has an identity")
+    val deadline = 5.seconds.fromNow
```
martinflorian-da: This is a bit overengineered given our analysis of past failures, so consider ignoring:

If you set this here you start counting even before the endpoint was reachable at all. So in some bad case, we start retrying, 5 seconds pass, endpoint becomes reachable, now we don't wait for any deadline and once again get the race condition.

So maybe have 2 retry loops for getId? One terminates as usual, and if the result is `None` we enter a fresh retry loop bounded by the 5 seconds?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
pasindutennage-da: Fixed

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
pasindutennage-da: Fixed.

Starting the 5s after the first call returned some value.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/setup/NodeInitializer.scala
```diff
@@ -109,39 +111,43 @@ class NodeInitializer(
...
+              case Some(deadline) if deadline.hasTimeLeft() =>
                 Future.failed(
                   Status.UNAVAILABLE
                     .withDescription(
-                      s"Node is still bootstrapping (status: $status). Retrying getId..."
+                      s"Node returned empty ID. Retrying to ensure it is not still bootstrapping... (Time left: ${deadline.timeLeft.toMillis}ms)"
                     )
                     .asRuntimeException()
                 )
+              case _ =>
```
martinflorian-da, line 149: AFAIU we should never be able to reach that; let's make this clear by crashing if we do.

martinflorian-da: uf nvm brain is still booting

PR #4495, opened 2026-03-18
[ci] Fix 4213, don’t log empty rejections

### LINKED ISSUE #4213: Investigate empty rejections logged on API requests
HttpRequestLogger logs rejections often.

For instance 

```
HTTP GET /api/scan/v0/dso from (redacted_ip): Rejected: 
```
Maybe this is a false positive. Or the logging needs to be improved.

ray-roestenburg-da:
Fixes #4213 

- The requestLogger is now at the top of the routes, in the dynamicRoute in HttpAdminService, where all the routes are combined.
- added tests for logging
- Added a rejectionHandler and logging for it

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLogger.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLogger.scala
@@ -4,7 +4,7 @@
 package org.lfdecentralizedtrust.splice.admin.api
 
 import org.apache.pekko.http.scaladsl.model.{ContentTypes, HttpEntity, RemoteAddress}
-import org.apache.pekko.http.scaladsl.server.{Directive0, RejectionHandler, RequestContext}
+import org.apache.pekko.http.scaladsl.server.{Directive0, RequestContext}
 import org.apache.pekko.http.scaladsl.server.Directives.*
 import com.digitalasset.canton.config.ApiLoggingConfig
 import com.digitalasset.canton.logging.{NamedLoggerFactory, NamedLogging}
@@ -39,33 +39,6 @@ object HttpRequestLogger {
     maxMetadataSize = loggingConfig.maxMetadataSize,
     loggerFactory = loggerFactory,
   )
-
-  /** Creates a RejectionHandler that logs the rejection status code, then delegates to the
-    * default handler to produce the appropriate HTTP response. This should be used at the
-    * top level of the route tree (via handleRejections) to seal the route — ensuring all
-    * rejections become Complete responses with their reason logged.
-    *
-    * Must be placed OUTSIDE the HttpRequestLogger directive so that the logger's
-    * mapResponse sees the rejection-converted responses too.
-    */
-  def loggingRejectionHandler(
-      loggingConfig: ApiLoggingConfig,
-      loggerFactory: NamedLoggerFactory,
-  )(implicit traceContext: TraceContext): RejectionHandler = {
-    val inst = new HttpRequestLogger(
-      loggingConfig.messagePayloads,
-      loggingConfig.maxMethodLength,
-      loggingConfig.maxStringLength,
-      loggingConfig.maxMetadataSize,
-      loggerFactory,
-    )
-    RejectionHandler.default.mapRejectionResponse { response =>
-      if (response.status.isFailure()) {
-        inst.logger.debug(s"HTTP request rejected with status code: ${response.status}")
-      }
-      response
-    }
-  }
 }
 
 final class HttpRequestLogger(
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/admin/http/HttpAdminService.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/admin/http/HttpAdminService.scala
@@ -12,8 +12,8 @@ import com.digitalasset.canton.logging.{ErrorLoggingContext, NamedLoggerFactory}
 import io.opentelemetry.api.trace.Tracer
 import org.apache.pekko.actor.ActorSystem
 import org.apache.pekko.http.scaladsl.Http
+import org.apache.pekko.http.scaladsl.server.{RejectionHandler, Route}
 import org.apache.pekko.http.scaladsl.server.Directives.*
-import org.apache.pekko.http.scaladsl.server.Route
 import org.lfdecentralizedtrust.splice.admin.api.HttpRequestLogger
 import org.lfdecentralizedtrust.splice.admin.api.TraceContextDirectives.withTraceContext
 import org.lfdecentralizedtrust.splice.environment.SpliceStatus
@@ -77,13 +77,10 @@ object HttpAdminService {
       withTraceContext { traceContext =>
         // The logger logs the request and uses mapResponse to log the response.
         // handleRejections (inside the logger) seals the route: rejections are
-        // converted to HTTP responses so mapResponse sees all outcomes.
-        // loggingRejectionHandler additionally logs the rejection status code.
+        // converted to HTTP responses so mapResponse sees all outcomes and logs
+        // exactly one "Responding with status code" per request.
         HttpRequestLogger(apiLoggingConfig, loggerFactory)(traceContext) {
-          handleRejections(
-            HttpRequestLogger
-              .loggingRejectionHandler(apiLoggingConfig, loggerFactory)(traceContext)
-          ) {
+          handleRejections(RejectionHandler.default) {
             encodeResponse(concat((commonAdminRoute +: routes.get())*))
           }
         }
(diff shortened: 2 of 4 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLogger.scala
```diff
@@ -103,12 +103,13 @@ final class HttpRequestLogger(
...
               )
 
-            case Rejected(rejections) =>
+            case Rejected(rejections) if rejections.nonEmpty =>
               if (rejections.contains(AuthorizationFailedRejection)) {
                 logger.debug(msg("Rejected: Unauthorized."))
               } else {
                 logger.debug(msg(s"""Rejected: ${rejections.mkString(",")}"""))
               }
+            case Rejected(_) => // do nothing, part of normal route resolution
```
moritzkiefer-da: This looks weird to me:
1. We still log the request afaik, we now just don't log a response. That seems rather confusing when debugging.
2. Afaiu pekko-http empty rejections is essentially a 404 i.e. no route handler matches, but I think we also saw logs where it looked like there should be a request handler, at least the url didn't look right. Do we actually understand what is causing those?

ray-roestenburg-da: 1. right.. let me have another look, maybe it has to be at the top of the tree where the routes are composed
2. yes no route handler matches, but in this case only for the route wrapped in the logger directive, so likely another route does handle it.

I'll have a closer look in any case.

ray-roestenburg-da: Maybe look into a rejectionHandler instead

ray-roestenburg-da: yeah Ok this needs a little more work to be completely correct.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLoggerTest.scala
```diff
@@ -0,0 +1,233 @@
...
+          pathPrefix("api" / "app") {
+            complete(StatusCodes.OK, "app")
+          },
+        )
+      }
+    }
+
+  "HttpRequestLogger" should {
+
+    "old style: log duplicate 'received request' for non-matching sibling" in {
```
isegall-da: Not sure that it makes sense to continue testing for the old style that we're trying to fix here.

ray-roestenburg-da: yeah that was a regression test and ensuring this is why we saw those `Rejected:` messages but can remove now

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLogger.scala
ray-roestenburg-da: Should be good now

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLogger.scala
```diff
@@ -44,6 +39,33 @@ object HttpRequestLogger {
...
+    val inst = new HttpRequestLogger(
+      loggingConfig.messagePayloads,
+      loggingConfig.maxMethodLength,
+      loggingConfig.maxStringLength,
+      loggingConfig.maxMetadataSize,
+      loggerFactory,
+    )
+    RejectionHandler.default.mapRejectionResponse { response =>
+      if (response.status.isFailure()) {
+        inst.logger.debug(s"HTTP request rejected with status code: ${response.status}")
```
moritzkiefer-da: why do we need both this and the "Responding with status code" log below?

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLoggerTest.scala
```diff
@@ -0,0 +1,144 @@
...
+      loggerFactory.assertLogsSeq(SuppressionRule.Level(Level.DEBUG))(
+        {
+          Get("/api/app") ~> route ~> check {
+            status shouldBe StatusCodes.OK
+            responseAs[String] shouldBe "app"
+          }
+        },
+        { logEntries =>
+          // Exactly one "received request"
+          val receivedLogs = logEntries.filter(_.message.contains("received request"))
```
moritzkiefer-da: let's use a forExactly(1 or something like that. Otherwise failures are impossible to debug if the list is empty

moritzkiefer-da: similar for all the instances below

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLoggerTest.scala
```diff
@@ -0,0 +1,144 @@
...
+          Get("/api/unknown") ~> route ~> check {
+            status shouldBe StatusCodes.NotFound
+          }
+        },
+        { logEntries =>
+          // One "received request"
+          val receivedLogs = logEntries.filter(_.message.contains("received request"))
+          receivedLogs should have size 1
+
+          // loggingRejectionHandler logs the rejection
```
moritzkiefer-da: so based on this we now log rejections twice once as "rejected with status code" and once with "responding with status code"? What do we gain from that?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLogger.scala
ray-roestenburg-da: These are different paths that don't bot fire for the same request. A request is rejected, or a response is matched and a response is created for it. both are needed for complete lifecycle logging. (note that `mapResponse` is now used instead of `mapRouteResult` which gave the 'strange' empty rejection list cause it did not get matched but went further along the route tree to find another match) But let me check the test code since. There might be some other reason why both end up logged.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLoggerTest.scala
ray-roestenburg-da: yeah let me fix this

ray-roestenburg-da: One or the other should happen

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLogger.scala
ray-roestenburg-da: ah it is because of handleRejections, I'll fix this.

ray-roestenburg-da: fixed.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLoggerTest.scala
ray-roestenburg-da: fixed

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/admin/api/HttpRequestLoggerTest.scala
ray-roestenburg-da: fixed

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/admin/http/HttpAdminService.scala
```diff
@@ -71,19 +74,27 @@ object HttpAdminService {
     )
     private val routes: AtomicReference[List[Route]] = new AtomicReference(List())
     private val dynamicRoute: Route = ctx => {
-      encodeResponse(concat((commonAdminRoute +: routes.get())*))(ctx)
+      withTraceContext { traceContext =>
```
moritzkiefer-da, line 77: What is this trace context used for? We seem to lose the more specific one in favor of the one here. I believe we also have a per request one so not sure what the per class one is used at all.

ray-roestenburg-da: `withTraceContext` parses the trace context from headers, (W3C Trace context), so it should just be done for any route?

ray-roestenburg-da: So if the UI (or another client) passes in the W3C trace context via headers, it is taken along on the server side

PR #4501, opened 2026-03-18
DAML changes for expiry trigger for expired AmuletAllocations

pasindutennage-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4500

### DIFF
--- a/nix/canton-sources.json
+++ b/nix/canton-sources.json
@@ -4,8 +4,8 @@
   "daml_release": "v3.3.0-snapshot.20250417.0",
   "enterprise_sha256": "sha256:0aqzklvbzn74afqbqq5dg3vy0df1wbnmrx2wdk0bfjmjaf541i79",
   "oss_sha256": "sha256:1byb6znh0h879vdb31awl28n50zfbm34vb8dgli6rm9a2z3534vl",
-  "canton_base_image_sha256": "sha256:485a28f687d5bca698b8a2660989e5c78f7922797bae307b4b721a44415bdf94",
-  "canton_participant_image_sha256": "sha256:68f636afa8d4178fc412e604ea4fe38818ed3fc6b4ae45f43f10a39b9d76f934",
-  "canton_mediator_image_sha256": "sha256:95c681f056961d938f0683a415dc45d5d5f7be425e170dc0872d3c8bbe45bb43",
-  "canton_sequencer_image_sha256": "sha256:8c978260991cfeafd66b4641d20f6927402870495b9f942142be537a738d025d"
+  "canton_base_image_sha256": "sha256:3d8a57adfc5f0c0c1aec80007677c23b7d4e3501d743f6bc5bc642e1e29fcbaf",
+  "canton_participant_image_sha256": "sha256:49a21a0bbbc005d2a5b117541ba8d07b2fd74383b700b7d66356cc6794584550",
+  "canton_mediator_image_sha256": "sha256:dfed91abf19f64e0e97c8a0b8367c516f5a16fbbfcddd5af653929ca47541564",
+  "canton_sequencer_image_sha256": "sha256:6b137c1da69e1ac13709dd64a18fd73eca14647723d586b0fe1bed50ffefb3f9"
 }


daml/splice-amulet/daml/Splice/AmuletAllocation.daml
```diff
@@ -47,6 +61,27 @@ template AmuletAllocation
             senderHoldingCids
             meta = emptyMetadata
 
+    -- Implemented on the concrete template to avoid a breaking change to the shared Allocation interface.
```
moritzkiefer-da:
```suggestion
    -- Implemented on the concrete template as you cannot upgrade interfaces so we cannot add a choice to the Allocation interface.
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1124,6 +1143,28 @@ template DsoRules with
     -- Amulet management driven directly by the DSO delegates
     ---------------------------------------------------------
 
+    -- Implemented directly in DsoRules to avoid a cyclic module dependency between AmuletRules and AmuletAllocation.
+    nonconsuming choice DsoRules_ExpireAmuletAllocations : DsoRules_ExpireAmuletAllocationsResult
+      with
+        expireRequest : DsoRules_ExpireAmuletAllocationsRequest
+        sv : Party
+      observer expireRequest.observers
```
moritzkiefer-da: This is not going to work. The validators don't vet dso-governance so you can't put them as choice observers here.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -324,6 +325,24 @@ data DsoRules_CreateBootstrapExternalPartyConfigStateInstructionResult = DsoRule
...
+  deriving (Eq, Show)
+
+data DsoRules_ExpireAmuletAllocationsResult = DsoRules_ExpireAmuletAllocationsResult
+  with
+    results : [AmuletAllocation_DsoExpireResult]
+  deriving (Eq, Show)  
+
+data DsoRules_ExpireAmuletAllocationsRequest = DsoRules_ExpireAmuletAllocationsRequest
+  with
+    expectedDso : Party
```
moritzkiefer-da: expectedDso doesn't really make sense. You use it if there is another contract you pass in that you want to check like AmuletRules. But if your implementation is fully on dso rules there is nothing to check.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1124,6 +1143,28 @@ template DsoRules with
     -- Amulet management driven directly by the DSO delegates
     ---------------------------------------------------------
 
+    -- Implemented directly in DsoRules to avoid a cyclic module dependency between AmuletRules and AmuletAllocation.
```
moritzkiefer-da: I don't understand this. `AmuletAllocation` does not seem to import `AmuletRules` from a quick look at the code. What's the cyclic dependency you're trying to avoid? I do think we need it on amulet rules (or a similar contract) if we want to use the choice observers

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1124,6 +1143,28 @@ template DsoRules with
...
+      observer expireRequest.observers
+      controller sv
+      do
+        _ <- getAndValidateSvParty this (Some sv)
+        
+        require ("expectedDso " <> show expireRequest.expectedDso <> " matches the actualDso " <> show dso) (expireRequest.expectedDso == dso)
+
+        individualResults <- forA expireRequest.inputs $ \input -> do
+          alloc <- fetchButArchiveLater (ForDso with dso) input.allocationCid
+          require "The controller must be the admin of the allocation's instrument" (alloc.allocation.transferLeg.instrumentId.admin == dso)
```
moritzkiefer-da:
```suggestion
          require "The dso must be the admin of the allocation's instrument" (alloc.allocation.transferLeg.instrumentId.admin == dso)
```
not sure what you mean by controller here, the controller is the sv party which isn't relevant

daml/splice-dso-governance/daml/Splice/DsoRules.daml
pasindutennage-da: Splice.AmuletRules import AmmuletAllocation imports Splice.Amulet.TwoStepTransfer imports Splice.AmuletRules

daml/splice-dso-governance/daml/Splice/DsoRules.daml
pasindutennage-da: whenever I use "dso" in a comment, the static-test fails, when it scans for occurences of "dso". How should I fix it?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: I see that's unfortunate. I think we can put the batch choice on ExternalPartyAmuletRules though instead which should avoid this.

Or alternatively we can just skip the choice observer optimization and assume that this is used sufficiently rarely that it's not worth it. Which isn't unreasonable either here.

I think I'd lean towards the ExternalPartyAmuletRules option just because it doesn't seem much more complex and may come in handy at some point.

@meiersi-da wdyt?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: Use `dsoParty`

daml/splice-dso-governance/daml/Splice/DsoRules.daml

daml/splice-dso-governance/daml/Splice/DsoRules.daml
pasindutennage-da: Implemented in ExternalPartyAmuletRules

PR #4534, opened 2026-03-19
list bulk-storage objects for ACS snapshot

isegall-da:
Part of #3429 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
@@ -2332,7 +2332,9 @@ class HttpScanHandler(
     }
   }
 
-  override def listBulkAcsSnapshotObjects(respond: ScanResource.ListBulkAcsSnapshotObjectsResponse.type)(
+  override def listBulkAcsSnapshotObjects(
+      respond: ScanResource.ListBulkAcsSnapshotObjectsResponse.type
+  )(
       atOrBeforeRecordTime: OffsetDateTime
   )(extracted: TraceContext): Future[ScanResource.ListBulkAcsSnapshotObjectsResponse] = {
     implicit val tc = extracted
--- a/apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/ScanAppReference.scala
+++ b/apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/ScanAppReference.scala
@@ -754,7 +754,9 @@ abstract class ScanAppReference(
   }
 
   @Help.Summary("List all objects in bulk storage for an ACS snapshot")
-  def getBulkAcsSnapshot(timestamp: CantonTimestamp): definitions.ListBulkAcsSnapshotObjectsResponse =
+  def getBulkAcsSnapshot(
+      timestamp: CantonTimestamp
+  ): definitions.ListBulkAcsSnapshotObjectsResponse =
     consoleEnvironment.run {
       httpCommand(
         HttpScanAppClient.GetBulkAcsSnapshot(timestamp)


apps/scan/src/main/openapi/scan.yaml
```diff
@@ -1617,6 +1617,30 @@ paths:
...
 
+  /v0/history/bulk/acs:
+    get:
+      tags: [ external, scan ]
+      x-jvm-package: scan
+      operationId: "getBulkAcsSnapshot"
+      description: |
+        Get download URLs and metadata for an ACS snapshot available for bulk download, at or before a certain record time.
+      parameters:
+        - name: "record_time"
```
ray-roestenburg-da: Minor comment to give 2 seconds of thought: maybe `at_or_before_record_time` is even a little more clear? someone calling it can't be confused what the record time field means.

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -1617,6 +1617,30 @@ paths:
         "500":
           $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/500"
 
+  /v0/history/bulk/acs:
+    get:
+      tags: [ external, scan ]
+      x-jvm-package: scan
+      operationId: "getBulkAcsSnapshot"
+      description: |
+        Get download URLs and metadata for an ACS snapshot available for bulk download, at or before a certain record time.
```
ray-roestenburg-da: The description makes me think that `/v0/history/bulk/acs-metadata` is maybe better, as well as `getBulkAcsSnapshotMetadata`, but leave it to you. (you don't get the actual acs but information about the acs)

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -3914,3 +3938,37 @@ components:
...
+            The list of bulk storage objects containing the ACS snapshot data.
+          type: array
+          items:
+            $ref: "#/components/schemas/BulkStorageObject"
+
+    BulkStorageObject:
+      type: object
+      required:
+        - url
+        - checksum
```
ray-roestenburg-da: `digest` is probably better?

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -3914,3 +3938,37 @@ components:
           type: array
           items:
             "$ref": "../../../../common/src/main/openapi/common-internal.yaml#/components/schemas/ContractWithState"
+
+    GetBulkAcsSnapshotResponse:
+      type: object
+      required:
+        - timestamp
+        - objects
```
ray-roestenburg-da: maybe `content_refs`

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -3914,3 +3938,37 @@ components:
           type: array
           items:
             "$ref": "../../../../common/src/main/openapi/common-internal.yaml#/components/schemas/ContractWithState"
+
+    GetBulkAcsSnapshotResponse:
+      type: object
+      required:
+        - timestamp
```
ray-roestenburg-da: is this not a `record_time`? Maybe better to use that to clarify that it is not some timestamp when it was uploaded, but the record_time for the acs snapshot, and that it relates to the at_or_before_record_time that you pass in.

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -3914,3 +3938,37 @@ components:
...
+        - objects
+      properties:
+        timestamp:
+          description: |
+            The timestamp at which the ACS snapshot was taken.
```
ray-roestenburg-da: Who decides this timestamp, is it locally defined? or is it record_time? Can we also give the record_time for the ACS snapshot? is there a difference? if the naming can prevent these questions from coming up that would be great. (or if there is a relation between timestamp and record_time, if they are the same or not, I think is all relevant)

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -3914,3 +3938,37 @@ components:
...
+        - url
+        - checksum
+      properties:
+        url:
+          description: |
+            The URL from which the bulk storage object can be downloaded.
+          type: string
+        checksum:
+          description: |
+            The checksum of the bulk storage object, for verification of integrity and consistency across SVs.
```
```suggestion
            The digest of the bulk storage object, for verification of integrity and consistency across SVs.
```

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -3914,3 +3938,37 @@ components:
...
+          description: |
+            The timestamp at which the ACS snapshot was taken.
+          type: string
+          format: date-time
+        objects:
+          description: |
+            The list of bulk storage objects containing the ACS snapshot data.
+          type: array
+          items:
+            $ref: "#/components/schemas/BulkStorageObject"
```
ray-roestenburg-da: I think I would call this something like a ..Ref because it's not the actual object but a ref to it?  Maybe BulkContentRef.
Anyway I'm a bit in the naming bikeshed. Feel free to ignore or give it some extra thought, since it's always hard to rename these later, and the less questions we get if things are more obvious the better.

apps/scan/src/main/openapi/scan.yaml
ray-roestenburg-da: Can we say something about the type of digest this is? probably sha-256 if I remember correctly

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2325,6 +2332,36 @@ class HttpScanHandler(
       }
     }
   }
+
+  override def getBulkAcsSnapshot(respond: ScanResource.GetBulkAcsSnapshotResponse.type)(
+      atOrBeforeTimestamp: OffsetDateTime
+  )(extracted: TraceContext): Future[ScanResource.GetBulkAcsSnapshotResponse] = {
+    implicit val tc = extracted
+    withSpan(s"$workflowId.getBulkAcsSnapshot") { _ => _ =>
+      val recordTimeTs = Codec.tryDecode(Codec.OffsetDateTime)(atOrBeforeTimestamp)
```
ray-roestenburg-da: so it is `atOtBeforeRecordTime` really, rather not mix if not needed?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2325,6 +2332,36 @@ class HttpScanHandler(
...
+      bulkStorage.getAcsSnapshotBulkStorage.fold(
```
ray-roestenburg-da: you could inline this blow up if not exists inside a `bulkStorage.getAcsSnapshotAtOrBefore` as a unifying interface, instead of making that somehting the caller needs to know about.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2325,6 +2332,36 @@ class HttpScanHandler(
...
+        )
+      )(acsSnapshotBulkStorage =>
+        acsSnapshotBulkStorage.getAcsSnapshotAtOrBefore(recordTimeTs).map {
+          case AcsSnapshotObjects(ts, objects) =>
+            ScanResource.GetBulkAcsSnapshotResponse.OK(
+              definitions.GetBulkAcsSnapshotResponse(
+                Codec.encode(ts),
+                objects.map { case ObjectKeyAndChecksum(key, checksum) =>
+                  definitions.BulkStorageObject(
+                    // TODO(#3429): for now we return just the key, but this should be mapped to a full url in scan
```
ray-roestenburg-da, line 2357: Is this going to be behind a flag, or will this API be usable, and users will get the wrong data?  (If visible we should at least document this, and maybe on the API set some alpha / beta warning, or at least not prod ready)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanStorageConfig.scala
```diff
@@ -49,26 +49,51 @@ case class ScanStorageConfig(
...
+  ): CantonTimestamp = {
+    val rtHours = upToRecordTime.toInstant.atZone(ZoneOffset.UTC).getHour
+    val hourForSnapshot = timesToDoSnapshot(periodHours)
+      .find(_ <= rtHours)
+      .getOrElse(
+        throw new RuntimeException(
+          s"Unexpectedly could not find a time for a snapshot before ${rtHours}"
+        )
+      )
+    atHour(upToRecordTime, 0, hourForSnapshot)
```
ray-roestenburg-da, line 88: Is there a difference between the scheduled time for the snapshot to be created and the record_time you would associate it with up to which the snapshot holds? Could this cause issues?

apps/scan/src/main/openapi/scan.yaml
ray-roestenburg-da: Ok so s3 calls it checksum probably historically but since we're creating a fresh API we might as well be more precise with "digest", but leave it up to you.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/BulkStorage.scala
```diff
@@ -36,34 +37,50 @@ class BulkStorage(
     with FlagCloseableAsync
     with RetryProvider.Has {
 
+  private val acsSnapshots: AtomicReference[Option[AcsSnapshotBulkStorage]] = new AtomicReference(
```
ray-roestenburg-da: This reads too complicated for me, why can't the BulkStorage be created with a reference to AcsSnapshotBulkStorage and UpdateHistoryBulkStorage via it's constructor?

ray-roestenburg-da: I would prefer a more straightforward BulkStorage apply that calls retryableservice(acsSnapshotStorage) and retryableservice(updateHistoryBulkStorage)  (just constructs everything) where retryableService is also just a method or an apply on AcsSnapshotBulkStorage, I don't think you need the atomics for this unless I'm misreading something.

apps/scan/src/main/openapi/scan.yaml
isegall-da: > you don't get the actual acs but information about the acs

Definitely a valid point.
Bikeshedding a bit: I don't think I like metadata either, because typically one would not think that they need to first get metadata of objects in order to know where to download them from.

So how about
`/v0/history/bulk/list-acs-objects`? Too verbose?! Or maybe `/v0/history/bulk/acs/list` ?
And `listBulkAcsSnapshotObjects`. The latter I like, the list-acs-objects a bit less....  So I think I'm leaning toward 
keeping `/v0/history/bulk/acs` and changing only the (internal) `listBulkAcsSnaphotObjects`, but then that's an improvement only internally..... WDYT?

ray-roestenburg-da: yeah rather keep it as `/v0/history/bulk/acs` then.

ray-roestenburg-da: the internal improvement is still nice I think.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
isegall-da: Went for "pre-alpha". 
The flag will be the SV configuring the s3 bucket in the config. Until then, this will just fail (though I need to double-check that it fails gracefully enough)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanStorageConfig.scala
isegall-da: > scheduled time for the snapshot 

The ACS snapshot code suggested wrongfully in the past that that's a thing, but it's not. Everything is record time. We take snapshots of record time 00:00, 03:00, 06:00, etc., and we save to bulk storage snapshots of record time 00:00 daily.

ray-roestenburg-da (approved):
Thanks!

PR #4546, opened 2026-03-20
Fix multi-hosted partyid validation in Feature App Proposal

pawelperek-da:
Fixes #4555 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
@@ -2172,7 +2172,7 @@ class HttpScanHandler(
           case Right(domain) => Future.successful(domain)
           case Left(error) =>
             Future.failed(
-              HttpErrorHandler.badRequest(s"Could not decode synchronizer ID: $error")
+              HttpErrorHandler.badRequest(s"Could not decode domain ID: $error")
             )
         }
         actual <- sequencerAdminConnection.getSequencerTrafficControlState(member)
@@ -2250,7 +2250,7 @@ class HttpScanHandler(
           case Right(domain) => Future.successful(domain)
           case Left(error) =>
             Future.failed(
-              HttpErrorHandler.badRequest(s"Could not decode domain ID: $error")
+              HttpErrorHandler.badRequest(s"Could not decode synchronizer ID: $error")
             )
         }
         party <- PartyId.fromProtoPrimitive(partyId, "partyId") match {


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2210,22 +2210,22 @@ class HttpScanHandler(
           topologySnapshot =
             TopologySnapshot.Effective, // Follow the usual Canton APIs to return effective and not sequenced state.
         )
-        participantId <- response.mapping.participantIds match {
+        _ <-
+          if (response.mapping.partyId == party) Future.unit
```
pawelperek-da, line 2271: This is a guard which prevents prefix matches, i.e. if you call this endpoint with the `digital-asset-2::1` it will not match `digital-asset-2::122abc...`, since this is a behaviour we expect in the Feature App Proposal.
@moritzkiefer-da wdyt, should we keep it here and break backwards compatibility (if someone called this endpoint with not full partyId) or move it to the SV handler proxy?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvOperatorHandler.scala
```diff
@@ -418,55 +418,33 @@ class HttpSvOperatorHandler(
...
-            } else {
+        }
+        dsoRules <- dsoStore.getDsoRules()
+        scanConnection <- scanConnectionF
+        participantIds <- scanConnection
+          .getPartyToParticipant(
+            dsoRules.domain,
+            party,
+          )
+          .recoverWith {
```
pawelperek-da: This might be unnecessary, I'm running integration tests to confirm that

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
moritzkiefer-da: seems ok to break, I'd argue it was a bug that it worked before

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -3617,13 +3615,15 @@ components:
           format: int64
     GetPartyToParticipantResponse:
       type: object
-      required: [ "participant_id" ]
+      required: [ "participant_ids" ]
```
moritzkiefer-da, line 3703: This won't fly, we can't just hardbreak the scan API here completely. Contrary to the discussion above about exact matching this is not some edge case but will hit every single consuming.

The two options here are either:

1. Add `participant_ids` as a new field. If there is only one, set `participant_id` and `participant_ids` to a singleton list. If there is more than one set `participant_ids`  and leave `participant_id`
2. Add a v1 version of the endpoint that supports multi-hosting and only has a `participant_ids` field, keep the exact current behavior for the current one.

I'd lean towards option 2 with the v1 endpoint.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2237,6 +2237,55 @@ class HttpScanHandler(
...
+      partyId: String,
+  )(extracted: TraceContext): Future[ScanResource.GetPartyToParticipantV1Response] = {
+    implicit val tc = extracted
+    withSpan(s"$workflowId.getPartyToParticipantV1") { _ => _ =>
+      for {
+        domain <- SynchronizerId.fromString(synchronizerId) match {
+          case Right(domain) => Future.successful(domain)
+          case Left(error) =>
+            Future.failed(
+              HttpErrorHandler.badRequest(s"Could not decode domain ID: $error")
```
moritzkiefer-da:
```suggestion
              HttpErrorHandler.badRequest(s"Could not decode synchronizer ID: $error")
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2237,6 +2237,55 @@ class HttpScanHandler(
...
+            TopologySnapshot.Effective, // Follow the usual Canton APIs to return effective and not sequenced state.
+        )
+        _ <-
+          if (response.mapping.partyId == party) Future.unit
+          else
+            Future.failed(
+              HttpErrorHandler.notFound(s"Party not found: $party")
+            )
+        participantIds <- response.mapping.participantIds match {
+          case Seq() =>
```
moritzkiefer-da: now that we have a list, any reason not to just return the list even if it's empty?

pawelperek-da: good point

Issue #4120, opened 2026-02-25
Add explicit map for additional packages that need to be unvetted


julientinguely-da:
In addition to #4119, we want the ability to unvet specific package versions. This can be done by adding an explicit map to SVs and validators config of package name -> Set(package version) and unvet those versions in `vetPackages` and `vetCurrentPackages`.

PR #4553, opened 2026-03-20
Add `additionalPackagesToUnvet` map to specify extra package to unvet

julientinguely-da:
fixes https://github.com/hyperledger-labs/splice/issues/4120

enabled only for SVs and SV validators

- [x] implemented integration tests
- [x] tested it on scratchnet
- [x] add docs 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/docs/src/sv_operator/sv_operations.rst
+++ b/docs/src/sv_operator/sv_operations.rst
@@ -855,13 +855,12 @@ Unvet unsecure package versions
 .. warning::
 
   Only set these configurations **after agreement with at least 2/3 of the SVs** that this is desired.
+  Vetting for the DSO party requires that all SVs have vetted so a single SV unvetting a package will make it unusable and must therefore not be done unless there is an agreement.
+  All supported packages not included in this configuration are vetted.
 
 This mechanism is primarily used to unvet buggy or insecure versions in case of major issues.
 
 To unvet supported packages, SVs (but not regular validators) must set the following environment variable in their SV and validator configurations.
-
-Once at least one SV applies an unvet configuration, the unvetting will take effect within a short period.
-All supported packages not included in this configuration will be vetted.
 
 Here an example of how to unvet specific versions of a package:
 


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -371,6 +371,8 @@ case class SvAppBackendConfig(
     maxVettingDelay: NonNegativeFiniteDuration = NonNegativeFiniteDuration.ofHours(24),
     // `latestPackagesOnly=true` is intended for LocalNet testing only and is not supported in production
     latestPackagesOnly: Boolean = false,
+    // Map of package name -> set of versions that should be explicitly unvetted
```
moritzkiefer-da, line 375: can we make this `Map[PackageName, Set[PackageVersion]]`? That seems much easier to understand

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/config/ValidatorAppConfig.scala
```diff
@@ -216,6 +216,7 @@ case class ValidatorAppBackendConfig(
     // `latestPackagesOnly=true` is intended for LocalNet testing only and is not supported in production
     latestPackagesOnly: Boolean = false,
     acsStoreDescriptorUserVersion: Option[Long] = None,
+    additionalPackagesToUnvet: Map[String, Set[String]] = Map.empty,
```
moritzkiefer-da: same here

cluster/configs/shared/scratchnet.yaml
```diff
@@ -20,6 +20,9 @@ svs:
       cometbft:
         retainBlocks: 2000 # corresponds to a retention period of ~2 hours
     svApp:
+      additionalPackagesToUnvet:
```
moritzkiefer-da: remove

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PackageVettingTrigger.scala
```diff
@@ -95,4 +102,24 @@ abstract class PackageVettingTrigger(
...
+      tc: TraceContext
+  ): Seq[LfPackageId] =
+    additionalPackagesToUnvet.toSeq.flatMap { case (packageName, versions) =>
+      versions.toSeq.flatMap { version =>
+        DarResourcesUtil.lookupPackageMetadata(
+          PackageName.assertFromString(packageName),
+          PackageVersion.assertFromString(version),
+        ) match {
+          case None =>
+            logger.warn(
```
moritzkiefer-da: The code looks right but the error message looks wrong. You are not checking whether the package is uploaded. You are checking whether it is one of the known packages + versions.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PackageVettingTrigger.scala
```diff
@@ -45,18 +48,20 @@ abstract class PackageVettingTrigger(
       amuletRules <- getAmuletRules()
       voteRequests <- getVoteRequests()
       dsoRules <- getDsoRules()
+      additionalPackageIdsToUnvet = resolvePackageIdsToUnvet(additionalPackagesToUnvet)
```
moritzkiefer-da: Let's say we have this scenario:

We have an upgrade that only changed dso governance. We definitely had those in the past.
I want to unvet that version. Will it correctly handle the fact that it cannot unvet the dependencies like amulet because the prior version (or the next one) still uses it?

moritzkiefer-da: ideally let's try to have a test for this. Not sure we have a recent example of dso governance but I think we had one that only changed wallet payments which may work in a test.

julientinguely-da: when manually testing on scratchnet, I first tried this scenario and it failed to unvet the amulet yes

julientinguely-da: @moritzkiefer-da not sure I follow exactly the need of this test. Here to unvet the dependencies you would need to explicitly define the packageName->versions.

moritzkiefer-da: > I first tried this scenario and it failed to unvet the amulet yes

Well if it fails we should fix it and test that the fix works?

julientinguely-da: Added this test case in `DependentAdditionalPackagesToUnvetIntegrationTest`.

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceConfig.scala
```diff
@@ -594,6 +595,11 @@ object SpliceConfig {
       ConfigReader.fromString(str =>
         PackageVersion.fromString(str).left.map(err => CannotConvert(str, "PackageVersion", err))
       )
+    implicit val additionalPackagesToUnvetReader
+        : ConfigReader[Map[PackageName, Set[PackageVersion]]] =
+      genericMapReader(str =>
```
moritzkiefer-da, line 600: Do we actually need this, I was expecting just defining a `Configreader[PackageName]` is enough for it to infer the map? Similar for writer ofc

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
```diff
@@ -0,0 +1,271 @@
...
+import org.lfdecentralizedtrust.splice.integration.EnvironmentDefinition
+import org.lfdecentralizedtrust.splice.integration.tests.SpliceTests.IntegrationTest
+import org.lfdecentralizedtrust.splice.util.{DarResourcesUtil, PackageUnvettingUtil}
+import org.slf4j.event.Level
+
+abstract class AdditionalPackagesToUnvetIntegrationTestBase
+    extends IntegrationTest
+    with PackageUnvettingUtil {
+
+  protected val additionalPackagesToUnvet1: Seq[DarResource]
```
moritzkiefer-da: maybe call it `additionalPackagesToUnvetSv1` and `additionalPackagesToUnvetSv1Local`? otherwise it sounds like sv1 and 2 which is not what you're using it as (although maybe that's also a question: any reason we don't just use two different SVs)

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
```diff
@@ -0,0 +1,271 @@
...
+    DarResources.wallet_0_1_16,
+    DarResources.amuletNameService_0_1_16,
+    DarResources.dsoGovernance_0_1_21,
+  ) :+ problematicDar
+
+  override val additionalPackagesToUnvet1: Seq[DarResource] = darsWithMissingDependency
+  override val additionalPackagesToUnvet2: Seq[DarResource] =
+    darsWithMissingDependency :+ missingDependency
+
+  "sv1 cannot unvet packages that still have dependencies" in { implicit env =>
```
moritzkiefer-da, line 97: I'm not sure this tests the actual behavior we want. Consider this scenario: We upgrade from 0.5.8 to 0.5.9 (0.5.9 introduced wallet 0.1.16 with minting delegation).

We do a vote to upgrade to the new Daml models and somehow it blows up so we want to roll back.

The only change was to vet the new wallet version so that's also the only thing I want to unvet.

Your test seems to do more, it also unvets amulet name service and dso governance versions for some reason? And then the vetting doesn't succeed and you also have to unvet a prior version of the wallet model which doesn't sound right.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
```diff
@@ -0,0 +1,271 @@
...
+      }
+      stopAllAsync(
+        sv1LocalBackend,
+        sv1ValidatorLocalBackend,
+      ).futureValue
+    }
+  }
+}
+
+class UpgradeAdditionalPackagesToUnvetIntegrationTest
```
moritzkiefer-da: Please add some comments on which cases each of those tests is trying to cover. The names aren't all that descriptive.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PackageVettingTrigger.scala
```diff
@@ -66,16 +74,25 @@ abstract class PackageVettingTrigger(
...
       )
+      vettedPackageIds = vettedPackages.flatMap(_.mapping.packages).map(_.packageId)
       unsupportedPackages = DarResourcesUtil.filterUnsupportedPackageVersions(
-        vettedPackages.flatMap(_.mapping.packages).map(_.packageId)
+        vettedPackageIds,
+        enableUnsupportedDarsUnvetting,
+        latestPackagesOnly,
+        additionalPackagesToUnvet,
+      )
+      // Unvet unsupported packages only if they are currently vetted to avoid Canton failing conformance checks
```
moritzkiefer-da: that sounds weird, wdym by failing conformance checks?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
moritzkiefer-da: In particular I was expecting that we actually do need logic that does not unvet dependencies if they are referenced from other packages that should not yet be vetted which is what is currently missing from this test.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PackageVettingTrigger.scala
```diff
@@ -95,4 +112,12 @@ abstract class PackageVettingTrigger(
       Future.unit
     }
   }
+
+  private def resolvePackagesToUnvetFromVettingState(
+      vettedPackageIds: Seq[PackageId],
+      unsupportedPackagesIds: Seq[PackageId],
+  ): Seq[DarResource] =
+    unsupportedPackagesIds
+      .filter(vettedPackageIds.toSet)
```
moritzkiefer-da:
```suggestion
      .filter(vettedPackageIds.toSet.contains(_))
```
looks a bit confusing to rely on the apply method of Set

moritzkiefer-da: also consider using `intersect`

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceConfig.scala
julientinguely-da: I tried with only `Configreader[PackageName]`, but didn't work.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
julientinguely-da: I want to use an exact copy of sv1 to simulate a node that is restarting.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
julientinguely-da: I see. I think unvetting amuletNameService and dsoGovernance is useless/misleading here, but I strongly believe the test is what we want.

What this test tests and should test:
- Failed path: I try to unvet `wallet_0_1_16` together with `walletPayments_0_1_15` => this fails because `wallet_0_1_15` has a depedency to `walletPayments_0_1_15` => in the log it manifsests with line.message should include regex s"TOPOLOGY_DEPENDENCIES_NOT_VETTED.*${problematicDar.packageId}"
- Success path: if I want to unvet `wallet_0_1_16` together with `walletPayments_0_1_15` I have to unvet `wallet_0_1_15` too

moritzkiefer-da: > What this test tests and should test:

The test we are missing is unvetting just wallet 0.1.16 and seeing that this works.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/PackageVettingTrigger.scala
julientinguely-da: modifying the comment

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
```diff
@@ -0,0 +1,280 @@
...
+        sv1LocalBackend,
+        sv1ValidatorLocalBackend,
+      )
+      eventually() {
+        val vettedPackageIds = getVettedPackageIds(
+          sv1LocalBackend.appState.participantAdminConnection,
+          synchronizerId,
+        )
+        vettedPackageIds should contain allElementsOf darsFromAnUpgrade.map(_.packageId)
+        vettedPackageIds should not contain DarResources.wallet_0_1_16.packageId
```
julientinguely-da, line 218: @moritzkiefer-da I've added the test that unvets wallet 0.1.16 here

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
```diff
@@ -0,0 +1,280 @@
...
+      ).futureValue
+    }
+  }
+}
+
+/** This test verifies that an SV can downgrade to the versions before an upgrade.
+  */
+class DowngradeSvPackagesIntegrationTest extends AdditionalPackagesToUnvetIntegrationTestBase {
+
+  private val darsFromAnUpgrade = Seq(
```
moritzkiefer-da: Not sure what these versions are. What are we trying to unvet?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
moritzkiefer-da: The test looks reasonable, I'm a bit confused why it works though. What code filters out dependencies that are still referenced by other packages so this doesn't fail?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
julientinguely-da: Just testing a normal downgrade, from release 0.5.9 to 0.5.8 to see it works

docs/src/sv_operator/sv_operations.rst
```diff
@@ -846,3 +846,21 @@ A subsequent re-ingestion can be triggered by incrementing the value, as shown i
...
       persistence:
         txLogStoreDescriptorUserVersion: 2
+
+.. _sv-unvet_unsercure_package_versions:
+
+Unvet unsecure package versions
+-------------------------------
+
+This is mainly used to downgrade to previous versions in case of major issues.
+Do not use it if not advised to do so; unvetting a package changes the vetting state for all SVs.
```
julientinguely-da: honestly, I don't really know how to best phrase this. It feels we should maybe add guards to avoid this to happen.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
julientinguely-da: It seems canton just checks that the packages from the new vetting state does not contain `17dca10fd8eb6a833be530fe9c6f9c2b7397a38c06e9c86d0679adc200b90e14`, which it doesn't. 

I think the test with WalletPayments_0_1_15 that is referenced in both wallets 15 and 16 makes sense to that regards.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
moritzkiefer-da: If we want to test that wouldn't it make more sense to run this test with unvetAllPackages beforehand and then use the initialPackages for 0.5.9 to bootstrap the network and then unvet them again? Otherwise it's not very obvious what we vet and why.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
```diff
@@ -142,15 +151,25 @@ class PackageWithDependencyIntegrationTest extends AdditionalPackagesToUnvetInte
...
     DarResources.dsoGovernance_0_1_23,
     DarResources.walletPayments_0_1_17,
     DarResources.wallet_0_1_18,
-    DarResources.amuletNameService_0_1_18,
     DarResources.amulet_0_1_17,
   )
 
-  override val additionalPackagesToUnvetSv1: Seq[DarResource] = darsFromAnUpgrade
+  // initialize network with the versions from the upgrade 0.5.16
+  override val initialPackageConfig: InitialPackageConfig = new InitialPackageConfig(
```
moritzkiefer-da, line 166: you probably want a `withNoVettedPackages` on all of these tests to make sure a prior test vetting something doesn't persist

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
```diff
@@ -142,15 +151,25 @@ class PackageWithDependencyIntegrationTest extends AdditionalPackagesToUnvetInte
...
+    DarResources.amuletNameService_0_1_18,
     DarResources.dsoGovernance_0_1_23,
     DarResources.walletPayments_0_1_17,
     DarResources.wallet_0_1_18,
-    DarResources.amuletNameService_0_1_18,
     DarResources.amulet_0_1_17,
   )
 
-  override val additionalPackagesToUnvetSv1: Seq[DarResource] = darsFromAnUpgrade
+  // initialize network with the versions from the upgrade 0.5.16
```
moritzkiefer-da, line 165: didn't we want 0.5.9 or something like that?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/DarResourcesUtil.scala
```diff
@@ -26,21 +39,37 @@ object DarResourcesUtil {
   def lookupAllPackageVersions(name: PackageName): Seq[DarResource] =
     packageResources.view.flatMap(_.all).toSeq.filter(_.metadata.name == name)
 
+  // TODO(hyperledger-labs/splice#4049): remove enableUnsupportedDarsUnvetting once it's on MainNet
```
moritzkiefer-da: don't understand what you mean with this comment

moritzkiefer-da: are you just saying to remove the feature flag once we have confidence it's not needed?

docs/src/sv_operator/sv_operations.rst
```diff
@@ -846,3 +846,30 @@ A subsequent re-ingestion can be triggered by incrementing the value, as shown i
       # Example to trigger re-ingestion of the TxLog store for the second time
       persistence:
         txLogStoreDescriptorUserVersion: 2
+
+.. _sv-unvet_unsercure_package_versions:
+
+Unvet unsecure package versions
+-------------------------------
+
+This mechanism is primarily used to downgrade to previous versions in the event of major issues.
```
moritzkiefer-da:
```suggestion
This mechanism is primarily used to unvet buggy or insecure versions in case of major issues.
```

docs/src/sv_operator/sv_operations.rst
```diff
@@ -846,3 +846,30 @@ A subsequent re-ingestion can be triggered by incrementing the value, as shown i
...
+
+To unvet supported packages, all SVs must set the following environment variable in their SV and validator configurations.
```
moritzkiefer-da:
```suggestion
To unvet supported packages, all SVs (but not regular validators) must set the following environment variable in their SV and validator configurations.
```

docs/src/sv_operator/sv_operations.rst
```diff
@@ -846,3 +846,30 @@ A subsequent re-ingestion can be triggered by incrementing the value, as shown i
...
+Only supported versions strictly after the minimal initialization version can be unvetted.
```
moritzkiefer-da: sounds a bit confusing, the other versions are already unvetted so it's not like you're doing any harm by putting them in there.

docs/src/sv_operator/sv_operations.rst
```diff
@@ -846,3 +846,30 @@ A subsequent re-ingestion can be triggered by incrementing the value, as shown i
...
+
+Once all SVs apply the identical unvet configuration, the unvetting will take effect within a short period (up to 30 minutes).
```
moritzkiefer-da: where is the 30min coming from?

Also I think it's technically not all SVs but a single SV unvetting is already enough to break things for better or for worse. Probably worth pointing that out here including something like "only set this after agreement with at least 2/3 of the SVs that this is desired"

docs/src/sv_operator/sv_operations.rst
```diff
@@ -846,3 +846,30 @@ A subsequent re-ingestion can be triggered by incrementing the value, as shown i
...
+If all SVs remove a supported package from this configuration, the package will be vetted again.
+
+Here an example of how to unvet specific versions of a package:
+
+  .. code-block:: yaml
+
+     additionalEnvVars:
+       - name: ADDITIONAL_CONFIG_ADDITIONAL_PACKAGES_TO_UNVET
+         value: |
+           canton.sv-apps.sv.additional-packages-to-unvet.splice-package-1 = ["0.1.16"]
```
moritzkiefer-da: probably easire to undertand if you use actual package names like splice-amulet and dso-governance

moritzkiefer-da (approved):
Nice work thanks!

docs/src/sv_operator/sv_operations.rst
```diff
@@ -846,3 +846,33 @@ A subsequent re-ingestion can be triggered by incrementing the value, as shown i
...
         txLogStoreDescriptorUserVersion: 2
+
+.. _sv-unvet_unsercure_package_versions:
+
+Unvet unsecure package versions
+-------------------------------
+
+.. warning::
+
+  Only set these configurations **after agreement with at least 2/3 of the SVs** that this is desired.
```
moritzkiefer-da, line 857: maybe move the reasoning in the warning here.

Issue #4060, opened 2026-02-20
Track start of activity record ingestion in activity_record_meta table (was Fetched traffic summaries should not be empty)


adetokunbo:
This should [not be possible](https://github.com/hyperledger-labs/splice/pull/4037#discussion_r2827054775)
This issue tracks handling of this situation due to the comment raised during review of #4037

meiersi-da:
Thanks for the issue. Please assign all of them to milestone [Traffic-Based App Rewards](https://github.com/hyperledger-labs/splice/milestone/42)

PR #4557, opened 2026-03-20
Workaround for missing traffic summaries

rautenrieth-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/7528

It moves the problem to the existing https://github.com/hyperledger-labs/splice/issues/4060, see explanation in https://github.com/hyperledger-labs/splice/issues/4060#issuecomment-4098140036.

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictStoreIngestion.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictStoreIngestion.scala
@@ -3,6 +3,7 @@
 
 package org.lfdecentralizedtrust.splice.scan.automation
 
+import com.daml.grpc.GrpcException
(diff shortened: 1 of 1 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictStoreIngestion.scala
```diff
@@ -146,6 +148,25 @@ class ScanVerdictStoreIngestion(
                 DbScanVerdictStore
                   .fromProtoWithCorrelation(proto, viewHashToViewIdByTime, logger)
               })
+              // TODO(#4060): handle missing traffic summaries more robustly. In particular,
+              // note that the whole call will fails if ANY of the requested traffic summaries are missing.
```
ray-roestenburg-da:
```suggestion
              // note that the whole call will fail if ANY of the requested traffic summaries are missing.
```

PR #4576, opened 2026-03-23
Avoid storing LSU state in memory

moritzkiefer-da:
[ci]

fixes #4302

Ended up going for using a file as an intermediate. Seems less of a hassle with no huge downside.

I repurposed the HDM field as we already have that set anyway and setting another one doesn't seem very helpful.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvDsoAutomationService.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvDsoAutomationService.scala
@@ -256,6 +256,9 @@ class SvDsoAutomationService(
             synchronizerNodeService.nodes,
             successorSynchronizerNode,
             store,
+            config.domainMigrationDumpPath.getOrElse(
+              throw new IllegalArgumentException("Domain migration dump path must be set for LSU")
+            ),
           )
         )
         registerTrigger(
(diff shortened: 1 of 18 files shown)


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvDsoAutomationService.scala
```diff
@@ -256,6 +256,9 @@ class SvDsoAutomationService(
             synchronizerNodeService.nodes,
             successorSynchronizerNode,
             store,
+            config.domainMigrationDumpPath.getOrElse(
+              throw new IllegalArgumentException("Domain migration dump path must be set for LSU")
+            ),
```
nicu-da, line 261: Any reason why we don't fallback to using a tmp dir?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/lsu/LogicalSynchronizerUpgradeTrigger.scala
```diff
@@ -51,6 +53,7 @@ class LogicalSynchronizerUpgradeTrigger(
 
   private val exporter =
     new LsuStateExporter(
```
nicu-da, line 55: should we add a check if the file already exists before calling export? This way if we have the state we can just re-use it

moritzkiefer-da: The problem is that we stream to the file so the file existing does not mean it's complete. And at that point it seem more of a hassle to try to ensure we don't accidentally try to use an incomplete file by first writing to an intermediate file or whatever.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvDsoAutomationService.scala
moritzkiefer-da: Just didn't seem worth it, we always set it in the helm chart, it's not even configurable. And tmp dirs are often relatively limited in size so that may cause issues if anyone actually relies on that.

nicu-da: wfm

PR #4578, opened 2026-03-23
Allow per-SV cloudSql configuration for apps-pg databases

martinflorian-da:
See https://github.com/hyperledger-labs/splice/issues/4558

Will backport to DevNet branch once approved

ais-7

### DIFF
--- a/cluster/configs/shared/base.yaml
+++ b/cluster/configs/shared/base.yaml
@@ -124,6 +124,8 @@ monitoring:
       sequencerRateLimits:
         rejectionRateThreshold: 0
         circuitBreakerStateThreshold: 0.5
+      gcpQuotas:
+        enabled: false
     # Alert on secrets appearing in logs (JWTs, Bearer tokens, passwords, etc.)
     # Key=value secrets (excluding masked values)
     # JWTs identified by base64 header prefix "eyJhbGc" (decodes to '{"alg')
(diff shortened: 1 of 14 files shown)


cluster/deployment/mock/config.yaml
```diff
@@ -87,6 +87,9 @@ pulumiProjectConfig:
     hasPublicInfo: true
     cloudSql:
       enabled: true
+  sv-runbook:
+    cloudSql:
+      enabled: true
```
martinflorian-da, line 92: flipped this to confirm the new flag also works for `sv`; happy to revert this if someone feels strongly about it

cluster/pulumi/canton-network/src/sv.ts
```diff
@@ -245,7 +245,7 @@ export async function installSvNode(
       `cn-apps-pg`,
       `cn-apps-pg`,
       config.version,
-      spliceConfig.pulumiProjectConfig.cloudSql,
+      svConfig.appsPg?.cloudSql ?? spliceConfig.pulumiProjectConfig.cloudSql,
```
mblaze-da, line 248: I have mixed feelings about config fallback logic deep in Pulumi code. I wonder is not having a fallback and `!include`ing a base config would be too much noise?

martinflorian-da: I think you are right to question this but let us pick this up in a follow-up PR; really my main agenda here was to get this fixed quickly so I can bump a db on DevNet.

Probably the thing to fix is getting rid of `pulumiProjectConfig` though? Why do we even have that?

PR #4626, opened 2026-03-24
[ci] return URLs instead of keys in list ACS objects

isegall-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanTimeBasedIntegrationTest.scala
@@ -5,11 +5,18 @@ import com.digitalasset.canton.{HasActorSystem, HasExecutionContext}
 import com.digitalasset.canton.config.NonNegativeFiniteDuration
 import com.digitalasset.canton.data.CantonTimestamp
 import org.apache.pekko.http.scaladsl.model.Uri
-import org.lfdecentralizedtrust.splice.codegen.java.splice.amulet.{Amulet, AppRewardCoupon, ValidatorRewardCoupon}
+import org.lfdecentralizedtrust.splice.codegen.java.splice.amulet.{
+  Amulet,
+  AppRewardCoupon,
+  ValidatorRewardCoupon,
+}
 import org.apache.pekko.util.ByteString
 import org.lfdecentralizedtrust.splice.codegen.java.splice.ans.AnsEntry
 import org.lfdecentralizedtrust.splice.config.ConfigTransforms
-import org.lfdecentralizedtrust.splice.config.ConfigTransforms.{ConfigurableApp, updateAutomationConfig}
+import org.lfdecentralizedtrust.splice.config.ConfigTransforms.{
+  ConfigurableApp,
+  updateAutomationConfig,
+}
 import org.lfdecentralizedtrust.splice.console.WalletAppClientReference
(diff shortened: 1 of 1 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
ray-roestenburg-da, line 2349: would prefer if the future failed was part of implementation of getting the acsSnapshotBulkStorage from bulkStorage. Just a nicety so I don't need to read that here. Since you're going to fail hard with the Future anyway

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2349,21 +2353,29 @@ class HttpScanHandler(
...
-              definitions.ListBulkAcsSnapshotObjectsResponse(
-                Codec.encode(ts),
-                objects.map { case ObjectKeyAndChecksum(key, digest) =>
-                  definitions.BulkStorageObjectRef(
-                    // TODO(#3429): for now we return just the key, but this should be mapped to a full url in scan
-                    key,
-                    digest,
-                  )
-                }.toVector,
+        publicUrl.fold(
```
ray-roestenburg-da, line 2356: Prefer that you instead pass the publicUrl as an argument to constructing the acsSnapshotBulkStorage, so you don't even have to pass it in here, (and if the acsSnapshotBulkStorage is configured it definitely needs a publicUrl so it can be non-optional there, now you have two things that can independently be set or not set, so more corner cases) getAcsSnapshotAtOrBefore could return publicUrl via AcsSnapshotObjects, this call site doesn't really need to know about it except for encoding to BulkStorageObjectRef, and maybe you want to neatly add encoding to ScanHttpEncodings

rautenrieth-da: Not sure I like adding the publicUrl to the store. Usually stores are just responsible for fetching and storing data, wiring up the stores to the API is the responsibility of the HttpHander. The constructed download URL also has to be in sync with the OpenAPI spec, which is anther thing the store should not know about.

Can we maybe just make publicUrl a required config? Or at least add a TODO to remove the optionality when the bulk storage code is ready.

isegall-da: Yes, it's defo planned to become mandatory together with the rest of s3 config. Just felt weird to make it mandatory before the rest

ray-roestenburg-da: maybe even easier to not fold and fail on the public url cause you already fail on the storage not being there and do a publicUrl.getOrElse(encodedKey) in the encoding for now, with a comment that publicUrl will not be optional later

isegall-da: As long as it's optional, I'd like to fail if one configured s3 but not publicUrl, to be more preventive. Both of them will be mandatory fairly soon, at which point this whole sequence of `fold`s will just go away, so not too worried about the code here temporarily being slightly less polished.

PR #4627, opened 2026-03-24
Publish docs from token-standard-v2-upcoming

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4617

### DIFF
--- a/.github/workflows/post_merge_token_standard_v2.yml
+++ b/.github/workflows/post_merge_token_standard_v2.yml
@@ -17,6 +17,7 @@ jobs:
     runs-on: ubuntu-24.04
 
     permissions:
+      contents: read    # to check out the code
       pages: write      # to deploy to Pages
       id-token: write   # to verify the deployment originates from an appropriate source
 


.github/workflows/post_merge_token_standard_v2.yml
```diff
@@ -0,0 +1,64 @@
...
+    runs-on: ubuntu-24.04
+
+    permissions:
+      pages: write      # to deploy to Pages
+      id-token: write   # to verify the deployment originates from an appropriate source
+
+    # Deploy to the github-pages environment
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
```
OriolMunoz-da, line 27: I do not understand what this is supposed to do, but it was also in post_merge... lmk if it can be removed

.github/workflows/post_merge_token_standard_v2.yml
OriolMunoz-da, line 1: Basically same as post_merge, but doesn't run ci because this branch is based on canton-3.5 so it probably won't always work AND would also spam people on CI monitoring rota

This seemed the most sensible approach of doing what I wanted to do, but I definitely should not be trusted when it comes to GHA stuff!

isegall-da: I don't think this does what you want it to. Look at the top few lines of this file, this runs only on main and release-line-* branches.
I think you just want to duplicate this file for canton-3.5, and delete the parts you don't want there.

isegall-da: OK, I'm stupid, that is what you did already!

isegall-da: So IIUC, you're killing the publish_docs from main, and replacing it with a publish_docs from 3.5? Is that on purpose? Why not just have both, to different pages?

isegall-da: (see https://github.com/actions/deploy-pages/tree/854d7aa1b99e4509c4d1b53d69b7ba4eaf39215a/
I think/guess that by setting artifact_name to something like github-pages-3.5, it will not overwrite the ones from main?! Admittedly, quite hard to understand from these docs)

.github/workflows/post_merge_token_standard_v2.yml
isegall-da: Me neither, literally copy-pasted from their README at https://github.com/actions/deploy-pages/tree/854d7aa1b99e4509c4d1b53d69b7ba4eaf39215a/

.github/workflows/post_merge_token_standard_v2.yml
OriolMunoz-da: > Why not just have both, to different pages?

Looks like different pages (as in, GitHub pages specifically) are not supported: https://github.com/orgs/community/discussions/47867

That being said I might still be able to host them under some subpath I guess, like https://hyperledger-labs.github.io/splice/token-standard-v2 by putting stuff in directories

OriolMunoz-da: > I think/guess that by setting artifact_name to something like github-pages-3.5, it will not overwrite the ones from main

The way I understand this, it just means that a different artifact will be deployed, not that both can be deployed at the same time because that's just not supported by GH pages.

Afaict, to put it into a subpath we'd need to build an artifact with the docs from both branches:
- Checkout main
- Build docs for it into /tmp
- Checkout ts-v2 branch
- Build docs for it into /tmp/ts-v2
- Publish artifact
- Deploy pages

On every commit to main / ts-v2. That doesn't sound too difficult I guess. Though it assumes that the upper-level docs will not prevent the lower-level ones from showing. No idea.

OriolMunoz-da: Another option would be to just deploy the docs from/to a different repo. From Simon:

> I'd even be OK to host it from your oriol-da fork of splice
> It's very temporary and used with trusted partners.

Though then we'd have to keep the fork up2date, but that doesn't sound too bad.

PR #4632, opened 2026-03-24
list updates bulk-storage objects

isegall-da:
Sorry, this ended up being a larger PR than I had thought it would be.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorageTest.scala
+++ b/apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorageTest.scala
@@ -447,7 +447,8 @@ class UpdateHistoryBulkStorageTest
           None,
         )
         .futureValue
-      res4.objects.map(_.key) should contain theSameElementsInOrderAs Seq(d21u0, d21u1, d22u0, d22u1)
+      res4.objects
+        .map(_.key) should contain theSameElementsInOrderAs Seq(d21u0, d21u1, d22u0, d22u1)
       res4.nextPageTokenO shouldBe None
 
       // limit too low for first folder


apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/ScanAppReference.scala
```diff
@@ -775,6 +775,24 @@ abstract class ScanAppReference(
       )
     }
 
+  @Help.Summary("List all objects in bulk storage with updates between given timestamps")
+  def getBulkUpdateHistory(
+      afterTimestamp: CantonTimestamp,
+      atOrBeforeTimestamp: CantonTimestamp,
```
isegall-da: I'm not sure I like these names here (and in the in the openapi etc below). I tried to be consistent with the `atOrBefore` that we use for snapshots, but it feels less natural here, when it's the start and and of a range. OTOH, just "start" and "end" omit the minor detail that it's a (start, end] range. On the third hand, not sure anybody cares too much about that detail in the range...

rautenrieth-da: In other places, we use `startExclusive` and `endInclusive` which sounds a bit more natural to me, but I still prefer being consistent with the `after`/`atOrBefore` that we use in scan.

ray-roestenburg-da: I agree, rangeStart, rangeEnd? or a range parameter, which has start and end?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
```diff
@@ -52,21 +56,70 @@ class S3BucketConnection(
   def listObjects: Future[ListObjectsResponse] =
     s3Client.listObjects(ListObjectsRequest.builder().bucket(bucketName).build()).asScala
 
-  def listObjectsSource(prefix: String): Source[S3Object, NotUsed] = {
-    val listRequest = ListObjectsV2Request
-      .builder()
-      .bucket(bucketName)
-      .prefix(prefix)
-      .build();
+  private def s3List(
```
ray-roestenburg-da: listS3ObjectRefs? or something like this? what are the strings coming out of this, all of the content of the s3 objects concatenated, is that not problematic in size? would help to have some comments here maybe.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
```diff
@@ -52,21 +56,70 @@ class S3BucketConnection(
...
+      .mapConcat(_.toList)
+      .runWith(Sink.seq[String])
   }
 
-  def readChecksum(key: String)(implicit ec: ExecutionContext): Future[String] = {
+  def listObjects(
+      prefix: String,
+      keyFilter: String => Boolean,
+      limit: Limit,
+  )(implicit tc: TraceContext, as: ActorSystem): Future[Seq[String]] =
```
ray-roestenburg-da, line 90: Ok so I guess the Seq[String] is never going to be too big for memory or something like that.

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -4061,6 +4091,53 @@ components:
...
+        - page_size
+      properties:
+        after_record_time:
+          description: |
+            The returned objects must include all updates with record time greater than start_record_time (but may also include updates before it).
+          type: string
+          format: date-time
+        at_or_before_record_time:
+          description: |
+            The returned objects must include all updates with record time at most up_to_or_before record time (but may also include updates after it).
```
ray-roestenburg-da: the up_to_or_before reads like a parameter but does not exist

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -4061,6 +4091,53 @@ components:
           items:
             $ref: "#/components/schemas/BulkStorageObjectRef"
 
+    ListBulkUpdateHistoryObjectsRequest:
```
ray-roestenburg-da, line 4094: Would it make sense to use Refs in that name somewhere, this is going to return the refs right, not the actual objects?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -145,7 +149,7 @@ class HttpScanHandler(
     initialRound: String,
     externalTransactionHashThresholdTime: Option[Instant] = None,
     updateHistoryMaxPageSize: Int,
-    publicUrl: Option[Uri],
+    publicUrlO: Option[Uri],
```
ray-roestenburg-da, line 152: nitpick: Line 150 doesn't do the O thing, we're unlikely concistent in this, personally I'm not a huge fan of adding the trailing O everywhere, but I will not die on that hill.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
```diff
@@ -52,21 +56,70 @@ class S3BucketConnection(
...
+      listRequest: ListObjectsV2Request,
+      pageToContent: ListObjectsV2Response => Iterable[String],
+      limit: Limit,
+  )(implicit tc: TraceContext, as: ActorSystem): Future[Seq[String]] = {
     Source
       .fromPublisher(
         s3Client.listObjectsV2Paginator(listRequest)
       )
-      .mapConcat(_.contents().asScala)
+      .fold(Seq.empty[String])((acc, page) => {
```
rautenrieth-da, line 74: This will still request the entire list, the limit is applied locally. Are we concerned about performance? The list of server-side objects will grow without bounds.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2440,45 +2451,88 @@ class HttpScanHandler(
     }
   }
 
+  private def getBulkStorage(): Option[(AcsSnapshotBulkStorage, UpdateHistoryBulkStorage, Uri)] = {
```
ray-roestenburg-da, line 2454: Nitpick: could at least create a case class for the thing you're returning. I still think it makes more sense if the handler just gets refs to the storage classes and it maybe blows up if it is not configured, even at this stage, but likely I just have a different preferences

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
isegall-da: I agree that s3List is a terrible name...
It's a utility method that takes a listRequest, and a function that extracts "something" (specifically, either the object keys or folder names) from every request, runs the entire list through that, sorts it, and takes the top `limit` elements. 
No actual content (despite the argument name...).
Will improve naming and documentation..

apps/scan/src/main/openapi/scan.yaml
isegall-da: Right, but I think that "list" already suggests that (as opposed to "get"). We can make it "get...ObjectRefs", but I'm not sure that's any less confusing.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
isegall-da: Yeah, I agree. Specifically here it was because then inside I pass publicUrl around, which is already the non-option one, so I thought I'd make the distinction, but not sure that's a good enough reason. The compiler will catch any confusion anyway.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
isegall-da: This is going away very soon, so I went for the minimum that just works for now.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -197,4 +201,157 @@ class UpdateHistoryBulkStorage(
...
+    def isFolderInRange(folder: String): Boolean = {
+      storageConfig.getStartAndEndTimestampsForFolder(folder) match {
+        case Left(err) =>
+          throw io.grpc.Status.INTERNAL
+            .withDescription(
+              s"Cannot parse folder name $folder, error: $err"
+            )
+            .asRuntimeException()
+        case Right((folderStart, folderEnd)) =>
+          folderStart < atOrBeforeRecordTime && folderEnd > afterRecordTime
```
rautenrieth-da, line 221:
```suggestion
          folderStart <= atOrBeforeRecordTime && folderEnd > afterRecordTime
```
?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorageTest.scala
```diff
@@ -298,6 +301,213 @@ class UpdateHistoryBulkStorageTest
...
+      val res1b = svc
+        .getUpdatesBetweenDates(
+          CantonTimestamp.tryFromInstant(Instant.parse("2015-10-10T00:00:00Z")),
+          CantonTimestamp.tryFromInstant(Instant.parse("2015-10-30T00:00:00Z")),
+          PageLimit.tryCreate(10),
+          res1.nextPageTokenO,
+        )
+        .futureValue
+      res1b.objects.map(_.key) shouldBe empty
+      res1b.nextPageTokenO shouldBe Some("2015-10-23T00:00:00Z-Migration-1-2015-10-24T00:00:00Z/")
```
rautenrieth-da, line 403: The second response return the exact same pagination token as the first, that's to signal that the store hasn't caught up yet? That feels a bit confusing to me, wouldn't it be better to return an `503 UNAVAILABLE` response?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorageTest.scala
```diff
@@ -298,6 +301,213 @@ class UpdateHistoryBulkStorageTest
...
+      // A smaller range within the data
+      val res2 = svc
+        .getUpdatesBetweenDates(
+          CantonTimestamp.tryFromInstant(Instant.parse("2015-10-21T16:00:00Z")),
+          CantonTimestamp.tryFromInstant(Instant.parse("2015-10-21T16:00:05Z")),
+          PageLimit.tryCreate(10),
+          None,
+        )
+        .futureValue
+      res2.objects.map(_.key) should contain theSameElementsInOrderAs allObjs.slice(2, 4)
```
rautenrieth-da: Suggestion: create named variables for the `allObjs` entries, something like
```scala
 val d20u0 = "2015-10-20T00:00:00Z-Migration-1-2015-10-21T00:00:00Z/updates_0.zstd"
 val d20u1 = "2015-10-20T00:00:00Z-Migration-1-2015-10-21T00:00:00Z/updates_1.zstd",
```
and then list them explicitly here instead of using `allObjs.slice`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -197,4 +201,157 @@ class UpdateHistoryBulkStorage(
...
+              if (folderLimit <= 0) {
+                Future.successful((folderAcc, folderLimit))
+              } else {
+                getUpdateObjectsInFolder(folder).map { folderObjs =>
+                  if (folderObjs.size > folderLimit) {
+                    // Folder would exceed the limit; omit it entirely (and stop adding more by making the limit 0)
+                    if (folderAcc.isEmpty) {
+                      throw io.grpc.Status.INVALID_ARGUMENT
+                        .withDescription(
+                          s"Limit of ${limit.limit} is too low to return any objects, even from a single folder of objects"
```
rautenrieth-da, line 320: Did you check whether this could become a problem if MainNet grows to very large sizes? We probably don't want to end up in situations where we need to return >1k elements from any API request.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
ray-roestenburg-da: thanks

apps/scan/src/main/openapi/scan.yaml
ray-roestenburg-da: yeah ok, it gets quite long.. I guess it's fine

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
ray-roestenburg-da: ok

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/ScanAppReference.scala
isegall-da: @rautenrieth-da we use after/atOrBefore for a single point in time, e.g. an acs snapshot. Doesn't feel right to me for a range.
I am leaning toward a simple start/end here, nobody cares enough about the exclusive/inclusive of the exact microsecond event at the edge of the range

ray-roestenburg-da: It's not possible to be consistent across every call site and across domains / components, I would say use what is most natural and has clear intent, here I also lean towards start and end

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
isegall-da: > are we concerned

Not in the near-term. 
ATM, I'm more concerned about different S3 providers having different (potentially undocumented) order guarantees in their responses, so reading the whole list is unavoidable. The only alternative is to keep a copy of the entire list in DB, which is not crazy, but has a complexity overhead of its own, so I'm not keen on adding that any time soon.

Note that this either lists through the list of folders (one per day), or objects with the range that the user requested (a few per day, but hopefully users usually don't query the whole history), so I think it should be fine for now. Finally, since it's only on the query side, it's easy enough to improve in the future.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: no, < is correct. If the folder ends at the start of the requested range, since they're both ( .. ], then their intersection is empty.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
isegall-da: I did.
A single object holds about 700K updates ATM. The only limit here is that the user-provided limit is large enough for a full day, as I didn't want to paginate within folders. 
So this will break when we have 700M updates a day, or 8000 TPS. Happy to say "we'll deal with it then"... :)

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorageTest.scala
isegall-da: > The second response return the exact same pagination token as the first, that's to signal that the store hasn't caught up yet?

That's the current implementation, yes. I agree that it's a bit unintuitive (hence the TODO to make sure it's well documented). I'm not sure I like the 503 option much better though, unless we initially don't return anything and fail with 503 for any request for which the end time is beyond what's been dumped. Otherwise, you get a different confusing scenario where you query something, get a result with a pagination token, then query the same thing with the token you just got, and it suddenly fails with a 503. That also seems quite confusing.

So I think the two options are:
- keep as is implemented right now, document it well
- 503 whenever end-time > latest-saved-time

The latter might not be that bad actually... The downside is that one cannot just blindly use "now" for the end time, and expect to just get the latest data, but that might be fine. But then maybe we need an endpoint that gets that latest allowed timestamp?

isegall-da: note that list-objects-for-snapshot also doesn't 503 on a too-late timestamp, it simply returns the latest (so its semantics is "latest saved that is at or before the requested timestamp")

ray-roestenburg-da: Really would not like 503 for this

ray-roestenburg-da: if you can't provide data I think not found would even be better

isegall-da: WDYM "not found"? 404? I still have the same concern:
I ask for a time range, get some data with a pagination token, now I ask the exact same query with the pagination token that I just received, and I get a 404? That sounds quite non-intuitive.

ray-roestenburg-da: hm.. yeah 4xx means you did something wrong as a client, which is not the case here.. but 5xx means something bad happened at the server really, so that's not great

ray-roestenburg-da: Can you return an empty result?

isegall-da: That's exactly the current behavior, which @rautenrieth-da thought was confusing

ray-roestenburg-da: I think that's fine actually, there is no failure, if you keep asking with the token everything will transparantly work, which I think makes sense.

rautenrieth-da: Yes, on second thought, the current behavior makes sense, not sure why I thought it not intuitive before. Sorry for the noise!

PR #4636, opened 2026-03-25
Add caching store for ScanRewardsReferenceStore

dfordivam:
Fixes https://github.com/hyperledger-labs/splice/issues/4635

Caching store for `ScanRewardsReferenceStore`
based on the existing `CachingScanStore` pattern.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/CachingScanRewardsReferenceStore.scala
```diff
@@ -0,0 +1,57 @@
...
+    override protected val loggerFactory: NamedLoggerFactory,
+)(implicit override protected val ec: ExecutionContext)
+    extends ScanRewardsReferenceStore
+    with NamedLogging {
+
+  private val featuredAppPartiesCache
+      : ScaffeineCache.TracedAsyncLoadingCache[Future, CantonTimestamp, Set[String]] =
+    ScaffeineCache.buildTracedAsync[Future, CantonTimestamp, Set[String]](
+      Scaffeine()
+        .maximumSize(2L),
```
rautenrieth-da, line 35: Can you please remind me why we need two entries? In https://github.com/hyperledger-labs/splice/pull/4420 (in particular, see commit https://github.com/hyperledger-labs/splice/pull/4420/changes/31e637c9ddfb08dc9c21b99404974861b6dfaf21), we are assuming that all activity records for round N are calculated before the first record for round N+1 is calculated. The design doc also says that we assign activity records to the oldest open round.

I'm not worried about the memory consumption, but rather whether I'm missing something.

meiersi-da: I was proposing two entries so that at in case we are processing batches in parallel we don't end up trashing the cache.

dfordivam: Yeah, if we do a `parTraverse*` on the computation, which we could in our case, then having two is necessary. In future there may even be some algo to distribute the round to either of open round, and having two cached would be useful for that.

PR #4624, opened 2026-03-24
lsu - fix preflights and sv runbook deployment 

nicu-da:
fix sv runbook sv app and scan config
fix preflights to account for physical synchronizers

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a logical synchronizer upgrade test is required (from canton-3.5), comment `/lsu_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #4641, opened 2026-03-25
force running SV stacks on pulumi down to properly maintain state in sv stacks on resets

mblaze-da:
Without this, after reset, the sv stacks required a refresh to recognize that their resources were deleted.


nicu-da:
sorry, should've mentioned that I've included it in the fixes here https://github.com/hyperledger-labs/splice/pull/4624

PR #4646, opened 2026-03-25
fix: fix problems found in dependencies for token-standard

krzysztofczyz-da:


token-standard/cli/package.json
```diff
@@ -20,20 +20,20 @@
...
+    "eslint": "8.57.0",
     "msw": "^2.7.3",
+    "prettier": "^3.5.3",
     "ts-node": "^10.9.2",
     "typescript": "^5.7.3",
-    "vitest": "^2.1.4",
-    "eslint": "8.57.0",
-    "@typescript-eslint/eslint-plugin": "^7.7.1",
-    "prettier": "^3.5.3"
+    "vitest": "^4.1.1"
```
stephencompall-DA, line 37: Noting the only non-reordering change here.

PR #4662, opened 2026-03-26
Add DB versions of Splice.Amulet.CryptoHash

adetokunbo:
First part the changes that address #4382
- split out because it's a specific change that can be reviewed on its own 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
@@ -40,21 +40,17 @@ class CryptoHashEquivalenceIntegrationTest extends IntegrationTest with WalletTe
   override def environmentDefinition: SpliceEnvironmentDefinition =
     EnvironmentDefinition
       .simpleTopology1Sv(this.getClass.getSimpleName)
-      .withAdditionalSetup { implicit env =>
-        sv1Backend.participantClient.upload_dar_unless_exists(darPath)
-      }
       .withManualStart
 
   "CryptoHash equivalence (Daml == SQL)" should {
 
     "set up environment" in { implicit env =>
-      startAllSync(
-        sv1Backend,
-        sv1ScanBackend,
-      )
+      startAllSync(sv1Backend)
       svParty = sv1Backend.getDsoInfo().svParty
+      svDb = sv1Backend.appState.storage
 
-      clue("Create CryptoHashProxy") {
+      clue("Upload dar and create CryptoHashProxy") {
+        sv1Backend.participantClient.upload_dar_unless_exists(darPath)
         val createArgs = new DamlRecord(
           java.util.List.of(
             new DamlRecord.Field("owner", new Party(svParty.toProtoPrimitive))
@@ -70,10 +66,6 @@ class CryptoHashEquivalenceIntegrationTest extends IntegrationTest with WalletTe
           .collectFirst { case ce: CreatedEvent => ce.getContractId }
           .getOrElse(fail("No CreatedEvent from CryptoHashProxy create"))
       }
-      scanDb = sv1ScanBackend.appState.storage match {
-        case db: DbStorage => db
-        case s => fail(s"non-DB storage configured: ${s.getClass}")
-      }
     }
 
     allTestCaseDefs.foreach { tcDef =>
@@ -81,7 +73,7 @@ class CryptoHashEquivalenceIntegrationTest extends IntegrationTest with WalletTe
         val damlHash = clue("Daml") { exerciseDaml(tcDef.op) }
         intermediates(tcDef.description) = damlHash
         clue("SQL should match Daml") {
-          val sqlHash = scanDb
+          val sqlHash = svDb
             .query(
               sql"""select #${toSqlExpr(tcDef.op, resolveRef)}""".as[String].head,
               "test.sqlHash",
@@ -129,7 +121,8 @@ class CryptoHashEquivalenceIntegrationTest extends IntegrationTest with WalletTe
 
   private var svParty: PartyId = _
   private var proxyContractId: String = _
-  private var scanDb: DbStorage = _
+  private var svDb: DbStorage = _
+
 }
 
 object CryptoHashEquivalenceIntegrationTest {


apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V065__app_reward_hash_functions.sql
```diff
@@ -0,0 +1,67 @@
...
+  AS $$ SELECT encode(extensions.digest(s, 'sha256'), 'hex') $$
+  LANGUAGE sql IMMUTABLE;
+
+-- Hash a list of already-hashed elements.
+-- Matches: hashListInternal ts = sha256(intercalate "|" (show(length ts) :: ts))
+CREATE FUNCTION crypto_hash_list(elems text[]) RETURNS text
+  AS $$
+    SELECT encode(extensions.digest(
+      array_to_string(
+        ARRAY[COALESCE(array_length(elems, 1), 0)::text] || COALESCE(elems, ARRAY[]::text[]),
```
rautenrieth-da: You could avoid the COALESCE calls by:
- Adding `RETURNS NULL ON NULL INPUT` to the function definition (IMO not silently handling NULL input is a good idea anyway)
- Using `cardinality()` instead of `array_length()`

Also, reading the docs for `CREATE FUNCTION`, adding a `PARALLEL SAFE` is probably a good idea, as it affects query planning. I would have expected that `IMMUTABLE` implies `PARALLEL SAFE`, but the [docs](https://www.postgresql.org/docs/current/parallel-safety.html) also say "all user-defined functions are assumed to be parallel unsafe unless otherwise marked."

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbCryptoHashFunctionsTest.scala
```diff
@@ -0,0 +1,208 @@
...
+  override protected def cleanDb(
+      storage: DbStorage
+  )(implicit traceContext: TraceContext): FutureUnlessShutdown[?] =
+    resetAllAppTables(storage)
+}
+
+object DbCryptoHashFunctionsTest {
+
+  /** Scala reimplementation of Daml's CryptoHash module (Splice.Amulet.CryptoHash).
+    * Used as the test oracle to verify plpgsql hash functions produce identical output.
```
meiersi-da: Please also add a corresponding test that calls the actual Daml functions.

I'd do so by adding a `Splice.Amulet.TestTemplates` module defining a `template CryptoHashProxy` that allows evaluating calls to crypto hashes.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V065__app_reward_hash_functions.sql
```diff
@@ -0,0 +1,67 @@
...
+-- Callers that store results in bytea columns (batch_hash, root_hash) must
+-- convert with decode(result, 'hex').
+
+-- ============================================================================
+-- Primitive hash functions matching Daml CryptoHash encoding
+-- ============================================================================
+
+-- Hash a text scalar: sha256(value) → 64-char hex string.
+-- Matches: instance Hashable Text where hash = hashText . id
+CREATE FUNCTION crypto_hash_text(s text) RETURNS text
```
rautenrieth-da:
```suggestion
CREATE FUNCTION daml_crypto_hash_text(s text) RETURNS text
```

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V065__app_reward_hash_functions.sql
```diff
@@ -0,0 +1,67 @@
+-- V065: plpgsql functions for Merkle tree hash computation over app reward batches.
+-- These replicate the Daml CryptoHash module (Splice.Amulet.CryptoHash) from PR #3964:
```
rautenrieth-da:
```suggestion
-- These replicate the Daml CryptoHash module (Splice.Amulet.CryptoHash)
```

By the time we merge this to main, the above PR should be merged.

rautenrieth-da (commented):
Looks good, I'm only slightly nervous about the test strategy. We are comparing the SQL output to a re-implementation of the daml code in scala, with only two hardcoded hashes manually copied from the daml output. I think it's still fine because we don't expect changes in #3964, but I'd like to hear input on this from @meiersi-da.

Unfortunately I don't think it's possible to execute arbitrary daml functions from scala code. We have an in-memory daml engine (`com.digitalasset.canton.util.TestEngine`) but that takes a command and produces a transaction.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V065__app_reward_hash_functions.sql
adetokunbo: Fixed, applied all suggestions, thanks!

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V065__app_reward_hash_functions.sql
adetokunbo: Fixed: Rename this and two others, PTAL

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V065__app_reward_hash_functions.sql
adetokunbo: Removed

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbCryptoHashFunctionsTest.scala
adetokunbo: Fixed.
Needed to 
- change the build files to allow `com.digitalasset.canton.util.TestEngine` to be run by splice tests [cdf0afe](https://github.com/hyperledger-labs/splice/pull/4662/commits/cdf0afe39f5a25a4d536b2f05b22020b70aaf1b9)
- inline the daml functions in the proxy class; they will be removed when this branch is merged with the actual daml changes (part of [80cc9d9](https://github.com/hyperledger-labs/splice/pull/4662/commits/80cc9d95c0119948007c0dd257e798ae8f24aa2f))

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/CryptoHashEquivalenceTest.scala
```diff
@@ -0,0 +1,262 @@
...
+  case class HashVariant(tag: String, fields: Seq[String]) extends HashOp
+  case class HashMintingAllowance(provider: String, amount: String) extends HashOp
+  case class HashBatchOfMintingAllowances(allowanceHashes: Seq[String]) extends HashOp
+  case class HashBatchOfBatches(childHashes: Seq[String]) extends HashOp
+
+  // -- Derive SQL expression from HashOp ----------------------------------------
+
+  def toSqlExpr(op: HashOp): String = op match {
+    case HashText(v) =>
+      s"daml_crypto_hash_text('${escapeSql(v)}')"
```
rautenrieth-da: For a test like this is good enough. For production code, you can avoid all unsafe interpolation and manual quoting using
```scala
(
  sql"select " ++ sql"daml_crypto_hash_text($v)" ++ ...
).toActionBuilder
```

daml/splice-amulet-test/daml/Splice/Testing/CryptoHashProxy.daml
```diff
@@ -0,0 +1,93 @@
...
+
+-- | Hash a BatchOfMintingAllowances variant.
+hashBatchOfMintingAllowances : [Text] -> Text
+hashBatchOfMintingAllowances allowanceHashes =
+  hashVariant "BatchOfMintingAllowances" [hashList allowanceHashes]
+
+-- | Hash a BatchOfBatches variant.
+hashBatchOfBatches : [Text] -> Text
+hashBatchOfBatches childHashes =
+  hashVariant "BatchOfBatches" [hashList childHashes]
```
meiersi-da, line 47: Please reuse all of them from the actual Splice.Amulet.CryptoHash module.

project/BuildCommon.scala
```diff
@@ -1265,6 +1266,10 @@ object BuildCommon {
       ) // to accommodate different daml repo coding style
       .settings(
         removeTestSources,
+        // Re-add TestEngine so Splice tests can run Daml code in-memory
+        Test / unmanagedSources := Seq(
+          (Test / sourceDirectory).value / "scala/com/digitalasset/canton/util/TestEngine.scala"
+        ),
```
meiersi-da: I'm wary of this dependency, as it is only available because we fork Canton.

I'd much prefer to just use the normal participant nodes we have available, and thereby be safe both from divergences in the behavior as well as be safe from having to be able to provision this dependency.

meiersi-da: @rautenrieth-da : wdyt?

rautenrieth-da: I had proposed to use the local engine in order to save test execution time, as I only know of 3 types of scala tests used in our code:
- Unit tests (no dependencies)
- Store tests (uses a Postgres database)
- Integration tests (starts a whole CN cluster, high overhead)

You're right though that adding a dependency on a Canton internal class is not ideal, and using the already running participant would be better. I don't immediately see how one could create a `ParticipantReference` without creating a whole splice environment, but maybe I'm missing something and worst case we can add an integration test with 0 CN apps or add the test to an already existing integration test.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/CryptoHashEquivalenceTest.scala
adetokunbo: Noted, but will leave as is for this test

daml/splice-amulet-test/daml/Splice/Testing/CryptoHashProxy.daml
adetokunbo: Now tracked using a more specific issue #4749
AFAIK `Splice.Amulet.CryptoHash` is not yet on main or any other branch we can rebase of yet

project/BuildCommon.scala
meiersi-da: I think if we don't start the SV and Validator apps, we get cheap startup and can use the participant nodes from the long-running Canton instance.

What we though might have to do is to factor out the code into a dedicated .dar package that has no script dependencies. I don't know yet, as I don't fully remember the constraints around uploading packages with Daml script deps to a node; and the extra constraints around our upgrading tests.

I'd suggest that we create a ticket to move away from the test engine in a later cleanup phase, if required. The goal being that we don't jeopardize our efforts to get rid of the Canton fork.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
```diff
@@ -0,0 +1,287 @@
...
+
+import java.io.File
+import java.security.MessageDigest
+import scala.jdk.CollectionConverters.*
+
+/** Three-way integration test: Daml (real participant) == SQL (scan Postgres) == Scala oracle.
+  *
+  * This complements the store-level CryptoHashEquivalenceTest (which uses
+  * Canton's in-memory TestEngine) by validating against a real ledger and
+  * the actual scan database.
```
rautenrieth-da: Update comment, the old `CryptoHashEquivalenceTest` has been removed.

project/BuildCommon.scala
adetokunbo: Fixed: 
I converted the test to an IntegrationTest (https://github.com/hyperledger-labs/splice/pull/4662/commits/d1356fea5a7345fff4a25a4673a9ae628169d694) so it uses the normal test infrastructure instead of TestEngine.  It's shared and it's not much slower (~60s vs ~30s on my machine)
I also reverted the BUILD dependency changes.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
```diff
@@ -0,0 +1,287 @@
...
+                actAs = Seq(svParty),
+                commands = Seq(new CreateCommand(templateId, createArgs)),
+              )
+          createTx.getEvents.asScala
+            .collectFirst { case ce: CreatedEvent => ce.getContractId }
+            .getOrElse(fail("No CreatedEvent from CryptoHashProxy create"))
+        },
+      )(
+        "CryptoHashProxy contract is created",
+        contractId => contractId should not be empty,
```
rautenrieth-da: `contractId` is immutable, you don't need `actAndCheck` for this check. `actAndCheck` is used for the following pattern:

1. You execute an action exactly once (no retries). The effect of the action is not immediately observable.
2. You repeatedly poll some endpoint until some condition passes, to make sure the effect of the action is as expected (the check part has automatic retries).

I would use a simple `clue()` for the above code.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
```diff
@@ -0,0 +1,287 @@
...
+/** Three-way integration test: Daml (real participant) == SQL (scan Postgres) == Scala oracle.
+  *
+  * This complements the store-level CryptoHashEquivalenceTest (which uses
+  * Canton's in-memory TestEngine) by validating against a real ledger and
+  * the actual scan database.
+  *
+  * Uses a shared environment so the dar is uploaded and proxy created once,
+  * then each hash function gets its own test.
+  */
+class CryptoHashEquivalenceIntegrationTest extends IntegrationTest with WalletTestUtil {
```
rautenrieth-da, line 36: This might need a tag like `@org.lfdecentralizedtrust.splice.util.scalatesttags.SpliceAmulet_0_1_17` to denote that this test relies on a specific amulet version.

@meiersi-da can you please help with the version?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
```diff
@@ -0,0 +1,287 @@
...
+
+  import CryptoHashEquivalenceIntegrationTest.*
+
+  override def environmentDefinition: SpliceEnvironmentDefinition =
+    EnvironmentDefinition
+      .simpleTopology1Sv(this.getClass.getSimpleName)
```
rautenrieth-da, line 42:
```suggestion
      .simpleTopology1Sv(this.getClass.getSimpleName)
      .withAdditionalSetup(implicit env => {
       sv1Backend.participantClientWithAdminToken.upload_dar_unless_exists(darPath)
      })
```

I don't think it matters in this particular case, but `withAdditionalSetup` is the common pattern for uploading dars in tests.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
```diff
@@ -0,0 +1,292 @@
...
+      .withManualStart
+
+  "Three-way CryptoHash equivalence (Daml, SQL, Scala)" should {
+    "set up CryptoHashProxy" in { implicit env =>
+      startAllSync(
```
rautenrieth-da: It looks like we are starting the SV nodes just to get a daml party. Is it possible to allocate a random party directly, so that we don't have to start any CN apps at all?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
```diff
@@ -0,0 +1,292 @@
...
+      }
+    }
+
+    allTestCases.foreach { tc =>
+      tc.description in { implicit env =>
+        val scalaHash = toScalaExpected(tc.op)
+
+        clue(s"Daml vs Scala for ${tc.description}") {
+          val (choiceName, choiceArg) = toDamlChoiceAndArg(tc.op)
+          actAndCheck(
```
rautenrieth-da: Same here, no need for `actAndCheck` for synchronous actions.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
```diff
@@ -0,0 +1,292 @@
...
+    val modName = Ref.ModuleName.assertFromString(moduleName)
+    dar.all
+      .collectFirst { case (pkgId, pkg) if pkg.modules.contains(modName) => pkgId }
+      .getOrElse(sys.error(s"Could not find package containing $moduleName"))
+      .toString
+  }
+
+  private val templateId = new Identifier(packageId, moduleName, "CryptoHashProxy")
+
+  // -- Scala oracle (mirrors Daml CryptoHash module) --------------------------
```
rautenrieth-da: Do we actually need the scala oracle now? We run all test cases with both the scala oracle and the daml implementation. The scala oracle is only used in this test, and if I understand correctly, we only care about the equivalence of the database functions and the daml implementation.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
meiersi-da: I think we actually want to just tag it to be always ignored in the upgrade integration tests. Need to have a look where the tag you mentioned is read, and used to ignore a test. We want to add a tag that is always ignored.

moritzkiefer-da: Haven't looked at the change to see whether it should be ignored but the tag to always ignore it is `org.lfdecentralizedtrust.splice.util.scalatesttags.NoDamlCompatibilityCheck`

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
adetokunbo: Fixed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
adetokunbo: As discussed earlier, this is part of the minimal configuration

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
adetokunbo: Fixed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
adetokunbo: Fixed, It works for this case too, so I've added it

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
adetokunbo: Fixed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
adetokunbo: I'm not sure.  
when either implementations change in future, does the oracle help determine where the error lies ?
If not, we can remove it.

meiersi-da: For this code the Daml code is the authority. So no, we don't need the oracle.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
adetokunbo: Fixed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CryptoHashEquivalenceIntegrationTest.scala
adetokunbo: Fixed

PR #4673, opened 2026-03-26
[ci] Fix flaky DevelopmentFundCouponIntegrationTest history assertions

martinflorian-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/7729

Three test cases in DevelopmentFundCouponIntegrationTest assert on
listDevelopmentFundCouponHistory in bare `clue` blocks without retry
logic, but the history data comes from the async tx log ingestion
pipeline. Different wallet stores process ledger events at different
speeds, so Bob's wallet tx log may not have finished ingesting the
rejection/claim/expiry event by the time the assertion runs.

The "Withdrawing" test already had the correct pattern: it places its
history assertions inside the `actAndCheck` check function, which
retries via `eventually()`. The other three tests did not.

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
@@ -399,19 +399,21 @@ class DevelopmentFundCouponIntegrationTest
     }
 
     clue("Collected coupon is listed in listDevelopmentFundCouponHistory as claimed") {
-      val expectedStatus = httpDef.ArchivedDevelopmentFundCoupon.Status.Claimed -> None
-      // As the fund manager, Alice can view the claimed coupon
-      assertListDevelopmentFundCouponHistoryStatuses(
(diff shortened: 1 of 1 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
```diff
@@ -399,19 +399,21 @@ class DevelopmentFundCouponIntegrationTest
...
-        beneficiary,
-        Seq(expectedStatus),
-      )
-      // As the beneficiary, Bob can view the claimed coupon
-      assertListDevelopmentFundCouponHistoryStatuses(
-        bobWalletClient,
-        beneficiary,
-        Seq(expectedStatus),
-      )
+      eventually() {
```
pasindutennage-da, line 402: the goal of this change was to avoid the race conditions, where the assertions failed simply because they are still in progress.

binding them inside eventually helps, but it also has timeUntilSuccess: FiniteDuration = 20.seconds

for all practical reasons 20s should be enough IMO, for this test, but if it flakes again, we might need to consider a higher timeout

PR #4689, opened 2026-03-26
another fix for leaking entities

isegall-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4516

retryProvider was still dropping it. Decided to just terminate&parse the entity earlier

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/AppConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/AppConnection.scala
@@ -148,8 +148,13 @@ abstract class BaseAppConnection(
 }
 
 object BaseAppConnection {
+  def truncate(json: Json, maxLength: Int = 100): String = {
+    val jsonString = json.noSpaces
+    if (jsonString.length > maxLength) jsonString.take(maxLength - 3) + "..."
+    else jsonString
+  }
   final class UnexpectedHttpJsonResponse(val statusCode: StatusCode, val content: Json)
-      extends Throwable(s"Unexpected Http Json Response (status $statusCode): $content")
+      extends Throwable(s"Unexpected Http Json Response (status $statusCode): ${truncate(content)}")
   final class UnexpectedHttpMalformedJsonResponse(val statusCode: StatusCode, val content: String)
       extends Throwable(
         s"Unexpected Http response, with json header but a malformed Json content (status $statusCode): $content"


ray-roestenburg-da (approved):

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/AppConnection.scala
```diff
@@ -86,9 +87,21 @@ abstract class BaseAppConnection(
...
+              .flatMap(s =>
+                io.circe.jawn.parseByteBuffer(s.asByteBuffer) match {
+                  case Right(value) =>
+                    Future.successful(
+                      new BaseAppConnection.UnexpectedHttpJsonResponse(response.status, value)
+                    )
+                  case Left(failure) =>
+                    Future.successful(
+                      new BaseAppConnection.UnexpectedHttpNonJsonResponse(response.status)
+                    )
```
nicu-da, line 103: This looks a bit dangerous to ignore the failure.
 Also the `UnexpectedHttpNonJsonResponse` is not really true, as the content type does claim json so this is actually a malformed json response. 
Shall we add a new response type? Malformed json and also capture the failure as part of the log?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/AppConnection.scala
```diff
@@ -133,8 +146,8 @@ abstract class BaseAppConnection(
 }
 
 object BaseAppConnection {
-  final class UnexpectedHttpJsonResponse(val statusCode: StatusCode, val entity: ResponseEntity)
-      extends Throwable(s"Unexpected Http Response (status $statusCode): $entity")
+  final class UnexpectedHttpJsonResponse(val statusCode: StatusCode, val content: Json)
+      extends Throwable(s"Unexpected Http Json Response (status $statusCode): $content")
```
nicu-da: I am a bit weary about logging the full content. I am mostly worried about the content being really huge, thus in most cases have serious side effects as if we log that then it might be dropped by some log collectors because it exceeds the size.

nicu-da: Can we maybe restrict the size and print just up to a certain limit? I think we have such support in canton for sure

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/AppConnection.scala
isegall-da: I don't think we ignore it. We return an object that indicates failures, and handle that caller side. Having said that, I don't mind logging the parse error here too.

nicu-da: I think we need to log either the parse error or the body, if not debugging this won't be easy

isegall-da: yup, adding it

PR #4690, opened 2026-03-26
Add featured app parties to load tester

rautenrieth-da:
Fixes https://github.com/DACH-NY/canton-network-internal/issues/4314

Tested on a scratchnet with the following block added to the cluster config:
```
loadTester:
  enable: true
  usersPerValidator: 10
  featuredUsersPerValidator: 3
  iterationsPerMinute: 60
  minRate: 0.95
  adaptiveScenario:
    enabled: true
    scaleDownStep: 2 # lower step as by default it reduces the load too fast
```
and verified that verdict ingestion observes featured app rights

### DIFF
--- a/load-tester/src/client/validator/validator.ts
+++ b/load-tester/src/client/validator/validator.ts
@@ -24,10 +24,10 @@ export class ValidatorClient {
 
   private token: string;
   /** Expected featured app status
-  * true: this party should be featured
-  * false: this party should not be featured
-  * undefined: do not attempt to modify the featured app status of this party
-  */
+   * true: this party should be featured
+   * false: this party should not be featured
+   * undefined: do not attempt to modify the featured app status of this party
+   */
   public featured: boolean | undefined;
   private _partyId: string | undefined;
 


cluster/pulumi/common/src/config/loadTesterConfig.ts
```diff
@@ -25,6 +25,8 @@ export const LoadTesterConfigSchema = z.object({
       iterationsPerMinute: z.number().default(60),
       maxVUs: z.number().optional(),
       adaptiveScenario: LoadTesterAdaptiveScenarioConfigSchema.prefault({}),
+      usersPerValidator: z.number().default(10),
+      featuredUsersPerValidator: z.number().default(0),
```
moritzkiefer-da: any reason to not set the default to what we also expect to use so we don't need to overwrite and have a higher chance that it's in sync between CILR and devnet?

load-tester/src/client/validator/validator.ts
```diff
@@ -146,6 +148,20 @@ export class ValidatorClient {
           })
           .then(resp => jsonStringDecoder(userStatusResponse, resp.body));
       },
+      selfGrantFeatureAppRight: (): void => {
+        this.http.post.success(
+          `${this.validatorBaseUrl}/api/validator/v0/wallet/self-grant-feature-app-right`,
+          null,
+          { retry: false, headers: this.headers() },
```
moritzkiefer-da, line 160: any reason we don't need to retry this? What happens if it fails due to some temp network error?

load-tester/src/utils.ts
```diff
@@ -50,29 +50,46 @@ export function pickTwoRandomUsers(validators: ValidatorConf[]): {
...
     const senderToken = userTokens[senderIndex];
     const recipientToken = userTokens[recipientIndex];
 
-    const adminClient = new ValidatorClient(walletBaseUrl, adminToken);
-    const senderClient = new ValidatorClient(walletBaseUrl, senderToken);
-    const receipientClient = new ValidatorClient(walletBaseUrl, recipientToken);
+    const senderFeatured = userFeatured[senderIndex];
+    const recipientFeatured = userFeatured[recipientIndex];
+
+    const adminClient = new ValidatorClient(walletBaseUrl, adminToken, undefined);
```
moritzkiefer-da, line 88: why do we need to have true|false|undefined for featured instead of just true|false?

rautenrieth-da: `undefined` is "don't try to change featured status", used for the admin party, I'll add a comment.

load-tester/src/client/validator/validator.ts
rautenrieth-da: Retries are implemented in `http.post.success`, and the load balancer will attempt to correct the onboarding/featured app status before each transfer (via `doIfOnboarded`), so it should handle even long network errors.

PR #4691, opened 2026-03-26
Add support to automatically copy votes between SVs [ci] (#3084)

### LINKED ISSUE #3084: Add support to automatically copy votes between SVs
Several SV operators run more than one SV node, and always want the votes to be in sync between them.
Would be nice to add support for `copyVotesFrom: <sv name>`, then we can define e.g. DA-2 to be our primary, and all votes on DA-2 get automatically copied by DA-1.

ericmann:
Adds a `copyVotesFrom` configuration option to `SvAppBackendConfig` that enables automatic copying of governance votes (VoteRequests) from a named source SV. This supports the use case where an operator runs multiple SV nodes and wants votes kept in sync between them.

The implementation follows the established `ScheduledTaskTrigger` pattern (modeled after `FollowAmuletConversionRateFeedTrigger`). When configured, a new `CopyVotesTrigger` polls for VoteRequests where the source SV has voted but this SV has not, and automatically casts a matching vote with the same accept/reject decision and reason.

**Configuration example:**
```yaml
copy-votes-from = "Digital-Asset-2"
```

**Changes:**
- `SvAppConfig.scala`: Add `copyVotesFrom: Option[String]` to `SvAppBackendConfig`
- `CopyVotesTrigger.scala`: New `ScheduledTaskTrigger` implementation
- `SvDsoAutomationService.scala`: Conditional trigger registration + `expectedTriggerClasses`
- `CopyVotesTimeBasedIntegrationTest.scala`: Integration tests covering accept/reject vote copying, idempotency, and no-op when source SV hasn't voted

Fixes #3084

#### PR Guidelines

#### Merge Guidelines


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/CopyVotesTrigger.scala
```diff
@@ -0,0 +1,141 @@
...
+class CopyVotesTrigger(
+    override protected val context: TriggerContext,
+    store: SvDsoStore,
+    connection: SpliceLedgerConnection,
+    sourceSvName: String,
+)(implicit
+    ec: ExecutionContext,
+    mat: Materializer,
+    tracer: Tracer,
+) extends ScheduledTaskTrigger[CopyVotesTrigger.Task]() {
```
isegall-da: Shouldn't this be a `OnAssignedContractTrigger` instead? Seems more direct.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CopyVotesTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,195 @@
...
+import org.lfdecentralizedtrust.splice.codegen.java.splice.dsorules.DsoRules_OffboardSv
+import org.lfdecentralizedtrust.splice.config.ConfigTransforms
+import org.lfdecentralizedtrust.splice.integration.EnvironmentDefinition
+import org.lfdecentralizedtrust.splice.integration.tests.SpliceTests.IntegrationTestWithIsolatedEnvironment
+import org.lfdecentralizedtrust.splice.sv.automation.singlesv.CopyVotesTrigger
+import org.lfdecentralizedtrust.splice.util.SvTestUtil
+
+import scala.jdk.CollectionConverters.*
+
+class CopyVotesTimeBasedIntegrationTest
```
isegall-da: Note that simtime (TimeBased) tests are more expensive than regular ones, the default should be wallclock. I don't see that you actually use simtime here, nor think that you need to, so better make this a wallclock-time test.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/CopyVotesTrigger.scala
ericmann: `OnAssignedContractTrigger` is a better fit here. I've refactored accordingly; the filtering logic now lives in `completeTask` and returns `TaskNoop` when the conditions aren't met.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CopyVotesTimeBasedIntegrationTest.scala
ericmann: Good call. Resolved.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/CopyVotesTrigger.scala
```diff
@@ -0,0 +1,110 @@
...
+          logger.warn(
+            s"This SV party '$thisSvParty' not found in DsoRules.svs; cannot copy votes"
+          )
+          Future.successful(TaskNoop)
+        case (true, Some(myName)) =>
+          val votes = task.payload.votes.asScala
+          val sourceVoteOpt = votes.get(sourceSvName)
+          val thisSvHasVoted = votes.contains(myName)
+          sourceVoteOpt match {
+            case Some(sourceVote) if !thisSvHasVoted =>
```
isegall-da: If the source SV actually changed the vote, e.g. from accept to reject, don't we want to copy the changed vote too?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/CopyVotesIntegrationTest.scala
```diff
@@ -0,0 +1,195 @@
...
+            else svConfig
+          }(config),
+        (_, config) => ConfigTransforms.withNoVoteCooldown(config),
+        (_, config) =>
+          ConfigTransforms.updateAutomationConfig(ConfigTransforms.ConfigurableApp.Sv)(
+            _.withPausedTrigger[CopyVotesTrigger]
+          )(config),
+      )
+
+  "CopyVotesTrigger copies votes from source SV" in { implicit env =>
```
isegall-da: Note that there's some startup overhead to every new test, so I'd suggest to merge the tests if possible.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/singlesv/CopyVotesTrigger.scala
```diff
@@ -0,0 +1,110 @@
...
+              val trackingCid = task.payload.trackingCid.toScala
+                .getOrElse(task.contractId)
+              for {
+                resolvedVoteRequest <- store.getVoteRequest(trackingCid)
+                reason = new Reason(
+                  sourceVote.reason.url,
+                  s"Copied from $sourceSvName: ${sourceVote.reason.body}",
```
isegall-da:
```suggestion
                  s"Automatically Copied from $sourceSvName: ${sourceVote.reason.body}",
```
nitpick, just to make it more explicit that this was performed by automation. Mostly in case something goes wrong I guess...

isegall-da:
@ericmann sorry, this is still failing the logs check, this time with:
```
java.lang.IllegalArgumentException: Invalid unique identifier `copy-vote-host-2:1234` with missing namespace.
```
One thing you could do, since all you care about is the votes being copied, is to disable the CloseVoteRequestTrigger trigger, which is the automation that acts on the votes and tries to execute the thing that was just voted on. That's the step that fails, but it's not interesting for what you are testing here (which is also why the test itself is not failing, and we see the failure only in logs).

PR #4728, opened 2026-03-30
add autobumps for splice-shared-gha

krzysztofczyz-da:
Proof that it works: [link](https://github.com/hyperledger-labs/splice/pull/4729)

Fixes #4244

### DIFF
--- a/.github/workflows/bump_splice_shared_gha.yml
+++ b/.github/workflows/bump_splice_shared_gha.yml
@@ -1,7 +1,7 @@
 name: Bump splice-shared-gha version
 on:
   schedule:
-    - cron: '30 */2 * * *'  # Bihourly everyday
+    - cron: '30 1 * * *'  # Runs at 1.30 everyday
   workflow_dispatch:
 
 jobs:
--- a/scripts/bump-splice-shared-gha.sh
+++ b/scripts/bump-splice-shared-gha.sh
@@ -17,7 +17,6 @@ fi
 git add -u
 echo "$submodule_name version is not up to date. Creating a PR..."
 
-git add "$submodule_name"
 updated_branch="$submodule_name-bump-$(date +%Y-%m)"
 git switch -c "${updated_branch}"
 git commit -m "[ci] bump $submodule_name version to the latest (auto-generated)" -s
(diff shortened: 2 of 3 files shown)


scripts/bump-splice-shared-gha.sh
```diff
@@ -0,0 +1,37 @@
...
+submodule_name="splice-shared-gha"
+
+git submodule update --init --force --remote -- "$submodule_name"
+
+if git diff --exit-code --quiet $submodule_name; then
+  echo "$submodule_name version is up to date."
+  exit 0
+fi
+
+git add -u
```
martinflorian-da, line 17: You both `add -u` and `add "$submodule_name". Do you want just one of these? Either look fine IMO.

.github/workflows/bump_splice_shared_gha.yml
```diff
@@ -0,0 +1,35 @@
+name: Bump splice-shared-gha version
+on:
+  schedule:
+    - cron: '30 */2 * * *'  # Bihourly everyday
+  workflow_dispatch:
```
martinflorian-da, line 5: That means we can trigger it manually, right? Can we add a short README line somewhere how to do it, in case we want a an update quickly?

.github/workflows/bump_splice_shared_gha.yml
```diff
@@ -0,0 +1,35 @@
+name: Bump splice-shared-gha version
+on:
+  schedule:
+    - cron: '30 */2 * * *'  # Bihourly everyday
```
martinflorian-da: I strongly suggest running this more rarely, maybe even just once per day. A full `[ci]` run is expensive.

scripts/bump-splice-shared-gha.sh
```diff
@@ -0,0 +1,37 @@
...
+echo "$submodule_name version is not up to date. Creating a PR..."
+
+git add "$submodule_name"
+updated_branch="$submodule_name-bump-$(date +%Y-%m)"
```
martinflorian-da, line 20: nor sure why we need a new branch every month but since it's consistent with what we do elsewhere I'm OK with keeping

scripts/bump-splice-shared-gha.sh
krzysztofczyz-da: Fixed, thanks :)

.github/workflows/bump_splice_shared_gha.yml
krzysztofczyz-da: This will rarely do a git push, but sure, changed to once a day

.github/workflows/bump_splice_shared_gha.yml
krzysztofczyz-da: Added some docs

.github/workflows/bump_splice_shared_gha.yml
martinflorian-da: Ahh right, good point. Still IMHO once per day with manual override sounds fine!

PR #4732, opened 2026-03-30
ValidatorPreflightIntegrationTest: wait longer for SV runbook connection

martinflorian-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/7711

We actually were waiting for the runbook to be up; but it sometimes also needs a bit longer for its sequencer to become usable.

[static]

(Static checks only because the other thing doesn't run preflights anyway)


OriolMunoz-da (approved):
ty

PR #4755, opened 2026-03-31
Add grafana dashboard for store perf test result

jagathweerasinghe-da:
[static]

Screenshot 2026-04-01 at 11 06 02


cluster/pulumi/infra/grafana-dashboards/splice-stores/store-ingestion-performance.json
```diff
@@ -0,0 +1,312 @@
...
+  },
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "graphTooltip": 2,
+  "id": null,
+  "links": [],
+  "panels": [
+    {
+      "title": "UpdateHistory Ingestion",
+      "description": "All metrics for UpdateHistoryIngestionPerformanceTest. Left axis: time (ms, s). Right axis: counts (items, batches).",
```
OriolMunoz-da: instead of copy-pasting every visualization, you can repeat them per test_name (or whatever the label is called)

See for example the [pulumi stacks dashboard](https://grafana.global.canton.network.digitalasset.com/d/QP_wDqDnz/pulumi-operator-stacks-dashboard?orgId=1&from=now-30m&to=now&timezone=UTC&var-stack=$__all&refresh=30s). The conditions at the bottom are not copy-pasted, but instead a `"repeat": "stack"` is  set:

Image

This has the advantage that ~if~ when we add other tests, we don't need to touch the dashboards too, it'll just be picked up automatically

cluster/pulumi/infra/grafana-dashboards/splice-stores/store-ingestion-performance.json
OriolMunoz-da, line 1: I am also not sure that we gain anything by having everything into a single visualization, as opposed to one for each:
- total items can be shown as a single number since it'll be stable each month
- total time with its own graph
- total batches in its own graph, it currently looks like "0" and that's not very useful
- avg time with its own graph; otherwise putting it together with total time means that it will look close to 0 all the time or what's happening now

Image

it looks like avg time ~= total time which is very confusing

jagathweerasinghe-da: I was planning to further restructure charts in a separate PR. But, now done. PTAL.

PR #4779, opened 2026-04-01
Enable traffic ingestion by default

rautenrieth-da:
Fixes #4752

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanEventHistoryIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanEventHistoryIntegrationTest.scala
@@ -110,14 +110,6 @@ class ScanEventHistoryIntegrationTest
     withClue("Mismatch between CompactJson and ProtobufJson update ids") {
       txIdsProtobuf shouldBe txIdsCompact
     }
-
-    withClue(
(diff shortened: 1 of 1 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanEventHistoryIntegrationTest.scala
```diff
@@ -110,14 +110,6 @@ class ScanEventHistoryIntegrationTest
     withClue("Mismatch between CompactJson and ProtobufJson update ids") {
       txIdsProtobuf shouldBe txIdsCompact
     }
-
-    withClue(
-      "Traffic summaries should not be served when serveAppActivityRecordsAndTraffic is disabled"
```
rautenrieth-da, line 115: This is reducing test coverage, as there is now no test that checks whether traffic and app activity record data is hidden from API responses if the flag is off.

However, I don't want to add an entire new integration test just for verify we won't have a regression in handling this feature flag, and want to have the feature flag turned on in as many tests as possible (we did find a bug after turning it on globally).

meiersi-da: yeah, no worries. unlikely to mess that up

PR #4782, opened 2026-04-01
Wait for stores in verdict ingestion

rautenrieth-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/7821.

Otherwise we get 
```
java.lang.RuntimeException: Using acsStoreId before it was assigned
```

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbTcsStore.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbTcsStore.scala
@@ -148,20 +148,22 @@ class DbTcsStore(
     earliestArchivedAtCache.get() match {
       case some @ Some(_) => Future.successful(some)
       case None =>
-        val storeId = acsStore.acsStoreId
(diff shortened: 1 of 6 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbTcsStore.scala
```diff
@@ -148,20 +148,22 @@ class DbTcsStore(
     earliestArchivedAtCache.get() match {
       case some @ Some(_) => Future.successful(some)
       case None =>
-        val storeId = acsStore.acsStoreId
-        val migrationId = acsStore.domainMigrationId
-        storage
-          .query(
-            sql"""SELECT MIN(archived_at) FROM #$archiveTableName
+        acsStore.waitUntilAcsIngested {
```
rautenrieth-da, line 151: This is making `getEarliestArchivedAt` safe to call without explicitly waiting for initialization. We use this in almost all `DbMultiDomainAcsStore` methods.

PR #4796, opened 2026-04-01
[ci] Drop nulls only using OmitNullString type

ray-roestenburg-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4744

## Omit null `external_transaction_hash`

New optional fields break BFT consensus.

**Fix:** `OmitNullString` type in OpenAPI spec → guardrail generates `Option[OmitNullString]` → custom encoders in `ScanJsonSupport` omit the key when `None`. Type-safe accessor prevents misuse.

**Review order:** `scan.yaml` → `OmitNullString.scala` → `ScanJsonSupport.scala` → `ScanHttpEncodings.scala` → `UpdateHistorySegmentBulkStorage.scala` → `UpdateHistoryOmitNullStringComplianceTest.scala`

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,6 +12,8 @@
     - [Backwards-compatible Daml changes](#backwards-compatible-daml-changes)
     - [Daml Numerics](#daml-numerics)
   - [Message Definitions](#message-definitions)
+  - [OpenAPI / Scan API Changes](#openapi--scan-api-changes)
+    - [Modifying BFT-consensus types](#modifying-bft-consensus-types)
   - [Config Parameters](#config-parameters)
   - [Code Layout](#code-layout)
   - [Domain Specific Naming](#domain-specific-naming)
@@ -85,7 +87,7 @@ Follow these conventions:
       update_history_transactions table.
     - Expose the hash through the v2/updates and v2/updates/{update_id}
       Scan endpoints.
-    - Gate hash inclusion behind a configurable threshold date for BFT-safety.
+    - Gate hash inclusion behind a configurable threshold record time for BFT-safety.
 
     Signed-off-by: Git-Hygienic-Dev <ghd@example.com>
     ```
@@ -152,6 +154,42 @@ Overall, please refer to the `wallet.tap` command implementation for the canonic
 * Use a plural name for `repeated` fields
 * Use `string` fields with a suffix `contract_id` to store contract ids
 * Use `string` fields with a suffix `party_id` to store party ids
+
+## OpenAPI / Scan API Changes
+
+### Modifying BFT-consensus types
+
+Update history types served by the Scan API are serialized to JSON and compared across SVs for BFT
+consensus.  **Any** schema change — adding, removing, or renaming a field — will cause JSON
+divergence between SVs running different code versions, breaking consensus.
+
+For any such change you must ensure a coordination mechanism is in place (e.g., a threshold record time
+gate) so that the new schema is not emitted before all SVs have deployed the updated code.
+
+#### Adding or removing required fields
+
+Adding a new required field means new-code SVs emit a key that old-code SVs do not; removing one
+has the opposite effect.  Both break JSON equality.  Gate the change behind a threshold record time or
+equivalent mechanism so that all SVs switch over simultaneously.
+
+#### Adding optional fields
+
+Optional fields are especially easy to add without realizing the consequences: circe emits `null`
+for `Option[_]` fields by default instead of omitting them, so new-code SVs emit `"field": null`
+while old-code SVs omit the key entirely.
+
+If you need to add an optional field to any type in the update history schema:
+
+1. **Prefer making the field required** with a sensible default — this avoids the problem entirely
+   (though you still need a gate for the new key itself).
+2. If the field must be optional, use `Option[OmitNullString]` (via `x-scala-type` in the OpenAPI
+   spec) and wire it into `ScanJsonSupport` using the `omitWhenNone` helper so the key is omitted
+   from JSON when `None`.
+3. Add the new type (if it is a new leaf type) to the compliance check in
+   `UpdateHistoryOmitNullStringComplianceTest` — this test verifies that no `Option[_]` fields
+   exist on BFT-consensus types unless they are `Option[OmitNullString]`.
+
+See `ScanJsonSupport.scala` for the full set of custom encoders and their documentation.
 
 ## Config Parameters
 


ray-roestenburg-da:
Updated to only drop nulls for responses containing `external_transaction_hash`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanJsonSupport.scala
```diff
@@ -0,0 +1,61 @@
...
+  // -- Internal backfilling response types --------------------------------------------------------
+
+  /** Internal: getUpdatesBefore */
+  implicit val getUpdatesBeforeResponseMarshaller: ToEntityMarshaller[GetUpdatesBeforeResponse] =
+    CirceJsonSupport.dropNullsMarshaller[GetUpdatesBeforeResponse]
+
+  /** Internal: getImportUpdates */
+  implicit val getImportUpdatesResponseMarshaller: ToEntityMarshaller[GetImportUpdatesResponse] =
+    CirceJsonSupport.dropNullsMarshaller[GetImportUpdatesResponse]
+}
```
isegall-da: Not 100% sure about these two. @rautenrieth-da should we include them?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/http/GeneratedScanRouteDropNullsTest.scala
```diff
@@ -0,0 +1,104 @@
...
+import org.scalatest.wordspec.AnyWordSpec
+
+import scala.collection.immutable.SortedMap
+import scala.concurrent.Future
+import scala.concurrent.duration.*
+
+/** Proves that the **generated** guardrail route code drops null values from JSON responses
+  * only for the specific response types that have a [[ScanJsonSupport]] marshaller.
+  */
+@SuppressWarnings(Array("com.digitalasset.canton.DiscardedFuture"))
```
isegall-da: This almost always a bad idea to suppress. Usually in tests we just do `.futureValue` to wait for the future. Is that not relevant here?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/http/GeneratedScanRouteDropNullsTest.scala
```diff
@@ -0,0 +1,104 @@
...
+        "/api/scan/v2/updates",
+        HttpEntity(ContentTypes.`application/json`, """{"page_size": 1}"""),
+      ) ~> route ~> check {
+        status shouldBe StatusCodes.OK
+        val json =
+          io.circe.parser.parse(responseBodyString).valueOr(e => fail(s"Invalid JSON: $e"))
+
+        // external_transaction_hash was None, so must be absent from the JSON
+        val txJson = json.hcursor.downField("transactions").downArray
+        txJson.downField("external_transaction_hash").focus shouldBe None
```
isegall-da, line 80: Hmmm... Does this actually test that it's absent? If you remove the import, does this fail?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/http/GeneratedScanRouteDropNullsTest.scala
isegall-da: (I can't immediately spot which future we're even discarding TBH, but I also can't see anything that I think should be discarded)

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/http/GeneratedScanRouteDropNullsTest.scala
isegall-da: Ah, I see below that when it's not absent, it's Some(Json.Null).

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/CirceJsonSupport.scala
```diff
@@ -0,0 +1,28 @@
...
+object CirceJsonSupport {
+  private val dropNullsPrinter: Printer =
+    Printer.noSpaces.copy(dropNullValues = true)
+
+  /** Creates a [[ToEntityMarshaller]] that encodes `A` to JSON with null values dropped.
+    */
+  def dropNullsMarshaller[A](implicit encoder: Encoder[A]): ToEntityMarshaller[A] =
+    Marshaller.withFixedContentType(MediaTypes.`application/json`) { value =>
+      val json = encoder(value).deepDropNullValues
+      HttpEntity(MediaTypes.`application/json`, dropNullsPrinter.print(json))
```
rautenrieth-da: Do we need `.deepDropNullValues`? It looks to me like `dropNullsPrinter` is already excluding all null values.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/ScanJsonSupport.scala
rautenrieth-da: Yes, we should include them. These endpoints are internal, used by the scan app itself for replicating history from other scans. 

`ScanHistoryBackfillingIntegrationTest` would fail if we somehow failed to read responses with missing optional fields.

isegall-da: Don't we do bft reads in backfilling? Won't those break if this is applied asynchronously across nodes?

rautenrieth-da: 1. We don't have any SVs that do backfilling at the moment (apart from sv runbook on CILR, and there we upgrade in sync)
2. The trigger would just emit a bunch of warnings and keep trying until the BFT reads succeed

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/CirceJsonSupport.scala
ray-roestenburg-da: dropNullPrinter (according to circe code) only drops nulls in objects.
deepDrop: ` Drop the entries with a null value if this is an object or array.`
dropNullValues: `   * Drop the entries with a null value if this is an object. Note that this does not apply recursively.`

But can check to be safe. In any case I'm more doubtful about this change. it will likely have unexpected side effects on reading existing data

ray-roestenburg-da: yes this actually failed in diff between bulk and API first, now that they both deepDrop, it matches.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/http/GeneratedScanRouteDropNullsTest.scala
ray-roestenburg-da: It's because of mocking. I think that's fine.

isegall-da: Not sure I'm convinced. We use mocking in all unit tests, and never need to add this. Can you be more explicit what futures we're dropping and why it can't be avoided?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/http/UpdateHistoryOmitNullStringComplianceTest.scala
```diff
@@ -0,0 +1,211 @@
...
+  * SVs that don't have the field omit the key entirely — causing JSON divergence.
+  *
+  * The only permitted optional type at this time is `Option[OmitNullString]`, which is handled by
+  * [[ScanJsonSupport]] to omit the key from JSON when `None`.
+  *
+  * Any other `Option[_]` field on these types will cause this test to fail, guiding the
+  * developer to either make the field required or use `OmitNullString` with the necessary
+  * wiring in `ScanJsonSupport`.
+  */
+class UpdateHistoryOmitNullStringComplianceTest extends AnyWordSpec with BaseTest {
```
isegall-da, line 46: This test will only catch new Optionals at the top of the object, right? Not nested. (otherwise I'd expect it to fail already, as I think we do have optional fields already, don't we?).
It's a good start of course, but let's make sure we have at least an issue to add tests that will catch all breaking changes to BFT endpoints.

isegall-da: Ah, you just ignore existing ones in allowedLegacyOptionFields. So this is a deep search?

I think that regardless I'd still like a test that will capture all breaking changes in BFT endpoints, e.g. if we decide to rename some field, or change return types, or whatever.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/http/GeneratedScanRouteDropNullsTest.scala
ray-roestenburg-da: We already do this, for instance:
```
@SuppressWarnings(Array("com.digitalasset.canton.DiscardedFuture"))
class BftScanConnectionTest
```

The discarded future is a `Future.succesful` in the mocked response, it's harmless in this case. I completely agree in other cases, you never want to discard a future normally. I removed the suppress warning and explicitly assign to `_`.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/http/UpdateHistoryOmitNullStringComplianceTest.scala
ray-roestenburg-da: It's not a deep search, but because I've added all the types that are currently in the update responses in openAPI spec (I found some more so I've added them), this is already really quite good I think. It does not cover if you add a new type to an exiting object, and then within that new type you add an optional field. I've added a note for that. I think that's good enough for now.

> I think that regardless I'd still like a test that will capture all breaking changes in BFT endpoints, e.g. if we decide to rename some field, or change return types, or whatever.

I feel that's beyond the scope of this issue (omitting nulls in a controlled manner, dealing with optional fields), I've added an issue for this: https://github.com/hyperledger-labs/splice/issues/4953

I've also added some doc in contributing.md

CONTRIBUTING.md
```diff
@@ -153,6 +155,42 @@ Overall, please refer to the `wallet.tap` command implementation for the canonic
...
 
+## OpenAPI / Scan API Changes
+
+### Modifying BFT-consensus types
+
+Update history types served by the Scan API are serialized to JSON and compared across SVs for BFT
+consensus.  **Any** schema change — adding, removing, or renaming a field — will cause JSON
+divergence between SVs running different code versions, breaking consensus.
+
+For any such change you must ensure a coordination mechanism is in place (e.g., a threshold date
```
isegall-da:
```suggestion
For any such change you must ensure a coordination mechanism is in place (e.g., a threshold record time
```

CONTRIBUTING.md
```diff
@@ -153,6 +155,42 @@ Overall, please refer to the `wallet.tap` command implementation for the canonic
...
+gate) so that the new schema is not emitted before all SVs have deployed the updated code.
```
isegall-da, line 167:
```suggestion
gate) so that every single update uses a consistent schema across all SVs.
```

CONTRIBUTING.md
```diff
@@ -153,6 +155,42 @@ Overall, please refer to the `wallet.tap` command implementation for the canonic
...
+
+#### Adding or removing required fields
+
+Adding a new required field means new-code SVs emit a key that old-code SVs do not; removing one
+has the opposite effect.  Both break JSON equality.  Gate the change behind a threshold date or
```
isegall-da:
```suggestion
has the opposite effect.  Both break JSON equality.  Gate the change behind a threshold record time or
```

CONTRIBUTING.md
```diff
@@ -153,6 +155,42 @@ Overall, please refer to the `wallet.tap` command implementation for the canonic
...
+while old-code SVs omit the key entirely.
+
+If you need to add an optional field to any type in the update history schema:
+
+1. **Prefer making the field required** with a sensible default — this avoids the problem entirely
+   (though you still need a gate for the new key itself).
+2. If the field must be optional, use `Option[OmitNullString]` (via `x-scala-type` in the OpenAPI
+   spec) and wire it into `ScanJsonSupport` using the `omitWhenNone` helper so the key is omitted
+   from JSON when `None`.
+3. Add the new type (if it is a new leaf type) to the compliance check in
```
isegall-da, line 188: Hmmm... Why is this needed? Isn't the whole point of the test that if the new type is correctly an OmitNullString, then the test does pass?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/http/UpdateHistoryOmitNullStringComplianceTest.scala
isegall-da: > I feel that's beyond the scope of this issue (omitting nulls in a controlled manner, dealing with optional fields), I've added an issue for this

Yes, definitely agree. Just wanted to make sure we create a followup issue for that.

CONTRIBUTING.md
ray-roestenburg-da: a completely new leaf type (that later on might contain a new optional type). (the case that is not covered cause it's not checking recursively, not when you are adding an optional string to an existing type)

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/http/GeneratedScanRouteDropNullsTest.scala
isegall-da: Thanks (I definitely prefer the explicit assigning to `_` in the places where it's explicitly identified as harmless over suppressing globally for the test, and risking future changes where "real" futures are unintentionally dropped)

PR #4797, opened 2026-04-01
Add ignore 

pasindutennage-da:


project/ignore-patterns/canton-standalone-sv123-reonboarding.ignore.txt
```diff
@@ -22,3 +22,6 @@ Sequencing result message timed out
 Could not send a time-advancing message
 
 Request failed for sequencer.* Is the server running.*Causes: finishConnect.*
+
+#Todo(DACH-NY/cn-test-failures/7808)
+The synchronizer Synchronizer 'global' failed the following topology transactions
```
moritzkiefer-da: any reason we're not just ignoring that globally? afaict Canton is just kinda stupid and these get retried and there is zero reason why they should be a warning.

Let's also make sure we report it to Canton

pasindutennage-da: Fixed

PR #4800, opened 2026-04-01
Add alert on too many featured app rights

rautenrieth-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4491


cluster/pulumi/infra/grafana-alerting/traffic_based_rewards_alerts.yaml
```diff
@@ -0,0 +1,62 @@
...
+          - refId: A
+            relativeTimeRange:
+              from: 600
+              to: 0
+            datasourceUid: prometheus
+            model:
+              editorMode: code
+              expr: max by (namespace)(pg_stat_user_tables_n_live_tup{relname="scan_rewards_reference_store_active"})
+              instant: true
+              intervalMs: 1000
```
meiersi-da, line 21: Is this the frequency of alert evaluation? We could bump that to 1 minute to save resources, or is that something we don't do?

cluster/pulumi/infra/grafana-alerting/traffic_based_rewards_alerts.yaml
```diff
@@ -0,0 +1,62 @@
...
+              type: threshold
+        dashboardUid: "00000039"
+        panelId: 74
+        noDataState: $NODATA
+        execErrState: Alerting
+        for: 1m
+        annotations:
+          __dashboardUid__: "00000039"
+          __panelId__: "74"
+          description: Scan rewards reference store on namespace {{ $labels.namespace }} has too many live rows. Check the number of active featured app right contracts.
```
meiersi-da:
```suggestion
          description: |
            Scan rewards reference store on namespace {{ $labels.namespace }} has too many live rows,
            which is likely because there are too many featured app right contracts.
            This may become a problem as the app activity record computation loads
            all featured app right contracts at round start into memory at once.
            
            Please investigate the reason for there being that many active featured
            app right contracts and determine the best course of action.
            (Note: bumping the alert limit may be OK, if the Scan app has enough memory available.)          
```

rautenrieth-da: Thanks, the detailed instructions on how to resolve the alert sound very useful!

cluster/pulumi/infra/grafana-alerting/traffic_based_rewards_alerts.yaml
rautenrieth-da: This number affects the resolution of data, but in this case it has no effect because the query is an instant query (it does not do any aggregation over time). 

I have bumped the evaluation frequency from 1m to 5min further above though.

PR #4831, opened 2026-04-02
Add dpm-sdk via nix

dasormeter:
#### PR Guidelines
- This PR lays the groundwork for making `dpm` available to splice repo, by installing the dpm sdk via nix and adding in the key damlc and daml-script components that are needed.

##### Follow-on Ideas
- A separate PR will cover changing the sbt DamlPlugin to make use of dpm.  However, merging that will need to wait on the daml projects within splice first upgrading to use a later 3.5.x sdk-version instead of the current 3.3.0-snapshot, which is too old to be supported by dpm.
- Another separate PR can also use the canton jar that is pulled in from dpm in place of whatever canton jar is being pulled in today -- this is purely optional since there is already a derivation that pulls canton jar from the canton.io site


nix/dpm-sdk.nix
```diff
@@ -0,0 +1,49 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+let
+  dpmVersion = "3.5.0-pre-snapshot.20260130.227.0.vdd0bd833";
```
isegall-da: Let's please move the version and the hashes to a standalone json file, as we do for other components. Will make upgrades much cleaner

nix/dpm.nix
```diff
@@ -0,0 +1,47 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+let
+  dpmPath = "europe-docker.pkg.dev/da-images/public/components/dpm";
+  dpmVersion = "1.0.10";
```
isegall-da: Same comment as above re placement of version-specific values

dasormeter: broke it out

nix/dpm-sdk.nix
dasormeter: moved

isegall-da:
Stole and extended in #5055

PR #4847, opened 2026-04-06
Traffic Based App Rewards: Compute Hashes

### LINKED ISSUE #4382: Scan app: Traffic-based App Rewards: compute hashes

adetokunbo:
Fixes #4382
Follows #4662 
Tracked in #4384

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/openapi/scan.yaml
+++ b/apps/scan/src/main/openapi/scan.yaml
@@ -4134,6 +4134,7 @@ components:
           description: Hex-encoded root hash
 
     GetRewardAccountingBatchResponse:
+      type: object
       oneOf:
         - "$ref": "#/components/schemas/RewardAccountingBatchOfBatches"
         - "$ref": "#/components/schemas/RewardAccountingBatchOfMintingAllowances"
@@ -4143,35 +4144,45 @@ components:
           BatchOfBatches: "#/components/schemas/RewardAccountingBatchOfBatches"
           BatchOfMintingAllowances: "#/components/schemas/RewardAccountingBatchOfMintingAllowances"
 
-    RewardAccountingBatchOfBatches:
+    BaseRewardAccountingBatch:
       type: object
       required:
         - batch_type
-        - child_hashes
       properties:
         batch_type:
           type: string
-          enum: [BatchOfBatches]
-        child_hashes:
-          type: array
-          items:
-            type: string
-          description: Hex-encoded child batch hashes
+          description: |
+            The type of batch.
+            BatchOfBatches: Contains child batch hashes.
+            BatchOfMintingAllowances: Contains party + amount pairs.
+
+    RewardAccountingBatchOfBatches:
+      type: object
+      allOf:
+        - "$ref": "#/components/schemas/BaseRewardAccountingBatch"
+        - type: object
+          required:
+            - child_hashes
+          properties:
+            child_hashes:
+              type: array
+              items:
+                type: string
+              description: Hex-encoded child batch hashes
 
     RewardAccountingBatchOfMintingAllowances:
       type: object
-      required:
-        - batch_type
-        - minting_allowances
-      properties:
-        batch_type:
-          type: string
-          enum: [BatchOfMintingAllowances]
-        minting_allowances:
-          type: array
-          items:
-            "$ref": "#/components/schemas/RewardAccountingMintingAllowance"
-          description: Party + amount pairs
+      allOf:
+        - "$ref": "#/components/schemas/BaseRewardAccountingBatch"
+        - type: object
+          required:
+            - minting_allowances
+          properties:
+            minting_allowances:
+              type: array
+              items:
+                "$ref": "#/components/schemas/RewardAccountingMintingAllowance"
+              description: Party + amount pairs
 
     RewardAccountingMintingAllowance:
       type: object
--- a/apps/scan/src/main/openapi/scan.yaml
+++ b/apps/scan/src/main/openapi/scan.yaml
@@ -1665,6 +1665,31 @@ paths:
         "404":
           $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/404"
 
+  /v0/reward-accounting-process/rounds/{round_number}/root-hash:
+    get:
+      tags: [internal, scan]
+      x-jvm-package: scan
+      operationId: "getRewardAccountingRootHash"
+      description: |
+        SV node internal API (CIP-0104, subject to change).
+        Returns the root hash computed for the specified round.
+      parameters:
+        - name: "round_number"
+          in: "path"
+          required: true
+          schema:
+            type: integer
+            format: int64
+      responses:
+        "200":
+          description: ok
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/GetRewardAccountingRootHashResponse"
+        "404":
+          $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/404"
+
   /v0/history/bulk/acs:
     get:
       tags: [ pre-alpha, scan ]
@@ -4061,3 +4086,16 @@ components:
         activity_records_count:
           type: integer
           format: int64
+
+    GetRewardAccountingRootHashResponse:
+      type: object
+      required:
+        - round_number
+        - root_hash
+      properties:
+        round_number:
+          type: integer
+          format: int64
+        root_hash:
+          type: string
+          description: Hex-encoded root hash
--- a/apps/scan/src/main/openapi/scan.yaml
+++ b/apps/scan/src/main/openapi/scan.yaml
@@ -4134,23 +4134,44 @@ components:
           description: Hex-encoded root hash
 
     GetRewardAccountingBatchResponse:
+      oneOf:
+        - "$ref": "#/components/schemas/RewardAccountingBatchOfBatches"
+        - "$ref": "#/components/schemas/RewardAccountingBatchOfMintingAllowances"
+      discriminator:
+        propertyName: batch_type
+        mapping:
+          BatchOfBatches: "#/components/schemas/RewardAccountingBatchOfBatches"
+          BatchOfMintingAllowances: "#/components/schemas/RewardAccountingBatchOfMintingAllowances"
+
+    RewardAccountingBatchOfBatches:
       type: object
       required:
         - batch_type
+        - child_hashes
       properties:
         batch_type:
           type: string
-          enum: [BatchOfBatches, BatchOfMintingAllowances]
+          enum: [BatchOfBatches]
         child_hashes:
           type: array
           items:
             type: string
-          description: Hex-encoded child batch hashes (present for BatchOfBatches)
+          description: Hex-encoded child batch hashes
+
+    RewardAccountingBatchOfMintingAllowances:
+      type: object
+      required:
+        - batch_type
+        - minting_allowances
+      properties:
+        batch_type:
+          type: string
+          enum: [BatchOfMintingAllowances]
         minting_allowances:
           type: array
           items:
             "$ref": "#/components/schemas/RewardAccountingMintingAllowance"
-          description: Party + amount pairs (present for BatchOfMintingAllowances)
+          description: Party + amount pairs
 
     RewardAccountingMintingAllowance:
       type: object
--- a/apps/scan/src/main/openapi/scan.yaml
+++ b/apps/scan/src/main/openapi/scan.yaml
@@ -1690,6 +1690,39 @@ paths:
         "404":
           $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/404"
 
+  /v0/reward-accounting-process/rounds/{round_number}/batches/{batch_hash}:
+    get:
+      tags: [internal, scan]
+      x-jvm-package: scan
+      operationId: "getRewardAccountingBatch"
+      description: |
+        SV node internal API (CIP-0104, subject to change).
+        Returns the contents of a reward batch identified by its hash.
+        The response is either a list of child batch hashes (for internal nodes)
+        or a list of minting allowances (for leaf nodes).
+      parameters:
+        - name: "round_number"
+          in: "path"
+          required: true
+          schema:
+            type: integer
+            format: int64
+        - name: "batch_hash"
+          in: "path"
+          required: true
+          schema:
+            type: string
+            description: Hex-encoded batch hash
+      responses:
+        "200":
+          description: ok
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/GetRewardAccountingBatchResponse"
+        "404":
+          $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/404"
+
   /v0/history/bulk/acs:
     get:
       tags: [ pre-alpha, scan ]
@@ -4099,3 +4132,33 @@ components:
         root_hash:
           type: string
           description: Hex-encoded root hash
+
+    GetRewardAccountingBatchResponse:
+      type: object
+      required:
+        - batch_type
+      properties:
+        batch_type:
+          type: string
+          enum: [BatchOfBatches, BatchOfMintingAllowances]
+        child_hashes:
+          type: array
+          items:
+            type: string
+          description: Hex-encoded child batch hashes (present for BatchOfBatches)
+        minting_allowances:
+          type: array
+          items:
+            "$ref": "#/components/schemas/RewardAccountingMintingAllowance"
+          description: Party + amount pairs (present for BatchOfMintingAllowances)
+
+    RewardAccountingMintingAllowance:
+      type: object
+      required:
+        - provider
+        - amount
+      properties:
+        provider:
+          type: string
+        amount:
+          type: string
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
@@ -2498,32 +2498,28 @@ class HttpScanHandler(
               case Some(batch: DbScanAppRewardsStore.BatchOfBatches) =>
                 ScanResource.GetRewardAccountingBatchResponse.OK(
                   definitions.GetRewardAccountingBatchResponse(
-                    batchType =
-                      definitions.GetRewardAccountingBatchResponse.BatchType.BatchOfBatches,
-                    childHashes = Some(
-                      batch.childHashes
-                        .map(_.toHex)
-                        .toVector
-                    ),
-                    mintingAllowances = None,
+                    definitions.RewardAccountingBatchOfBatches(
+                      batchType =
+                        definitions.RewardAccountingBatchOfBatches.BatchType.BatchOfBatches,
+                      childHashes = batch.childHashes.map(_.toHex).toVector,
+                    )
                   )
                 )
               case Some(batch: DbScanAppRewardsStore.BatchOfMintingAllowances) =>
                 ScanResource.GetRewardAccountingBatchResponse.OK(
                   definitions.GetRewardAccountingBatchResponse(
-                    batchType =
-                      definitions.GetRewardAccountingBatchResponse.BatchType.BatchOfMintingAllowances,
-                    childHashes = None,
-                    mintingAllowances = Some(
-                      batch.allowances
+                    definitions.RewardAccountingBatchOfMintingAllowances(
+                      batchType =
+                        definitions.RewardAccountingBatchOfMintingAllowances.BatchType.BatchOfMintingAllowances,
+                      mintingAllowances = batch.allowances
                         .map(a =>
                           definitions.RewardAccountingMintingAllowance(
                             provider = a.provider,
                             amount = a.amount.toString,
                           )
                         )
-                        .toVector
-                    ),
+                        .toVector,
+                    )
                   )
                 )
             }
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
@@ -2499,8 +2499,7 @@ class HttpScanHandler(
                 ScanResource.GetRewardAccountingBatchResponse.OK(
                   definitions.GetRewardAccountingBatchResponse(
                     definitions.RewardAccountingBatchOfBatches(
-                      batchType =
-                        definitions.RewardAccountingBatchOfBatches.BatchType.BatchOfBatches,
+                      batchType = "BatchOfBatches",
                       childHashes = batch.childHashes.map(_.toHex).toVector,
                     )
                   )
@@ -2509,8 +2508,7 @@ class HttpScanHandler(
                 ScanResource.GetRewardAccountingBatchResponse.OK(
                   definitions.GetRewardAccountingBatchResponse(
                     definitions.RewardAccountingBatchOfMintingAllowances(
-                      batchType =
-                        definitions.RewardAccountingBatchOfMintingAllowances.BatchType.BatchOfMintingAllowances,
+                      batchType = "BatchOfMintingAllowances",
                       mintingAllowances = batch.allowances
                         .map(a =>
                           definitions.RewardAccountingMintingAllowance(
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
@@ -2463,7 +2463,7 @@ class HttpScanHandler(
               ScanResource.GetRewardAccountingRootHashResponse.OK(
                 definitions.GetRewardAccountingRootHashResponse(
                   roundNumber = rootHash.roundNumber,
-                  rootHash = rootHash.rootHash.toByteArray.map("%02x".format(_)).mkString,
+                  rootHash = rootHash.rootHash.toHex,
                 )
               )
           }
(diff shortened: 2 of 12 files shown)


apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
```diff
@@ -614,6 +614,344 @@ class DbScanAppRewardsStoreTest
...
+          )
+        )
+        _ <- store.insertAppRewardPartyTotals(
+          (0 to 2).map(i =>
+            AppRewardPartyTotalT(historyId, roundNumber, i, BigDecimal(s"${i + 1}.0"))
+          )
+        )
+        _ <- store.computeRewardHashes(roundNumber, batchSize = 100)
+        batchHashes <- store.getAppRewardBatchHashesByRound(roundNumber)
+        rootHash <- store.getAppRewardRootHashByRound(roundNumber)
```
dfordivam: Perhaps the basic stuff like rootHash could be checked for all scenarios. Factoring out this and hashsize checks, into a separate API and apply to all tests would improve readibility

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -463,6 +476,90 @@ class DbScanAppRewardsStore(
...
+              and round_number = $roundNumber
+              and batch_level = ${parentLevel - 1}
+              and party_seq_num_begin_incl >= $beginIncl
+              and party_seq_num_end_excl <= $endExcl
+            order by party_seq_num_begin_incl
+      """.as[Array[Byte]],
+      "appRewards.lookupBatchByHash.children",
+    ).map(_.map(ByteString.copyFrom))
+
+  private def getLeafAllowances(
```
dfordivam, line 592: Add comment about coalesce

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -269,6 +275,13 @@ class DbScanAppRewardsStore(
     )
   }
 
+  /** DBIO action to read the total activity weight for a round. */
+  private def getAppActivityRoundTotalWeightAction(roundNumber: Long) =
+    sql"""select total_round_app_activity_weight
+          from #${Tables.appActivityRoundTotals}
+          where history_id = $historyId and round_number = $roundNumber
+    """.as[Long].head
```
dfordivam, line 328: Perhaps comment on why `head` is ok here (I assume that caller has to be cautious in using this?) The alternate would be to return option and throw on the caller side.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -576,52 +693,185 @@ class DbScanAppRewardsStore(
...
+    runUpdate(
+      computeRewardTotalsAction(roundNumber, params)
+        .map(_ => logger.debug(s"Computed reward totals for round $roundNumber."))
+        .transactionally,
+      "appRewards.computeRewardTotals",
+    )
+  }
+
+  /** DBIO action for computing reward totals. */
+  private def computeRewardTotalsAction(
```
dfordivam, line 764: Here too, some more comments would be appreciated.

dfordivam (commented):
Posting some comment, still doing the review of sqls

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
adetokunbo: Please suggest other scenarios, I'm not sure what that means in this context

dfordivam: I meant 
    "computeRewardHashes — exact boundary: 4 parties, batchSize=2"
"computeRewardHashes — 3 levels: 9 parties, batchSize=2" in
etc

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: FYI: this PR's change turns an existing var containing SQL into a private function returning a DBIO, so that it can be used by computeAndStoreRewards.  The SQL performs various tasks, and was reviewed for an earlier PR.  

Please explain what you would like clarified, and I'll expand the comment

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
dfordivam: actually now I recall this won't pass CI, `head` is not allowed by static checks.

adetokunbo: I'm looking at this now.  FYI on github this has passed CI, including the static check.

adetokunbo: Fixed, added a comment explaining why this is safe

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
dfordivam: Ok, it wasn't obvious from this diff, but now I see

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -463,6 +477,97 @@ class DbScanAppRewardsStore(
...
+              and party_seq_num_end_excl <= $endExcl
+            order by party_seq_num_begin_incl
+      """.as[Array[Byte]],
+      "appRewards.lookupBatchByHash.children",
+    ).map(_.map(ByteString.copyFrom))
+
+  /** Retrieve minting allowances for a leaf batch range.
+    *
+    * Uses a LEFT JOIN from activity totals to reward totals because parties
+    * whose reward fell below the threshold have no row in `app_reward_party_totals`.
```
meiersi-da: This sounds off. It sounds like we are numbering the parties that had activity instead of numbering the ones that have rewards, could that be the case?

I might have accidentally caused that with a premature optimization to use party_seq_num for activity totals already.

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -1665,6 +1665,64 @@ paths:
         "404":
           $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/404"
 
+  /v0/reward-accounting-process/rounds/{round_number}/root-hash:
+    get:
+      tags: [internal, scan]
+      x-jvm-package: scan
+      operationId: "getRewardAccountingRootHash"
+      description: |
+        SV node internal API (CIP-0104, subject to change).
```
rautenrieth-da, line 1674: @meiersi-da if it's SV node internal, should it have authorization? Or do we expect SV operators to protect this via IP allowlisting?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -622,6 +624,76 @@ class DbScanAppRewardsStore(
...
+              )
+            ), 'hex')
+          from (
+            select
+              a.app_provider_party_seq_num as seq_num,
+              a.app_provider_party as party,
+              r.total_app_reward_amount::text as amount,
+              (a.app_provider_party_seq_num / $batchSize) as batch_num
+            from #${Tables.appActivityPartyTotals} a
+            left join #${Tables.appRewardPartyTotals} r
```
rautenrieth-da: From our discussion:

- We don't need to include parties with 0 rewards.
- Only use the reward party totals table. If possible, structure queries such that they only depend on the immediately preceding stage.
- In this case, can we move `app_provider_party_seq_num` to the rewards table (only numbering the above-threshold parties), and thus avoid the join?

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -4057,3 +4115,46 @@ components:
...
+
+    GetRewardAccountingBatchResponse:
+      type: object
+      required:
+        - batch_type
+      properties:
+        batch_type:
+          type: string
+          enum: [BatchOfBatches, BatchOfMintingAllowances]
+        child_hashes:
```
rautenrieth-da: `child_hashes` and `minting_allowances` are mutually exclusive, have you considered using `oneOf` (it's used in a few other places in scan.yaml)?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2435,6 +2435,101 @@ class HttpScanHandler(
...
+      respond: ScanResource.GetRewardAccountingRootHashResponse.type
+  )(roundNumber: Long)(extracted: TraceContext): Future[
+    ScanResource.GetRewardAccountingRootHashResponse
+  ] = {
+    implicit val tc = extracted
+    withSpan(s"$workflowId.getRewardAccountingRootHash") { _ => _ =>
+      appRewardsStoreO match {
+        case None =>
+          Future.successful(
+            ScanResource.GetRewardAccountingRootHashResponse.NotFound(
```
rautenrieth-da, line 2450: A 404 response feels wrong in this case, consider using
```scala
Future.failed[ScanResource.GetRewardAccountingRootHashResponse](
  Status.UNIMPLEMENTED
    .withDescription("Reward accounting is not enabled on this node")
    .asRuntimeException()
)
```
It looks like we're not always consistent in how we handle similar errors, so I don't have a strong opinion here.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -576,52 +701,191 @@ class DbScanAppRewardsStore(
...
+    */
+  private def insertRootHash(
+      roundNumber: Long
+  ) =
+    sql"""insert into #${Tables.appRewardRootHashes}(history_id, round_number, root_hash)
+          select history_id, round_number, batch_hash
+          from #${Tables.appRewardBatchHashes}
+          where history_id = $historyId and round_number = $roundNumber
+            and batch_level = (
+              select max(batch_level) from #${Tables.appRewardBatchHashes}
```
rautenrieth-da, line 991: There is no index on the batch level. Is there some mechanism that keeps the table small, so that we can afford a full table scan? For example, are we planning on removing batches from the table once the rewards are issued?

Alternatively, you can have `aggregateToRoot` return the level of the root.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
```diff
@@ -614,6 +614,344 @@ class DbScanAppRewardsStoreTest
...
+        //             → level 1: [[0,1],[2,3]], [[4,5],[6,7]], [[8]] = 3 batches
+        //             → level 2: [[[0,1],[2,3]],[[4,5],[6,7]]], [[[8]]] = 2 batches
+        _ <- store.insertAppActivityPartyTotals(
+          (0 to 8).map(i =>
+            AppActivityPartyTotalT(
+              historyId,
+              roundNumber,
+              1000000L * (i + 1),
+              i,
+              s"party$i::provider",
```
dfordivam: I would like the tests to have multiple activity records with varying weights, for the same party for a round.

dfordivam: nevermind, I see that your starting point is totals per round.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
```diff
@@ -614,6 +614,344 @@ class DbScanAppRewardsStoreTest
...
+            AppRewardPartyTotalT(historyId, roundNumber, i, BigDecimal(s"${i + 1}.0"))
+          )
+        )
+        _ <- store.computeRewardHashes(roundNumber, batchSize = 2)
+        rootHash <- store.getAppRewardRootHashByRound(roundNumber)
+        latestRound <- store.lookupLatestRoundWithRewardComputation()
+      } yield {
+        rootHash shouldBe defined
+        rootHash.value.rootHash.size shouldBe 32
+        latestRound.value shouldBe roundNumber
```
rautenrieth-da: Would it make sense to include these checks in all other tests, instead of only checking the root hash in a single scenario?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
```diff
@@ -614,6 +614,344 @@ class DbScanAppRewardsStoreTest
...
+      }
+    }
+
+    "computeRewardHashes — below-threshold party hashes same as explicit zero amount" in {
```
rautenrieth-da: This test is not needed anymore if we exclude below-threshold parties in `insertLeafBatches`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -576,52 +701,191 @@ class DbScanAppRewardsStore(
...
-    //              and prevent writes if the assertion fails.
     runUpdate(
-      insertRewardTotals
-        .map(_ =>
-          logger.debug(
-            s"Computed reward totals for round $roundNumber."
-          )
-        )
+      action
+        .map(_ => logger.debug(s"Computed reward hashes for round $roundNumber."))
```
rautenrieth-da: This will only log once every 10min, I think this can be an INFO level log. If you have readily available details (number of leaves, number of levels, ...), include them in the message. Such details are very useful for investigating issues during testing.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -463,6 +477,97 @@ class DbScanAppRewardsStore(
...
+      roundNumber: Long,
+      batchHash: ByteString,
+  )(implicit tc: TraceContext): Future[Option[(Int, Int, Int)]] = {
+    import storage.DbStorageConverters.setParameterByteArray
+    runQuerySingle(
+      sql"""select batch_level, party_seq_num_begin_incl, party_seq_num_end_excl
+            from #${Tables.appRewardBatchHashes}
+            where history_id = $historyId
+              and round_number = $roundNumber
+              and batch_hash = ${batchHash.toByteArray}
```
rautenrieth-da: We're converting `ByteString -> Array[Byte]` here, and `Array[Byte] -> ByteString` when reading the hash in `getResultAppRewardBatchHash`. Ideally, we would only convert types when reading/writing HTTP requests/responses.

Would it make sense to create a named case class for the hash, where the value is stored as `Array[Byte]` (or whatever results in the least amount of conversions), and with methods to convert it to/from the HTTP representation?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
rautenrieth-da: > head is not allowed by static checks

I think this only applies to collections (List, for example). We use `.head` on slick actions in many places.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
dfordivam: Thanks, I was looking around where we noted down this comment.😅

After carefully going through the schema now I understood the issue. We can't use `LEFT JOIN` with the guarantee of always getting the `party`, and using `JOIN` is a code smell (or worse). So in this case moving `party` to `app_reward_party_totals`, and having `app_provider_party_seq_num` in it is the better choice

This would also mean that `seq_num_begin` `seq_num_end` would be contiguous, ie end=begin+batch_size.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
dfordivam: Actually I had a discussion yesterday with Tim about the use of 404 in the response of `GET` for these endpoints, more specifically about the scenario where the data is not yet calculated, vs it cannot be provided. But for the case of no `appRewardsStoreO`  `UNIMPLEMENTED` may be even better.

I guess its fine for now to keep using 404 for both not yet calculated and it cannot be provided scenarios, and let the sv-app retry this endpoint and subsequently fallback to BFT read. Perhaps as part of BFT read implementation it would be better to have discussion about this.

apps/scan/src/main/openapi/scan.yaml
dfordivam: I guess its fine as it is? its not a security issue right, this being a straightforward single row lookup.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
dfordivam: As there are two other places with `batch_level` query, it makes sense to add the index. (`insertNextLevel` and `getChildBatchHashes`)

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
dfordivam: One scenario currently missing in tests, and discussed yesterday in call, a round with no app-activity should still get a root hash with empty `BatchOfMintingAllowances`

apps/scan/src/main/openapi/scan.yaml
meiersi-da: It actually has to be accessible to all other SV operators to handle SV onboarding. Thus making it public is the easiest option; and it seems fine given that the method should be performant as @dfordivam said.

adetokunbo:
> An issue I noticed; `computeAndStoreRewards` is the API exposed via `ScanAppRewardsStore`, but there are no unit test(s) for it. I know its not specifically related to this PR, but perhaps a test could be added in`DbScanAppRewardsStoreTest` now since all the parts are in-place

IIRC, this was discussed at the planning meeting for this feature branch.
My interpretation of the discussion was that it was decided to unit test all the parts of this function, but not the function itself and instead use an integration test, which is as you may have noted, is planned for the following PR

In such a unit test, one is really just checking the interaction between computeAndStoreRewards and the functions it calls that have been themselves unit-tested
- either one attempts to cover all the cases that the called functions cover. which involves duplicating existing test cases and test data;
- a subset with specific combinations of test data; maybe 1 case ?
- or one uses a technique like mocking to directly confirm the interaction between computeAndStoreRewards and the function it calls; some call this is dubious technique; others claim that's its the only value such a test case provides. AFAICT is not used elsewhere in the splice codebase

@dfordivam: What do you have in mind, any of these or something else ?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
adetokunbo: Fixed: The unittests have been refactored, improving their readability, PTAL. Also a test for the missing scenario was added in 584012c42

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed: This has been addressed.  Note that the schema changed [here](bc074a175) as suggested in other [review comments 
](https://github.com/hyperledger-labs/splice/pull/4847#discussion_r3043904903)

apps/scan/src/main/openapi/scan.yaml
adetokunbo: Noted, no action taken

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed.  See bc074a175 and e0160a274

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
adetokunbo: Noted, at the moment this is left as-is

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
adetokunbo: Fixed: Added in 4ecefa2a0

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
adetokunbo: Fixed.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed: 2cbb56cd6

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
```diff
@@ -614,6 +607,209 @@ class DbScanAppRewardsStoreTest
...
+        latestRound <- store.lookupLatestRoundWithRewardComputation()
+      } yield {
+        latestRound.value shouldBe roundNumber
+      }
+    }
+
+    "computeRewardHashes — round with no rewarded parties produces empty root hash" in {
+      for {
+        (store, batchHashes, rootHash) <- setupAndComputeHashes(
+          partyCount = 1,
```
dfordivam: may be this could be 0 also, or may be one more test with this as 0?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed, added in 1a7d985c3

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed: see 333702aa5

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
adetokunbo: Fixed: see b220e45bf

adetokunbo:
> > What do you have in mind, any of these or something else ?
> 
> Well naturally "it depends on context". One scenario likely not covered by integration tests is the throw by `assertCompleteActivity`, another would be calling it on round already computed. Another one is whether `RewardComputationInputs` is being properly used, but because this is so critical we will have to write an integration test for it. So unit test at this API level can be ignored.

Added these tests

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -576,52 +708,234 @@ class DbScanAppRewardsStore(
...
+              and batch_level = $currentLevel
+          ) sub
+          group by batch_num
+    """.asUpdate
+
+  /** The hash of a BatchOfMintingAllowances containing no parties.
+    *
+    * Used for rounds where no parties are above the reward threshold.
+    * Inserting this as a level-0 batch and root hash marks the round as
+    * computed, so that ComputeAppRewardsTrigger does not retry it, and
```
dfordivam: `ComputeAppRewardsTrigger` has been mentioned twice in this file, should be `RewardComputationTrigger` ?

apps/scan/src/main/openapi/scan.yaml
adetokunbo: Fixed: see cb37261ef

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Fixed

adetokunbo:
@rautenrieth-da, I think all the comments have been addressed, PTAL

rautenrieth-da (approved):
Thanks! The PR looks good to me now, the merge conflicts should be easy to resolve.

PR #4852, opened 2026-04-06
support ephemeral storage requests&limits in gha runners

isegall-da:
Part of https://github.com/DACH-NY/cn-test-failures/issues/7861
(the actual values are in cn-internal)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/common/src/config/configSchema.ts
+++ b/cluster/pulumi/common/src/config/configSchema.ts
@@ -40,6 +40,7 @@ export const SingleResourceSchema = z
   .object({
     memory: z.string().optional(),
     cpu: z.string().optional(),
+    ephemeralStorage: z.string().optional(),
   })
   .optional();
 


cluster/pulumi/gha/src/runners.ts
nicu-da, line 24: don't think this does what we need. this is basically just a type indicator for typescript, it has no effect on the underlying object. we need to modify the actual schema in `SingleResourceSchema`

isegall-da: Argh. I think we need both. This is the resources that get copied to the actual k8s objects, so pretty sure we do need this, but indeed we also need the one that parses our configs

isegall-da: (added)

nicu-da: ideally yes, we can in theory get away with not adding it to the type we define here because that's just informative, it will not strip away any extra fields you pass afaik

isegall-da: gotta love typescript...

PR #4857, opened 2026-04-06
localnet: add support for pv35

isegall-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/LocalNetFrontendIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/LocalNetFrontendIntegrationTest.scala
@@ -26,17 +26,17 @@ class LocalNetFrontendIntegrationTest
   override def environmentDefinition: SpliceEnvironmentDefinition =
     EnvironmentDefinition
       .fromResources(Seq("localnet-topology.conf"), this.getClass.getSimpleName)
-      .addConfigTransforms((_, conf) =>
-        conf.copy(
-          remoteParticipants = conf.remoteParticipants.map { case (p, config) =>
-            p -> config.copy(
-              adminApi = config.adminApi,
-              ledgerApi = config.ledgerApi,
-              token = Some(AuthUtil.testToken(AuthUtil.testAudience, "ledger-api-user", "unsafe")),
-            )
-          }
-        )
-      )
+//      .addConfigTransforms((_, conf) =>
+//        conf.copy(
+//          remoteParticipants = conf.remoteParticipants.map { case (p, config) =>
+//            p -> config.copy(
+//              adminApi = config.adminApi,
+//              ledgerApi = config.ledgerApi,
+//              token = Some(AuthUtil.testToken(AuthUtil.testAudience, "ledger-api-user", "unsafe")),
+//            )
+//          }
+//        )
+//      )
       .updateTestingConfig(
         _.focus(_.participantsWithoutLapiVerification).replace(
           Set(
@@ -212,7 +212,7 @@ class LocalNetFrontendIntegrationTest
       val participantClient = new ParticipantClientReference(
         env,
         appProviderParticipant.name,
-        appProviderParticipant.config,
+        appProviderParticipant.config.copy(token = Some(token)),
       )
       val hash = participantClient.upload_dar_unless_exists(
         "apps/app/src/test/resources/nuck-example-main-0.0.1.dar"


cluster/compose/localnet/conf/canton/app.conf
```diff
@@ -30,6 +30,9 @@ canton {
     # a large number of participants.
     timeouts.processing.verify-active = 40.seconds
     timeouts.processing.slow-future-warn = 20.seconds
+
+    alpha-version-support = yes
```
isegall-da: TODO: remove this.
If we merge before pv35 is stable, make this condition
If we wait for it to be stable, just drop this

cluster/compose/localnet/conf/canton/app.conf
```diff
@@ -91,6 +94,8 @@ _participant {
     }
     # Bump ACS pruning interval to make sure ACS snapshots are available for longer
     journal-garbage-collection-delay = 24h
+
+    alpha-version-support = yes
```
isegall-da: TODO: remove this.
If we merge before pv35 is stable, make this condition
If we wait for it to be stable, just drop this

isegall-da:
@moritzkiefer-da @meiersi-da 
Still need to decide whether we merge this already or wait for a stable PV, but mind taking a look already please?

cluster/compose/localnet/conf/canton/app.conf
```diff
@@ -81,7 +84,7 @@ _participant {
   }
 
   parameters {
-    initial-protocol-version = 34
+    # initial-protocol-version = 34
```
moritzkiefer-da: can we just delete it instead of uncommenting?

cluster/compose/localnet/conf/canton/app.conf
moritzkiefer-da: >  If we wait for it to be stable, just drop this

How about, just adding documentation on how to set this? Otherwise we have this whole mess of a discussion we had around PV dev/35 for the next PV dev again. I'd much rather just never set it by default but document how to enable it so we can solve this once and forall.

cluster/compose/localnet/conf/canton/app.conf
moritzkiefer-da: remove?

cluster/compose/localnet/conf/canton/app.conf
moritzkiefer-da: remove and document

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/LocalNetFrontendIntegrationTest.scala
```diff
@@ -179,4 +191,220 @@ class LocalNetFrontendIntegrationTest
...
+          .singleRequest(
+            Post(s"$url/v2/packages")
+              .withHeaders(Seq(Authorization(OAuth2BearerToken(token))))
+              .withEntity(entity)
+          )
+          .futureValue
+      response.status should be(StatusCodes.OK)
+
+      val grpcHost = "grpc-ledger-api.localhost"
+      val channel = io.grpc.ManagedChannelBuilder.forAddress(grpcHost, port).usePlaintext().build()
```
moritzkiefer-da: can we just add a remote participant in the env and then use that for all these commands to remove all this boilerplate?

moritzkiefer-da (approved):
lgtm minus the discussion on alpha version support

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/LocalNetFrontendIntegrationTest.scala
isegall-da: Hmm... For some reason I convinced myself that that doesn't work, but I can't think of any good reason why not given that I did get it to work this way, so will definitely give it a try again.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/LocalNetFrontendIntegrationTest.scala
```diff
@@ -179,4 +191,220 @@ class LocalNetFrontendIntegrationTest
...
+          userId = "ledger-api-user",
+        )
+      ).party
+
+      val createCommand = Command.of(
+        Command.Command.Create(
+          CreateCommand(
+            templateId = Some(
+              Identifier(
+                packageId = "22ce2c5b30d288f6aa48094d9775618851fec952c86cf1b7904a2eaaac27190d",
```
paulbrauner-da: Nit: I think you should be fine with `#package-name` which would make it easier to update this test to future versions of the dar.

isegall-da: With a `#` prefix? (I tried just package name, that did not work, was not aware that there's a `#` trick...)

isegall-da: @paulbrauner-da Trying `#nuck-example-main` here fails with: 
```
The submitted command has a field with invalid value: Invalid field package_id: non expected character 0x23 in Daml-LF Package ID "#nuck-example-main"
```

Am I missing something?

paulbrauner-da: Sorry for the probably misleading advice. I'm not familiar with the client libraries you're using so my advice is maybe wrong. The ledger API [supports](https://docs.digitalasset.com/build/3.4/reference/lapi-proto-docs.html#com-daml-ledger-api-v2-identifier) the `#{package_name}` syntax:

> The field is overloaded to also be able to contain the package-name of the Daml package. This is supported if the entity referenced is either an interface or template. When representing the Daml package-name, the encoding is of form `#<package-name>` where `#` (not a valid package-id character) is used as a discriminator for signalling a package-name encoding.

I'm not sure how that translates to the client library you're using.

paulbrauner-da: I looked at `com.daml.ledger.api.v2.value.Identifier` and it seems to be directly generated from the proto so I am not sure what's going on here.

paulbrauner-da: @isegall-da this is not important, I wouldn't consider this a blocker at all for this PR.

@tudor-da do you have an idea why `#<package-name>` is not working here?

isegall-da: Definitely not. I also ended up reusing existing logic to extract the hash from the dar, so I avoided hard-coding the hash that way.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/LocalNetFrontendIntegrationTest.scala
```diff
@@ -179,4 +193,125 @@ class LocalNetFrontendIntegrationTest
...
+        userId = "ledger-api-user",
+        optTimeout =
+          None, // Otherwise, optionallyAwait fails with permission issues reading from unrelated parties
+      )
+
+      participantClient.ledger_api.commands.submit(
+        actAs = Seq(alice),
+        commands = Seq(
+          Command.of(
+            Command.Command.ExerciseByKey(
```
moritzkiefer-da, line 276: I guess we don't run the java codegen for the models? if we did this seems needlessly convoluted but probably not worth setting it up just for this

build-tools/splice-localnet-compose.sh
```diff
@@ -21,16 +21,27 @@ export IMAGE_TAG
...
 DOWN_COMMAND=( stop )
+CANTON_PROTOCOL_VERSION=""
+
+function usage() {
+    echo "Usage: $SCRIPTNAME <start|stop> [-D] [-M] [-u] [-p <protocol_version>]"
+    echo ""
+    echo "Options:"
+    echo "  -D                        Completely tear down the localnet (using 'docker compose down') instead of just stopping the containers (using 'docker compose stop')"
+    echo "  -M                        Start the localnet with the 'multi-sync' profile enabled"
+    echo "  -u                        Enable unstable Canton protocol versions"
```
moritzkiefer-da: add a big disclaimer that this is unsafe and must only be used for temporary testing environments that can be reset

docs/src/app_dev/testing/localnet.rst
```diff
@@ -249,3 +249,17 @@ The additional synchronizer has the following characteristics:
...
+
+
+
+The protocol version used in LocalNet can be configured by setting the ``CANTON_PROTOCOL_VERSION``
+environment variable to the required version prior to launching LocalNet. Non-stable protocol versions can be used for
+early testing, but require explicit opt-in. To enable that, export also a
+``ALPHA_PROTOCOL_VERSION_ENV=$LOCALNET_DIR/env/alpha-protocol-version.env`` environment variable.
+
+.. warning::
+   Non-stable protocol versions are unreleased versions that are under development, and are
```
moritzkiefer-da, line 264: Insufficiently scary, I think the important part to get across is that you give up any option of keeping your network long running and being able to upgrade as you generally do not have data continuity on these versions.

docs/src/app_dev/testing/localnet.rst
```diff
@@ -249,3 +249,17 @@ The additional synchronizer has the following characteristics:
 - It simulates a **private synchronizer**.
 - Both the ``app-provider`` and ``app-user`` participants are cross-connected
   to the Global Synchronizer and the ``app-synchronizer``.
+
+Using Non-Default Protocol Versions
+-----------------------------------
+
+
+
+The protocol version used in LocalNet can be configured by setting the ``CANTON_PROTOCOL_VERSION``
```
moritzkiefer-da:
```suggestion
The protocol version used in the LocalNet synchronizer can be configured by setting the ``CANTON_PROTOCOL_VERSION``
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/LocalNetFrontendIntegrationTest.scala
isegall-da: exactly. I'm not even sure we can right now given the compiler version.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/LocalNetFrontendIntegrationTest.scala
tudor-da: > @tudor-da do you have an idea why #<package-name> is not working here?

That's odd. The only reason I think of why you'd get this is if you set the `#package-name` for recordIds (there indeed we don't support by-package-name notation)

tudor-da: @isegall-da is it working if you set `#package-name` only in the command's template-id's package-id?

isegall-da: yes, indeed, in the template ID it works, I originally tried setting it everywhere.

PR #4886, opened 2026-04-07
Bump default threshold on `ACS Commitment Checkpoint Delay` alert

martinflorian-da:
For the existing release lines I will override the config on internal

[static]


moritzkiefer-da (approved):
thanks!

cluster/pulumi/infra/grafana-dashboards/canton/acs-commitment-performance.json
```diff
@@ -315,7 +315,7 @@
         "type": "prometheus",
         "uid": "prometheus"
       },
-      "description": "Lag behind the timestamp the commitment checkpoint was created and the timestamp of the checkpoint",
+      "description": "Lag behind the timestamp the commitment checkpoint was created and the timestamp of the checkpoint; periodic lag spikes around during commitment computation are expected",
```
isegall-da:
```suggestion
      "description": "Lag behind the timestamp the commitment checkpoint was created and the timestamp of the checkpoint; periodic lag spikes around the time when commitments are computed are expected",
```

isegall-da: (or just "during commitment computation", just drop the "around")

martinflorian-da: thanks for catching that, fixing

PR #4919, opened 2026-04-08
Multi-fetch allocations & LockedAmulet

OriolMunoz-da:


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala
```diff
@@ -214,6 +214,31 @@ final class DbMultiDomainAcsStore[TXE](
       .value
   }
 
+  override def lookupContractsById[C, TCid <: ContractId[?], T](
```
meiersi-da, line 217: add comment explaining what is returned. I assume it is all the contracts that exist with one of the ids passed in?

meiersi-da (approved):
Nice work! Thanks.

PR #4965, opened 2026-04-10
Implement account authorization for TestTokenV2

meiersi-da:
### Motivation

- validate APIs and their integration with token standard wallets
- deliver 
 
### Outcome

- extended the v2 API's to support `AllocationInstruction_Accept`
- implemented full-fledged set of account policies, and show that a token standard wallet can use them through the v2 APIs without any special support from the registry

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


token-standard/examples/splice-test-token-v2/daml/Splice/Testing/Tokens/TestTokenV2.daml
```diff
@@ -133,40 +136,139 @@ template TokenAllocationV2 with
...
       view = V2.TransferInstructionView with
         originalInstructionCid = None
         transfer
-        availableActions = Map.fromList
-          [ (jointAuthorizers transfer.sender, [V2.TIA_Withdraw])
-          , (jointAuthorizers transfer.receiver, [V2.TIA_Accept, V2.TIA_Reject])
-          ]
-          -- TODO(#4542): add support for joint authorization steps
+        availableActions =
+          let extract account = (account, accountConfigFromId account.id) in
```
meiersi-da: Note: storing the authz config in the account id is beneficial here, as it allows to compute it as part of the view where we can't read from the ledger.

token-standard/splice-token-standard-test-v2/daml/Splice/Tests/TestBatchedTokenTransfer.daml
```diff
@@ -87,6 +86,9 @@ test_v2_transfer = do
       mkTransfer = WalletClientV2.mkTransferFactory_TransferV2' registries
       checkMetadata = True
 
+selfCustodied : Party -> V2.Account
+selfCustodied party = V2.Account with owner = party; provider = None; id = "oiap/self"
```
meiersi-da: This was introduced because `basicAccount party` does currently not create a valid account. However that is problematic for V1 wallets. We should also allow the basic account by special casing it in account-id parsing.

bame-da:
> - should we introduce a V2.TransferInstruction_Confirm choice to confirm a transfer and provide the input holdings for it?
After discussion I think the main case for a new choice is that the second party confirming the transfer may want to overwrite/set the input holding Cids. But if that use case does arise, one could use the choice context to do so. I'd therefore keep the API simple and stick with _Accept.

> - are we fine with the aspect that this implementation makes the authorization flow on the receiver side visible to the sender?
> - I @meiersi-da would argue that this seems OK, as registries can decide on their own whether they want to do so. If they build their own UI, they can use their own delegation workflows to execute the token standard choices with joint authority.
I agree with that assessment.

> - can we use this approach for the reference token with privacy, or do we need to change something? If yes, then what needs changing?
In the reference token I'd keep things simple and not let the second "confirmer" override the input holdings. But if you would like to validate that that does work, I'm also not against doing so.

token-standard/examples/splice-test-token-v2/daml/Splice/Testing/Tokens/TestTokenV2/AccountConfig.daml
```diff
@@ -0,0 +1,140 @@
...
+-- String conversion
+--------------------
+
+-- | Account configs are encoded in the account id as a string with the format:
+-- "o{ownerConfig}p{providerConfig}/"
+-- where {ownerConfig} and {providerConfig} are strings with the following format:
+-- "{i}{a}"
+-- where "i" is present if the party can initiate an action and "a" is present if the party must approve an action.
+--
+-- Note: storing the config in the account id is a convenient option, but not a requirement.
```
bame-da: The downside you present for storing account config on-chain is real, but I think the downside of string-encoding it on every holding is worse. If you store it on-chain, you have the "lag" between the config and the actionMap on open workflows. But open workflows are short-lived and few.
But if you store it per holding, you have the same "lag" with holdings, which are long-lived and potentially many. In particular, if I have 1000 holdings, I need to cycle 1000 contracts to change account config.

Furthermore, this pattern "pollutes" the accountId which I would like to use as a general reference to the account with technical information.

token-standard/examples/splice-test-token-v2/daml/Splice/Testing/Tokens/TestTokenV2.daml
bame-da: We are going to have to store the available actions on the workflow (instruction) contracts, yes. But I'm uneasy about making it part of the Account object.

token-standard/splice-api-token-transfer-events-v2/daml.yaml
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) 2025 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-sdk-version: 3.3.0-snapshot.20250502.13767.0.v2fc6c7e2
+sdk-version: 3.4.11
```
moritzkiefer-da, line 4: is there some rebase missing? why are we bumping the sdk version here?

token-standard/splice-token-standard-test-v2/daml/Splice/Tests/TestTokenV2/UnitTest_AccountConfig.daml
```diff
@@ -0,0 +1,404 @@
...
+    PartyConfig with canInitiate = False; mustApprove = True
+
+-- sc for self-custodual/basic
+sc : MkAccountConfig
+sc id admin owner provider =
+  let acc = ic id admin owner provider
+   in acc with account.provider = None
+
+test_availableTransferInstructionActions : Script ()
+test_availableTransferInstructionActions = do
```
moritzkiefer-da, line 83: maybe just me but I find this notation impossible to understand

moritzkiefer-da (approved):
Thanks!

I've reviewed the API changes and they made sense.

I skimmed the test code but did not review in detail (realistically that is gonna take me several days to understand). I'm quite worried about the complexity, it doesn't seem unreasonable to claim that we're exceeding the complexity of the actual splice Daml models in test code here which feels very questionable. That said: I'm ok with merging it provided we treat this as "internal" (as internal as OSS code can be) test code and not a reference implementation. In particular, I'd like to reserve the right that if we discover that maintenance of this code takes significant resources we can either aggressively change it or even fully rip it out.

token-standard/splice-api-token-transfer-events-v2/daml.yaml
meiersi-da: It was a botched merge in an earlier PR. The SDK bump was not propagated to the newly introduced API.

meiersi-da:
> Thanks!
> 
> I've reviewed the API changes and they made sense.
> 
> I skimmed the test code but did not review in detail (realistically that is gonna take me several days to understand). I'm quite worried about the complexity, it doesn't seem unreasonable to claim that we're exceeding the complexity of the actual splice Daml models in test code here which feels very questionable. That said: I'm ok with merging it provided we treat this as "internal" (as internal as OSS code can be) test code and not a reference implementation. In particular, I'd like to reserve the right that if we discover that maintenance of this code takes significant resources we can either aggressively change it or even fully rip it out.

@bame-da : are you OK with this proviso?

token-standard/splice-token-standard-test-v2/daml/Splice/Tests/TestTokenV2/UnitTest_AccountConfig.daml
meiersi-da: It's definitely not just you. I found it OK'ish once I looked at the fully expanded function with proper names.

bame-da: Do you think layout and a comment at the top would help that effectively turn it into a table layout?

moritzkiefer-da: I think what would help the most is passing around some records with easy to understand field names 

so replace

```
tc a v [e1] j a [vaia] [e1s] True,
```

by something like

```
tc {
  authorizers = a,
  provider =  v,
  …
}
```

3 different lists and a boolean + a bunch of single letter variable names are just inherently hard to understand unless you fully remember the definition each time.

Issue #2856, opened 2025-10-24
Make mediator verdict ingestion crash-fault tolerant 


rautenrieth-da:
Both UpdateHistory and DbScanVerdictStore insert rows into the database where they need to read the auto-generated value of the `row_id` column for the row just inserted. Normally we would use `INSERT ... RETURNING` for that, but that does not work with `INSERT ... ON CONFLICT DO NOTHING` which we usually use for handling idempotency. Instead, before inserting the row, we manually check if the row exists. See [UpdateHistory](https://github.com/hyperledger-labs/splice/blob/ee48e6154b0b8da122623396f8efc258805ac21e/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala#L2217-L2219) and [DbScanVerdictStore](https://github.com/hyperledger-labs/splice/blob/ee48e6154b0b8da122623396f8efc258805ac21e/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanVerdictStore.scala#L216-L220)

The DbScanVerdictStore query that checks if rows already exists is one of the most expensive queries on CILR right now. Try to rewrite the code using the following idea:

```scala
for {
  // Try to insert the row. This will fail due to unique constraint violation of `scan_verdict_hi_ui` if the row already exists
  result <- sql"""insert ... returning row_id""".asTry

  row_id <- result match {
    // Almost always, the insert will succeed
    case Success(row_id) => DBIOAction.successful(row_id)
    // If it does not, try to read the already existing row or fail if it doesn't exist
    case Failure(_) => readRowId()
  }
} yield row_id
```

This DBIOAction should still be idempotent, even if one of the queries might fail.

If that is not possible for some reason, check why DbScanVerdictStore checks for the existence of all items to be inserted while UpdateHistory only checks for the existence of the first one.

OriolMunoz-da:
> checks if rows already exists is one of the most expensive queries on CILR right now.

Note it was fixed in https://github.com/hyperledger-labs/splice/pull/2829/files

> check why DbScanVerdictStore checks for the existence of all items to be inserted while UpdateHistory only checks for the existence of the first one.

I can answer that. see the description in: https://github.com/hyperledger-labs/splice/pull/2726

rautenrieth-da:
> Note it was fixed in https://github.com/hyperledger-labs/splice/pull/2829/files

[...]

PR #4968, opened 2026-04-10
Refactor scan verdict ingestion

rautenrieth-da:
Contributes to https://github.com/hyperledger-labs/splice/issues/2856

This PR refactors verdict ingestion from a Trigger to a RetryingService. This makes sure we never miss any verdicts due to giving up on retries.


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/RetryingService.scala
```diff
@@ -146,6 +148,39 @@ abstract class RetryingService(
     // Healthy if there's an active instance
     currentServiceInstance.get().exists(_.isActive)
 
+  @SuppressWarnings(Array("org.wartremover.warts.Var"))
```
rautenrieth-da, line 151: All pausing code moved from UpdateIngestionService to the base class.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -0,0 +1,206 @@
...
+      logger.info(s"Streaming verdicts starting from $ingestionStart")
+      new ServiceWithGuaranteedShutdown(
+        source = batchSource(mediatorClient.streamVerdicts(Some(ingestionStart))),
+        map = processWhenUnpaused,
+        retryProvider = retryProvider,
+        loggerFactory = loggerFactory.append("subsClient", this.getClass.getSimpleName),
+      )
+    }
+
+  private def process(batch: Seq[v30.Verdict])(implicit
```
rautenrieth-da, line 93: The code in this method was copied almost 1:1 from the old ingestion trigger.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/metrics/ScanMediatorVerdictIngestionMetrics.scala
```diff
@@ -31,21 +31,23 @@ class ScanMediatorVerdictIngestionMetrics(metricsFactory: LabeledMetricsFactory)
...
-  )(MetricsContext.Empty)
-
-  val restartErrors: Meter = metricsFactory.meter(
-    MetricInfo(
-      name = prefix :+ "restart_errors",
-      summary = "Count of ingestion restart errors",
-      qualification = Traffic,
-    )
-  )(MetricsContext.Empty)
+  val latency: Timer =
```
rautenrieth-da, line 34: Removed error metrics (the base class for ingestion services doesn't track them), added some useful performance metrics.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ManualSignatureIntegrationTest.scala
```diff
@@ -24,6 +25,12 @@ class ManualSignatureIntegrationTest
       .withManualStart
       .withoutAliceValidatorConnectingToSplitwell
       .withSequencerConnectionsFromScanDisabled()
+      .addConfigTransforms((_, config) =>
+        updateAllScanAppConfigs((_, config) =>
+          // Sequencer is returning TRAFFIC_CONTROL_DISABLED when looking up traffic summaries.
```
moritzkiefer-da, line 30: Any reason why we don't enable traffic control in this test?
Also unclear to me why changing ingestion changes this.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/TriggerTestUtil.scala
```diff
@@ -51,16 +52,27 @@ trait TriggerTestUtil { self: BaseTest =>
...
       scan.automation.services[UpdateIngestionService].foreach(_.pause().futureValue)
       codeBlock
     } finally {
-      logger.info(s"Resuming ingestion for ${scan.name}")
+      logger.info(s"Resuming update ingestion for ${scan.name}")
       scan.automation.services[UpdateIngestionService].foreach(_.resume())
     }
   }
+
+  def pauseScanVerdictIngestionWithin[T](scan: ScanAppBackendReference)(codeBlock: => T): T = {
```
moritzkiefer-da, line 66: any reason why we're not using `setTriggersWithin` here?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -0,0 +1,206 @@
...
+    for {
+      _ <- waitForStores()
+      ingestionStart <- getIngestionStart()
+    } yield {
+      logger.info(s"Streaming verdicts starting from $ingestionStart")
+      new ServiceWithGuaranteedShutdown(
+        source = batchSource(mediatorClient.streamVerdicts(Some(ingestionStart))),
+        map = processWhenUnpaused,
+        retryProvider = retryProvider,
+        loggerFactory = loggerFactory.append("subsClient", this.getClass.getSimpleName),
```
moritzkiefer-da, line 89: subsClient sounds quite generic, what does the full logger name look like? is it obvious that it's mediator ingestion?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -0,0 +1,206 @@
...
+      val trafficSummariesF: Future[Seq[DbScanVerdictStore.TrafficSummaryT]] =
+        sequencerTrafficClientO match {
+          case Some(sequencerTrafficClient) =>
+            sequencerTrafficClient
+              .getTrafficSummaries(sequencingTimes)
+              .map(_.map { proto =>
+                DbScanVerdictStore
+                  .fromProtoWithCorrelation(proto, viewHashToViewIdByTime, logger)
+              })
+              // TODO(#4060): handle missing traffic summaries more robustly. In particular,
```
moritzkiefer-da, line 116: @rautenrieth-da I assume that we're considering this a blocking issue for actually turning this on for rewards? This seems non-deterministic otherwise as the batches are not consistent across nodes probably?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -0,0 +1,206 @@
...
+                val errorDetails = ErrorDetails.from(statusProto)
+                val errorCodeId = errorDetails
+                  .flatMap {
+                    case ed: ErrorDetails.ErrorInfoDetail =>
+                      Some(ed.errorCodeId)
+                    case _ => None
+                  }
+                  .headOption
+                  .getOrElse("none")
+                if (errorCodeId == "NO_EVENT_AT_TIMESTAMPS")
```
moritzkiefer-da: usually we reference the corresponding Canton type instead of hardcoding the error code id as a string so if it changes we get a compile error (or it adjusts automatically)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -0,0 +1,206 @@
...
+                  Future.successful(Seq.empty)
+                else
+                  Future.failed(ex)
+              }
+          case _ =>
```
moritzkiefer-da:
```suggestion
          case None =>
```
maybe a bit clearer

I also have a question: When do we set this to None? Is this just there because it's still in development and we want to be able to turn off the feature flag?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -0,0 +1,206 @@
...
+      batch: Seq[v30.Verdict]
+  )(implicit traceContext: TraceContext): Future[Unit] = {
+    // If paused, this step will backpressure the source
+    waitForResume().flatMap { _ =>
+      ingestionMetrics.latency.timeFuture(process(batch))
+    }
+  }
+
+  private def batchSource[T](source: Source[T, NotUsed]): Source[Seq[T], NotUsed] =
+    source.groupedWithin(
```
moritzkiefer-da: We use `source.batch` for `UpdateIngestionService`. Here we use `source.groupedWithin`. Why are they different? IIRC for update ingestion the logic was that if we're fast enough to not accumulate anything we don't really care any we'd rather get the lower latency from `batch` and under load they collapse to more or less the same anyway, cc @OriolMunoz-da in case he remembers more of the reasoning.

If we do keep them different let's add a clear comment why this is desired

moritzkiefer-da (approved):
very nice thank you!

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ManualSignatureIntegrationTest.scala
rautenrieth-da: According to @julientinguely-da this test is running an old version of Canton and might be removed soon, so I didn't spend much time investigating. The test is also relatively short, so it's possible the old trigger-based ingestion didn't get to the point where it would give up on retries and log an error/warning.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/TriggerTestUtil.scala
rautenrieth-da: `setTriggersWithin` is for subclasses of `Trigger`, ingestion services are subclasses of `RetryingService` which is not a trigger.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
rautenrieth-da: Yes, definitely a blocking issue. The end goal is that scan always has a contiguous range of verdicts (it's only fine to miss some during SV onboarding).

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
rautenrieth-da: @dfordivam do you remember why you picked `. groupedWithin`? Was it to make sure we don't produce too many small batches?
One difference between UpdateHistory is that for verdict ingestion, we do one sequencer API call for each batch (in addition to one SQL transaction).

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
rautenrieth-da: Exactly, we haven't planned the removal of feature flags yet.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
rautenrieth-da: The logger name will include `subsClient=ScanVerdictIngestionService`:
```
2026-04-13T23:05:51.423Z [⋮] ERROR - o.l.s.e.ServiceWithGuaranteedShutdown:ValidatorSequencerConnectionIntegrationTest/config=4b56fc4/scan=sv1Scan/subsClient=ScanVerdictIngestionService (---) - Fatally failed to handle element java.lang.RuntimeException: Forcing retry after batch processing
	at org.lfdecentralizedtrust.splice.scan.automation.ScanVerdictIngestionService.$anonfun$processWhenUnpaused$3(ScanVerdictIngestionService.scala:197)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:470)
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
dfordivam: I actually don't remember why this was picked. Doing batched ingestion was necessary, but there wasn't a good reason for preferring `groupedWithin`. Also at the time verdict ingestion was added its only use-case was events endpoint, so lower latency in ingestion wasn't really a priority.
Now that we are doing sequencer API call, it seems that `groupedWithin` is a more preferred option?
IIUC we would want lower latency only at the round closing time, other than this its fine to have some lag.

rautenrieth-da: Switched to use `.batch` here as well.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ManualSignatureIntegrationTest.scala
meiersi-da: @julientinguely-da : what's the condition that must hold for us to remove this test? Could we remove it now?

Asking as @dfordivam and @mrdziuban are working on making app record ingestion the default.

julientinguely-da: We're keeping it because _in theory_ someone could still recover from key their ancient node. Not against removing it though, @moritzkiefer-da wdyt?

moritzkiefer-da: if we want to remove it we need to discuss with Canton who asked for this to be added. So far it didn't really seem to get in the way so keeping it seemed ok.

PR #4991, opened 2026-04-12
Add trigger to SV app to create ParticipantSynchronizerPermission for Validator

### LINKED ISSUE #4988: Add trigger to SV app that creates ParticipantSynchronizerPermission topology txs based on ValidatorPermission Daml contracts

pasindutennage-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4988

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
@@ -87,7 +87,9 @@ class PermissionedSynchronizerIntegrationTest
                     store = decentralizedSynchronizerId,
                     filterUid = aliceParticipantId.filterString,
                   )
-              permissions.map(_.item.permission) should contain(ParticipantPermission.Submission)
+              permissions.map(_.item.permission) should contain(
+                ParticipantPermission.Submission
+              ) // reports errors in log for premature checks
             }
           }
         },
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
@@ -67,6 +67,7 @@ class PermissionedSynchronizerIntegrationTest
     }
 
     withClue("Grant validator permission to Alice and verify visibility across all SVs") {
+
       val aliceParticipantId = aliceValidatorBackend.participantClient.id
 
       val allSvValidators =
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
@@ -4,11 +4,13 @@
 package org.lfdecentralizedtrust.splice.integration.tests
 
 import com.digitalasset.canton.admin.api.client.data.OnboardingRestriction.RestrictedOpen
+import com.digitalasset.canton.logging.SuppressionRule
 import com.digitalasset.canton.topology.transaction.ParticipantPermission
 import org.lfdecentralizedtrust.splice.config.ConfigTransforms
 import org.lfdecentralizedtrust.splice.integration.EnvironmentDefinition
 import org.lfdecentralizedtrust.splice.integration.tests.SpliceTests.IntegrationTest
 import org.lfdecentralizedtrust.splice.util.{ProcessTestUtil, WalletTestUtil}
+import org.slf4j.event.Level
 
 class PermissionedSynchronizerIntegrationTest
     extends IntegrationTest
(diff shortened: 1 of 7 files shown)


pasindutennage-da:
CI failure of PermissionSynchronizer test is just because of logs-containing-problems, test it self passes

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
```diff
@@ -67,51 +66,45 @@ class PermissionedSynchronizerIntegrationTest
...
+            clue(s"Checking active topology state on ${svValidator.name}") {
+              val permissions =
+                svValidator.participantClient.topology.participant_synchronizer_permissions
+                  .list(
+                    store = decentralizedSynchronizerId,
+                    filterUid = aliceParticipantId.filterString,
+                  )
+              permissions.map(_.item.permission) should contain(
+                ParticipantPermission.Submission
+              ) // reports errors in log for premature checks
```
pasindutennage-da: because it comes true only eventually, we have some annoying logs. But test eventually passes

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/GrantValidatorPermissionTrigger.scala
```diff
@@ -0,0 +1,73 @@
...
+        if (payload.isRevoked) {
+          logger.info(
+            s"Skipping ValidatorPermission for $participantId because the contract is marked as revoked."
+          )
+          Future.unit
+        } else {
+          logger.info(
+            s"Proposing ParticipantSynchronizerPermission (Submission) for $participantId on synchronizer $synchronizerId"
+          )
+          participantAdminConnection.proposeMapping(
```
moritzkiefer-da: use `ensureTopologyMapping`

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/GrantValidatorPermissionTrigger.scala
```diff
@@ -0,0 +1,73 @@
...
+        splice.validatorpermission.ValidatorPermission,
+      ]
+  )(implicit tc: TraceContext): Future[TaskOutcome] = {
+    val payload = task.payload
+    val synchronizerId = task.domain
+    val participantId = ParticipantId.tryFromProtoPrimitive(payload.validatorParticipantId)
+    for {
+      _ <-
+        if (payload.isRevoked) {
+          logger.info(
```
moritzkiefer-da, line 58: maybe add a TODO to actually handle revocation?

moritzkiefer-da (commented):
Nice thanks! Mostly looks good but let's switch over to the right topology primitives

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/GrantValidatorPermissionTrigger.scala
pasindutennage-da: I used `ensureTopologyMappingO`, because `ensureTopologyMapping` has the following

```
check.andThen(_.leftMap(_.some)),
      {
        case Some(value) => update(value)
        case None =>
          throw Status.INTERNAL
            .withDescription("Update was called without an existing state, should be impossible")
            .asRuntimeException()
      }
```

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/GrantValidatorPermissionTrigger.scala
pasindutennage-da: Done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
```diff
@@ -67,18 +68,39 @@ class PermissionedSynchronizerIntegrationTest
...
-        sv.participantClient.topology.participant_synchronizer_permissions
-          .propose(
-            decentralizedSynchronizerId,
-            aliceParticipantId,
-            permission = ParticipantPermission.Submission,
+
+      val allSvValidators =
+        Seq(sv1ValidatorBackend, sv2ValidatorBackend, sv3ValidatorBackend, sv4ValidatorBackend)
+
+      loggerFactory.assertEventuallyLogsSeq(SuppressionRule.LevelAndAbove(Level.ERROR))(
```
moritzkiefer-da, line 78: what error are we suppressing here? `assertEventuallyLogsSeq` with the assertion just being `succeed` also seems off. You need an eventually if it may fail initially but if you just succeed always you clearly don't need it.

I'd also recommend to actually assert on the expected error so the test already answers the question on what we are suppressing

moritzkiefer-da (approved):
nice thanks!

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
pasindutennage-da: Fixed by removing clue

PR #5015, opened 2026-04-13
exclude daml interfaces from compilation

isegall-da:
Also tested locally (to make sure some CI caching is not hiding something) with: `find -name ".daml" -type d -exec rm -rf {} \; && sbt bundle`

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/DarResources.scala
isegall-da, line 1: The changes in this file are actually no longer needed, we do also have a `-current.dar` file, but still felt nicer not to use it given that for apis we must never upgrade. Happy to revert if anyone disagrees.

daml/splice-util-featured-app-proxies/daml.yaml
isegall-da, line 1: all these changes to daml.yaml's are actually not really required anymore either, given that the damlPlugin copies the dars to the .daml/dist directory, and even copied it also as `-current.dar`. But also here, since we're depending on the stable APIs from daml/dars, seems slightly nicer to make that explicit.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/DarResources.scala
isegall-da: Argh, given that it's failing the static check, this is probably not worth changing the static checks for, and we should just revert it. Leaving it here for now in case you want to convince me not to.

project/DamlPlugin.scala
```diff
@@ -114,42 +128,65 @@ object DamlPlugin extends AutoPlugin {
...
+            // Derive the -current.dar name by stripping the version suffix
+            val darName = darFile.getName.stripSuffix(".dar")
+            val projectName = darName.replaceAll("-[0-9]+\\.[0-9]+\\.[0-9]+$", "")
+            val currentDar = outputDirectory / s"$projectName-current.dar"
+            IO.copyFile(darFile, currentDar)
+            Seq(destFile, currentDar)
           }
-        }
-        val log = streams.value.log
+        } else {
```
isegall-da: I initially wanted to add an assertion here that any package named *-api-* must have a version 1.0.0, but I couldn't find an easy enough way to get the daml version of the package, so ended up dumping the idea.

isegall-da:
Argh, I also see now that `splice-clean` deletes the stuff from docs that are no longer generated. Might need to move the committed version somewhere else and have sbt copy it to the target location instead...

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/DarResources.scala
moritzkiefer-da: tbc: this is a generated file. You definitely cannot modify this. So if anything you need to modify the generator and at that point just not doing it seems easier

daml/splice-api-featured-app-v1/docs/index.rst
```diff
@@ -0,0 +1,14 @@
+..
+   Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+..
+   SPDX-License-Identifier: Apache-2.0
+
+splice-api-featured-app-v1 docs
```
moritzkiefer-da, line 6: why do we need to commit these instead of regenerating them? We're not deleting the source so why does `damlc docs` not work?

project/DamlPlugin.scala
```diff
@@ -114,42 +128,65 @@ object DamlPlugin extends AutoPlugin {
...
-
-        def buildOrder(fst: File, snd: File): Boolean = {
-          def indexOf(file: File): Int = {
-            val asString = file.toString
-            buildDependencies.indexWhere(asString.contains(_))
+
+        if (skipBuild) {
+          // Skip compilation and use pre-built DARs
+          if (prebuiltDars.isEmpty) {
+            log.warn(s"damlSkipBuild is true but no prebuilt DAR files were specified")
```
moritzkiefer-da: any reason why this is only a warn and not a hard error?

Maybe turning it around: why do we even need the skipBuild flag and this is not jut tied to whteher prebuiltDars is set or not?

project/DamlPlugin.scala
```diff
@@ -164,9 +201,10 @@ object DamlPlugin extends AutoPlugin {
                   damlLanguageVersions.value,
                 )
               }.toSet
-          }
+            }
 
-        cache(allDamlFiles.get.toSet ++ damlProjectFiles.get.toSet ++ dependencies).toSeq
+          cache(allDamlFiles.get.toSet ++ damlProjectFiles.get.toSet ++ dependencies).toSeq
```
moritzkiefer-da, line 181: this seems like probably it should include prebuilt dars?

project/DamlPlugin.scala
```diff
@@ -114,42 +128,65 @@ object DamlPlugin extends AutoPlugin {
...
-            false
-          } else if (sndIdx == -1) {
-            true
-          } else {
-            fstIdx < sndIdx
+          // Copy pre-built DARs into the output directory so downstream tasks find them there.
+          // Also create a -current.dar copy to match the convention of buildDamlProject,
+          // which downstream tasks (damlCodeGeneration, copyDarResources) rely on.
+          IO.createDirectory(outputDirectory)
+          prebuiltDars.flatMap { darFile =>
```
moritzkiefer-da: I believe caching works properly but did you double check that rerunning it does not end up copying the dar everytime?

project/DamlPlugin.scala
```diff
@@ -52,6 +52,18 @@ object DamlPlugin extends AutoPlugin {
...
       settingKey[Boolean]("Read config from daml.yaml")
+    val damlSkipBuild =
+      settingKey[Boolean](
+        "Skip Daml compilation and use pre-built DARs instead. Pre-built DARs will be copied into the damlDarOutput, so that downstream tasks e.g. codegen don't need to be aware of whether we built the dars or not."
+      )
+        .withRank(
+          KeyRanks.Invisible
+        ) // suppresses unused warnings in canton-community-app where we exclude damlBuild altogether
+    val damlPrebuiltDars =
+      settingKey[Seq[File]]("pre-built DAR files to use when damlSkipBuild is true")
```
moritzkiefer-da: is there ever a case where we want multiple dars? if not, any reason why this is not an `Option` instead?

moritzkiefer-da (approved):
Nice thanks! Not really clear to me why we need the change to the docs but otherwise lgtm

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/DarResources.scala
isegall-da: Doh of course

daml/splice-api-featured-app-v1/docs/index.rst
isegall-da: It is not generating docs for the types, my guess was that it uses some other build artifacts that damlc produces other than the final dar, and at that point I stopped trying. Do you think we should dig deeper and see if that can be solved and the docs generation can be fixed?

project/DamlPlugin.scala
isegall-da: Good idea

project/DamlPlugin.scala
isegall-da: I consciously decided to not care, copying a couple of dars seems quick enough to not bother, no? Happy to test explicitly if you disagree

project/DamlPlugin.scala
isegall-da: Did not end up needing more than one, no. It just seemed like damlPlugin supports generating more than one dar, so seemed more consistent to support also more than one precompiled dars

daml/splice-api-featured-app-v1/docs/index.rst
moritzkiefer-da: my understanding is that it uses the sources not the dar. given that we don't delete them I would have hoped it just continues working. Seems at least worth understanding a bit better why it fails but don't mind checking it in all that much either.

project/DamlPlugin.scala
moritzkiefer-da: I think it's worth testing. Mainly because if something relies on modification times instead of hashes which e.g. make does and maybe also SBT in some cases this can cascade. You make a seemingly irrelevant change to a DAR but then everything that depends on it gets rebuilt.

isegall-da: Good point, will do

daml/splice-api-featured-app-v1/docs/index.rst
isegall-da: So I think I get it:

> it uses the sources not the dar.

That is correct, but also for dependencies. The build phase copies the source code of dependencies into `.daml/dependencies/...`, and the docsgen phase then uses that to generate docs. Since we skip the build phase, we don't have the source of dependencies available, and then e.g. docsgen for splice-api-token-burn-mint-v1fails with:
```
Could not find module ‘Splice.Api.Token.MetadataV1’
It is not a module in the current program, or in any known package.
```

I don't immediately see any cli flags for `damlc docs` that would allow working around this, and copying the dependencies ourselves doesn't seem worth it, so I'm falling back to committing the docs.

daml/splice-amulet-name-service-test/daml.yaml
```diff
@@ -12,7 +12,7 @@ data-dependencies:
 - ../splice-amulet-name-service/.daml/dist/splice-amulet-name-service-current.dar
 - ../splice-util/.daml/dist/splice-util-current.dar
 - ../splice-wallet-payments/.daml/dist/splice-wallet-payments-current.dar
-- ../splice-api-featured-app-v1/.daml/dist/splice-api-featured-app-v1-current.dar
+- ../dars/splice-api-featured-app-v1-1.0.0.dar
```
meiersi-da, line 15: @isegall-da : this change seems to not be strictly necessary as the Daml plugin copies the prebuilt .dar to the expected location. Should I apply this kind of change on the TS2 branch as well, or just leave it as-is? Note that this change is also not required in `DEVELOPMENT.md`.

isegall-da: Yeah, I think I ended up being a bit inconsistent there. I think I did this part of the change before landing on the solution where we just copy the prebuilt dars to the dist directories, and that indeed it's not really required now. So happy to leave as-is too.

PR #5030, opened 2026-04-14
Add v1 holdings summary endpoint without holding fees

### LINKED ISSUE #4094: Revisit holding fees in scan summary endpoint
The current holding fees in that endpoint don't make sense and having total available coin deduct holding fees is just wrong.

While expiry still exists I don't think summing up holding fees expresses it in any meaningful way. So imho the best option is to introduce a new version of the summary endpoint that just drops all the holding fees fields.

Maybe need to see on where we end up with scan bulk storage and which APIs survive but on the other hand it's also a trivial change and has caused actual confusion for users so maybe just doing it is also fine. Could be a good first issue for someone.

samsondav:
## Summary

- Introduces `/v1/holdings/summary` scan endpoint that drops all holding fee fields (`accumulated_holding_fees_unlocked`, `accumulated_holding_fees_locked`, `accumulated_holding_fees_total`, `total_available_coin`) and the `as_of_round` request parameter, as these don't express meaningful aggregate values
- Deprecates `/v0/holdings/summary`
- Adds corresponding v1 endpoint to the validator scan-proxy

Closes #4094

## Test plan

- [x] Verify `apps-scan/compile`, `apps-validator/compile`, `apps-app/compile`, and `apps-app/Test/compile` all pass
- [x] New assertions in `ScanTimeBasedIntegrationTest` verify v1 response contains correct coin totals matching v0 and handles `atOrBefore` record time matching — executed in CI
- [x] Verify v0 endpoint still works unchanged (backwards compatible) — v0 handler code preserved, only deprecation metadata added
- [x] Verify v1 response schema has only `party_id`, `total_unlocked_coin`, `total_locked_coin`, `total_coin_holdings` per summary entry — confirmed from Guardrail-generated `HoldingsSummaryV1` case class


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2020,6 +2021,63 @@ class HttpScanHandler(
...
+    implicit val tc: TraceContext = extracted
+    withSpan(s"$workflowId.getHoldingsSummaryAtV1") { _ => _ =>
+      val HoldingsSummaryRequestV1(
+        migrationId,
+        recordTime,
+        recordTimeMatch,
+        partyIds,
+      ) = body
+
+      // Round 0 is used as a placeholder since the v1 endpoint does not compute holding fees.
```
isegall-da: Looking at the computation in `SpliceUtil.foldingFee()`, I'm not convinced that a placeholder round 0 returns the correct result here.

isegall-da: Ah, I think it does, because it does:
```
amulet.amount.initialAmount.min(...)
```
So it would never be larger than the initialAmount, and with round 0 placeholder, it will be >= initialAmount.
Can you double check that, and add to the comment here why this is correct?

isegall-da: And I guess your unit test below confirms that too.

isegall-da:
Thanks @samsondav . Can you please add release notes to `release_notes_upcoming.rst` too?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
samsondav: Pushed f5eddf4d1.

I confirmed the round-0 question by tracing through the code: `SpliceUtil.holdingFee` is only consumed by `HoldingsSummary.addAmulet`/`addLockedAmulet` for the `accumulatedHoldingFees*` and `totalAvailableCoin` fields. The v1 response only exposes `totalUnlockedCoin`, `totalLockedCoin`, and `totalCoinHoldings`, all of which are direct sums of `amulet.amount.initialAmount` and never call `holdingFee`. So `asOfRound` is unused for the v1 path regardless of the value passed; 0 is just a sentinel.

(For the record, your `.min(initialAmount)` reasoning would not actually hold at round 0 — `currentRound - createdAt.number` is negative there, so the `.min` would return the negative value, not `initialAmount`. But it doesn't matter, because the v1 fields don't go through that code path at all.)

I expanded the comment in `HttpScanHandler.scala` to spell this out.

samsondav:
Added release notes for the new endpoint and the v0 deprecation in f5eddf4d1.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
isegall-da: Nice, thanks for checking.

isegall-da:
@samsondav can you fix the DCO please too?

samsondav:
Resolved the merge conflict with upstream and signed off all commits to fix DCO. Pushed e03986a5c. DCO check is now passing.

isegall-da (approved):
Thanks, and apologies for dropping the ball on this.

PR #5034, opened 2026-04-14
Fix usage of runSupervised

rautenrieth-da:
The previous implementation looks broken:

1. The `errorLogMessagePrefix` parameter is passed by value, so the `if (retryProvider.isClosing)` condition is evaluated only once, when the service starts.
2. The code for `PekkoUtil.runSupervised` contains a section to "avoid errors on shutdown", but that relies on the `isDone` parameter which we never set and it defaults to `_ => false`

It does mean that instead of printing "Ignoring failure to handle transaction, as we are shutting down" on shutdown, we will print the more confusing "Fatally failed to handle element (encountered after the graph is completed)". I don't see a way around that without changing PekkoUtil from the Canton fork.

Relevant change: https://github.com/DACH-NY/canton-network-node/pull/16890

Concrete example where the current code fails:

1. Verdict ingestion is running
2. RetryProvider initiates shutdown
3. Verdict ingestion stream processing (the async map call presumably started before shutdown was initiated) fails, because it uses `DbMultiDomainAcsStore.waitUntilRecordTimeReached` which throws an exception when the RetryProvider is shutting down.
4. We get a spurious log error


isegall-da (approved):
Definitely looks sensible, thanks

PR #5036, opened 2026-04-15
Bump canton to 3.5.0-snapshot.20260414.18683.0.v153dde21

OriolMunoz-da:


moritzkiefer-da (approved):
newer is always better

PR #5039, opened 2026-04-15
Ensure BulkStorageTests work with newer amulet versions

OriolMunoz-da:
Stumbled upon this in the token standard branch after merging. The default is 'current' , which means the package_id changes everytime amulet gets bumped.


isegall-da:
hmmm... not sure about this. Shouldn't this work for the latest package?

isegall-da:
Argh, it's probably failing the hash checks if the package_id changes though??

OriolMunoz-da:
> Argh, it's probably failing the hash checks if the package_id changes though??

exactly, because the hashes are hardcoded in the test

moritzkiefer-da:
> Argh, it's probably failing the hash checks if the package_id changes though??

yep exactly the package ids are included in the hash (correctly fully so imho) but that means if the version changes the hash also changes

isegall-da (approved):
TIHI.

PR #5042, opened 2026-04-15
lsu - continue verdict ingestion from successor 

nicu-da:
fixes https://github.com/hyperledger-labs/splice/issues/4161

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/mediator/MediatorVerdictsClient.scala
```diff
@@ -57,12 +59,44 @@ final class MediatorVerdictsClient(
...
+        done.onComplete {
+          case Success(_) =>
+            if (completePromise.trySuccess(None)) {
+              logger.info(
+                "Mediator verdicts stream completed without a Complete message"
+              )
+            }
+          case Failure(ex) =>
+            logger.warn(
+              s"Mediator verdicts stream terminated with an error",
```
nicu-da: @ray-roestenburg-da adding this logs basically gets flagged in all the tests. 
What was the expectation previously on how we would log failures for the ingestion? Is there a case as to why we wouldn't log such issues?

canton/community/synchronizer/src/main/protobuf/com/digitalasset/canton/mediator/admin/v30/mediator_inspection_service.proto
```diff
@@ -31,7 +31,21 @@ message VerdictsRequest {
 }
 
 message VerdictsResponse {
-  Verdict verdict = 1;
```
nicu-da, line 34: The API actually changed so this will fail temporarily during the upgrade until both apps are on the same version. Not ideal but also I don't think it's that big of a problem to actually have canton revert this and add a parallel api (that was the plan initially actually)

moritzkiefer-da: tbc it fails on the 0.6.0 upgrade briefly but not for the LSU?

nicu-da: yup, just for the 0.6.0 upgrade until both scan and mediator are on the same version

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -82,8 +93,55 @@ class ScanVerdictIngestionService(
       ingestionStart <- getIngestionStart()
     } yield {
       logger.info(s"Streaming verdicts starting from $ingestionStart")
+      val currentSource = currentMediatorClient.streamVerdicts(Some(ingestionStart))
```
moritzkiefer-da: How does this work in this scenario:

1. We stream mediator verdicts on the current synchronizer. Last verdict is at T1.
2. We switchover at upgrade time T2.
3. We consume at least one verdict on the successor at T3.
4. We now restart the SV app.
5. Afaict we will now try to get a verdict stream on the old synchronizer starting from T3. We seem to rely here on this giving us the LSU complete message. Will it do that or do we get an empty stream?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -82,8 +93,55 @@ class ScanVerdictIngestionService(
...
+            completedWithCompleteF.tryFailure(ex)
+          }(ec)
+          NotUsed
+        }
+        .concat(
+          Source
+            .futureSource(
+              completedWithCompleteF.future.flatMap {
+                case Some(_) =>
+                  getIngestionStart().map { successorIngestionStart =>
```
moritzkiefer-da: is it save to reuse the ingestion values across physical synchronizers? I don't remember what exactly they contain

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
nicu-da: I am pretty sure we do get that because we have a restart of the sv app as part of the test, double checking logs as well though

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/mediator/MediatorVerdictsClient.scala
moritzkiefer-da: at a minimum it needs to be guarded by retryProvider.isClosing. But it feels a bit off to me to log this here anyway, usually we log a level higher up on retries exceeding instead of logging every single network blip that forces a reconnet as a warning.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
nicu-da: the `successorIngestionStart`? it's just a timestamp of the last verdict we have in the db, so it should map out cleanly

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -51,14 +52,24 @@ class ScanVerdictIngestionService(
...
-  private val mediatorClient =
+  private val currentMediatorClient =
     new MediatorVerdictsClient(
       config.synchronizerNodes.current.mediator,
       this,
       grpcClientMetrics,
       loggerFactory,
     )(ec, esf)
 
+  private lazy val successorMediatorClientO =
```
moritzkiefer-da, line 71: I would expect that we can actually test for this now in our LSU test. In particular check that we now continue receiving all verdicts properly

moritzkiefer-da (approved):
Nice thanks! A few questions on details and seems worthwhile to extend the tests

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
nicu-da: Yes, good point, adding this.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/mediator/MediatorVerdictsClient.scala
ray-roestenburg-da: agree, it should not be logged on every single network blip.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
nicu-da: Actually we had failures because of this, will check if there's an announcement and compare the upgrade time with the init from time.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/mediator/MediatorVerdictsClient.scala
nicu-da: > at a minimum it needs to be guarded by retryProvider.isClosing.

Why is that? That would close the stream normally, right? Without an exception


Anywho, I will log this at info for now.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
nicu-da: added

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
nicu-da: Extended the checks, please take a look

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/mediator/MediatorVerdictsClient.scala
moritzkiefer-da: Have not double checked what we do for the mediator but in a bunch of other cases we force clone the grpc channel first to ensure that any requests abort immediately which iirc makes it fail with UNAVAILABLE

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -80,10 +94,84 @@ class ScanVerdictIngestionService(
     for {
       _ <- waitForStores()
       ingestionStart <- getIngestionStart()
+      announcements <- synchronizerNodes.current.sequencerAdminConnection
+        .listLsuAnnouncements(synchronizerId)
+      currentPsid <- synchronizerNodes.current.sequencerAdminConnection
+        .getPhysicalSynchronizerId()
     } yield {
-      logger.info(s"Streaming verdicts starting from $ingestionStart")
+      val pastUpgradeAnnouncement = announcements.find { announcement =>
```
moritzkiefer-da, line 107: Trying to convince myself this also works for roll-forward:

It seems like we essentially just assume that we already caught up before we initiate the roll-forward as  that has only current and legacy. That seems pretty reasonable

moritzkiefer-da (approved):
thanks!

nicu-da:
@ray-roestenburg-da I had to touch quite a bit of the ingestion stream, can you please also take a look

ray-roestenburg-da:
@nicu-da Maybe you want to ask @rautenrieth-da as well to look at this, he's worked on the traffic stuff that's new for me in here.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/mediator/MediatorVerdictsClient.scala
```diff
@@ -57,12 +59,45 @@ final class MediatorVerdictsClient(
...
         stub.verdicts,
       )
-      .mapConcat(_.verdict)
-      .mapMaterializedValue(_ => NotUsed)
+      .takeWhile(
+        response => {
+          response.payload match {
+            case v30.VerdictsResponse.Payload.Complete(complete) =>
+              logger.info(s"Received Complete message from mediator verdicts stream: $complete")
+              completePromise.trySuccess(Some(complete)).discard
```
ray-roestenburg-da, line 74: Can you add a unit test for handling the complete?

nicu-da: The value compared to all the mocking required makes it not worthy

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
nicu-da: Sounds good, will still need to see how we handle the DR, will integrate with your test after it's merged

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -51,14 +56,24 @@ class ScanVerdictIngestionService(
...
-  private val mediatorClient =
+  private lazy val currentMediatorClient =
     new MediatorVerdictsClient(
       config.synchronizerNodes.current.mediator,
       this,
       grpcClientMetrics,
       loggerFactory,
     )(ec, esf)
 
+  private lazy val successorMediatorClientO =
```
ray-roestenburg-da, line 71: So you always create it (if Some passed in) even only to close it if it is never used really? (lazy val). Though it is easy I guess to access from the future this way.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -76,84 +91,155 @@ class ScanVerdictIngestionService(
...
+            }
+          case None =>
+            logger.info(s"Streaming verdicts starting from $ingestionStart")
+            val currentSource =
+              streamVerdictsAndBatchWithTraffic(
+                ingestionStart,
+                currentMediatorClient,
+                synchronizerNodes.current.sequencerTrafficClient,
+              )
+            val completedWithCompleteF = Promise[Option[v30.VerdictsResponse.Complete]]()
```
ray-roestenburg-da, line 140: I don't know why you are doing this here and not for the stream source created at 116, maybe add some comments?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -8,38 +8,43 @@ import com.daml.grpc.adapter.ExecutionSequencerFactory
...
+import org.lfdecentralizedtrust.splice.scan.ScanSynchronizerNode
+import org.lfdecentralizedtrust.splice.scan.sequencer.SequencerTrafficClient
 
-import scala.concurrent.{ExecutionContextExecutor, Future}
+import scala.concurrent.{ExecutionContextExecutor, Future, Promise}
 
 /** Ingestion for verdict ingestion store. */
 class ScanVerdictIngestionService(
     config: ScanAppBackendConfig,
+    synchronizerNodes: LocalSynchronizerNodes[ScanSynchronizerNode],
```
ray-roestenburg-da, line 46: In general I could use some comments on how this is used and what sequence this class needs to handle to deal with LSU

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
ray-roestenburg-da: Is this because the mediator will signal a completion signaling that there will be a successor or something like that?

nicu-da: > Is this because the mediator will signal a completion signaling that there will be a successor or something like that?

I think this would make more sense to read the docs on the canton side, as this is just the client for what canton has implemented.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
nicu-da: sure, let me add a comment at the top level on how we deal with LSU-s here as well

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/mediator/MediatorVerdictsClient.scala
```diff
@@ -57,12 +59,45 @@ final class MediatorVerdictsClient(
...
+      .watchTermination() { (_, done) =>
+        done.onComplete {
+          case Success(_) =>
+            if (completePromise.trySuccess(None)) {
+              logger.info(
+                "Mediator verdicts stream completed without a Complete message"
+              )
+            }
+          case Failure(ex) =>
+            if (!managedChannel.isClosing)
```
ray-roestenburg-da, line 93: How do you know that this will happen before the failure (is there anything to enforce that managedChannel.isClosing  when the failure occurs, otherwise the logging is possibly a bit flaky?

ray-roestenburg-da (approved):
Thanks, added a few comments, looks good to me.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/mediator/MediatorVerdictsClient.scala
nicu-da: Not sure what you mean, we added this just to prevent logging if the failure is cause by the channel closing. If we get here and it's caused by the channel closing then this should return true.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
nicu-da: Pretty much, worst case scenario it just creates a channel that is not used and then it closes it.

ray-roestenburg-da: yeah it's fine.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/mediator/MediatorVerdictsClient.scala
ray-roestenburg-da: So if the managedChannel is closing but the ex is not caused by managedChannel (which you can't know really) you just don't log it and don't know about the exception?

nicu-da: If it's closing then everything else is closing as well so we don't really care about the failures.

nicu-da:
merging this but @rautenrieth-da if you have any other comments feel free to add  and i will fix in another PR

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/mediator/MediatorVerdictsClient.scala
ray-roestenburg-da: ok

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
```diff
@@ -76,84 +95,155 @@ class ScanVerdictIngestionService(
...
-  ): Future[Unit] = {
-    if (batch.isEmpty) {
-      logger.error(
-        "Received empty batch of verdicts to ingest. This is never supposed to happen."
-      )
-      Future.successful(())
-    } else {
+  private def streamVerdictsAndBatchWithTraffic(
+      ingestionStart: CantonTimestamp,
+      successorMediatorClient: MediatorVerdictsClient,
```
rautenrieth-da, line 192:
```suggestion
      mediatorClient: MediatorVerdictsClient,
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/LogicalSynchronizerUpgradeIntegrationTest.scala
```diff
@@ -651,6 +651,20 @@ class LogicalSynchronizerUpgradeIntegrationTest
...
+              scan.appState.eventStore.verdictStore
+                .maxVerdictRecordTime(0)
+                .futureValue
+                .value
+                .toInstant should be > beforeBobActivityTimestamp
+            }
+          }
+        }
+      }
+
```
rautenrieth-da, line 667: How about a check here that restarting scan (while it still has the same `ScanSynchronizerNodesConfig` config) doesn't break anything? Or at least, add a check that the old mediator behaves like we expect in scan? I.e., that the verdict stream at the current time immediately returns a `Complete`.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/LogicalSynchronizerUpgradeIntegrationTest.scala
```diff
@@ -625,6 +623,8 @@ class LogicalSynchronizerUpgradeIntegrationTest
         sv1ScanBackend.startSync()
       }
 
+      // this also ensures that sv1 ingested verdicts after the restart
+      val beforeBobActivityTimestamp = Instant.now()
```
rautenrieth-da, line 627: Shouldn't this use the sequencer clock or something? It probably doesn't matter in practice, it just feels weird to compare timestamps from different clocks.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/LogicalSynchronizerUpgradeIntegrationTest.scala
nicu-da: We restart scan a few lines above.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/LogicalSynchronizerUpgradeIntegrationTest.scala
nicu-da: realistically for the test it doesn't really matter

PR #5046, opened 2026-04-15
Include NUCK setup and latest version in TOKEN_STANDARD_V2_DEVNET.md

OriolMunoz-da:


token-standard/TOKEN_STANDARD_V2_DEVNET.md
OriolMunoz-da, line 7: already updated, because apparently we can do whatever we want with GH pre-releases

moritzkiefer-da (approved):
thanks!

PR #5060, opened 2026-04-16
Add lookupOpenMiningRoundByNumber

adetokunbo:
Adds this to the SV app rewards branch since it's only going to be used by this branch until it is merged into main

This change anticipates the use of this function on this branch after the computation branch is merged into main, and then merged into here.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V067__scan_rewards_reference_round_index.sql
+++ b/apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V067__scan_rewards_reference_round_index.sql
@@ -1,0 +1,11 @@
+-- Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Index on round column for efficient lookups by round number.
+create index scan_rewards_reference_store_active_round
+    on scan_rewards_reference_store_active (store_id, migration_id, round)
+    where round is not null;
+
+create index scan_rewards_reference_store_archived_round
+    on scan_rewards_reference_store_archived (store_id, migration_id, round)
+    where round is not null;
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
@@ -18,11 +18,21 @@ import org.lfdecentralizedtrust.splice.scan.store.ScanRewardsReferenceStore
 import org.lfdecentralizedtrust.splice.store.{Limit, TcsStore}
 import org.lfdecentralizedtrust.splice.store.db.{
   AcsArchiveConfig,
+  AcsQueries,
+  AcsTables,
   DbAppStore,
   DbTcsStore,
   StoreDescriptor,
 }
-import org.lfdecentralizedtrust.splice.util.{ContractWithState, TemplateJsonDecoder}
+import org.lfdecentralizedtrust.splice.store.db.AcsQueries.SelectFromAcsTableResult
+import org.lfdecentralizedtrust.splice.util.{
+  Contract,
+  ContractWithState,
+  PackageQualifiedName,
+  TemplateJsonDecoder,
+}
+import org.lfdecentralizedtrust.splice.util.FutureUnlessShutdownUtil.futureUnlessShutdownToFuture
+import slick.jdbc.canton.ActionBasedSQLInterpolation.Implicits.actionBasedSQLInterpolationCanton
 
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -62,7 +72,9 @@ class DbScanRewardsReferenceStore(
         )
       ),
     )
-    with ScanRewardsReferenceStore {
+    with ScanRewardsReferenceStore
+    with AcsTables
+    with AcsQueries {
 
   override def waitUntilInitialized: Future[Unit] = multiDomainAcsStore.waitUntilAcsIngested()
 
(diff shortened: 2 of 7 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
```diff
@@ -137,4 +149,38 @@ class DbScanRewardsReferenceStore(
       tc: TraceContext
   ): Future[Seq[ContractWithState[OpenMiningRound.ContractId, OpenMiningRound]]] =
     tcsStore.listAllContractsAsOf(OpenMiningRound.COMPANION, asOf)
+
+  override def lookupOpenMiningRoundByNumber(
```
dfordivam, line 153: This should at least make use of `acsStore.waitUntilAcsIngested`.
It may be even better to wait if the round being looked up has not been ingested yet? we could compare max round in active table to assess that, and use `acsStore.waitUntilRecordTimeReached`?

If doing wait here is not possible, then it may be better to return something other than `None`?

If `None` is an acceptable response in such cases, (because the trigger would try again after a while?), then it should be made clear in comments.

dfordivam: The scenario would happen when scan has restarted and is catching up, The `CalculateRewardsV2` would come from `ScanStore` which may be ahead of `DbScanRewardsReferenceStore`.
Currently it seems that the only thing the trigger would do when receiving `None` is retry?
it won't give up right?
so doing retry at the trigger level may be fine

rautenrieth-da: Definitely add `acsStore.waitUntilAcsIngested`, the call might throw an exception otherwise ("using a storeId before it was initialized").

IMO it's fine to have ACS stores return data for the current state of the store (i.e., return `None` if the mining round has not been ingested yet), and let callers handle the fact that different stores might be lagging behind by arbitrary amounts of time.

As for triggers, after `retrieveTasks()` returns a task, the trigger will retry the task a few times, but if it keeps failing the trigger will eventually give up and retrieve the next task.

As long as you make sure `retrieveTasks()` is implemented such that it properly resumes from the persisted state, you should be fine.

I would still make sure we are not failing tasks just because some store is not caught up. Tasks that fail even after retries produce log warnings, and we alert on those in production. So either have `retrieveTasks()` only return tasks that have a reasonable chance of succeeding, or successfully complete a task if some data is missing.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V067__scan_rewards_reference_round_index.sql
```diff
@@ -0,0 +1,11 @@
...
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Index on round column for efficient lookups by round number.
+create index scan_rewards_reference_store_active_round
+    on scan_rewards_reference_store_active (store_id, migration_id, round)
+    where round is not null;
+
+create index scan_rewards_reference_store_archived_round
+    on scan_rewards_reference_store_archived (store_id, migration_id, round)
+    where round is not null;
```
rautenrieth-da, line 11: I'm always nervous when we're creating indexes on non-empty tables in flyway migrations, because migrations are blocking application startup. AFAICT, this table should be sufficiently small that the index creation completes in a second, so it's fine.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
adetokunbo: Fixed: Thanks for all the feedback here, I've added `waitUntilAcsIngested` as recommended

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V067__scan_rewards_reference_round_index.sql
adetokunbo: Noted, thanks!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
```diff
@@ -137,4 +149,47 @@ class DbScanRewardsReferenceStore(
...
+  )(implicit
+      tc: TraceContext
+  ): Future[Option[Contract[OpenMiningRound.ContractId, OpenMiningRound]]] = {
+    val storeId = multiDomainAcsStore.acsStoreId
+    val migrationId = multiDomainAcsStore.domainMigrationId
+    val pqn = PackageQualifiedName.fromJavaCodegenCompanion(OpenMiningRound.COMPANION)
+    val columns = SelectFromAcsTableResult.sqlColumnsCommaSeparated()
+    val query =
+      sql"""(
+         select #$columns
```
rautenrieth-da, line 173: Query plan looks good on CILR. We only have 600+4600 rows there, but I don't expect any issues with large datasets: the new `scan_rewards_reference_store_active_round` only contains mining rounds, so it should be preferred over the existing index on `(store_id, migration_id, package_name, template_id_qualified_name, created_at)`.

project/BuildCommon.scala
```diff
@@ -267,8 +267,8 @@ object BuildCommon {
             "splice-util-token-standard-wallet-test-daml/clean",
             "splice-util-batched-markers-daml/clean",
             "splice-util-batched-markers-test-daml/clean",
-            "splice-featured-app-api-v1-daml/clean",
-            "splice-featured-app-api-v2-daml/clean",
+            "splice-api-featured-app-v1-daml/clean",
+            "splice-api-featured-app-v2-daml/clean",
```
rautenrieth-da: Is this intentional? It looks unrelated to this PR.

rautenrieth-da (approved):
Thanks! PR looks good to me, just make sure to address the comment on the change in `BuildCommon.scala`.

project/BuildCommon.scala
adetokunbo: Yes: this is a fix for `sbt clean-splice`, which does not work without this

dfordivam: I confirmed that `sbt clean-splice` works for me on your branch, without this fix. Please confirm if this is really necessary.

adetokunbo: Here's why the change is here:

After switching to a new branch or rebasing the current branch, I usually use

`sbt clean && sbt clean-splice && sbt compile`

to make sure the dev environment is pristine

Without this fix, the command `sbt clean-splice` fails

adetokunbo: After a discussion with @dfordivam, I'm reverting this.  I'll try and find out why my dev environment seems to need it.

PR #5088, opened 2026-04-17
Add read path store performance tests

jagathweerasinghe-da:
- Related issue: https://github.com/hyperledger-labs/splice/issues/4790
- This builds the basic setup for read-path performance tests for updates. 
- In Future PRs, we will add further tests/metrics to assess the other application-level processing steps
- Test run: https://github.com/hyperledger-labs/splice/actions/runs/24560551098/job/71807820127
Screenshot 2026-04-17 at 15 20 07


OriolMunoz-da:
nits (starting review now, but i have nowhere to put this): 
image
that's not seconds. I would propose that you use Grafana's units (you already do, judging by the "ms" there) and remove it from the label

image
This is a bit confusing, because the green line is under the blue one, maybe just split it

OriolMunoz-da:
btw @ray-roestenburg-da , this might be useful for testing improvements to `makeConsistentAcrossSvs`

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/BaseStorePerformanceTest.scala
```diff
@@ -0,0 +1,187 @@
...
+        closeContext,
+      ) match {
+        case storage: DbStorage => storage
+        case storageType => throw new RuntimeException(s"Unsupported storage type $storageType")
+      }
+    // Suppress Flyway ClassPathScanner warnings about unloadable test jars
+    org.slf4j.LoggerFactory
+      .getLogger("org.flywaydb.core.internal.scanner.classpath.ClassPathScanner")
+      .asInstanceOf[ch.qos.logback.classic.Logger]
+      .setLevel(ch.qos.logback.classic.Level.ERROR)
```
OriolMunoz-da: why is this needed now? what jars are we loading that we weren't before? 😕

ray-roestenburg-da:
We should def have some tests for txs with large number of nodes.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/StoreReadPerformanceTest.scala
```diff
@@ -0,0 +1,172 @@
...
+        metricsFile,
+        json,
+        StandardOpenOption.CREATE,
+        StandardOpenOption.TRUNCATE_EXISTING,
+      )
+      println(s"Wrote metrics to $metricsFile")
+    } catch {
+      case e: Exception =>
+        println(s"Failed to write metrics file for $testName: ${e.getMessage}")
+    }
```
OriolMunoz-da: this looks the same as for write tests, and the part above largely has the same metrics except for 1 or 2. Unless I'm missing something, DRY

.github/workflows/performance_tests.yml
```diff
@@ -94,3 +94,81 @@ jobs:
...
+
+      - name: Setup
+        id: setup
+        uses: ./splice-shared-gha/.github/actions/tests/common_test_setup
+        with:
+          cache_version: 8
+          test_name: oss
+          target: 'oss'
+
+      - name: Run read performance tests
```
OriolMunoz-da, line 121: since the setup is the same, why don't we just run the read performance tests after the write tests?
Thus we can avoid setting up nix (which takes a while) and thus also avoid consuming unnecessary resources

OriolMunoz-da (approved):
thanks! no major comments

jagathweerasinghe-da:
> nits (starting review now, but i have nowhere to put this): image that's not seconds. I would propose that you use Grafana's units (you already do, judging by the "ms" there) and remove it from the label
> 
> image This is a bit confusing, because the green line is under the blue one, maybe just split it

Done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/BaseStorePerformanceTest.scala
jagathweerasinghe-da: {"@timestamp":"2026-04-17T10:06:43.247Z","message":"Skipping unloadable jar file: file:/__w/splice/splice/target/bg-jobs/sbt_6051fcf9/job-1/target/0b3eb9bd/770df5d0/apps-app_2.13-0.1.0-SNAPSHOT-tests.jar (/__w/splice/splice/target/bg-jobs/sbt_6051fcf9/job-1/target/0b3eb9bd/770df5d0/apps-app_2.13-0.1.0-SNAPSHOT-tests.jar)","logger_name":"o.f.c.i.s.c.ClassPathScanner","thread_name":"scala-execution-context-global-407","level":"WARN"}

[Get this, do you have any idea why?](https://github.com/hyperledger-labs/splice/actions/runs/24559326566/job/71803728207)

.github/workflows/performance_tests.yml
jagathweerasinghe-da: Though about this.  Let me do that change in a future PR after looking into the other tests we add there.

PR #5096, opened 2026-04-17
fix the task completion marker

krzysztofczyz-da:
This fixes https://github.com/DACH-NY/cn-test-failures/issues/8078

We've had the task completion marker `touch $@` before finishing the task, which created the issue in the first place.


moritzkiefer-da (approved):
makes a lot of sense thanks!

PR #5103, opened 2026-04-17
Ensure that npm install does not change the package-lock.json file in CI

krzysztofczyz-da:
Fixes #4945

Green [example](https://github.com/hyperledger-labs/splice/actions/runs/24568818370?pr=5103) 🟢 
Red [example](https://github.com/hyperledger-labs/splice/actions/runs/24569893937?pr=5103) 🔴

### DIFF
--- a/.github/workflows/pulumi_npm.yml
+++ b/.github/workflows/pulumi_npm.yml
@@ -15,12 +15,14 @@ jobs:
             cd cluster/pulumi
             npm install
             if ! git diff --quiet || [ -n "$(git ls-files --others --exclude-standard)" ]; then
-              echo ":red_circle: Build produced changes:" >> $GITHUB_STEP_SUMMARY
-              git status --short >> $GITHUB_STEP_SUMMARY
-              echo 'Please run `cd $(git rev-parse --show-toplevel)/cluster/pulumi && npm install` and commit the changes in `package-lock.json`' >> $GITHUB_STEP_SUMMARY
+              # shellcheck disable=SC2129
+              echo ":red_circle: Build produced changes:" >> "$GITHUB_STEP_SUMMARY"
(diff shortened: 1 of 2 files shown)


.github/workflows/pulumi_npm.yml
```diff
@@ -0,0 +1,28 @@
...
+        - shell: bash
+          run: |
+            cd cluster/pulumi
+            npm install
+            if ! git diff --quiet || [ -n "$(git ls-files --others --exclude-standard)" ]; then
+              # shellcheck disable=SC2129
+              echo ":red_circle: Build produced changes:" >> "$GITHUB_STEP_SUMMARY"
+              git status --short >> "$GITHUB_STEP_SUMMARY"
+              # shellcheck disable=SC2016
+              echo 'Please run `cd $(git rev-parse --show-toplevel)/cluster/pulumi && npm install` and commit the changes in `package-lock.json`' >> "$GITHUB_STEP_SUMMARY"
```
mblaze-da, line 22: Why do we need `$(git rev-parse --show-toplevel)` if this only run in `splice`?

mblaze-da (approved):
Thanks!

mblaze-da:
Does current precommit setup reliably trigger dependency installation? If not we could also introduce a deliberate install in precommit hooks to check this.

.github/workflows/pulumi_npm.yml
krzysztofczyz-da: Well, `git rev-parse --show-toplevel` is so that wherever in the repo you are it will just work 😅

krzysztofczyz-da:
> Does current precommit setup reliably trigger dependency installation?

pulumi_tests does that for `pre-commit` - this PR is mostly for dependabot-created PRs (and we don't run pre-commit in CI 😿 )

PR #5107, opened 2026-04-17
Use streamContinuationToken in activeContracts

OriolMunoz-da:
Fixes https://github.com/hyperledger-labs/splice/issues/4846

[preflight](https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/64885/workflows/9fa12fcf-f918-48d4-b65a-d67f67f91be4)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
```diff
@@ -98,14 +100,27 @@ class BaseLedgerConnection(
...
-        streamContinuationToken = None,
-      )
-    )
+  )(implicit
+      tc: TraceContext
+  ): Source[BaseLedgerConnection.ActiveContractsItem, NotUsed] = {
+    // The source can restart due to expired auth, so we have to restart it from where we left off
+    val streamContinuationToken = new AtomicReference(Option.empty[ByteString])
+    val activeContractsRequest = {
+      RestartSource.onFailuresWithBackoff(RestartSettings(0.5.seconds, 1.second, 0.5))(() => {
```
OriolMunoz-da: approach (meaning, AtomicReference usage) copied from PekkoUtil.restartSource. Unfortunately that one restarts also when the stream completes, which is not what we want here

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
```diff
@@ -98,14 +100,32 @@ class BaseLedgerConnection(
...
```
martinflorian-da: I wonder if the retries are not a bit too aggressive; if the participant has some overload issue, wouldn't 1 second retries be a bit too frequent?

martinflorian-da: Also: Will this retry forever now? What if we have an actual hard problem? Will we get an error/fail log eventually?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
```diff
@@ -98,14 +100,32 @@ class BaseLedgerConnection(
...
+        val streamContinuationTokenValue = streamContinuationToken.get()
+        logger.info(
+          s"Starting active contracts stream with continuation token ${streamContinuationTokenValue
+              .map(_.toStringUtf8)}"
```
martinflorian-da: Did you look at the logs to confirm that `.toStringUtf8` does something useful here?

OriolMunoz-da: there's an assertion for it in the test, though tbf that was a synthetic token

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
OriolMunoz-da: >  Will this retry forever now?

yes, except when the stream is terminated ~(though _when_ it gets terminated is not entirely clear to me)~ => when UpdateIngestionService is terminated

> Will we get an error/fail log eventually?

the test had some warnings from akka on every restart with the exception

> I wonder if the retries are not a bit too aggressive

will make that configurable and increase it

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
meiersi-da: The token is a bytestring is it not? Then `.toStringUtf8` is risky, as it is a partial function. Not all bytestrings are valid UTF-8 strings.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
```diff
@@ -98,14 +98,32 @@ class BaseLedgerConnection(
...
-        eventFormat = Some(eventFormat),
-        streamContinuationToken = None,
-      )
-    )
+      restartSettings: RestartSettings,
+  )(implicit
+      tc: TraceContext
+  ): Source[BaseLedgerConnection.ActiveContractsItem, NotUsed] = {
+    // The source can restart due to expired auth, so we have to restart it from where we left off
+    val lastItem = new AtomicReference(Option.empty[lapi.state_service.GetActiveContractsResponse])
```
OriolMunoz-da, line 106: approach (meaning, AtomicReference usage) copied from PekkoUtil.restartSource. Unfortunately that one restarts also when the stream completes, which is not what we want here
The same is true for PekkoRetryingService

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/environment/ActiveContractsRestartTest.scala
```diff
@@ -0,0 +1,165 @@
...
+        )
+
+      callCount.get() shouldBe 2
+      result
+        .collect { case ActiveContractsItem.ActiveContract(contract) =>
+          contract.createdEvent.getContractId
+        } should be(List("1", "2", "3"))
+    }
+
+    "fail the stream if the failure is not in the Source" in {
```
OriolMunoz-da, line 135: and therefore let the client retry

OriolMunoz-da: in the case of `UpdateIngestionService` this is: just start over, which was the existing behavior
related to https://github.com/canton-network/splice/pull/5107#discussion_r3195644147, we might want to instead restart from where we last failed (via the same continuation token), thoughts? 

(and if so, I'll make a tech debt for it)

OriolMunoz-da:
@martinflorian-da @isegall-da i can haz review? :prettyplease:

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/UpdateIngestionService.scala
```diff
@@ -113,7 +113,17 @@ class UpdateIngestionService(
...
     ingestionSink.ingestAcsStreamInBatches(
-      batchSource(connection.activeContracts(filter, offset)),
+      batchSource(
+        connection.activeContracts(
+          filter,
+          offset,
+          RestartSettings(
+            config.ingestion.activeContractsMinBackoff.underlying,
+            config.ingestion.activeContractsMaxBackoff.underlying,
+            config.ingestion.activeContractsRandomFactor,
```
martinflorian-da, line 123: I don't understand why we need this, as opposed to just passing 0, but don't mind it either.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/environment/ActiveContractsRestartTest.scala
martinflorian-da: I'd say I'm not sufficiently convinced that it is important to restart from scratch here to ask you to even create a tech debt. Not opposed, but it doesn't seem like something we'd prioritize without additional evidence it has good ROI.

martinflorian-da (approved):
LGTM! Thanks and thanks for your patience 🙇

OriolMunoz-da:
will merge on Monday since I'm off the rest of the week and this should be on cilr for a few days ideally before a release

isegall-da (approved):
Apologies for the very late review, dropped the ball on this one. LGTM too.

PR #5119, opened 2026-04-20
Traffic-Based App Rewards: Integrate the Reward Computation Trigger with the necessary daml contracts

adetokunbo:
Fixes #4383
Fixes #4570

This integrates the RewardComputationTrigger with the Daml contracts from which it obtains the values it uses in its calculations

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
```diff
@@ -63,7 +75,28 @@ class TrafficBasedRewardsTimeBasedIntegrationTest
         )(config)
       )
 
+  import TrafficBasedRewardsTimeBasedIntegrationTest.*
+
   "App activity records are created for featured app parties" in { implicit env =>
+    clue("Set rewardConfig on AmuletConfig so OpenMiningRounds have CIP-104 fields") {
```
dfordivam: I have some commits to set this, I have pushed on [this branch](https://github.com/obsidiansystems/canton-splice/tree/dfordivam/rewardConfig-test-set)

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
```diff
@@ -207,6 +240,7 @@ class TrafficBasedRewardsTimeBasedIntegrationTest
       val totals = sv1ScanBackend.getRewardAccountingActivityTotals(earliest)
       totals.value.roundNumber shouldBe earliest
       totals.value.activityRecordsCount should be > 0L
+      totals.value.totalAppActivityWeight should be > 0L
```
dfordivam: now we should be able to compare with actuals right?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -59,17 +62,38 @@ class RewardComputationTrigger(
...
+              case None =>
+                logger.debug(
+                  s"OpenMiningRound for round $roundNumber not yet ingested, waiting."
+                )
+                Seq.empty
+              case Some(contract) =>
+                RewardComputationInputs.fromOpenMiningRound(contract.payload) match {
+                  case None =>
+                    logger.debug(
+                      s"Round $roundNumber is a pre-CIP-104 round (missing rewardConfig or trafficPrice), skipping."
```
dfordivam: why would the trigger need to retry in this case? 
what should happen for rounds with no traffic-price,batch-size and threshold data?
In my opinion the totals calcs could be ignored for them.

dfordivam: or perhaps I did not understand the above comment properly, as the log says "skipping", does `Seq.empty` cause a skip or retry?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
```diff
@@ -395,3 +429,28 @@ class TrafficBasedRewardsTimeBasedIntegrationTest
     new allocationv1.Allocation.ContractId(allocation.contractId.contractId)
   }
 }
+
+object TrafficBasedRewardsTimeBasedIntegrationTest {
+
+  // Use traffic-based app rewards (CIP-0104), not on-ledger coupon counting.
+  private val rewardMintingVersion = RewardVersion.REWARDVERSION_TRAFFICBASEDAPPREWARDS
```
dfordivam: Ideally we should test that calcs are producing totals even when we are on featured-app rewards, and there is no dry run. It could be an almost identical copy of this test, and we could get rid of that test eventually when traffic-based rewards are default.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: this return value does not cause a retry, but rather a skip; I've updated the comment to further clarify

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
adetokunbo: Thanks!  I've updated the test to use `ConfigTransforms.withRewardConfig`

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
dfordivam: Discussed with Tim, I will enhance the integration test further for various scenarios, as we would otherwise end up in lots of conflicts in this test.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
dfordivam: according to the logic if no round totals have been computed, the `earliestCompleteO` will always be picked, and if this round does not have the `OpenMiningRound.rewardConfig` data, how will the trigger ever gonna skip this round and move to the next round? the `lookupLatestRoundWithRewardComputation` will always return `None`.

```
    "return the earliest complete round when nothing has been computed" in {
      nextRound(round3, round10, noRound) shouldBe Some(3L)
    }
```

dfordivam: even though the debug log says skipping, are we really skipping it, how?

adetokunbo: IIUC, by returning `Seq.empty`.   That signals to this trigger type to poll again without doing anything.   An alternative is to fail by throwing an exception; then the trigger will retry n times before shutting itself down.

dfordivam: we could fetch the `lookupOpenMiningRoundTriple` from scan-store, and first confirm if the current open rounds have `rewardConfig`, ie the daml has been updated to fill these. As once the daml is vetted and is in mainnet, all subsequent rounds would have `rewardConfig`, and at that point onwards it makes sense for this trigger to run. Doing this check should be fairly low on resources, as the open rounds are cached in store.

dfordivam: we could even add/expose this lookup api in `ScanRewardsReferenceStore` to keep dependency on a single store.

dfordivam: To be clear, I meant that at the start of trigger we confirm the latest / current open mining round has `rewardConfig`, if it does only then we run the subsequent `RewardComputationTrigger.nextRound` calcs. Though this isn't enough on its own, as we still need to skip all rounds between the `earliestRoundWithCompleteAppActivity` and the first round with with `rewardConfig`. But doing that skip could be done in a simple loop fetching all rounds one after another.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
dfordivam, line 254: In my local testing on this branch the trigger computes totals for round 6, but no further rounds are calculated even when the trigger is run multiple times at this point. Could you confirm in this test multiple invocations actually perform computations for subsequent rounds.

dfordivam: I guess this is expected as there is no round 8 currently. so round 7 won't get calculated. nevermind let me fix the test and continue debug.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: what about having an api `earliestRoundWithRewardConfig`? The result will determine whether to skip directly, and what to pass to nextRound E.g,
- None: skip
-  Some earliestWithRC: 
   - earliestComputableRound = max(earliestRoundWithCompleteAppActivity, Some earliestWtihRC)
   - use earliestComputableRound instead of earliestRoundWithCompleteAppActivity in the call to nextRound

does this have the same effect as what you are suggesting @dfordivam

dfordivam: Ok, adding an `earliestRoundWithRewardConfig` lookup in `ScanRewardsReferenceStore` may be good. I am ok with implementing this without requiring any DB modifications, ie confirm latest round has `rewardConfig`, then do a linear lookup of rounds in a loop to find the earliest (ie keep it simple.)
And if we make sure we call this API only when lookupLatestRoundWithRewardComputation is None, then it would in principle be only a one use API, and could be removed in future.

Also lets take a moment to consider the activity meta table effects on this.
The meta table would be used to constrain only the `latestRoundWithCompleteAppActivity` and `earliestRoundWithCompleteAppActivity` , so it should work with the above right?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -59,17 +62,41 @@ class RewardComputationTrigger(
         earliestCompleteO <- appActivityStore.earliestRoundWithCompleteAppActivity()
         latestCompleteO <- appActivityStore.latestRoundWithCompleteAppActivity()
```
dfordivam, line 72: In my testing today I found that a round with empty activity isn't handled by `latestRoundWithCompleteAppActivity`, hence stopping the calcs at that round.

dfordivam: rather its `assertCompleteActivity` which is probably causing the stop. It seems that by using less than round, more than round, (as opposed to round - 1, + 1) would allow the handling of rounds with zero activity.

dfordivam: note that its fine to create an issue and address this later, perhaps even after the dry-run milestone

adetokunbo: given you have clearer context, do you mind opening the issue ?

dfordivam: Ok added https://github.com/hyperledger-labs/splice/issues/5186

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: >  meta table would be used to constrain only the latestRoundWithCompleteAppActivity and earliestRoundWithCompleteAppActivity , so it should work with the above right?

that's correct

adetokunbo: I've justed added: 6a390efbb4da7eebc7cef35a7d67cd81c5ec7057 which should resolve this.
PTAL @dfordivam

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -59,17 +62,44 @@ class RewardComputationTrigger(
...
-        latestCompleteO,
-        latestComputedO,
-        batchSize,
-        inputs,
-      )
+        // While no rewards have been computed yet, skip pre-CIP-104 rounds
+        // by finding the earliest round with rewardConfig set.
+        earliestComputableO <- latestComputedO match {
+          case Some(_) => Future.successful(earliestCompleteO)
+          case None => earliestRoundWithRewardConfig(earliestCompleteO, latestCompleteO)
```
dfordivam: I just observed a bug introduced by this change. The latestComplete could in principle be less than earliestCompleteO. which results in earliestComputableO being less than earliestCompleteO

dfordivam: which points out to a test missing for such a scenario

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
```diff
@@ -203,10 +227,22 @@ class TrafficBasedRewardsTimeBasedIntegrationTest
...
       val totals = sv1ScanBackend.getRewardAccountingActivityTotals(earliest)
       totals.value.roundNumber shouldBe earliest
       totals.value.activityRecordsCount should be > 0L
+      totals.value.activePartiesCount shouldBe expectedProviders.size.toLong
+      totals.value.totalAppActivityWeight should be > 0L
+      // The total weight must be at least as large as the traffic cost from the
+      // test's known settlement, since that settlement contributes activity records
+      // to the round (other background transactions may also contribute).
+      val roundTrafficCost = trafficCostByRound(earliest)
+      totals.value.totalAppActivityWeight should be >= roundTrafficCost
```
dfordivam: Actually this assertion is not very useful. I tried to do exact totals calcs for a round on my branch, though it gets complicated as I have added multiple transfers in a round. So I will get rid of this assertion on my branch, and rely on store tests for the totals calcs

adetokunbo: Noted:  I'll leave this here for now, unless indicate you want it removed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Fixed, PTAL

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -59,17 +62,44 @@ class RewardComputationTrigger(
...
+          case Some(roundNumber) =>
+            rewardsReferenceStore.lookupOpenMiningRoundByNumber(roundNumber).map {
+              case None =>
+                logger.debug(
+                  s"OpenMiningRound for round $roundNumber not yet ingested, waiting."
+                )
+                Seq.empty
+              case Some(contract) =>
+                RewardComputationInputs.fromOpenMiningRound(contract.payload) match {
+                  case None =>
```
dfordivam: We have an issue here, the round checked here will actually never get skipped and will block the trigger to process further rounds.
I modified my integ test to add, and then remove `rewardConfig` from amulet config, and confirmed that trigger gets stuck.
We have to make sure we skip here as the mainnet has been running reward computation, and that will cause `latestCompleteO` to be `Some` even in absence of  `rewardConfig`.

dfordivam: Actually using `earliestRoundWithRewardConfig(Some roundNumber, latestCompleteO)` here should work right?
Perhaps we could re-structure the code such that finding the correct round with config is done at both places with single generalized API

dfordivam: Based on response by Simon yesterday on the doc, we dont need to handle this scenario right now. So lets make sure we truncate the tables on feature branch

adetokunbo: Noted, is that something that ought  to be done a follow-up PR?

dfordivam: yes, but could be done at the very end. And we already have a migration on the feature branch, so we could add truncate to existing migration and avoid file renaming

adetokunbo:
@rautenrieth-da, this is the PR that integrates the RewardComputationTrigger into this feature branch.  PTAL

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -113,46 +186,20 @@ object RewardComputationTrigger {
...
       earliestCompleteO: Option[Long],
       latestCompleteO: Option[Long],
       latestComputedO: Option[Long],
-      batchSize: Int,
-      inputs: RewardComputationInputs,
-  ): Seq[Task] =
+  ): Option[Long] =
     (earliestCompleteO, latestCompleteO) match {
-      case (Some(earliestComplete), Some(latestComplete)) =>
+      case (Some(earliestComplete), Some(latestComplete)) if earliestComplete <= latestComplete =>
```
rautenrieth-da: I think `earliestComplete <= latestComplete` is always true here, when both of them are defined:

`latestCompleteO` may be a round from before the completeness boundary (this will happen after bumping the store version and before ingesting a full round on the new store version).

However, if `earliestCompleteO` is defined then we have a complete round after the completeness boundary and thus `latestComplete` will also be after the completeness boundary and thus higher than `earliestComplete`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -59,17 +62,44 @@ class RewardComputationTrigger(
...
+              case None =>
+                logger.debug(
+                  s"OpenMiningRound for round $roundNumber not yet ingested, waiting."
+                )
+                Seq.empty
+              case Some(contract) =>
+                RewardComputationInputs.fromOpenMiningRound(contract.payload) match {
+                  case None =>
+                    logger.debug(
+                      s"Round $roundNumber missing rewardConfig or trafficPrice, skipping."
```
rautenrieth-da: We return no tasks here and don't update any data, so the next trigger iteration will do the exact same work again. I think the trigger would get stuck in this case processing the same round over and over.

We would need something like inserting data into the `app_reward_root_hashes` table that persists the fact that this trigger has processed the round but found that no rewards are to be calculated.

rautenrieth-da (commented):
This trigger tries to run the reward computation for each round, starting from the earliest round with the rewardConfig set. This results in a gapless history of reward data, but can  potentially lead to a lot of lag and useless computation when a SV node is catching up. 

We can change the whole logic for what round to process to a simple "Fetch the oldest active `CalculateRewardsV2` contract, and calculate the root hash its round".

Why does this work?
- We do not need a gapless history for the reward data. We only need it for processing rewards. Once all rewards are processed, the data will still be persisted in the UpdateHistory of `CalculateRewardsV2` and `ProcessRewardsV2` contracts. We therefore don't expect to ever need to make the internal scan API for reward processing public.
- The reward processing trigger _may_ read data from other scan instances if its own scan instance doesn't have the data, so we don't need to implement any backfilling:
  - When turning a `CalculateRewardsV2` into a `ProcessRewardsV2`, we need to make sure the voting on the root hash is BFT safe. If the round is after the completeness boundary of an SV, the SV should use its own scan data. If the round is before (e.g., because the SV has just onboarded or bumped their store version), the SV will do a BFT read from other SVs to avoid deadlocking the consensus.
  - When breaking down `ProcessRewardsV2` contracts, each SV may read reward data from any single other scan instead of its own. There is no BFT read required - since we have already voted on the root hash, the `ProcessRewardsV2_ProcessBatch` choice would immediately fail if called with malicious data. We would only have a problem if ALL nodes that computed the root hash went offline before the processing of rewards is finished.

This is all very complex, let's discuss this in our next call.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: This guard was added in eba909722, PTAL

E.g., with just 2 consecutive rounds (e.g., rounds 10 and 11), `earliestRoundWithCompleteAppActivity` returns 11 (min+1) and `latestRoundWithCompleteAppActivity` returns 10 (max-1), so `earliestComplete` > `latestComplete`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: The trigger _will_ get stuck in this situation, and if IIUC, in practice that will trigger alerts.
AFAIK, `RewardConfig` should never be `None` once this branch is merged; ideally there should have been an assertion upstream of the trigger (e.g. in daml) if this changes

dfordivam: I had a discussion about this [earlier](https://github.com/canton-network/splice/pull/5119#discussion_r3172569581).
Though the log message should definitely be modified, as the "skipping" is confusing to the code readers and the operators seeing such logs. How about some explicit error message like, this should never have happened, and a comment above this clarifying why we never expect it to happen.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
dfordivam: I actually hit this bug in my integration test and reported it. Perhaps a comment here would be good.

dfordivam: Actually this bug has not been fixed via this, as the `earliestRoundWithRewardConfig(earliestCompleteO, latestCompleteO)` can still return the lower value. I hit the issue again after rebasing on your latest branch, without my local fix.
Nevertheless we are planning to get rid of this logic altogether so its fine.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
rautenrieth-da: > AFAIK, RewardConfig should never be None once this branch is merged

Don't we want to be able to enable and disable dry run during testing? I thought only for the real rewards we expect to switch exactly once from on-ledger reward calculation to CIP-104 reward calculation.

dfordivam: we can  enable and disable dry run while keeping `rewardConfig` as `Some`. Of course this requires care during voting as the daml does accept setting  `rewardConfig` to `None`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Noted.

adetokunbo:
> This trigger tries to run the reward computation for each round, starting from the earliest round with the rewardConfig set. This results in a gapless history of reward data, but can potentially lead to a lot of lag and useless computation when a SV node is catching up.
> 
> We can change the whole logic for what round to process to a simple "Fetch the oldest active `CalculateRewardsV2` contract, and calculate the root hash its round".
> 
>  ....
> 
> This is all very complex, let's discuss this in our next call.

@dfordivam , @rautenrieth-da the trigger is now simplified as discussed, PTAL

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: The re-implementation changes this to an assertion with reporting language that avoids this confusion

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Noted.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
```diff
@@ -78,6 +79,12 @@ trait ScanRewardsReferenceStore extends AppStore {
       tc: TraceContext
   ): Future[Option[Contract[OpenMiningRound.ContractId, OpenMiningRound]]]
 
+  /** List active CalculateRewardsV2 contracts, sorted by round number ascending.
+    */
+  def listActiveCalculateRewardsV2(limit: Limit = defaultLimit)(implicit
```
dfordivam, line 84: It seems that we don't need this API on rewards reference store, as we only ever need the active contracts, they are available in scan-app's ACS store.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -56,20 +61,54 @@ class RewardComputationTrigger(
...
-        earliestCompleteO,
-        latestCompleteO,
-        latestComputedO,
-        batchSize,
-        inputs,
-      )
+        // For each eligible round until the configured parallelism limit, look
+        // up OpenMiningRound and extract inputs. The trigger will handle the
+        // remainder in subsequent polls.
+        tasks <- Future.traverse(eligible.take(context.config.parallelism)) { roundNumber =>
```
dfordivam: why did we add parallelism here. This seems to be fairly simple  DB lookup for task creation.
Also I think `MonadUtil.parTraverseWithLimit` is the correct API to use

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -91,9 +130,9 @@ class RewardComputationTrigger(
   override protected def isStaleTask(
       task: RewardComputationTrigger.Task
   )(implicit tc: TraceContext): Future[Boolean] =
-    appRewardsStore
-      .lookupLatestRoundWithRewardComputation()
-      .map(_.exists(_ >= task.roundNumber))
+    rewardsReferenceStore.multiDomainAcsStore
+      .lookupContractById(CalculateRewardsV2.COMPANION)(task.calculateRewardsId)
+      .map(_.isEmpty)
```
dfordivam: missing check for already computed totals for this round

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: IIUC `parTraverseWithLimit` will compute everything in parallel?  We really only want the first `n` of the parallelism limit.  I've actually pushed another commit with a fix that adds a helper to address this PTAL

As to why we add parallelism here, this emulating how the parallelism config is used in `ExpiresRewardsCouponsTrigger`, with a similar idea: to avoid contention in any scenario where there are large number of potential tasks.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: If that check happens when the tasks are filtered (as it does), does it also need to happen here?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
adetokunbo: Do you mean adding the `CalculateRewardsV2` contract filter to the main ScanStore instead? The trigger would need a new dependency on it, and the contract filter addition would still be needed there. I put it in `ScanRewardsReferenceStore` since the trigger already depends on it and it's purpose-built for reward computation reference data — but happy to move it if you prefer.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
dfordivam: this would be called only when `completeTask` fails, before doing the retry. If  `completeTask` fails ever happened after total were actually calculated, then  `completeTask` would be called again and it would fail again, resulting in a loop. This is my understanding, but I will let @rautenrieth-da to also provide clarifications on this.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
dfordivam: The ScanStore should be ingesting all dso visible contracts, if it isn't we should definitely add it there.
Here for the reference store I don't see a need to keep a reference of all these contracts. (as in, this contract is not meant to be "referenced")

dfordivam: I looked at ScanStore, its the scan updateHistory which probably ingests all visible contracts, for ScanStore adding a filter would be necessary. I still feel that using ScanStore is correct choice.

If we add to reference store, then we would end up keeping all these contracts.
Also if we do need to add to reference store, then lets try to make use of underlying DbMultiDomainAcsStore lookup APIs, (as opposed to writing the API ourselves)

dfordivam: @adetokunbo Actually after some thought I can see that there is some merit in keeping `CalculateRewardsV2` in reward reference store. As we keep its archival time. It could be used for metrics in future.
So lets keep this change, just re-use the DbMultiDomainAcsStore lookup API

adetokunbo: Fixed: it's now using DbMultiDomainAcsStore

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Fixed: sorry, I misunderstood this point before: obviously if the work is already done, the task should be stale; and there may be a gap between the filtering in `retrieveTasks` that generates a Task, and when `isStaleTask` is invoked on it that allows this to happen

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -88,12 +105,40 @@ class RewardComputationTrigger(
...
       }
 
+  private def buildTask(
+      roundNumber: Long,
+      contractId: CalculateRewardsV2.ContractId,
+  )(implicit tc: TraceContext): Future[Option[RewardComputationTrigger.Task]] =
+    rewardsReferenceStore.lookupOpenMiningRoundByNumber(roundNumber).map {
+      case None =>
+        logger.debug(
+          s"OpenMiningRound for round $roundNumber not yet ingested, waiting."
```
dfordivam: Feels like this should be an error, as discussed presence of activity-records for a round is a guarantee that we have OpenMiningRound in the store already ingested

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -56,20 +61,32 @@ class RewardComputationTrigger(
...
-        earliestCompleteO,
-        latestCompleteO,
-        latestComputedO,
-        batchSize,
-        inputs,
-      )
+        // Look up OpenMiningRound for eligible rounds and extract inputs,
+        // collecting up to the parallelism limit. Rounds whose OpenMiningRound
+        // is not yet ingested are skipped without counting toward the limit.
+        tasks <- RewardComputationTrigger.takeFirstN(
```
dfordivam: The `PollingParallelTaskExecutionTrigger` seems to be doing its own management of tasks parallelism. How about we let it handle it.

dfordivam: The default parallelism value is 4, so it makes sense to not limit here and let `PollingParallelTaskExecutionTrigger` process all rounds (4 at a time). Having 4 pending rounds in a bit unrealistic, but anyways handling them in one run of this trigger is still better than multiple runs.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Fixed: that makes sense

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Make sense:  as it now asserts that the `OpenMiningRound` must be present, takeFirstN ceases to do anything useful

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -88,12 +105,40 @@ class RewardComputationTrigger(
...
       task: RewardComputationTrigger.Task
   )(implicit tc: TraceContext): Future[Boolean] =
-    appRewardsStore
-      .lookupLatestRoundWithRewardComputation()
-      .map(_.exists(_ >= task.roundNumber))
+    for {
+      contractGone <- rewardsReferenceStore.multiDomainAcsStore
+        .lookupContractById(CalculateRewardsV2.COMPANION)(task.calculateRewardsId)
+        .map(_.isEmpty)
+      alreadyComputed <- appRewardsStore
```
dfordivam, line 123: We can avoid this DB read if contract is gone

adetokunbo: Fixed.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -56,20 +61,26 @@ class RewardComputationTrigger(
       Future.successful(Seq.empty)
     } else
       for {
+        // List active CalculateRewardsV2 contracts, ascending by round
+        // and filter out
+        // - rounds where the rewards are already computed
+        // - rounds with incomplete activity
+        candidates <- rewardsReferenceStore.listActiveCalculateRewardsV2()
```
rautenrieth-da, line 67: FYI, there is a `OnAssignedContractTrigger` base class that makes it easy to write triggers that do something for each new contract of a given template. I think it's still better to use `PollingParallelTaskExecutionTrigger` as you do now, because `OnAssignedContractTrigger` might give up on tasks that are failing, whereas the infinite polling loop makes sure we never skip rounds.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -610,19 +610,19 @@ class DbScanAppRewardsStore(
...
-      """.as[Option[Long]].headOption.map(_.flatten),
-      "appRewards.lookupLatestRoundWithRewardComputation",
-    )
+  ): Future[Set[Long]] = {
+    if (rounds.isEmpty) Future.successful(Set.empty)
+    else {
+      runQuery(
+        (sql"""select round_number from #${Tables.appRewardRootHashes}
+               where history_id = $historyId
+                 and """ ++ inClause("round_number", rounds)).toActionBuilder
```
rautenrieth-da, line 621: Query plan should be good due to 
```
    constraint app_reward_root_hashes_pkey primary key (history_id, round_number)
```
and the fact that the input sequence has at most 1000 elements.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
```diff
@@ -215,4 +216,13 @@ class DbScanRewardsReferenceStore(
...
     } yield result.headOption.map(contractFromRow(OpenMiningRound.COMPANION)(_))
   }
+
+  override def listActiveCalculateRewardsV2(limit: Limit = defaultLimit)(implicit
+      tc: TraceContext
+  ): Future[Seq[Contract[CalculateRewardsV2.ContractId, CalculateRewardsV2]]] =
+    waitUntilInitialized.flatMap { _ =>
+      multiDomainAcsStore
+        .listContracts(CalculateRewardsV2.COMPANION, limit)
+        .map(_.map(_.contract).sortBy(_.payload.round.number))
```
rautenrieth-da: We're sorting client-side after taking a db-side limit, this can result in unexpected outcomes.
If you really want to guarantee that the method returns the first N contracts in round order, you have to sort in the db query, i.e., use
```scala
storage.query(
  selectFromAcsTableWithState(
    ...,
    orderLimit = ...,
  )
)
```
and check that the resulting query uses the right index.

Also, there can be two CalculateRewardsV2 contracts for the same round, so for an ordering that is guaranteed to be stable across SVs, you would need to additionally order by something like the dry run flag or the contract id. This doesn't look relevant in our case, so I would not bother with it.

Alternatively, you can not sort the contracts, and accept the fact that if there are many rounds with outstanding reward calculation, they will be processed in a random order.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
rautenrieth-da, line 67:
```suggestion
        candidates <- rewardsReferenceStore.listActiveCalculateRewardsV2(PageLimit.tryCreate(???))
```

By default, list-active-contract queries will return up to 1000 contracts, so in worst case we would create up to 1000 trigger tasks here.

Since the trigger processes tasks in parallel (by default 4 at a time), you could in theory run into situations where round `N` is calculated but round `N - 999` is not yet finished. If we limit the number of tasks to 4, we would get more predictable behavior, at the cost of slower catch up speed.

I'm fine with either choice.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -56,20 +61,26 @@ class RewardComputationTrigger(
...
-        batchSize <- Future.successful(RewardComputationTrigger.placeholderBatchSize)
-      } yield RewardComputationTrigger.nextTask(
-        earliestCompleteO,
-        latestCompleteO,
-        latestComputedO,
-        batchSize,
-        inputs,
-      )
+        // Look up OpenMiningRound for each eligible round and extract inputs.
+        tasks <- Future.traverse(eligible)(r => buildTask(r, roundToContract(r)))
```
rautenrieth-da: A few general comments on the `retrieveTasks()` method:

1. Move looking up the round contract and building inputs to `completeTask()`. Once a round passes the point where you compute `eligible`, we don't expect these steps to fail due to some store being behind, so it's just work that has to be done as part of the task. Then you also don't need the `Future.traverse` to create the tasks.
2. There can be two `CalculateRewardsV2` contracts for the same round - one for the dry run and one for real rewards. `roundToContract` and thus the task will randomly contain one of the two contracts, and `isStaleTask` _might_ not work correctly. I think it's better if the task just contains the round number, and the staleness check has to be a custom query that checks if an active CalculateRewardsV2 for a given round exists.

With this, the code can be simplified:

```scala

final case class Task(round: Long)

// The N oldest CalculateRewardsV2 contracts
candidates <- rewardsReferenceStore.listActiveCalculateRewardsV2(PageLimit.tryCreate(N))

// List of rounds with at least one CalculateRewardsV2 contract (still ordered)
candidateRounds = candidates.map(c.payload.round.number.toLong).distinct

// Out of the above, the set of rounds with already calculated rewards
alreadyCalculated <- appRewardsStore.roundsWithComputedRewards(
  candidateRounds
)

// List of rounds for which we can calculate rewards, still sorted by round number
tasks = candidateRounds.filter(round => {
  !alreadyCalculated.contains(round) &&
  earliestCompleteO.exists(_ <= round) &&
  latestCompleteO.exists(_ >= round)
}).map(round => Task(round))
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
```diff
@@ -88,12 +99,43 @@ class RewardComputationTrigger(
...
-      .map(_.exists(_ >= task.roundNumber))
+    rewardsReferenceStore.multiDomainAcsStore
+      .lookupContractById(CalculateRewardsV2.COMPANION)(task.calculateRewardsId)
+      .flatMap {
+        case None => Future.successful(true)
+        case Some(_) =>
+          appRewardsStore
+            .roundsWithComputedRewards(Seq(task.roundNumber))
+            .map(_.nonEmpty)
+      }
```
rautenrieth-da:
```suggestion
    for {
      contractExists <- rewardsReferenceStore
              .listActiveCalculateRewardsV2ForRound(task.round)
          .map(_.nonEmpty)
      
      alreadyComputed <- appRewardsStore
        .roundsWithComputedRewards(Seq(task.round))
        .map(_.nonEmpty)
    } yield !contractExists || alreadyComputed
```

It should be rare that a task becomes stale. I would optimize for readability here, like you had it in your previous version of this method.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/RewardComputationInputs.scala
```diff
@@ -94,14 +96,44 @@ final case class RewardComputationInputs(
 
 object RewardComputationInputs {
   private val scale: Numeric.Scale = Numeric.Scale.assertFromInt(10)
-  private[rewards] val zero: Numeric = fromLong(0L)
+  private[scan] val zero: Numeric = fromLong(0L)
+  private val defaultDevelopmentFundPercentage: BigDecimal = BigDecimal("0.05")
```
rautenrieth-da, line 101: Can you please add a comment for where this number is coming from?

rautenrieth-da (commented):
Thanks! I like this version where the trigger is driven by active CalculateRewardsV2 contracts. Left a few small comments, mostly around handling multiple CalculateRewardsV2 contracts for the same round.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/rewards/RewardComputationInputs.scala
adetokunbo: Fixed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Fixed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Thanks, both points here have been resolved; that there might be two valid `CalculateRewardsV2` and its impact on staleness was something that I had missed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Initially, I added a limit here, but it caused `TrafficBasedAppRewardsTimeBasedIntegrtaionTest` to fail.
I looked at options for updating the test, but rather than change it, since it's ok to avoid a limit, I'd rather make fewer changes to the integration test.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
adetokunbo: Fixed: changed back to ordering on the db-side.  IIRC, there was a further point to that, related to ensuring that the the scan instances of different SV's all perform this order in the same sequence, which will avoid a certain class of  distributed node ordering issue

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
adetokunbo: Noted

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/RewardComputationTrigger.scala
adetokunbo: Agreed

adetokunbo:
@rautenrieth-da I've addressed the comments, PTAL

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
```diff
@@ -526,10 +514,13 @@ class DbScanAppRewardsStoreTest
           Seq("bob::provider"),
           Seq(500000L),
         )
+        zeroThresholdInputs = testInputs.copy(
+          appRewardCouponThreshold = RewardComputationInputs.zero
```
rautenrieth-da: 1. There is `Numeric.zero`, can you use it directly instead of the custom helper?
2. `testInputs` already has a zero `appRewardCouponThreshold`, so this line does nothing. Make sure we have at least one test with a non-zero threshold and activity that falls below the threshold.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
```diff
@@ -445,59 +444,48 @@ class DbScanAppRewardsStoreTest
         result <- store.aggregateActivityTotals(roundNumber).failed
       } yield {
         result.getMessage should include("Incomplete app activity")
-        result.getMessage should include(s"round ${roundNumber + 1} exists=false")
+        result.getMessage should include("later round exists=false")
       }
     }
 
-    // -- lookupLatestRoundWithRewardComputation ------
+    // -- roundsWithComputedRewards ------
```
rautenrieth-da: nit: "should" blocks can be nested, you can turn these comments into actual structure in the test, e.g.:

```scala
"DbScanAppRewardsStore" should {
  "roundsWithComputedRewards" should {
    "return empty set for empty input" in { ... }
    "return correct subset" in { ... }
  }
}
```

This is mostly cosmetic, but can also make navigation easier in some IDEs.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
adetokunbo: Fixed, Numeric.zero is now used here, and a new test with nonZero appRewardCouponThreshold discriminates has been added.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
adetokunbo: This now fixed for the tests added by this PR; did you also want this fixed for pre-existing tests ?

adetokunbo:
@rautenrieth-da,  those comments have been addressed, PTAL

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
rautenrieth-da: I would prefer if the style was consistent at least within the file. We can do that in a separate PR though.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
```diff
@@ -445,115 +444,136 @@ class DbScanAppRewardsStoreTest
...
+          // alice has high activity, bob has low activity
+          _ <- insertActivityRecord(
+            historyId,
+            roundNumber,
+            Seq("alice::provider", "bob::provider"),
+            Seq(5000000L, 50000L),
+          )
+          nonZeroThresholdInputs = testInputs.copy(
+            appRewardCouponThreshold = RewardComputationInputs.fromBigDecimal(BigDecimal("0.5"))
+          )
```
rautenrieth-da, line 546: The test is failing as is.

```suggestion
          // totalIssuanceForFeaturedAppRewards is 0.45, and alice gets almost all of it. Only alice will therefore be above the threshold of 0.4

          nonZeroThresholdInputs = testInputs.copy(
            appRewardCouponThreshold = RewardComputationInputs.fromBigDecimal(BigDecimal("0.4"))
          )
```

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
adetokunbo: Noted

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
adetokunbo: Fixed

adetokunbo:
@rautenrieth-da , sorry about last issue with the test; I thought I had run the store tests locally before asking for the review,   I've updated the test and run the tests locally, PTAL

rautenrieth-da (approved):
Thanks!

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbScanAppRewardsStoreTest.scala
adetokunbo: Consistent style tracked by #5687

PR #5120, opened 2026-04-20
implement participant migration procedure to enable LSU

mblaze-da:
The migration procedure is done in two steps:
1. Import participant's DB resources in the `sv` stacks and prepare them to be removed from the `sv-canton` stacks. To do this:
    - deploy once with `synchronizerMigration.active.migrateParticipantsFromSvCantonToSv` set to `true`
2. Remove participant's DB resources from the `sv-canton` stacks, down participant chart from `sv-canton` and bring it up from the `sv` stack. Restarting the participant is also necessary due to DB user's password change. The password cannot be migrated. To do this:
    - prepare post migration config (these changes are permanent)
      - remove `synchronizerMigration.active.migrateParticipantsFromSvCantonToSv`
      - set `synchronizerMigration.active.enableLogicalSynchronizerDeploymentMode` to `true`
      - make sure that `synchronizerMigration.frozenMigrationId` is set
    - deploy once to finish migrating


cluster/pulumi/common-sv/src/synchronizer/participant.ts
```diff
@@ -101,7 +115,8 @@ function installParticipantChart(
       host: db.address,
       secretName: db.secretName,
       // the following will not be needed when the MIGRATION_ID gets removed from defaults
-      databaseName: `participant${migrationSuffix(migration?.id, '_')}`,
+      databaseName:
+        participant?.legacyDatabaseName ?? `participant${migrationSuffix(migration?.id, '_')}`,
```
moritzkiefer-da: not sure I understand why we need this. Can we not rely on the frozenMigrationId field?

cluster/pulumi/common-sv/src/synchronizer/participant.ts
```diff
@@ -48,28 +49,41 @@ export type ParticipantArgs = {
     id: DomainMigrationIndex;
     isStillRunning: boolean;
   };
+  migratingDatabaseInstanceName?: string;
+  migratingDatabaseSecretName?: string;
+  yieldManagement?: boolean;
```
moritzkiefer-da: do we use yieldManagement for more than setting retainOnDelete? If not maybe just call it retainOnDelete?

cluster/pulumi/common/src/postgres.ts
```diff
@@ -51,62 +53,71 @@ export interface Postgres extends pulumi.Resource {
...
       );
     }
-    this.zone = zoneFromEnv;
+    const zone = zoneFromEnv;
+
+    const databaseInstanceImportOpts =
+      existingInstanceName !== undefined
+        ? {
+            import: existingInstanceName,
+            ignoreChanges: ['userLabels'],
```
moritzkiefer-da, line 115: why do these change?

moritzkiefer-da:
> The migration procedure looks as follows:

Can you describe what happens in each step? It's a bit tricky to follow in the code. My understanding is:

1. Step 1 sets retain on delete on the DB. I think you are also assuming that in this case the sv stack gets deployed and can import the DB from the sv-canton outputs? If so, that seems like an essential step. I think the participant still gets deployed from sv-canton?
2. Step 2 (and step 3, afaict this is only one PR so not sure why it's 2 steps?) then actually removes theDB from the sv-canton stack which is fine due to retainOnDelete moves the participant chart to be deployed from sv?

mblaze-da:
> > The migration procedure looks as follows:
> 
> Can you describe what happens in each step? It's a bit tricky to follow in the code. My understanding is:
> 
>     1. Step 1 sets retain on delete on the DB. I think you are also assuming that in this case the sv stack gets deployed and can import the DB from the sv-canton outputs? If so, that seems like an essential step. I think the participant still gets deployed from sv-canton?
> 
>     2. Step 2 (and step 3, afaict this is only one PR so not sure why it's 2 steps?) then actually removes theDB from the sv-canton stack which is fine due to retainOnDelete moves the participant chart to be deployed from sv?

@moritzkiefer-da your analysis is pretty much spot on. I've updated the PR description to make it more clear.

cluster/pulumi/common-sv/src/synchronizer/participant.ts
mblaze-da: You're right.

cluster/pulumi/common/src/secrets.ts
```diff
@@ -214,9 +214,18 @@ export function svCometBftGovernanceKeySecret(
 
 export function installPostgresPasswordSecret(
   ns: ExactNamespace,
-  password: pulumi.Output<string>,
-  secretName: string
+  password: pulumi.Input<string>,
+  secretName: string,
+  existingSecretName?: string,
+  yieldManagement: boolean = false
```
nicu-da: Why the `yieldManagement` naming?

cluster/pulumi/deployment/src/stacks/splice.ts
```diff
@@ -69,7 +69,8 @@ export function* getSpliceStacksFromMainReference(): Generator<StackFromRef> {
   }
   if (
     deploymentConf.projectsToDeploy.has('sv') &&
-    DecentralizedSynchronizerUpgradeConfig.active.enableLogicalSynchronizerDeploymentMode
+    (DecentralizedSynchronizerUpgradeConfig.active.enableLogicalSynchronizerDeploymentMode ||
+      DecentralizedSynchronizerUpgradeConfig.active.migrateParticipantsFromSvCantonToSv)
```
nicu-da: Let's avoid doubling these conditions and just enforce the the migrateParticipantFromSvCantonToSv can be enabled only if enableLogicalSynchronizerDeploymentMode is also enabled

moritzkiefer-da (approved):
Thanks! lgtm minus the frozen migration id change

nicu-da (commented):
Thanks! Nothing else to add compared to @moritzkiefer-da 's comments. Lets make sure we use the frozen migration id to avoid needing any other config changes for the migration.

cluster/pulumi/common-sv/src/synchronizer/participant.ts
mblaze-da: It is used both on `retainOnDelete` and disabling the Pulumi `protect` option. The idea behind the name was that this flag in some sense makes the current stack yield the management of these resources. It does not yield it completely as a hypothetical modification afterwards would still be applied but it stops managing deletion. Now that I had to explain this I'm starting to lean towards something more specific like `retainDbResourcesOnDelete`.

cluster/pulumi/common/src/postgres.ts
mblaze-da: Because in the `sv` stack I generally tried to avoid using migration ID if it's not necessary. For example the participant instance is just called `participant` instead of `participant-0`. This also means that the `migration_id` label is not present in the `sv` stack deployment. To get the resource import to work I have to ignore this change as resource state and definition must be identical during the import. During the subsequent deployment (step 2) this label is actually removed.

cluster/pulumi/common/src/secrets.ts
mblaze-da: I renamed this.

nicu-da (approved):

PR #5124, opened 2026-04-20
use bigquery config instead of env var

isegall-da:
Apparently we already have a scanApp.bigQuery config schema, but we never actually used it. Instead, we used an envrc var to enable bigquery, and hard-coded values in the static config. This cleans that up. 

Tested on a scratchnet.
Also confirmed that without the bigquery config in mock the mock expected files dont actually change.

Requires updating configs in cn-internal, which I will do once approved and merged here.
I also think that I'll backport this to 0.5.18, to isolate issues in this from the coming 0.6.0 upgrade. Objections?

(This is a pre-step to moving parts of the bigquery pulumi code out of canton-network stack, into its own stack. At least the parts that will not trigger reingesting everything)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/common-sv/src/svConfigs.ts
+++ b/cluster/pulumi/common-sv/src/svConfigs.ts
@@ -89,7 +89,7 @@ export const standardSvConfigs: StaticSvConfig[] = (
               privateKey: svCometBftSecrets[0].validatorPrivateKey,
               publicKey: 'H2bcJU2zbzbLmP78YWiwMgtB0QG1MNTSozGl1tP11hI=',
             },
-        },
+          },
         },
       ]
     : [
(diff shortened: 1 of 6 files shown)


cluster/pulumi/common-sv/src/svConfigs.ts
```diff
@@ -95,9 +90,6 @@ export const standardSvConfigs: StaticSvConfig[] = (
               publicKey: 'H2bcJU2zbzbLmP78YWiwMgtB0QG1MNTSozGl1tP11hI=',
             },
           },
-          ...(sv1ScanBigQuery
-            ? { scanBigQuery: { dataset: 'mainnet_da2_scan', prefix: 'da2' } }
-            : {}),
```
stephencompall-DA, line 100: This should be fine as long as the config doesn't break when you try to distinguish between sv-1 (which should have this config) and other svs (which shouldn't).

stephencompall-DA (approved):

cluster/pulumi/common-sv/src/svConfigs.ts
isegall-da: I think it would not break, but indeed on scratchnet I tested it only on a single-sv network. Will extend the manual test on a scratchnet to four svs to confirm, and the best test will be just previews on devnet&mainnet when we backport this and adapt the configs accordingly.

moritzkiefer-da (approved):
thanks looks reasonable and a backport makes sense to me

PR #5138, opened 2026-04-21
SplitwellIntegrationTest Increase timeout for GroupRequest Archival

pasindutennage-da:
Fix https://github.com/DACH-NY/cn-test-failures/issues/7865

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SplitwellIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SplitwellIntegrationTest.scala
@@ -51,7 +51,7 @@ class SplitwellIntegrationTest
 
           // Wait for request to be archived and therefore either the group to be created or
           // the request to be rejected.
-          eventually() {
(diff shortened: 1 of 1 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/SplitwellIntegrationTest.scala
```diff
@@ -51,7 +51,7 @@ class SplitwellIntegrationTest
 
           // Wait for request to be archived and therefore either the group to be created or
           // the request to be rejected.
-          eventually() {
+          eventually(timeUntilSuccess = 40.seconds) {
```
pasindutennage-da, line 54: By default eventually has a timeout of 20s

In this test case, multiple retries from the duplicate GroupRequest often hits the circuit breaker, that enforces an additional 30s timeout before next allowed request.

40s seem like a good balance

Please see https://github.com/DACH-NY/cn-test-failures/issues/7865 for analysis

moritzkiefer-da (approved):
thx

julientinguely-da (approved):
thanks

Issue #5148, opened 2026-04-21
Make sure that `multi-participant`s inherit the upstream `canton-participant` config


martinflorian-da:
They currently don't, leading to confusing situations such as missing this: https://github.com/DACH-NY/canton/pull/31798

Suggestion by @moritzkiefer-da 

> Maybe a clean option:
>
> - Split the config in Canton in two parts. The participant config and the part that then includes it under canton.participants.participant
> - In splice we inherit from that, overwrite the part that sets canton.participants.participant and include the other for each of the 10 nodes

Specifically, the goal is to reuse the `participants.participant` defaults for each of our multi-participants.

PR #5149, opened 2026-04-21
`multi-participant`: use bumped participant encryption key cache defaults

martinflorian-da:
Apply https://github.com/DACH-NY/canton/pull/31798 because https://github.com/hyperledger-labs/splice/issues/5148 exists

This should fix CILR tomorrow and DevNet on Thursday... setting reminder to myself to check.

[static]


moritzkiefer-da (approved):
thx, don't know if the cost impact is big enough to justify backporting to devnet

PR #5161, opened 2026-04-22
Fix alert for number of featured app rights

rautenrieth-da:
Fixes https://github.com/DACH-NY/canton-network-internal/issues/4712

Also changes alerting interval from 1min to 60min, and switches to `noDataState: OK` (as used by most alerts). The underlying metric (number of featured app rights) is not expected to change at a high frequency.


moritzkiefer-da (approved):
thanks!

julientinguely-da (approved):
thanks

PR #5168, opened 2026-04-22
Fix UnsupportedPackageVettingIntegrationTest when some other test vets beforehand

OriolMunoz-da:
[ci]
As found in the token standard branch


moritzkiefer-da (approved):
thx!

PR #5174, opened 2026-04-22
Remove grpcurl from canton images

krzysztofczyz-da:


cluster/compose/localnet/docker/canton/Dockerfile
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2025, Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: 0BSD
+ARG BASE_IMAGE
+
+FROM $BASE_IMAGE
+USER root
+RUN curl -L https://github.com/vadimi/grpc-client-cli/releases/download/v1.24.3/grpc-client-cli_linux_x86_64.tar.gz | tar -C /usr/local/bin -xz
```
moritzkiefer-da, line 7: @isegall-da can we use this now? I thought Ed had some concerns?

cluster/images/canton/Dockerfile
```diff
@@ -8,5 +8,6 @@ FROM europe-docker.pkg.dev/da-images/public-all/docker/canton-base:${canton_vers
 
 COPY additional-config.conf /app/
 USER root
```
moritzkiefer-da, line 10: this doesn't actually remove it from the participant image and other stuff does it? And those still have the grpcurl based health check

cluster/compose/localnet/docker/canton/Dockerfile
moritzkiefer-da: probably we want to use https://github.com/grpc-ecosystem/grpc-health-probe instead which seems like a less dodgy "app from random guy" thing

cluster/images/canton/Dockerfile
krzysztofczyz-da: You're right, I confused the way it's done in canton and the way it's done in splice 😕

cluster/compose/localnet/docker/canton/Dockerfile
isegall-da: He definitely told us not to. @krzysztofczyz-da I'm surprised to still see this here. I thought we agreed that we're confirming that we use grpcurl only in non-prod, and that we'll add it there explicitly.

isegall-da: Or did we conclude that we need it in prod? This requires a discussion on what we replace it with, if at all. We can't just ignore the CISO's guidance.

moritzkiefer-da: >  I thought we agreed that we're confirming that we use grpcurl only in non-prod, and that we'll add it there explicitly.

I think sadly this only holds for k8s. For docker compose we do rely on the docker health checks I believe which are grpcurl based. So we do need an alternative. grpc-health-probe looks like a pretty trustyworthy one though (that was literally the recommendation in k8s before they added builtin support).

isegall-da: ACK. I'm happy to bring that option to Ed et al as our current plan on record if we think we can indeed use it.

isegall-da: (https://daholdings.slack.com/archives/C0AVD9FM7CG/p1777549449182359?thread_ts=1776862329.427069&cid=C0AVD9FM7CG)

krzysztofczyz-da:
This won't be needed anymore, closing

PR #5178, opened 2026-04-22
cncluster command for generating ignore list for ExpiredAmuletTransferInstructionTrigger

rautenrieth-da:
The list already has 180 entries on DevNet, and requires updating from time to time until all expired transfer instructions will be processed. 

The command will likely become useless after all expired instructions are archived, the PR is for documentation purposes.

### DIFF
--- a/build-tools/cncluster
+++ b/build-tools/cncluster
@@ -2470,6 +2470,61 @@ function subcmd_psql() {
       --dbname="$DB_NAME"
 }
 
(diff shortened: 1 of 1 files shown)


moritzkiefer-da:
fwiw what I ended up doing was to just get all senders and receivers out of the DB and then filter them by vetting state. That catches everyone that has unvetted it in one go without repeatedly needing to adjust the list.

You're then left with the the ones that are actually unresponsive. I just did that manually as it was just a handful.

PR #5190, opened 2026-04-23
add a guard condition to the sv project so that even when it somehow gets run the semantics are consistent with stack guards in the deployment project

mblaze-da:


nicu-da (approved):
Thx!

PR #5197, opened 2026-04-23
add DBs from the sv stacks to the list of Pulumi managed DBs in the purge_unused_dbs script

mblaze-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/8164


nicu-da (approved):
looks sensible, thanks!

PR #5199, opened 2026-04-23
Ignore splitwell slow connects in BootstrapTest

OriolMunoz-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/8158


pawelperek-da (approved):
LGTM

PR #5243, opened 2026-04-27
Update Token Standard Devnet version to 0.6.1-snapshot.20260427.2678.0.v86ff9ad5

OriolMunoz-da:
https://token-std-v2-dev.global.canton.network.digitalasset.com/version

[static]

I'm uploading the release bundle now, the link doesn't need to change.


meiersi-da (approved):
ty

PR #5248, opened 2026-04-27
fix: update dependabot config

krzysztofczyz-da:


.github/dependabot.yml
```diff
@@ -93,7 +93,7 @@ updates:
           - "minor"
           - "patch"
   - package-ecosystem: "npm"
-    directory: "/token-standard"
+    directory: "/token-standard/cli"
```
pawelperek-da, line 96: Why is it necessary? shouldn't `/token-standard` cover both `/token-standard` and `/token-standard/cli`?

krzysztofczyz-da: Idk, but dependabot sees packages under `/token-standard/cli` 🤔 
I.e. [here](https://github.com/hyperledger-labs/splice/security/dependabot/67)

pawelperek-da (approved):
LGTM

PR #5255, opened 2026-04-27
Create GH issue when store performance thresholds breached

jagathweerasinghe-da:
- This PR only focuses on communication of threshold breaches via GH issues. For that we used some fix thresholds in store-perf-thresholds.json
- The next PRs addresses the definition and proper calculation of the thresholds. 
- Test GH Isssue: https://github.com/DACH-NY/cn-test-failures/issues/8277
- Breach summary example: https://github.com/canton-network/splice/actions/runs/25444103583
- Breach summary includes the threshold and the collected metrics for the test


cluster/pulumi/observability/grafana-alerting/store_ingestion_performance_alerts.yaml
```diff
@@ -0,0 +1,126 @@
...
+              refId: B
+          - refId: C
+            datasourceUid: __expr__
+            model:
+              type: threshold
+              expression: B
+              conditions:
+                - evaluator:
+                    type: gt
+                    params: [300]
```
OriolMunoz-da: My main worry is, with an example:
- Ingestion test now takes ~200s
- A day later somebody does a change that makes ingestion 25% slower (that's very significant)
- Now the test takes 250s, which would still be < 300, we wouldn't alert, and likely nobody would notice

This is fine as an approach (per the design doc, "the CI run fails if the performance is below expected throughputs"), but then:
- These thresholds need to be tight enough that we'd detect big swings, but not flake (too often, I guess). They might be already, I cannot tell from this PR only
- We need to document our expectations to contributors wrt these tests (DEVELOPMENT.md/CONTRIBUTING.md, maybe?). From the design doc: "We expect that most changes won’t affect ingestion performance and for the ones that do the PR author runs the performance test. We can still switch to a more restrictive approach if we see that it is necessary."
- Likewise we should document for us what to do if these alerts get triggered (debug/check commits/bump the threshold here...)
- We should document (a comment here is fine) how/when these thresholds were set for reference. E.g. "performance tests take 280s~ as of yyyy-mm-dd"

cluster/pulumi/observability/grafana-alerting/store_ingestion_performance_alerts.yaml
OriolMunoz-da, line 1: btw, do you know where do the alerts from the splice cluster end up?

cluster/pulumi/observability/grafana-alerting/store_ingestion_performance_alerts.yaml
```diff
@@ -0,0 +1,126 @@
...
+        title: Scan Store Ingestion Total Time Too High
+        condition: C
+        data:
+          - refId: A
+            relativeTimeRange:
+              from: 86400
+              to: 0
+            datasourceUid: prometheus
+            model:
+              expr: splice_perf_ingestion_total_time_ns{test="ScanStoreIngestionPerformanceTest"} / 1e9
```
OriolMunoz-da: is there no easy way to have a single alert instead of a separate one? I'd like to not have to add an alert when we add a test, which is easy to forget

I can imagine that the different threshold per test makes it difficult, but maybe there's a way to do that

jagathweerasinghe-da: Simplified the rules, so that the changes needed when adding a new test is minimum.

cluster/pulumi/observability/grafana-alerting/store_ingestion_performance_alerts.yaml
jagathweerasinghe-da: It ends up in team-canton-network-internal-ci
./cluster/deployment/splice/.envrc.vars
`export SLACK_ALERT_NOTIFICATION_CHANNEL_FULL_NAME="team-canton-network-internal-ci"
`

OriolMunoz-da: pretty sure that that channel is mostly ignored in favor of the test failures dashboard, you might want to ask for opinions on #internal

cluster/pulumi/observability/grafana-alerting/store_ingestion_performance_alerts.yaml
jagathweerasinghe-da: Let's add the documentation to PERFORMANCE.md as suggested by the design doc.

jagathweerasinghe-da: Working on determining the threshold. We need to take into the datasize/batchsize as well as the dataset changes over time.

cluster/pulumi/observability/grafana-alerting/store_ingestion_performance_alerts.yaml
jagathweerasinghe-da: @OriolMunoz-da 

As per the internal discussion we had, let's go with the GH issue. Thanks for bringing this point:

There are two routes to create a GH issue:
1. GH Action->Prometheous->Grafana Alerts->GH Issue
2. GH Action->GH Issue

I would like go ahead with the second approach, which makes the integration clean and simple. WDYT?

OriolMunoz-da: yeah, 2 seems easiest

jagathweerasinghe-da:
@OriolMunoz-da PTAL. Please check the updated summary of the PR. This PR only focuses on GH issue creation on threshold breach.

.github/workflows/performance_tests.yml
```diff
@@ -69,6 +69,20 @@ jobs:
         run: |
           python3 scripts/performance/push_store_ingestion_perf_metrics.py
 
+      - name: Upload ingestion metrics artifact
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: store-ingestion-perf-metrics-${{ github.run_id }}
+          path: /tmp/store-ingestion-perf-metrics
+          if-no-files-found: ignore
```
OriolMunoz-da: when would we not have files?

scripts/performance/check_perf_thresholds.py
```diff
@@ -0,0 +1,173 @@
...
+        data = json.load(f)
+    if not isinstance(data, dict):
+        raise ValueError(f"thresholds file {path} must be a JSON object")
+    return {k: v for k, v in data.items() }
+
+
+def iter_metric_files(dirs: Iterable[Path]) -> Iterable[Path]:
+    for d in dirs:
+        if not d.is_dir():
+            print(f"warning: metrics dir {d} does not exist, skipping", file=sys.stderr)
```
OriolMunoz-da: also not clear here when it wouldn't exist and why it's fine to skip

scripts/performance/check_perf_thresholds.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    breaches: list[str] = []
+    files_seen = 0
+
+    for f in iter_metric_files(metric_dirs):
+        files_seen += 1
+        try:
+            data = json.loads(f.read_text())
+        except Exception as e:
+            print(f"warning: cannot parse {f}: {e}", file=sys.stderr)
```
OriolMunoz-da: do these warnings get captured by the log checker? otherwise this seems that it should be reported as a failure - but we have the `|| true` in the GHA definition

scripts/performance/check_perf_thresholds.py
```diff
@@ -0,0 +1,173 @@
...
+            continue
+
+        test_name = data.get("test_name") or f.stem
+        rules = thresholds.get(test_name)
+        if not rules:
+            print(f"info: no thresholds configured for '{test_name}', skipping ({f.name})")
```
OriolMunoz-da: I'm not convinced that this shouldn't be an error

OriolMunoz-da (approved):
approach looks good, thanks! just some questions about error handling

scripts/performance/check_perf_thresholds.py
jagathweerasinghe-da: made it an explicit error

scripts/performance/check_perf_thresholds.py
jagathweerasinghe-da: made it an explicit error

scripts/performance/check_perf_thresholds.py
jagathweerasinghe-da: gone, left over from tests

PR #5082, opened 2026-04-16
Wrap annotation field in a check for empty annotations

georgeorellanaDA:
#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a logical synchronizer upgrade test is required (from canton-3.5), comment `/lsu_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [x] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [x] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

Fixes #5079 by checking if default-annotations is empty before templating the annotations field

Current state
Screenshot 2026-04-16 at 2 34 25 PM

After changes
Screenshot 2026-04-16 at 2 34 41 PM

Issue #5079, opened 2026-04-16
Bug: Empty Annotations Block using 0.5.18 when disabling Reloader


georgeorellanaDA:
When using the splice-participant:0.5.18 helm chart with `enableReloader=false` set in a values.yaml file, a bug appears in the templating of the helm chart:

```bash
Pulling ghcr.io/digital-asset/decentralized-canton-sync/helm/splice-participant:0.5.18
Comparing release=participant-curr, chart=/Users/georgeorellana/Library/Caches/helmfile/oci__ghcr_io/digital-asset/decentralized-canton-sync/helm/splice-participant/0.5.18/splice-participant, namespace=lab-client
lab-client, participant-curr, Deployment (apps) has been added:
+ # Source: splice-participant/templates/participant.yaml
+ # Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+ # SPDX-License-Identifier: Apache-2.0
+
+ apiVersion: apps/v1
+ kind: Deployment
+ metadata:
+   annotations:
+
+   labels:
+     app: participant-curr
+     app.kubernetes.io/instance: participant-curr
+     app.kubernetes.io/managed-by: Helm
+     app.kubernetes.io/name: participant-curr
+     app.kubernetes.io/version: 0.5.18
+     helm.sh/chart: splice-participant-0.5.18
+   name: participant-curr
+   namespace: lab-client
```

The blank annotations field causes issues in rendering differences between helm charts and should not be rendered at all if `splice-util-lib.default-annotations` is empty

georgeorellanaDA:
I have a branch for not generating the annotations field on an empty default-annotations but I do not have contribution access to the repo

moritzkiefer-da:
> I have a branch for not generating the annotations field on an empty default-annotations but I do not have contribution access to the repo

you can just make a PR from a fork

georgeorellanaDA:
> > I have a branch for not generating the annotations field on an empty default-annotations but I do not have contribution access to the repo
> 
> you can just make a PR from a fork

Of course, just created ~~#5081~~ #5082

PR #4063, opened 2026-02-20
Add reloader helm release

krzysztofczyz-da:
This PR installs [reloader](https://github.com/stakater/Reloader/tree/master) to our clusters.

You can check the values [here](https://github.com/stakater/Reloader/blob/master/deployments/kubernetes/chart/reloader/README.md).

Fixes https://github.com/DACH-NY/canton-network-internal/issues/2365

#### Cluster Testing
- [x] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [x] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [x] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [x] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [x] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [x] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #5283, opened 2026-04-28
fix(charts): restore Stakater Reloader annotation rendering

### LINKED ISSUE #5282: Regression in 0.6.1: PR #5082 silently drops Stakater Reloader annotation in all charts
## Summary

PR #5082 (merged into `main` on 2026-04-17, shipped in **Splice 0.6.1**) introduced a regression that silently strips the `reloader.stakater.com/auto: "true"` annotation from every Deployment in 14 charts, **even when `enableReloader` is true** (the default).

In our deployment we observed this on `validator-app`, `participant`, `ans-web-ui`, and `wallet-web-ui` Deployments after upgrading from 0.5.18 → 0.6.1. ConfigMap/Secret reloads no longer trigger pod restarts, defeating the feature added in #4063 (originally documented as enabled by default).

## How the regression was introduced

Issue #5079 reported a *cosmetic* problem: when an operator sets `enableReloader: false`, the chart renders an empty `annotations:` block, which causes noise in `helm diff` output.

PR #5082 attempted to wrap the annotations block in an `if` check, but did so in a way that breaks the helper:

\`\`\`yaml
{{- $annotations := include \"splice-util-lib.default-annotations\" . }}    # was: .Values
{{- if $annotations }}
  annotations:
    {{- $annotations .Values | nindent 4 }}                                # was: include ... .Values
{{- end }}
\`\`\`

Two issues:

1. **Wrong context passed to helper.** The helper at `cluster/helm/splice-util-lib/templates/_helpers.tpl` reads:

   \`\`\`go-template
   {{- if (.enableReloader | default false) -}}
   reloader.stakater.com/auto: \"true\"
   {{- end -}}
   \`\`\`

   It expects \`.enableReloader\` to resolve directly. The previous call site passed \`.Values\` (which has \`.enableReloader\` at top level). The new call site passes \`.\` (full chart context) — \`.enableReloader\` is undefined there, the default \`false\` kicks in, and the helper returns an empty string. The \`if $annotations\` check is then always false, so the whole annotations block is dropped.

2. **Invalid template syntax.** \`{{- $annotations .Values | nindent 4 }}\` treats the captured string \`$annotations\` as a callable (Go template function-call syntax). Strings are not callable in Go templates — this is effectively a no-op for the \`.Values\` argument and only outputs the (empty) string.

Net effect: the annotation is silently absent for *all* users on 0.6.1, regardless of \`enableReloader\` setting.

## Reproduction

\`\`\`bash
helm pull oci://ghcr.io/digital-asset/decentralized-canton-sync/helm/splice-validator --version 0.6.1
tar xzf splice-validator-0.6.1.tgz
helm template foo splice-validator/ --set migration.id=1 \\
  | grep -B2 -A6 \"kind: Deployment\" | grep -A1 \"metadata:\"
# Expected: annotations: { reloader.stakater.com/auto: \"true\" } on each Deployment
# Actual:   annotations block absent
\`\`\`

Compare with 0.5.18 — the annotation is rendered as expected.

## Affected charts (all from PR #5082)

- \`splice-participant/templates/participant.yaml\`
- \`splice-validator/templates/validator.yaml\`
- \`splice-validator/templates/ans-web-ui.yaml\`
- \`splice-validator/templates/wallet-web-ui.yaml\`
- \`splice-sv-node/templates/sv.yaml\`
- \`splice-sv-node/templates/sv-web-ui.yaml\`
- \`splice-scan/templates/scan.yaml\` (two Deployments)
- \`splice-domain/templates/domain.yaml\`
- \`splice-global-domain/templates/mediator.yaml\`
- \`splice-info/templates/deployment.yaml\`
- \`splice-load-tester/templates/load-tester.yaml\`
- \`splice-party-allocator/templates/party-allocator.yaml\`
- \`splice-splitwell-app/templates/splitwell.yaml\`
- \`cn-docs/templates/docs.yaml\`

## Suggested fix

Keep the \`if\` wrap (so #5079 stays fixed) but pass \`.Values\` to the helper and emit the captured string directly:

\`\`\`yaml
{{- $annotations := include \"splice-util-lib.default-annotations\" .Values }}
{{- if $annotations }}
  annotations:
    {{- $annotations | nindent 4 }}
{{- end }}
\`\`\`

This both restores the 0.5.18 behaviour for the default \`enableReloader: true\` case and continues to suppress the empty block when \`enableReloader: false\`.

PR follows.

SLoeuillet:
## Summary

PR #5082 (shipped in **Splice 0.6.1**) wrapped the annotations block in an `if` check to fix the cosmetic issue reported in #5079 (empty `annotations:` block when `enableReloader: false`). The fix passed the wrong context to the helper and used invalid Go template syntax to emit the captured value, with the result that the `reloader.stakater.com/auto: "true"` annotation is **silently dropped from every affected Deployment**, regardless of the `enableReloader` value. This breaks the Stakater Reloader integration introduced as a default in #4063 for all 0.6.1 users.

Detailed root-cause analysis is in the linked issue: #5282.

The two bugs in the prior form:

1. `include "splice-util-lib.default-annotations" .` passes the full chart context. The helper reads `.enableReloader` directly, but that key lives at `.Values.enableReloader`, so the helper always returns empty and the `if $annotations` guard always fails.
2. `{{- $annotations .Values | nindent 4 }}` treats the captured string as a callable. Strings are not callable in Go templates; the `.Values` argument is silently discarded.

This PR:
- Restores the `.Values` context to the helper call (matches 0.5.18 behaviour and the helper's expectation).
- Emits the captured string directly (`{{- $annotations | nindent N }}`).
- Preserves the `if $annotations` wrap so the empty-block fix from #5079 still holds when `enableReloader: false`.

The change is applied identically across the 14 chart templates that PR #5082 touched.

## Test plan

- [x] Adds a `helm-unittest` case in `splice-validator/tests/validator_test.yaml` covering both branches:
  - default settings (`enableReloader: true`) → `metadata.annotations["reloader.stakater.com/auto"]` equals `"true"`
  - `enableReloader: false` → `metadata.annotations` does not exist (no empty block)
- [x] Verified locally on a real cluster: applying this fix to a 0.6.1 deployment of `splice-validator` causes the reloader annotation to reappear on `validator-app`, `ans-web-ui`, `wallet-web-ui`, and `participant` Deployments.

Closes #5282.
Refs #5079, #5082, #4063.

### DIFF
--- a/cluster/helm/cn-docs/templates/docs.yaml
+++ b/cluster/helm/cn-docs/templates/docs.yaml
@@ -7,10 +7,10 @@ kind: Deployment
 metadata:
   name: docs
   namespace: {{ .Release.Namespace }}
-  {{- $annotations := include "splice-util-lib.default-annotations" . }}
+  {{- $annotations := include "splice-util-lib.default-annotations" .Values }}
   {{- if $annotations }}
   annotations:
-    {{- $annotations .Values | nindent 4 }}
+    {{- $annotations | nindent 4 }}
   {{- end }}
   labels:
     {{- include "splice-util-lib.default-labels" (set . "app" "docs") | nindent 4 }}
--- a/cluster/helm/splice-domain/templates/domain.yaml
+++ b/cluster/helm/splice-domain/templates/domain.yaml
@@ -4,10 +4,10 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  {{- $annotations := include "splice-util-lib.default-annotations" . }}
+  {{- $annotations := include "splice-util-lib.default-annotations" .Values }}
   {{- if $annotations }}
   annotations:
-    {{- $annotations .Values | nindent 4 }}
+    {{- $annotations | nindent 4 }}
   {{- end }}
   labels:
     {{- include "splice-util-lib.default-labels" (set . "app" .Release.Name) | nindent 4 }}
--- a/cluster/helm/splice-global-domain/templates/mediator.yaml
+++ b/cluster/helm/splice-global-domain/templates/mediator.yaml
@@ -6,10 +6,10 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  {{- $annotations := include "splice-util-lib.default-annotations" . }}
+  {{- $annotations := include "splice-util-lib.default-annotations" .Values }}
   {{- if $annotations }}
   annotations:
-    {{- $annotations .Values | nindent 4 }}
+    {{- $annotations | nindent 4 }}
   {{- end }}
   labels:
     {{- include "splice-util-lib.default-labels" (set . "app" $mediatorLabel) | nindent 4 }}
--- a/cluster/helm/splice-info/templates/deployment.yaml
+++ b/cluster/helm/splice-info/templates/deployment.yaml
@@ -6,10 +6,10 @@ kind: Deployment
 metadata:
   name: {{ .Release.Name }}
   namespace: {{ .Release.Namespace }}
-  {{- $annotations := include "splice-util-lib.default-annotations" . }}
+  {{- $annotations := include "splice-util-lib.default-annotations" .Values }}
   {{- if $annotations }}
   annotations:
-    {{- $annotations .Values | nindent 4 }}
+    {{- $annotations | nindent 4 }}
   {{- end }}
 spec:
   replicas: 1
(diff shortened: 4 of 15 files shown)


krzysztofczyz-da:
Hi,

Please fix the [DCO](https://github.com/hyperledger-labs/splice/pull/5283/checks?check_run_id=73352290136), we will approve the CI run 😄 

In general LGTM

SLoeuillet:
@krzysztofczyz-da sorry, didn't check the comments since this morning. Fixed

krzysztofczyz-da (approved):
LGTM 🎉

PR #5293, opened 2026-04-28
Add scan http encodings encode() to read perf test

jagathweerasinghe-da:
- We already have the read performance test for UpdateHistory.getUpdate()
- This adds ScanHttpEncodigns.encode() test
- With both UpdateHistory.getUpdate() and ScanHttpEncodigns.encode(), we get a good understanding of the latency of /v2/updates/{update_id}

Grafana:

Screenshot 2026-04-28 at 16 37 33


OriolMunoz-da (approved):
lgtm, thanks!

PR #5294, opened 2026-04-28
Update runbook for roll forward LSU

moritzkiefer-da:
[static]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/docs/src/sv_operator/sv_logical_synchronizer_upgrade.rst
+++ b/docs/src/sv_operator/sv_logical_synchronizer_upgrade.rst
@@ -69,21 +69,6 @@ Disaster Recovery through Roll-Forward LSU
 In case of a disaster that causes the current physical synchronizer to become unavailable, an LSU can be used as a roll-forward recovery mechanism.
 The procedure is similar to a regular LSU but because the current physical synchronizer is unusable, the coordination through a vote and topology transactions is not possible and instead validators and super validators need to manually initiate the upgrade.
 
-Concretely, the procedure is as follows:
-
-1. The old physical synchronizer is deemed broken and the last sequenced message was at record time R.
-2. Super validators deploy successor nodes. Depending on the issue, the successor nodes may be configured with older image and protocol versions if the issue is limited to the new version.
-3. Super validators configure their SV app app to transfer the topology and traffic state from the old physical synchronizer to the successor nodes.
-
-   .. todo:: add more details once this is implemented
-
-4. Super Validators configure their scan to indicate to validators that they should roll forward to a new synchronizer.
-
-5. Validator app automation picks up that configuration and initiates a manual roll-forward LSU to the new synchronizer.
-
-Note that depending on how exactly the old synchronizer failed, validators may desynchronize if some validators have observed a transaction before the failure while others have not.
-In that case, the participant will produce ACS mismatches that should be resolved using the `standard ACS mismatch resolution process <https://docs.digitalasset.com/operate/3.4/howtos/troubleshoot/commitments.html>`_ after migrating to the new physical synchronizer.
-
 This procedure can also be used for recovering from a failed LSU. There are two relevant cases:
 
 1. The LSU did not get cancelled before the upgrade time but no Daml transactions and topology transactions were able to be sequenced on the successor physical synchronizer after the upgrade time. In this case, the original successor synchronizer can be thrown away and replaced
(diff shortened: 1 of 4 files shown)


docs/src/sv_operator/sv_logical_synchronizer_upgrade.rst
```diff
@@ -69,29 +69,89 @@ Disaster Recovery through Roll-Forward LSU
...
+     newPhysicalSynchronizerSerial: NEW_PHYSICAL_SYNCHRONIZER_SERIAL # Must be agreed between SVs, usually existing (broken) synchronizer serial + 1
+     newPhysicalSynchronizerProtocolVersion: NEW_PHYSICAL_SYNCHRONIZER_PROTOCOL_VERSION # Must be agreed between SVs, usually existing (broken) synchronizer serial + 1
+     exportTimes:
+       topologyExportTime: TOPOLOGY_EXPORT_TIME # Must be agreed between SVs
+       trafficExportTime: TRAFFIC_EXPORT_TIME # Must be agreed between SVs
+       upgradeTime: UPGRADE_TIME # Must be agreed between SVs
+
+5. Validators initiate the :ref:`procedure <validator_roll_forward_lsu>` on their side.
+
+Recovery from a failed LSU where nothing got sequenced
```
moritzkiefer-da, line 128: I know we had that discussion several times beforehand. But I'm once again wondering if automating this through scan is really a good idea. The benefit is clear, if we do hit it and can recover that way it's nicer for validators as they have to do nothing and there is much less downtime. But the downsides also seem real (in no particular order):

1. It's an extra procedure that needs documenting, testing, …. And it only works in a relatively niche case where really nothing got sequenced other than test messages on the new synchronizer.
2. We're adding the option for SVs to trigger a relatively unsafe repair procedure automatically on all nodes. This seems like it's just gonna ask for people to raise concerns and push around blame. At least if they're the ones initiating it and it goes wrong they explicitly did something.

nicu-da: I'm not strictly opposed to no doing it this way, but the main point for me is failing to see the downside of exposing it just for the case where the new sync can be thrown away. 

> It's an extra procedure that needs documenting, testing, …. And it only works in a relatively niche case where really nothing got sequenced other than test messages on the new synchronizer.

How is it different from documenting the changes required to do a rollforward through the validator app config without specifying the config? 

> We're adding the option for SVs to trigger a relatively unsafe repair procedure automatically on all nodes. This seems like it's just gonna ask for people to raise concerns and push around blame. At least if they're the ones initiating it and it goes wrong they explicitly did something.

But this will require that the announcement is in the topology state, the upgrade time was reached, and the participant reported psid is still the old one. I think the difference is that I don't see as a repair procedure but as a regular LSU, as it will be exactly the same thing in the participant just that it's triggered manually. 
We could in theory have this disabled by default.

moritzkiefer-da: > How is it different from documenting the changes required to do a rollforward through the validator app config without specifying the config?

You don't get around specifying this since you do need the manual option anyway if something did get sequenced. So it just seems like an extra option.

>  think the difference is that I don't see as a repair procedure but as a regular LSU

fair given that we have the code and docs already now let's just keep it.

docs/src/validator_operator/validator_disaster_recovery.rst
```diff
@@ -456,3 +461,44 @@ for initial onboarding of the party, i.e.,
...
+
+.. code::
+
+    val existingPhysicalSynchronizerId = participant.synchronizers.list_connected().find(_.synchronizerAlias == "global").head.physicalSynchronizerId
+    participant.synchronizers.perform_manual_lsu(
+      existingPhysicalSynchronizerId,
+      newPhysicalSynchronizerId,
+      upgradeTime = None,
+      sequencerSuccessors,
+    )
```
nicu-da, line 491: any reason not to have this configurable in the validator app in the future?

nicu-da (approved):
thanks!

docs/src/validator_operator/validator_disaster_recovery.rst
moritzkiefer-da: If we want that (for ease of operations over console access I would assume), I think I'd rather push for getting a config option in the participant for that. I don't really see why the validator app should have to do anything here.

PR #5295, opened 2026-04-28
Ignore Gave up waiting for .* as we are shutting down

OriolMunoz-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/8193


pawelperek-da (approved):
LGTM

PR #5296, opened 2026-04-28
Public Sequencer: Add scan endpoint to lookup if a given participant id is permissioned 

pasindutennage-da:
Fix https://github.com/hyperledger-labs/splice/issues/5286

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
@@ -141,6 +141,15 @@ class PermissionedSynchronizerIntegrationTest
 
     }
 
+    clue("Scan returns the ParticipantSynchronizerPermission") {
+      eventually() {
+        sv1ScanBackend.getParticipantSynchronizerPermission(
+          decentralizedSynchronizerId.toProtoPrimitive,
+          aliceValidatorBackend.participantClient.id.toProtoPrimitive,
+        ) shouldBe true
+      }
+    }
+
     actAndCheck(
       "Start Alice validator", {
         aliceValidatorBackend.start()
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
@@ -9,6 +9,7 @@ import org.lfdecentralizedtrust.splice.codegen.java.splice.validatorlicensereque
 import org.lfdecentralizedtrust.splice.config.ConfigTransforms
 import org.lfdecentralizedtrust.splice.integration.EnvironmentDefinition
 import org.lfdecentralizedtrust.splice.integration.tests.SpliceTests.IntegrationTest
+import org.lfdecentralizedtrust.splice.scan.admin.api.client.commands.HttpScanAppClient.SynchronizerPermissionState
 import org.lfdecentralizedtrust.splice.sv.config.SvOnboardingConfig
 import org.lfdecentralizedtrust.splice.sv.util.{SvOnboardingToken, SvUtil}
 import org.lfdecentralizedtrust.splice.util.{ProcessTestUtil, SvTestUtil, WalletTestUtil}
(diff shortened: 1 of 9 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
```diff
@@ -141,6 +141,15 @@ class PermissionedSynchronizerIntegrationTest
 
     }
 
+    clue("Scan returns the ParticipantSynchronizerPermission") {
```
moritzkiefer-da: can we check that it failed beforehand?

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -191,6 +191,42 @@ paths:
...
 
+  /v0/domains/{domain_id}/participants/{participant_id}/synchronizer-permission:
+    get:
+      tags: [external, scan]
+      x-jvm-package: scan
+      operationId: "getParticipantSynchronizerPermission"
+      description: |
+        Check if the ParticipantSynchronizerPermission topology transaction is confirmed.
+      parameters:
+        - name: "domain_id"
```
moritzkiefer-da:
```suggestion
        - name: "synchronizer_id"
```

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -4141,6 +4177,14 @@ components:
             synchronizer in bytes.
           type: integer
           format: int64
+    GetParticipantSynchronizerPermissionResponse:
+      type: object
+      required: ["is_permissioned"]
```
moritzkiefer-da: This seems a bit misleading. You only check if it exists but for example if loginAfter is set you still return it is permissioned. Instead of just returning a boolean I would return an (optional) subset of what is in `ParticipantSynchronizerPermission`. Probably only the actual login_after is relevant

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
```diff
@@ -2537,6 +2539,46 @@ class HttpScanHandler(
...
+            )
+        }
+
+        permissions <- participantAdminConnection.listParticipantSynchronizerPermission(
+          domain,
+          participant.filterString,
+        )
+      } yield {
+        val isPermissioned = permissions.exists { result =>
+          result.mapping.permission == Submission
```
moritzkiefer-da: Isn't this field deprecated and not used? Not sure why we would check it here, it doesn't define whether you're permissioned to connect to the synchronizer.

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
```diff
@@ -573,6 +574,7 @@ class ValidatorApp(
               for {
                 dsoInfo <- scanConnection.getDsoInfo()
                 sponsorPartyId = dsoInfo.svPartyId
+                _ <- waitForTopologyPermission(scanConnection, synchronizerId, participantId)
```
moritzkiefer-da, line 578: Do we hit this codepath in the test? It seems like you want to change the test start the validator app and then permission it and then wait for it to initialize.

moritzkiefer-da (approved):
thanks

apps/scan/src/main/openapi/scan.yaml
pasindutennage-da: Fixed

apps/scan/src/main/openapi/scan.yaml
pasindutennage-da: Fixed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
pasindutennage-da: Fixed

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/ValidatorApp.scala
pasindutennage-da: Fixed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
pasindutennage-da: Fixed

PR #5306, opened 2026-04-29
Prevent DamlCIUpgradeVotePreflightTest from sending empty votes

OriolMunoz-da:
mostly whiteline changes, you might wanna ignore when reviewing

Fixes https://github.com/DACH-NY/cn-test-failures/issues/8197

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/upgrades/DamlCIUpgradeVotePreflightTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/upgrades/DamlCIUpgradeVotePreflightTest.scala
@@ -1,5 +1,6 @@
 package org.lfdecentralizedtrust.splice.integration.tests.upgrades
 
+import org.lfdecentralizedtrust.splice.codegen.java.splice.amuletconfig.PackageConfig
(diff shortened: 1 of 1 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/upgrades/DamlCIUpgradeVotePreflightTest.scala
```diff
@@ -38,152 +38,168 @@ class DamlCIUpgradeVotePreflightTest
...
+        validatorLifecycle_current.toString,
+        wallet_current.toString,
+        walletPayments_current.toString,
+      )
+
+      val sv1ScanClient = scancl("sv1Scan")
+      val currentPackageConfig =
+        sv1ScanClient.getAmuletRules().payload.configSchedule.initialValue.packageConfig
+
+      if (targetPackageConfig == currentPackageConfig) {
```
OriolMunoz-da, line 61: we could also instead change the script in internal that does the diff

personally I prefer to check here, because
1) does not involve bash scripting
2) if we ever add a new field to PackageConfig, the compiler will complain here but not in the bash script

pawelperek-da (approved):
nice condition

PR #5225, opened 2026-04-24
 replace the usage of svs.default with !include tags to get better merging semantics

mblaze-da:
I'll remove `svs.default` implementation in a separate PR. I should probably backport but I'm pretty sure the automated one will fail on conflicts.

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a logical synchronizer upgrade test is required (from canton-3.5), comment `/lsu_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #5316, opened 2026-04-29
remove svs.default

mblaze-da:
After https://github.com/hyperledger-labs/splice/pull/5225 and https://github.com/DACH-NY/canton-network-internal/pull/4761 got merged we can finally get rid of the dedicated `svs.default` merging logic.


martinflorian-da (approved):
deletes code so must be good; thanks!

Issue #5141, opened 2026-04-21
Test DA base images


isegall-da:
https://github.com/DACH-NY/canton/pull/30782

isegall-da:
https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/62176/workflows/c784e450-c05c-4bbb-9ca9-1bf386d8dec2

isegall-da:
Failed, need to mount a writable /tmp directory.

another test run: https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/62235/workflows/38df3b4f-76b2-4200-a5f8-efc2aa3d608a

isegall-da:
https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/62239/workflows/dd1d1e76-3fca-4e4e-9e2c-f37696772606

PR #5325, opened 2026-04-30
Change grpcurl to grpc_health_probe

krzysztofczyz-da:


cluster/images/canton-participant/Dockerfile
```diff
@@ -5,10 +5,15 @@ ARG canton_version
...
 
 COPY additional-config.conf /app/
 COPY target/logback.xml /app/
 COPY R__repeatable-migration.sql /app/repeatable-migrations/
 USER root
+RUN rm -f /bin/grpcurl
 RUN chown -R nonroot:nonroot /app/additional-config.conf /app/repeatable-migrations/R__repeatable-migration.sql
 USER nonroot
+
+HEALTHCHECK --start-period=10m \
```
isegall-da, line 20: I don't remember, is this simply ignored when running in k8s? 
This can also be defined in the docker-compose file, right? Any reason to prefer one over the other?

cluster/images/canton-mediator/Dockerfile
```diff
@@ -5,9 +5,11 @@ ARG canton_version
 ARG image_sha256
 
 FROM europe-docker.pkg.dev/da-images/public-all/docker/canton-mediator:${canton_version}@${image_sha256}
+COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.38 /ko-app/grpc-health-probe /bin/grpc_health_probe
```
isegall-da, line 9: Do we actually use it? 
(I don't mind too much adding it, we'll just have it in the base anyway, but mostly trying to understand)

cluster/images/canton-participant/Dockerfile
```diff
@@ -5,10 +5,15 @@ ARG canton_version
 ARG image_sha256
 
 FROM europe-docker.pkg.dev/da-images/public-all/docker/canton-participant:${canton_version}@${image_sha256}
+COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.38 /ko-app/grpc-health-probe /bin/grpc_health_probe
```
isegall-da, line 9: Can you add a TODO(https://github.com/hyperledger-labs/splice/issues/5141) in all places where you add this to remove it once we use base images that have it in them?

cluster/images/canton-participant/Dockerfile
krzysztofczyz-da: Yes it's ignored in k8s.
I'm defining HC here, because the base image defines one as well, so an override seems fitting 😄

cluster/images/canton-mediator/Dockerfile
krzysztofczyz-da: I'm just adding it wherever I'm removing it 😄

cluster/images/canton-participant/Dockerfile
krzysztofczyz-da: Added to both added lines to make it clearer

moritzkiefer-da (commented):
Thanks! Changes lgtm but I would suggest we just merge the Canton side first and then only change the localnet file. No point in overwriting this on our side.

krzysztofczyz-da:
This won't be needed anymore -> relevant changes added to #5375

PR #5349, opened 2026-05-04
Add support for configMap as extraVolumes for splice-participant's helm chart

jstoja:
Hello,

This PR is adding support for configMap as extraVolumes for splice-participant's helm chart.
This can be used to mount a bootstrap script for example.

Best,
J


martinflorian-da:
Thanks @jstoja and sorry that this got buried for so long.

What is your use case for this change beyond boostrap scripts?

For bootstrap scripts, we currently document the following approach (docs will be moved to new docs site soon): https://docs.dev.sync.global/deployment/configuration.html#custom-bootstrap-scripts

jstoja:
Hello Martin,

I didn't know about this way of overriding it. Makes sense now.

Thanks for highlighting this feature.

Best,
J

jstoja:
Hello @martinflorian-da,

While implementing TLS on the ledger-api and admin-api, we want to mount a CA certificate.
I know your setup calls for using istio but that's not something we use.

Would you find this use case viable to enable mounting config-maps?

Best,
Julien

martinflorian-da:
> While implementing TLS on the ledger-api and admin-api, we want to mount a CA certificate.

Note that this is not well-tested in the splice context (i.e., we can't offer support) and that exposing the admin-api outside of a trusted private network is strongly discouraged for security reasons.

> Would you find this use case viable to enable mounting config-maps?

In principle, yes.

If you want to move forward with this PR, please also add a short Helm test so we have *some* regression testing (I don't think we want to add this for our more elaborate cluster tests soon): https://github.com/canton-network/splice/blob/b8b743f89867fbb911ee753257ec9aee526b90be/TESTING.md?plain=1#L450

isegall-da:
@jstoja do you still find this PR valuable? Please add the requested test if so, or close it if it's stale.

PR #5350, opened 2026-05-04
Use ADDITIONAL_CONFIG_PERSISTENCE to override storage configuration i…

jstoja:
Hello,

This PR adds `ADDITIONAL_CONFIG_PERSISTENCE` to override storage configuration in `splice-partitipant`'s helm chart.
This is using the same environment variable and templating as in `splice-validator` and allows the override of the storage `user` field.

Best,
J


martinflorian-da (changes requested):
Thanks!

There are bugs in this as is; how did you validate your changes? I suggest having a look at https://github.com/canton-network/splice/blob/main/TESTING.md#helm-checks and adding a test or two.

I also see that the `mediator.yaml` chart might also benefit from a similar improvement. While we are at it... that might also just be a nice thing to factor into a shared helper in `splice-util-lib`.

PR #5355, opened 2026-05-04
Update tstd docs with 0.6.3-snapshot.20260504.2741.0.v681c5054

OriolMunoz-da:
[static]

Cluster is deploying

Release bundle already updated: https://github.com/digital-asset/decentralized-canton-sync/releases/tag/token-standard-v2-upcoming

regarding the contract-mode removal: 
image

### DIFF
--- a/token-standard/TOKEN_STANDARD_V2_DEVNET.md
+++ b/token-standard/TOKEN_STANDARD_V2_DEVNET.md
@@ -1,6 +1,6 @@
 # Token Standard V2 DevNet
 
-Splice version: `0.6.1-snapshot.20260427.2678.0.v86ff9ad5`
+Splice version: `0.6.3-snapshot.20260504.2741.0.v681c5054`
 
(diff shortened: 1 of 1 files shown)


token-standard/TOKEN_STANDARD_V2_DEVNET.md
```diff
@@ -41,5 +41,4 @@ using the [guide for DevNet](https://hyperledger-labs.github.io/splice/validator
        canton.parameters.non-standard-config=true
        canton.participants.participant.parameters.alpha-version-support=true
        canton.participants.participant.parameters.initial-protocol-version=35
-       canton.participants.participant.parameters.engine.contract-state-mode=NUCK
```
meiersi-da, line 44: Can you add a note that this was removed? I suspect other's will have this override configured, and will be surprised by the problems.

meiersi-da (approved):
Thanks.

token-standard/TOKEN_STANDARD_V2_DEVNET.md
OriolMunoz-da: as discussed, will include this in the slack & forum posts

PR #5380, opened 2026-05-05
improve stack reference mocking in expected files generator to account for participant migration related references

mblaze-da:
This change was needed to make expected files generation for prod clusters happy. The reason is that participant migration introduces a dependency from the `sv` project to `sv-canton` in form of `participantDatabaseId` and `participantDatabaseSecretName` stack outputs. They need to be properly mocked when generating expected files during the migration.

### DIFF
--- a/cluster/pulumi/common/src/dump-config-common.ts
+++ b/cluster/pulumi/common/src/dump-config-common.ts
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 import * as pulumi from '@pulumi/pulumi';
 import * as path from 'path';
-import * as sinon from 'sinon';
 import { setMocks } from '@pulumi/pulumi/runtime/mocks';
 
 import {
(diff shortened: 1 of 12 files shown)


mblaze-da:
# [backport]
- [x] release-line-0.6.2
- [x] release-line-0.5.18

cluster/pulumi/common/src/dump-config-common.ts
```diff
@@ -392,3 +382,25 @@ export async function initDumpConfig(): Promise<void> {
...
+
+export type StackOutputsProvider = (
+  project: string,
+  stack: string
+) => Partial<Record<string, any>> | undefined;
+
+export const infraStackOutputsProvider: StackOutputsProvider = (project: string) => {
+  return project === 'infra'
+    ? {
+        istioDashboardVersions: '1234',
```
nicu-da, line 398: do we no longer use the other outputs? the ingress ones

nicu-da (approved):

cluster/pulumi/common/src/dump-config-common.ts
mblaze-da: I searched for uses and did not find any.

PR #5381, opened 2026-05-05
exclude sv-runbook sv from the remaining voting scripts

mblaze-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/8254
analogous to https://github.com/canton-network/splice/pull/5317

### DIFF
--- a/cluster/scripts/vote-for-lsu.sh
+++ b/cluster/scripts/vote-for-lsu.sh
@@ -110,7 +110,7 @@ for ((i=2; i<=DSO_SIZE; i++)); do
 done
 
 # extra SVs from all the config.yaml files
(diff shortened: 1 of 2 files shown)


nicu-da (approved):
wfm! 
though why is the runbook not generating a proper subdomain? 
Compared to the linked issue, here it would make sense for the runbook to vote as well.

mblaze-da:
> why is the runbook not generating a proper subdomain?

Sadly, I don't know an answer to that question of the top of my head but I don't believe it ever did.

> Compared to the linked issue, here it would make sense for the runbook to vote as well.

Since `sv` didn't have an explicit `svs` entry before I don't believe it was ever included in the vote. If it makes sense to include it we need to figure out the subdomain for it.

PR #5407, opened 2026-05-06
Use fsgroup to grant nonroot permissions to volumes

rautenrieth-da:
Like https://github.com/canton-network/splice/pull/2696, but for the sequencer and mediator.

Noticed because the `/persistent-data` mount for debug data is not writeable by the java process.

[cluster test](https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/63934/workflows/588b13a8-9cad-419d-b5e1-bca5e80c255a)


ray-roestenburg-da (approved):
LGTM

PR #5421, opened 2026-05-06
Reintroduce infinite scroll in Vote History

### LINKED ISSUE #5335: Reintroduce infinite scroll
Reintroduce [infinite scroll](https://github.com/hyperledger-labs/splice/pull/4871) without database performance regression. The issue described by Moritz:

```
So what is happening is:

1. We switched the sort order from effective at to event_number.
2. Even if you don't set after, in some cases postgres now decides to use the primary key index that only includes event_number instead of the vote type. It depends on table stats and whatever but we hit that relatively frequently on CILR.
3. There is no index that includes both the vote type and event_number so if you set after even in the best case where you hit the index for the vote type, you end up doing a linear scan over the event_number until you get to after.
4. If you hit the wrong index, those queries take something like > 45min based on what we've seen on CILR and the SV UI is completely unusable as you can't see or create any votes.
```

pawelperek-da:
Fixes #5335 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V067__scan_txlog_vote_index.sql
+++ b/apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V067__scan_txlog_vote_index.sql
@@ -1,0 +1,7 @@
+-- Supports listVoteRequestResultsQuery: ORDER BY entry_number DESC LIMIT N
+-- with optional cursor (entry_number < $after). Without this index, the
+-- planner picks an Index Scan Backward over the primary key on entry_number,
+-- filtering out non-vote rows one by one
+create index scan_txlog_store_sid_en_vot
+    on scan_txlog_store (store_id, entry_number desc)
+    where entry_type = 'vot';
--- a/apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V067__scan_txlog_vote_index.sql
+++ b/apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V067__scan_txlog_vote_index.sql
@@ -1,7 +1,0 @@
--- Supports listVoteRequestResultsQuery: ORDER BY entry_number DESC LIMIT N
--- with optional cursor (entry_number < $after). Without this index, the
--- planner picks an Index Scan Backward over the primary key on entry_number,
--- filtering out non-vote rows one by one
-create index scan_txlog_store_sid_en_vot
-    on scan_txlog_store (store_id, entry_number desc)
-    where entry_type = 'vot';
--- a/apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/ScanAppReference.scala
+++ b/apps/app/src/main/scala/org/lfdecentralizedtrust/splice/console/ScanAppReference.scala
@@ -828,7 +828,8 @@ abstract class ScanAppReference(
       effectiveFrom: Option[String],
       effectiveTo: Option[String],
       limit: BigInt,
-  ): Seq[DsoRules_CloseVoteRequestResult] = {
+      pageToken: Option[BigInt] = None,
+  ): (Seq[DsoRules_CloseVoteRequestResult], Option[BigInt]) = {
     consoleEnvironment.run {
       httpCommand(
         HttpScanAppClient.ListVoteRequestResults(
@@ -838,6 +839,7 @@ abstract class ScanAppReference(
           effectiveFrom,
           effectiveTo,
           limit,
+          pageToken,
         )
       )
     }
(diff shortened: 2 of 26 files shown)


pawelperek-da:
`WHERE store_id = <id> AND entry_type = 'vot' ORDER BY entry_number DESC LIMIT 20`

With index:
```                                                                                                                         
  Limit (cost=0.43..16.48 rows=20) (actual time=0.021..0.028 rows=20)                                                                              
    Buffers: shared hit=7                                                                          
    ->  Index Scan using scan_txlog_store_sid_en_vot                                                                                               
          Index Cond: (store_id = <id>)                                                                                                               
  Execution Time: 0.061 ms
 ```
 
 Without index:
 ```
  Limit (cost=0.57..1016.17 rows=20) (actual time=444896.993..444897.012 rows=20)                                                                  
    Buffers: shared hit=824632 read=6616311 written=658                                                                                            
    ->  Index Scan Backward using scan_txlog_store_pkey                                                                                            
          Filter: ((store_id = <id>) AND (entry_type = 'vot'::text))                                                                                  
          Rows Removed by Filter: 217500000                                                                                                        
  Execution Time: 444897.026 ms
  ```

**Comment by pawelperek-da** [comment]
`WHERE store_id = <id> AND entry_type = 'vot' AND entry_number < 1000000 ORDER BY entry_number DESC LIMIT 20`

With index:
```
Limit (cost=0.43..17.77 rows=20) (actual time=0.023..0.045 rows=20)                                                                              
    Buffers: shared hit=10                                                                         
    ->  Index Scan using scan_txlog_store_sid_en_vot                                                                                               
          Index Cond: ((store_id = <id>) AND (entry_number < 1000000))
  Execution Time: 0.051 ms 
```

Without index:
```
Limit (cost=0.57..1039.85 rows=20) (actual time=0.909..1.243 rows=20)                                                                            
    Buffers: shared hit=2 read=9                                                                                                                   
    ->  Index Scan Backward using scan_txlog_store_pkey                                                                                            
          Index Cond: (entry_number < 1000000)         
          Filter: ((store_id = <id>) AND (entry_type = 'vot'::text))                                                                                  
  Execution Time: 1.249 ms
```

File: apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V067__scan_txlog_vote_index.sql [hunk 1]
```diff
@@ -0,0 +1,7 @@
+-- Supports listVoteRequestResultsQuery: ORDER BY entry_number DESC LIMIT N
+-- with optional cursor (entry_number < $after). Without this index, the
+-- planner picks an Index Scan Backward over the primary key on entry_number,
+-- filtering out non-vote rows one by one
+create index scan_txlog_store_sid_en_vot
```
pawelperek-da: I'd love someone more experienced in DBs to take a look at this migration

ray-roestenburg-da: Can you share the explain plan for this (is the above comment from doing an explain plan on a relatively large db?) My only concern is the time it takes to build the index for existing records. If that takes too long we'll need to do it outside of a flyway migration.

rautenrieth-da: There is sadly no "explain create index". If I'm not sure if index creation is fast enough, what I usually do is run 

```sql
BEGIN;
create index ... ;
ROLLBACK;
```

on CILR (don't run it if you already know it will be too slow, as it does add load to a database that is being used). 

In this case, it took >5min, so we definitely should add it to `SqlIndexInitializationTrigger`.

ray-roestenburg-da: yes I checked sizes on prod, this has to move to `SqlIndexInitializationTrigger`. I meant an explain plan on the query, referring to the comment above the create index 
```
-- ... Without this index, the
-- planner picks an Index Scan Backward over the primary key on entry_number,
-- filtering out non-vote rows one by one"
```

ray-roestenburg-da: Thanks for the tips!

ray-roestenburg-da (changes requested):
Requires index to be added in SqlIndexInitializationTrigger

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V067__scan_txlog_vote_index.sql
rautenrieth-da: Ah of course, sorry I misread. What I do in that case is 

```sql
BEGIN;

-- create index
create index ... ;

-- check if query plan looks sensible
explain select ... ;

-- measure actual performance of query 
explain (analyze, buffers) select ... ;

-- undo index creation
ROLLBACK;
```

The annoying part is that if index creation takes too long, you need to run this from `cncluster psql` or similar.

ray-roestenburg-da (approved):
Thanks

PR #5423, opened 2026-05-06
Log and skip duplicate verdicts

rautenrieth-da:
As a workaround for https://github.com/DACH-NY/cn-test-failures/issues/8281#issuecomment-4392978743


meiersi-da (approved):
Thank you!

PR #5424, opened 2026-05-07
helm, splice-info: Add status.json

giner:
This adds `/runtime/status.json` with status for:
- mediator
- scan
- sv

What the result looks like can be viewed by following the link https://info.sv.dev.global.canton.network.sv-nodeops.com/runtime/status.v2.json 

`status.json` is refreshed every 60 seconds.

This change also:
- removes log generation after repeated failures (this can be added later if needed)
- renames container `content-runtime-updater` to shorter `updater`


isegall-da:
Thanks @stas-sbi 
looks good, but I think we need to add tests for it in `DsoPreflightIntegrationTest`.

giner:
@isegall-da, thank you. Added.

giner:
Pushed fixes for `static_test` and `deployment_test`. I'm not sure why docker and docker compose tests fail.

isegall-da (approved):

PR #5444, opened 2026-05-07
Define and create perf test thresholds

jagathweerasinghe-da:
- thresholds are calculated from the past data considering a noise_margin and a safety_margin
- we will keep monitoring the breaches and adjust the thresholds as needed


.github/store-perf-thresholds.json
```diff
@@ -1,21 +1,52 @@
 {
   "SvDsoStoreIngestionPerformanceTest": {
-    "splice_perf_ingestion_total_time_ns": 600000000000
+    "_comment_avg_item_time_ns": "median*(1+ noise_margin + safety_margin) of last 30d. noise_margin=(p95-median)/median, safety_margin=10%. calculated on 2026-May-07",
+    "splice_perf_ingestion_avg_item_time_ns": {
+      "max": 2500000
+    },
+    "_comment_total_time_ns": "hard max of 1h",
+    "splice_perf_ingestion_total_time_ns": {
+      "max": 3600000000000
```
OriolMunoz-da: 1h for 1h of mainnet updates is too permissive unless I'm misunderstanding this
Having a quick look at the dashboards, you could set everything to 5m except UpdateHistory to 15m

jagathweerasinghe-da: The performance regression is expected to be detected from the *_avg_item_time_ns, which accounts for the amount of items. We can of course add more checks later, e.g. p99 max.

The *_total_time_ns depends on the amount of data in 1h, which depends on the network dynamics. We can set a value as you mention, but we could see breaches more often. The reason why we started with 1h, is based on the discussion that '1h of data should not take more than 1h for ingestion'. Hope this clarifies the intension.

OriolMunoz-da: > The reason why we started with 1h, is based on the discussion that '1h of data should not take more than 1h for ingestion'

yeah, what I'm saying is that that's already too high a limit, if we hit that we're already underwater. I'd feel more comfortable, even if that means more breaches, if we set it to 10/20m at most (UpdateHistory definitely would need at least 20m atm)

jagathweerasinghe-da: Updated w/ comment. PTAL

OriolMunoz-da (approved):
ty

PR #5469, opened 2026-05-08
Release notes for 0.6.3

martinflorian-da:
See canton-network/splice#4956

Actual changes only in https://github.com/canton-network/splice/pull/5469/changes/28d1107689e27b36730f5ed2b7f9f7fe31c4f0bd

### DIFF
--- a/docs/src/release_notes.rst
+++ b/docs/src/release_notes.rst
@@ -30,7 +30,7 @@
           Note that the ``app_activity_records`` data already provided for events during this period is correct
           and the network explorers who have ingested this data should keep a copy of it.
 
-        - Added ``/v1/holdings/summary`` endpoint that drops the ``accumulated_holding_fees_unlocked``,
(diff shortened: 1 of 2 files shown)


isegall-da:
Argh, my comment on the commit does not show up here?
See https://github.com/canton-network/splice/commit/28d1107689e27b36730f5ed2b7f9f7fe31c4f0bd#r184714767

isegall-da (approved):
Thanks.
Please note my comment on the change commit, which doesn't show up in the main PR UI apparently...

martinflorian-da:
@isegall-da I see no comment :cry:

isegall-da:
> @isegall-da I see no comment 😢

I hate this......
I wrote on the "required" vs "requires":

I don't love either TBH. It's unclear whether the operator needs to do anything.
I think that "required" suggested that we do it for them in the app, which I think is what we meant, right? "requires" sounds like we're asking them to do something.

So maybe "To support this, the scan app will also clear ..." ?

PR #5472, opened 2026-05-08
Fix format and lint for root cluster/pulumi sources

### LINKED ISSUE #3334: fix formatting & linting for cluster/pulumi package sources
Formatting and linting for source files that are directly under the top level package defined by `cluster/pulumi/package.json` does not work. What is more, when a file there is introduced or changed the `scripts/fix-ts.py` script would pick it up and fail with
```
[error] `cluster/pulumi/package.json` script entry `format:fix` is invalid:
   The command `npm run format:fix --workspaces --if-present` does not include a `--` marker to end the option arguments
```
which is currently worked around by ignoring those sources in `.pre-commit-config.yaml`.

samsondav:
## Summary

Fixes #3334.

The root `cluster/pulumi` `format:fix` / `lint:fix` / `check` scripts only delegated to `--workspaces`, so `scripts/fix-ts.py` couldn't run on root `.ts` files: `npm run format:fix -- <file>` rejected the positional arg. To keep pre-commit working, `.pre-commit-config.yaml` excluded `/pulumi/[^/]+.ts$`, which left the root files uncovered.

This PR:

- Runs `prettier` / `eslint` directly on `cluster/pulumi/*.ts` in the root scripts, then delegates to workspaces with `--if-present`.
- Drops the `/pulumi/[^/]+.ts$` exclude from the pre-commit hook.
- Reformats and lints the existing root `.ts` files now in scope. Mostly import-block ordering; two `as any` casts replaced with typed local shapes; one `.then(success, error)` rewritten as `async` / `try` / `catch` to satisfy `promise/prefer-await-to-then`.

The root `check` script runs prettier and eslint on the root once, then delegates to each workspace's own `check` (which itself runs format / lint / type-check). The naive form `format:check && lint:check && check --workspaces` would have run workspace format/lint twice.

## Test plan

- [x] `python3 scripts/fix-ts.py cluster/pulumi/*.ts` reformats nothing (all files already clean)
- [x] `npm run check` from `cluster/pulumi` passes (root + every workspace)
- [x] `source .envrc.vars; source cluster/deployment/mock/.envrc.vars; make cluster/pulumi/test -j8`


isegall-da:
Thanks @samsondav and apologies for the delay, many people are out this week. I'll try to take a look a bit later today.

isegall-da (approved):
Nice, thank you

Issue #546, opened 2025-06-05
Avoid painful Canton fork bumps


canton-network-da:
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#18789). Original author: @moritzkiefer-da ]

Canton Network started out with a Canton fork. While that got us started quickly, bumping the fork has become increasingly painful especially now that 3.3 unified community and enterprise types which removed all the abstract classes that we used to extend with our splice specific environments. We fixed that by hacking those abstract classes back into the fork but that's a mess.

So far our plan for getting rid of the fork was to wait until we magically turn everything into libraries and can replace the fork. In practice, none of that has happened and we're just stuck with wasting our time on bumping several thousand lines of code every week, resolving hundreds of merge conflicts and hope that nothing important broke. And I think if we look at the problem, it's clear why that's the case: We have an enormous ill-defined API surface so it's not even clear where to start librarifying.

If we look at our actual usage we can split our fork usage in a few different components:

### Isolated Libraries

A few fairly isolated libraries that are also separated libraries in  Canton or in the Daml repo. That includes logging, metrics, error types and similar stuff. We patch essentially none of that and the only reason we don't consume them as proper libraries from the Canton repo already is because they are not published.

### Canton API client code used in our prod code

This includes everything from protobuf files to generated code from that up to the various Command classes that give slightly more high level wrappers around protobuf types. This is primarily about the admin APIs although we currently also consume the Java bindings that way.

This is the code that forces a fork bump when we really just want to bump the binaries as some API changed and we have to consume it or a new API got added that we want to use. We patch this a bit but not aggressively.

### Canton console code used in tests

This is very similar to the API client code but only used in tests. In lots of places we already replaced this by calls to our own connection classes but we still rely on it for direct ledger API submissions and similar. We do patch it a bit but mostly structured it by extending it instead of patching it.

### Canton App framework

This is everything upwards in the dependency chain from canton-community-base. It's hard to describe what this does but I'd roughly say "go from a config file to a set of running components defined in those" (including in in integration tests).
This is the part that is heavily patched. E.g., we have http instead of grpc serves, we obviously have different node types, our shutdown logic works differently, retries don't quite match, …. It is also the part that constantly conflicts. Note that we benefit very little from updating it though. Most of the changes Canton makes either do nothing for us (e.g. their shutdown redesign) or gets in the way (e.g. the unification of community and environment).

## Requirements

So if we look at the above, the requirements seem to be primarily the following:

[...]

PR #5480, opened 2026-05-08
remove unneeded divergence from Canton sources

stephencompall-DA:
For DACH-NY/canton#32877, part of #546, there were aspects of `CantonConfig`/`Environment` usage where the upstreaming would be far nicer, or in some cases could only be upstream-compatible, with a different approach to our own divergence. So those changes are collected here.

- remove `SharedCantonConfig#name`
- move `Environment#createConsole` back where it was
- restore `SharedCantonConfig#mergeDynamicChanges`
  - restore `Environment#pokeOrUpdateConfig`
- some statically empty `Map`s
- override `startupConfigFileMonitoring` to disable instead of just deleting from the fork
- remove unused `environmentFactory` addition, which is accessible elsewhere
- restore `TestEnvironment#actualConfig`'s original implementation
- use `Base*` pattern for some generalized classes with type aliases to avoid many, many downstream renames in Canton test code
- drop unneeded tparam from `Runner`

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceConfig.scala
```diff
@@ -306,6 +301,9 @@ case class SpliceConfig(
     import writers.*
     ConfigWriter[SpliceConfig].to(this).render(SpliceConfig.defaultConfigRenderer)
   }
+
+  override def mergeDynamicChanges(newConfig: SpliceConfig) =
+    this // dynamic changes not supported
```
nicu-da, line 308: should we throw instead?

nicu-da (approved):
Nice, thanks!

apps/app/src/main/scala/org/lfdecentralizedtrust/splice/config/SpliceConfig.scala
stephencompall-DA: No, because no-op is the semantically matching counterpart to what `CantonConfig` does (these are statically empty lists for `SpliceConfig`):

https://github.com/canton-network/splice/blob/bce98c28600e253157098a2418d0488cc0e7e66a/canton/community/app-base/src/main/scala/com/digitalasset/canton/config/CantonConfig.scala#L678-L689

ray-roestenburg-da (approved):

ray-roestenburg-da:
Nice, thanks

PR #5490, opened 2026-05-11
Integration tests: ignore `onboard/validator` timeout warnings

martinflorian-da:
Fixes DACH-NY/cn-test-failures#8111

Calls to this have retries; if the retries were not enough, we'll get a louder error.

[ci]

### DIFF
--- a/project/ignore-patterns/canton_network_test_log.ignore.txt
+++ b/project/ignore-patterns/canton_network_test_log.ignore.txt
@@ -159,3 +159,6 @@ Waiting for.*has not completed after
 Gave up waiting for.*as we are shutting down
 
 Ryuk has been disabled
+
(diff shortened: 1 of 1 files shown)


moritzkiefer-da:
Do we know why we still get those timeouts? I was hoping PV 35 fixes those issues.

moritzkiefer-da (approved):
thx

martinflorian-da:
> Do we know why we still get those timeouts? I was hoping PV 35 fixes those issues.

Tbh I didn't bother digging very deep but based on statistics - they have gotten more frequent rather than less.

If something *has* gotten slow this is the wrong signal to keep though IMHO; and if it's the only thing that "breaks" - who cares?

Note that https://github.com/DACH-NY/cn-test-failures/issues/8335 is a bit similar.

PR #5491, opened 2026-05-11
GCS authentication via wif for reading test data

jagathweerasinghe-da:
-with this history dump can be read even after public access of the bucket is removed
-test run: https://github.com/canton-network/splice/actions/runs/25722355553/job/75526407685


OriolMunoz-da (commented):
I trust this works - but I will defer to @isegall-da to approve this because of the possibility of leaking secrets and he's more familiar with this

isegall-da (approved):
Looks good, thanks

PR #5249, opened 2026-04-27
Rename "Validator faucet cap" UI label to "Validator liveness reward cap"

samsondav:
## Summary

Aligns user-facing terminology with the Canton Coin whitepaper, which uses "liveness rewards" rather than "faucet" for the per-round validator rewards. Fixes #958.

- `apps/sv/frontend`: relabel the two `optValidatorFaucetCap` rows in the SV governance minting curve diff view ("Validator faucet cap" -> "Validator liveness reward cap").

## Scope

Per the original issue author guidance ("No changes to Daml code; just UI for now"), this is intentionally minimal:

- **No Daml changes.** The final diff leaves `splice-amulet` untouched to avoid triggering a Daml package/version change.
- **No on-chain ABI changes.** The `ValidatorLivenessActivityRecord` template already exists alongside the deprecated `ValidatorFaucetCoupon`; the wallet OpenAPI already exposes both endpoint sets. Renaming `FaucetState`, `optValidatorFaucetCap`, `optIssuancePerValidatorFaucetCoupon`, etc. would break on-chain contract compatibility for fields still in active templates and is out of scope.
- **No public API renames.** The Scan OpenAPI endpoints (`/v0/validators/validator-faucets`, `/v0/top-validators-by-validator-faucets`) are unchanged here. Adding a parallel "liveness" set (mirroring what was done for the wallet API) would be a reasonable follow-up but is a separate concern from a label rename.
- **No internal Scala/test renames.** Many internal classes (e.g. `ReceiveFaucetCouponTrigger`, `listValidatorFaucetCouponsOnDomain`) still process the deprecated `ValidatorFaucetCoupon` template; their names accurately describe what they do. They can be revisited if/when the deprecated template is removed.

## Test plan

- [x] `git diff --check upstream/main HEAD`
- [x] `npm run format:check -- src/utils/buildAmuletConfigChanges.ts` from `apps/sv/frontend`
- [x] `npm run lint:check -- src/utils/buildAmuletConfigChanges.ts` from `apps/sv/frontend`
- [x] Reviewed the final PR diff: it is limited to the two user-facing SV frontend label strings.

PR #5519, opened 2026-05-13
Fix wallet delegation tests by scoping button lookup to active dialog

martinflorian-da:
Fixes DACH-NY/cn-test-failures#8344

Fix is stolen from #5249, thanks @samsondav

I removed things from the original fix proposal that don't seem required to fix above issue, based on my local testing.

[ci]


moritzkiefer-da (approved):
looks plausible, for my own education why is the scoping required? Do we have multiple proceed buttons?

martinflorian-da:
> Do we have multiple proceed buttons?

My understanding is that yes... But tbc my confidence here comes mainly from empirics. It was extremely flaky for me locally without this fix, and now it doesn't fail anymore.

moritzkiefer-da:
wfm don't think I really understand why we have multiple proceed buttons but if it's less flaky that seems good :shrug:

pawelperek-da (approved):
LGTM

PR #5522, opened 2026-05-13
change the HTTP status code for UnknownContractSynchronizers error to let the client know that they should retry during validator onboarding

mblaze-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/8335


martinflorian-da (approved):
Maybe reference the issue that this was about?

Apart from that: looks reasonable, thanks! If we classify errors such as `UnhandledException` and `TemplatePreconditionViolated` as `Conflict` (then retry on that) then adding this here is certainly reasonable.

PR #5523, opened 2026-05-13
Remove update history from validator app

moritzkiefer-da:
fixes #4880

[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletTxLogAcsIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletTxLogAcsIntegrationTest.scala
@@ -55,6 +55,12 @@ class WalletTxLogAcsIntegrationTest
         localToStart.foreach(_.waitForInitialization())
       }
 
+      clue("Start SV1 validator and onboard SV1 as an end user") {
+        sv1ValidatorBackend.start()
+        sv1ValidatorBackend.waitForInitialization()
+        onboardWalletUser(sv1WalletClient, sv1ValidatorBackend)
+      }
+
       // Note: this test uses an SV party, as it's easier to create
       // amulets for parties hosted on an SV participant.
       val sv1UserParty = sv1Backend.getDsoInfo().svParty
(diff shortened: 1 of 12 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletTxLogAcsIntegrationTest.scala
```diff
@@ -55,6 +55,12 @@ class WalletTxLogAcsIntegrationTest
         localToStart.foreach(_.waitForInitialization())
       }
 
+      clue("Start SV1 validator and onboard SV1 as an end user") {
```
moritzkiefer-da, line 58: This is actually a change: If you onboard your party to the validator you will only get tx history after you onboarded it. Previously update history initialized at latest pruned offset and we backfilled from there. I'm leaning towards just accepting this. It doesn't change anything for existing users and for new users it seems very unlikely you actually hit this in practice in a way that matters. And it's only tx history, balances are computed from the acs so they're correct

rautenrieth-da (approved):
Thanks! Change looks good, just two comments:
- We have automatically truncated history data before in https://github.com/canton-network/splice/blob/main/apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V051__truncate_sv_splitwell_history.sql. Should be easy to duplicate and modify, but we have to be sure to not accidentally delete scan history by this.
- What's the long term plan for letting 3rd parties create explorers that include private transaction data? That would have been possible by giving them access to the validator update history. Will they have to consume the ledger API directly? Or do we not want to support this?

moritzkiefer-da:
> What's the long term plan for letting 3rd parties create explorers that include private transaction data? That would have been possible by giving them access to the validator update history. Will they have to consume the ledger API directly? Or do we not want to support this?

consume the ledger API. We really do not want to duplicate things the ledger API already provides.

> Should be easy to duplicate and modify, but we have to be sure to not accidentally delete scan history by this.

Yeah it's not that hard, I think for now I'm not convinced it's worth the effort though either so I think I would just ignore it until we get sufficient evidence that it's worth doing.

PR #5527, opened 2026-05-13
Implement accept / withdraw / reject of transfers v2

OriolMunoz-da:
Fixes https://github.com/canton-network/splice/issues/5415

Fun fact: the accept was using v1, it just happened to work because they use the same template :facepalm:


meiersi-da (approved):
Thank you! 🙏

PR #5532, opened 2026-05-13
Fix the right apps at end of RollForwardLsuDRIntegrationTest

martinflorian-da:
Fixes DACH-NY/cn-test-failures#8419

Corrects #5489 ; the apps stopped there are not running anymore here...

[ci]


moritzkiefer-da (approved):
we should fire your reviewer

martinflorian-da:
> we should fire your reviewer

We wouldn't have to if we had fired me one year ago! I can't even get static checks to pass apparently...

PR #5533, opened 2026-05-13
Prototype SV governance voter flow

ericmann:
## Summary

This PR is a prototype for maintainer review and discussion, not necessarily a final upstream design. It makes the proposed Phase 1 SV governance-voter model concrete for discussion under https://github.com/canton-foundation/canton-dev-fund/pull/223, especially Milestone 1: Governance-Voting Identity and CIP.

Phase 1 preserves the existing one-vote-per-SV semantics. A governance voter is modeled as an alternate signer for a represented SV's vote, not as a new voting unit or source of additional vote weight.

Changes included:

- Add a hardcoded governance-voter action taxonomy via `isGovernanceVoterAction`, with new `ActionRequiringConfirmation` constructors rejected by default until explicitly reviewed.
- Extend `Vote` with `castBy` and `castByRole` so vote records distinguish the represented SV from the party/authority path that signed the vote.
- Add `SvGovernanceVoter`, an SV-declared binding template without a contract key; downstream paths fetch it by contract ID and validate the one-active-binding assumption at the workflow level.
- Add `DsoRules_CastGovernanceVote`, which validates the binding, represented SV, governance-voter signer, signer role, and action allowlist before writing the vote into the represented SV's existing vote slot.
- Add Daml tests for the action taxonomy, vote-slot/tally semantics, binding lifecycle and authorization, and governance-voter casting/overwrite scenarios.
- Add an SV-operator documentation note describing the prototype scope, proposed allowlist, attribution semantics, binding invariant, and explicit contract-ID submission shape.

(Internal) Design IDs covered: GV-001, GV-002, GV-003, GV-004, GV-005, GV-006.

## Notes For Reviewers

- `SRARC_OffboardSv` is included in the proposed Phase 1 allowlist so reviewers can evaluate it explicitly; this should be validated through maintainer/CIP review because it is a high-impact membership action.
- `governanceVoter == sv` is allowed for bootstrap/self-voting, while `governanceVoter == dso` is rejected.
- The binding is unilateral/SV-declared in this prototype. If maintainers prefer bilateral consent, this can move toward a Propose-Accept shape in a follow-up revision.
- The binding intentionally has no contract key in this prototype. The one-active-binding-per-SV rule is treated as an invariant to validate rather than a template-key constraint.

## Test Plan

- `direnv exec . sbt "splice-dso-governance-test-daml/damlTest"`
- `direnv exec . sbt damlDarsLockFileUpdate`


daml/splice-dso-governance-test/daml/Splice/Scripts/TestGovernance.daml
```diff
@@ -34,6 +35,293 @@ import Splice.Testing.Utils
...
+  -- Rotation to the DSO party is rejected.
+  submitMustFail sv1 $ exerciseCmd rotateResult.bindingCid RotateGovernanceVoter with
+    newGovernanceVoter = dso
+
+  -- Non-SV parties cannot rotate or clear the binding.
+  submitMustFail otherParty $ exerciseCmd rotateResult.bindingCid RotateGovernanceVoter with
+    newGovernanceVoter = otherParty
+  submitMustFail otherParty $ exerciseCmd rotateResult.bindingCid ClearGovernanceVoter
+
+  submit sv1 $ exerciseCmd rotateResult.bindingCid ClearGovernanceVoter
```
isegall-da: Does it make sense to support clearing?  Then no one can vote.
I'm thinking maybe as part of onboarding, always create self-voting, and only support rotation. "clearing" is then replaced by rotating back to the operator.

daml/splice-dso-governance-test/daml/Splice/Scripts/TestGovernance.daml
```diff
@@ -34,6 +35,293 @@ import Splice.Testing.Utils
...
+        reason = Reason with url = ""; body
+        optCastAt = None
+  let operatorVote accept body = Vote with
+        sv = sv1
+        castBy = sv1
+        castByRole = VCR_Operator
+        accept
+        reason = Reason with url = ""; body
+        optCastAt = None
+  let requestOffboardSv = submit (actAs sv2 <> readAs dso) $ exerciseCmd dsoRulesCid DsoRules_RequestVote with
```
isegall-da, line 213: I think we agreed that for non-operational votes, it's the governance voter who should also create vote requests, not the operator. Seems to make more sense to me.

daml/splice-dso-governance-test/daml/Splice/Scripts/TestGovernance.daml
```diff
@@ -34,6 +35,293 @@ import Splice.Testing.Utils
...
+  case Map.lookup sv1 request.votes of
+    None -> fail "represented SV vote missing"
+    Some vote -> do
+      vote.sv === sv1
+      vote.castBy === governanceVoter1
+      vote.castByRole === VCR_GovernanceVoter
+      vote.accept === False
+
+  passTime (minutes 1)
+  operatorResult <- submit (actAs sv1 <> readAs dso) $ exerciseCmd dsoRulesCid DsoRules_CastVote with
```
isegall-da: I don't think this should be allowed. 
governance votes should be voted on only by governance parties
operational votes only by operator

Don't see the argument for allowing operators to override the governance voters.

daml/splice-dso-governance/daml/Splice/DSO/GovernanceVoter.daml
```diff
@@ -0,0 +1,41 @@
...
+data ClearGovernanceVoterResult = ClearGovernanceVoterResult
+  deriving (Eq, Show)
+
+template SvGovernanceVoter
+  with
+    dso : Party
+    sv : Party
+    governanceVoter : Party
+  where
+    signatory sv
```
isegall-da: Thinking out loud:
I wonder if we need a two-phase workflow here where both the voter and the operator agree to this. I think not, and it's fine for the operator to unilaterally assign the voter without their consent (as you've implemented it), so just raising the option for discussion.

daml/splice-dso-governance-test/daml/Splice/Scripts/TestGovernance.daml
isegall-da: > no one can vote

That is, assuming we get rid of "operator may override voter" as I propose below.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -768,25 +801,72 @@ template DsoRules with
...
+        requestCid : ContractId VoteRequest
+        bindingCid : ContractId SvGovernanceVoter
+        vote : Vote
+      controller vote.castBy
+      do
+        -- validate vote parameters
+        requireWellformedVote config vote
+        binding <- fetch bindingCid
+        require "Binding dso must match rules dso" (binding.dso == dso)
+        require "Vote SV must match binding SV" (vote.sv == binding.sv)
```
Ian-avro: maybe my misunderstanding but can two SvGovernanceVoter contracts existing for the same SV?

daml/splice-dso-governance/daml/Splice/DSO/GovernanceVoter.daml
Ian-avro: I think it's fine as is - it's fair to assume the SV and govVoter have an implicit offchain relationship

isegall-da: Yes, I agree. Just wanted to bring it up for explicit discussion.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
isegall-da: Good point. 
I think that is it is right now, yes.
What I'm uncertain is whether that should indeed be supported... I think I don't see a good reason why not. Neither options was an explicit requirement I believe.

Opened a discussion here: https://daholdings.slack.com/archives/C08PC19QZA9/p1778706853632719

daml/splice-dso-governance-test/daml/Splice/Scripts/TestGovernance.daml
ericmann: Good call; operator override removed.

daml/splice-dso-governance-test/daml/Splice/Scripts/TestGovernance.daml
ericmann: Good strategy. Will also update the design doc to match.

daml/splice-dso-governance-test/daml/Splice/Scripts/TestGovernance.daml
ericmann: You're absolutely right and I missed that details - corrected on latest.

daml/splice-dso-governance/daml/Splice/DSO/GovernanceVoter.daml
ericmann: Resolved as-is — no code change. As Ian notes, the SV and governance-voter party are assumed to have an offchain relationship in Phase 1, so unilateral SV declaration is acceptable. The design doc records this as a Phase 1 choice with Propose-Accept called out explicitly as a possible follow-on if maintainers want on-ledger consent.

That said, good to flag for the record and discussion :-)

daml/splice-dso-governance/daml/Splice/DsoRules.daml
ericmann: To answer @Ian-avro's question, currently yes. As to whether or not it should be - I'll err on "don't see a good reason why not" for the moment.

`RotateGovernanceVoter` is consuming so rotation never duplicates, but direct `create SvGovernanceVoter` on top of an existing binding does. The cast choice trusts the presented `bindingCid` and checks `vote.castBy == binding.governanceVoter`, so any active binding can authorize a cast — into the represented SV's single vote slot, under the same per-SV cooldown.

Something we can always revisit later if there's a solid reason we want a strict 1:1.

isegall-da: If you do allow more than one, then supporting archival of them does make sense, but I think I'd go for minimalistic unless proven to be required otherwise. Start with one, extend if needed in the future. As @waynecollier-da wrote on Slack, a single party does not imply a single user.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -124,6 +125,26 @@ data AnsEntryContext_ActionRequiringConfirmation
...
   deriving (Eq, Show)
 
+-- | Phase 1 governance-voter action taxonomy.
+--
+-- This is deliberately an explicit allowlist. New action constructors are not
+-- governance-voter eligible unless they are reviewed and added here.
+isGovernanceVoterAction : ActionRequiringConfirmation -> Bool
+isGovernanceVoterAction action = case action of
+  ARC_DsoRules with dsoAction -> case dsoAction of
+    SRARC_OffboardSv _ -> True
```
moritzkiefer-da, line 144: Is that deliberate? I was expecting this to be an operational concern as it often needs to be acted upon quickly.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -124,6 +125,26 @@ data AnsEntryContext_ActionRequiringConfirmation
...
+    SRARC_GrantFeaturedAppRight _ -> True
+    SRARC_RevokeFeaturedAppRight _ -> True
+    SRARC_SetConfig _ -> True
+    SRARC_UpdateSvRewardWeight _ -> True
+    SRARC_CreateUnallocatedUnclaimedActivityRecord _ -> True
+    _ -> False
```
moritzkiefer-da: I would still recommend to not have a wildcard here and and instead explicitly match on all constructors. That way when a new one gets added you're forced to think about how it should be treated which is much better than getting the wrong default.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -391,8 +418,8 @@ template VoteRequest
       -- ^ The reason for requesting the execution of the action. Typically a reference to some off-ledger justification.
     voteBefore : Time
       -- ^ The time before which votes are accepted, and SHOULD be submitted.
-    votes : Map.Map Text Vote
-      -- ^ The votes cast by current or previous SVs. These may be previous SVs in case
+    votes : Map.Map Party Vote
```
moritzkiefer-da: That doesn't work, it's not a backwards compatible change.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -424,10 +451,17 @@ data VoteRequestOutcome
...
+  | VCR_GovernanceVoter
+  deriving(Eq, Show)
+
 --
--- | A vote cast by an SV.
+-- | A vote cast for an SV.
 data Vote = Vote with
-    sv : Party -- ^ The SV party used to submit the vote.
+    sv : Party -- ^ The represented SV whose vote slot is updated.
+    castBy : Party -- ^ The party that signed the vote command.
```
moritzkiefer-da: these also need to be optional

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -753,13 +795,64 @@ template DsoRules with
...
           trackingCid = None
           targetEffectiveAt
         return DsoRules_RequestVoteResult with ..
 
+    -- Governance-voter path for opening a vote request.
+    --
+    -- Controlled by the governance-voter party authenticated via the binding;
+    -- the represented SV is taken from the binding. The initial vote is
+    -- recorded against the represented SV's slot with VCR_GovernanceVoter.
+    nonconsuming choice DsoRules_RequestGovernanceVote : DsoRules_RequestGovernanceVoteResult
```
moritzkiefer-da: do we really need a new choice or can we just add biningCid as an optional argument to the vote choice and then in the choice branch on whether it's a governance or non-governance choice?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -753,13 +795,64 @@ template DsoRules with
...
+      with
+        governanceVoter : Party
+        bindingCid : ContractId SvGovernanceVoter
```
moritzkiefer-da: nit: Naming seems confusing, you call it `binding` but then the template is called `SvGoveranceVoter`

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -753,13 +795,64 @@ template DsoRules with
...
+        action : ActionRequiringConfirmation
+        reason : Reason
+        voteRequestTimeout : Optional RelTime
+        targetEffectiveAt : Optional Time
+      controller governanceVoter
+      do
+        require "Action must be governance-voter eligible" (isGovernanceVoterAction action)
+        binding <- fetch bindingCid
```
moritzkiefer-da: never use fetch. Use one of the checked fetches that also validates the DSO for you and probably in this case you can maybe even make the governance voter the owner and then use `ForOwner`

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -753,13 +795,64 @@ template DsoRules with
...
+          votes = Map.fromList [(binding.sv, initialVote)]
+          trackingCid = None
+          targetEffectiveAt
+        return DsoRules_RequestGovernanceVoteResult with ..
+
 
     -- Note that this choice can be used to both cast the initial vote, and update a vote.
+    --
+    -- Operator path: only for *operational* actions. Governance-voter eligible
+    -- actions are cast exclusively via DsoRules_CastGovernanceVote.
```
moritzkiefer-da: same question here, it's not clear to me why we need two separate choices

daml/splice-dso-governance/daml/Splice/DSO/GovernanceVoter.daml
```diff
@@ -0,0 +1,47 @@
...
+    observer dso, governanceVoter
+
+    ensure
+      sv /= dso
```
moritzkiefer-da: not sure why we need those assertions, I don't think we do that anywhere else.

daml/splice-dso-governance/daml/Splice/DSO/GovernanceVoter.daml
moritzkiefer-da: That means the node operator has fully control though, they can just change the governance voter to themselves again. So this doesn't really protect against anything.

daml/splice-dso-governance/daml/Splice/DSO/GovernanceVoter.daml
```diff
@@ -0,0 +1,47 @@
...
+-- on governance-voter actions for the represented SV.
+--
+-- The represented SV is the sole signatory, so the implicit per-signatory
+-- Archive choice still lets the SV unilaterally archive any of its bindings.
+-- This is self-harm only (the SV temporarily loses the ability to cast on
+-- governance-voter actions) and is trivially recoverable by creating a new
+-- self-binding. Hard enforcement of the "every SV has at least one active
+-- binding" invariant is deferred; the Phase 1 expectation is that the SV
+-- workflow uses RotateGovernanceVoter rather than direct archive.
+template SvGovernanceVoter
```
moritzkiefer-da, line 25: I'm rather nervous about the signatories on that. Having the SV as the signatory means we also need the SV as a confirmer for any transaction involving this which is often not what you want. If you look at the exiting contracts for votes, prices, … you see that they are all very careful to only have the DSO as a signatory. Is there anything that stops us from doing the same here and create this through dso rules? You could even then do the initial creation with confirmation and deduplication to avoid duplicates.

daml/splice-dso-governance/daml/Splice/DSO/GovernanceVoter.daml
```diff
@@ -0,0 +1,47 @@
...
+  with
+    dso : Party
+    sv : Party
+    governanceVoter : Party
+  where
+    signatory sv
+    observer dso, governanceVoter
```
moritzkiefer-da: I'm wondering if you really need the governanceVoter as an observer or whether you can read it through scan or something like that. That way you would fully avoid any potential issues with the node potentially unvetting the dar and breaking something. It could be that we're ok with that but if we are we need a careful analysis of that.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
isegall-da: Yes, we discussed this. offboarding is one of those where it is very unclear what the right choice is. 
Some offboardings are because the node is broken, some because of a governance change. 

For now, let's ignore which action is governance vs operational, that will be a separate discussion in parallel to the design here, that should not affect design choices.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: What staleness are we willing to accept? Let's say at time X someone creates a vote at this point SvGovernanceVoter is still pointing to party A. When the vote gets executed it now points to party B.

I'd lean towards saying this needs to get revalidated similarly to how we don't allow the votes of an offboarded SV to still have an effect.

daml/splice-dso-governance/daml/Splice/DSO/GovernanceVoter.daml
ericmann: Good call — switched in ea2cddf70.

daml/splice-dso-governance/daml/Splice/DSO/GovernanceVoter.daml
ericmann: Dropped the `ensure` clause in 1bf30494a.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
ericmann: Reverted in 9b521d095

daml/splice-dso-governance/daml/Splice/DsoRules.daml
ericmann: Done in 9b521d095

daml/splice-dso-governance/daml/Splice/DsoRules.daml
ericmann: Done (SRARC_* side only) in 4c8cc217b

daml/splice-dso-governance/daml/Splice/DsoRules.daml
ericmann: Done in 4a90855a8

daml/splice-dso-governance/daml/Splice/DsoRules.daml
ericmann: Done in 4a90855a8

daml/splice-dso-governance/daml/Splice/DsoRules.daml
ericmann: Per @isegall-da's reply: parking the operational/governance taxonomy debate; keeping `SRARC_OffboardSv -> True` for now.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
ericmann: Done in 52c51ea87

daml/splice-dso-governance/daml/Splice/DsoRules.daml
ericmann: Done in 6d6033ca0

daml/splice-dso-governance/daml/Splice/DsoRules.daml
ericmann: Done in e06033d1a.

isegall-da:
@ericmann let's please not bloat this PR with backend code, and keep it limited to Daml. 
We should create a feature branch to which we merge PRs gradually as they are reviewed, not aggregate more and more changes in a single PR.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -409,6 +452,12 @@ data DsoRules_CloseVoteRequestResult = DsoRules_CloseVoteRequestResult with
     offboardedVoters : [Text] -- ^ SVs that voted but were offboarded before the vote completed.
     abstainingSvs : [Text] -- ^ SVs that did not vote.
     outcome : VoteRequestOutcome  -- ^ The final result of the vote.
+    staleBindingVoters : Optional [Text]
+      -- ^ SVs whose governance-voter-cast votes were dropped because the
+      -- binding they were cast under was no longer authoritative at close
+      -- time. `None` when the caller did not supply `currentBindings` and
```
moritzkiefer-da, line 458: why does the caller get a choice in doing that or not? That seems like an easy thing to exploit

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -424,16 +473,37 @@ data VoteRequestOutcome
...
+    castByRole : Optional VoteCastRole
+      -- ^ Attribution metadata: the authority path used to cast the vote.
+      -- Set by the server when recording a vote; `None` on contracts that
+      -- pre-date this field.
+    bindingCid : Optional (ContractId SvGovernanceVoter)
+      -- ^ The SvGovernanceVoter binding under which this vote was cast on the
+      -- governance-voter path. Set by the server when recording a vote;
+      -- `None` on votes cast via the operator path and on contracts that
+      -- pre-date this field. Used at close-vote time to detect votes cast
+      -- under a binding that has since been rotated, so the vote can be
```
moritzkiefer-da, line 505: Why do we need to bind by contract id? Isn't it sufficient that there is an SvGovernanceVoter contract that says I'm allowed to vote as a given SV even if that has a different contract id?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -722,68 +792,169 @@ template DsoRules with
...
         requireWellformedReason config reason
         now <- getTime
+        (requesterName, initialVote) <- case bindingCid of
+          None -> do
+            require "Action is governance-voter eligible; pass `bindingCid` to take the governance-voter path"
+              (not (isGovernanceVoterAction action))
+            name <- case requester `Map.lookup` svs of
+              None -> fail "Requester is not an SV"
+              Some info -> pure info.name
+            let v = Vote with
```
moritzkiefer-da, line 835: might be easier to read if you introduce a helper function to create this vote with parameters for the fields that are actually different between the two code paths. That way it's easy to see what is shared and what is not.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -814,6 +985,17 @@ template DsoRules with
...
         amuletRulesCid : Optional (ContractId AmuletRules)
         sv : Optional Party
+        currentBindings : Optional [ContractId SvGovernanceVoter]
+          -- ^ Optional list of currently-active SvGovernanceVoter bindings.
+          -- When `Some cids`, each is `fetchChecked` (so the caller cannot
+          -- sneak in stale CIDs) and a `binding.sv -> ContractId` map is
+          -- built; any active-SV vote with `vote.bindingCid = Some cid`
+          -- where the live binding for `vote.sv` does not equal `cid` is
+          -- dropped from the tally and reported in `staleBindingVoters`.
+          -- The caller is trusted to pass the complete set of live bindings.
```
moritzkiefer-da, line 995: I don't see why this is a valid assumption.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1003,6 +1253,37 @@ template DsoRules with
...
+    -- confirmation quorum, so a single SV operator cannot unilaterally rotate
+    -- their own binding.
+    nonconsuming choice DsoRules_RotateGovernanceVoter : DsoRules_RotateGovernanceVoterResult
+      with
+        sv : Party
+        currentBindingCid : ContractId SvGovernanceVoter
+        newGovernanceVoter : Party
+      controller dso
+      do
+        require "Rotation target is not the DSO party" (newGovernanceVoter /= dso)
```
moritzkiefer-da, line 1274: feels like a weird assertion, generally we assert on things that really are required for correctness.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -945,6 +1156,45 @@ template DsoRules with
                 require "SV of duplicate matches" (duplicate.sv == vote.sv)
         return DsoRules_GarbageCollectAmuletPriceVotesResult
 
+    nonconsuming choice DsoRules_GarbageCollectSvGovernanceVoters : DsoRules_GarbageCollectSvGovernanceVotersResult
+      -- ^ Garbage-collect SvGovernanceVoter bindings for removed SVs and
+      -- duplicate bindings that may be left behind by dev-net re-onboarding
```
moritzkiefer-da, line 1161: how can you get duplicates on sv reonboarding? SV reonboarding changes party ids but keeps the name. Your `SvGovernanceVoter` contract seems to be tied to the party not the sv name so I would expect it not to be duplicate. What am I missing?

May still be a useful safeguard to have the option of merging duplicates if something goes wrong but then we should update the comment to claim it's just a safeguard.

ericmann:
## Superseded — migrating to the feature fork under the amended M1 scope

This prototype is being retired in favor of a cleaner, rescoped implementation on our feature fork, **[canton-network/splice-sv-voting-dapp](https://github.com/canton-network/splice-sv-voting-dapp)**.

### Why

The M1 deliverable is being amended via [canton-foundation/canton-dev-fund#414](https://github.com/canton-foundation/canton-dev-fund/pull/414), which decouples *the ability to split votes by voter type* (governance vs. operational) from *actually splitting the votes*. M1 now lands **the flexibility, not the votes** — i.e. the on-ledger infrastructure for vote classification with **no change to existing voting behavior**. The behavioral pieces in this prototype (strict operator-path rejection, vote-binding staleness, GC) are intentionally dropped for now and deferred to a later phase.

### Replacement work (stacked PRs on the fork)

The changes here were re-ported (not rebased — the branches had diverged too far) and split so contract and backend changes review independently:

1. **[#2 — Governance-voter Daml infrastructure](https://github.com/canton-network/splice-sv-voting-dapp/pull/2)**: `SvGovernanceVoter` binding, vote attribution (`castBy`/`castByRole`), `SRARC_RotateGovernanceVoter`, unified request/cast choices. Dormant in Phase 1.
2. **[#3 — Config field classification](https://github.com/canton-network/splice-sv-voting-dapp/pull/3)** (stacked on #2): `ConfigFieldClassification` model, default-classification map, `SRARC_SetConfigFieldClassification` reclassification vote, `ConfigFieldClassifications` singleton.
3. **[#4 — SV app store queries](https://github.com/canton-network/splice-sv-voting-dapp/pull/4)** (stacked on #3): read-only `SvDsoStore` observability for the dormant state, with store tests.

Once these land on the fork and the amendment is approved, we'll target upstreaming them into Splice. Closing this PR to consolidate the work there.

PR #4724, opened 2026-03-30
CIP-104: Daml for rewards accounting V2 and sv-app UI fixes

dfordivam:
Fixes: https://github.com/hyperledger-labs/splice/issues/4804

This is the base branch for all the work to be done as part of https://github.com/hyperledger-labs/splice/issues/4802

This branch consists of 
- DAML changes
  - https://github.com/hyperledger-labs/splice/pull/3964
  - Add `trafficPrice`, `rewardConfig` to `SummarizingMiningRound`
  - Add `appRewardCouponThreshold : Decimal` to `RewardConfig`
  - Compute featured app rewards from activity totals in `runNextIssuance` and `runNextIssuanceD`
- Version bumps, + dars
- SV UI fixes to allow 'rewardConfig' change voting

- Fixes done as part of https://github.com/canton-network/splice/pull/5193

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a logical synchronizer upgrade test is required (from canton-3.5), comment `/lsu_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/hyperledger-labs/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/hyperledger-labs/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #5552, opened 2026-05-15
SV-app add CalculateRewardsTrigger and ProcessRewardsTrigger

### LINKED ISSUE #4844: SV-app add CalculateRewardsTrigger

dfordivam:
Fixes: #4844 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -0,0 +1,241 @@
...
+
+  override def completeTask(
+      task: Task
+  )(implicit tc: TraceContext): Future[TaskOutcome] = {
+    val round = task.calculateRewards.payload.round.number
+    getRootHash(round).flatMap {
+      case None =>
+        Future.successful(
+          TaskSuccess(
+            s"waiting for scan to compute root hash for CalculateRewardsV2 round $round, will retry"
```
meiersi-da: Using `TaskSuccess` does not work for this:
- it retries immediately ==> busy looping
- it skews the metrics reported for task completion

Given that we expect Scan to typically be ready, I'd suggest to raise gRPC error `PRECONDITION_FAILED`, which is considered a retriable error by our trigger infra, and triggers a retry with exponential backoff.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -0,0 +1,241 @@
...
+        val action = startProcessingRewardsAction(
+          task.calculateRewards.contractId,
+          rootHash,
+        )
+        for {
+          queryResult <- store.lookupConfirmationByActionWithOffset(svParty, action)
+          taskOutcome <- queryResult match {
+            case QueryResult(_, Some(_)) =>
+              Future.successful(
+                TaskSuccess(
```
meiersi-da: These tasks must be filtered as part of `retrieveTasks`, as otherwise, the trigger will busy loop.

Have a look at your logs for all your triggers to double-check their behavior.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -0,0 +1,241 @@
...
+        } yield taskOutcome
+    }
+  }
+
+  override def isStaleTask(task: Task)(implicit
+      tc: TraceContext
+  ): Future[Boolean] =
+    store.multiDomainAcsStore
+      .lookupContractById(CalculateRewardsV2.COMPANION)(task.calculateRewards.contractId)
+      .map(_.isEmpty)
```
meiersi-da: it's also stale if the confirmation already exists

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -0,0 +1,241 @@
...
+        upgradesConfig,
+        context.clock,
+        context.retryProvider,
+        loggerFactory,
+        retryConnectionOnInitialFailure = false,
+      )
+      .transformWith {
+        case Failure(ex) =>
+          Future.failed(
+            new RuntimeException("Failed to connect to scan for root hash lookup", ex)
```
meiersi-da: This will likely be noisy in production. Scan can be down. Best is probably to just let the errors bubble up and be handled by our retry infrastructure.

Also: it seems that you are creating a fresh scan connection per task. Is this what's done in other parts of the codebase as well?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -0,0 +1,241 @@
...
+  private def getRootHash(round: Long)(implicit tc: TraceContext): Future[Option[Hash]] =
+    withScanConnection { conn =>
+      conn.getRewardAccountingRootHash(round).map {
+        case GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashOk(ok) =>
+          Some(new Hash(ok.rootHash))
+        case GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashUndetermined(_) =>
+          None
+        case GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashCannotProvide(_) =>
+          throw new RuntimeException(
+            s"Scan cannot provide root hash for round $round"
```
meiersi-da: there seems to be a missing TODO reference here. This is where we'd want to use the fallback or not?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
```diff
@@ -0,0 +1,219 @@
...
+
+  override def completeTaskAsDsoDelegate(
+      task: ProcessRewardsV2Contract,
+      controller: String,
+  )(implicit tc: TraceContext): Future[TaskOutcome] = {
+    if (task.payload.dryRun != isDryRun) {
+      Future.successful(
+        TaskSuccess(
+          s"Skipping ProcessRewardsV2 for round ${task.payload.round.number} with dryRun=${task.payload.dryRun}"
+        )
```
meiersi-da: I worry that the logs from this are confusing, and the metrics will also look inflated. 

I'd suggest to add an optonal `contractFilter` to the `OnAssignedContractTrigger` that is immediately applied to the `Pekko.Source` that's used to stream work. Thereby avoiding this noise.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
```diff
@@ -0,0 +1,219 @@
...
+        upgradesConfig,
+        context.clock,
+        context.retryProvider,
+        loggerFactory,
+        retryConnectionOnInitialFailure = false,
+      )
+      .transformWith {
+        case Failure(ex) =>
+          Future.failed(
+            new RuntimeException("Failed to connect to scan for batch lookup", ex)
```
meiersi-da: Same comments as for the other trigger apply.

meiersi-da: also wrt connection instantiation.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
```diff
@@ -691,6 +691,140 @@ class DbSvDsoStore(
...
+  ]] =
+    for {
+      result <- storage
+        .query(
+          selectFromAcsTableWithState(
+            DsoTables.acsTableName,
+            acsStoreId,
+            domainMigrationId,
+            splice.amulet.rewardaccountingv2.ProcessRewardsV2.COMPANION,
+            orderLimit = sql"""order by reward_round limit ${sqlLimit(limit)}""",
```
meiersi-da: Do we have an index that would make this sorting efficient? If not, do we need this sorting here?

There might be many coupons in prod, and we wouldn't want this query to become a problem.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
```diff
@@ -46,6 +55,28 @@ class TrafficBasedRewardsTimeBasedIntegrationTest
     with ExternallySignedPartyTestUtil
     with TokenStandardTest {
 
+  // We run this test in multiple modes to confirm that various triggers work
```
meiersi-da: As discussed. Aim to reduce the number of modes required for good enough test coverage.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
```diff
@@ -87,70 +133,193 @@ class TrafficBasedRewardsTimeBasedIntegrationTest
...
+              // with CalculateRewardsV2 to have a root hash.
+              if (dryRunEnabled) {
+                clue("Archive dry-run CalculateRewardsV2 for rounds 0..5 via sv admin API") {
+                  sv1Backend.archiveDryRunRewardAccountingContracts((0L to 5L).toSeq)
+                }
+              } else {
+                // Bootstrapping a network with mintingVersion set to trafficBasedAppRewards
+                // is in principle not supported, as the round 0 will never have
+                // activity totals/root-hash calculated, and its CalculateRewardsV2 cannot be processed.
+                // So the only way to handle this in test is via direct archive.
```
meiersi-da: This is actually a problem when we reset TestNet the next time.

What can we do so that we don't need to skip these first five rounds? I'd have expected anyways that we only need to skip Round 0, due to our requirement for there to be activity for the round one lower. 

We can probably use a similar approach as done here:
https://github.com/canton-network/splice/blob/08c039548dda1be4e77dfe19fc921f6ca5d80b6e/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/ScanAggregator.scala#L175-L179

i.e., if no app activity metadata record exists at all and we are the first SV, then we insert it with the `initialRound` and `recordTime = unixEpochStart`.

meiersi-da: And it's also a problem for all our tests once we make traffic-based app rewards the default.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/AmuletConfigUtil.scala
```diff
@@ -150,10 +150,11 @@ trait AmuletConfigUtil extends TestCommon {
...
             eventually() {
-              sv.listVoteRequests()
-                .filter(
-                  _.payload.trackingCid == voteRequestCid.contractId
-                ) should have size 1
+              sv.listVoteRequests().filter { vr =>
+                vr.contractId == voteRequestCid.contractId ||
+                (vr.payload.trackingCid.isPresent &&
+                  vr.payload.trackingCid.get == voteRequestCid.contractId)
+              } should have size 1
```
meiersi-da, line 157: Why was this change required? Not debating the change per-se, but perhaps there's something deeper hidden as to why we needed the change.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/AmuletConfigUtil.scala
```diff
@@ -171,7 +172,6 @@ trait AmuletConfigUtil extends TestCommon {
                 sv.lookupVoteRequest(voteRequestCid.contractId)
                   .payload
                   .votes should have size voteCount
-                sv.listVoteRequests() shouldBe empty
```
meiersi-da, line 174: Why is it no longer empty?

meiersi-da (commented):
Nice work! Thanks a lot! Great to see of this coming together @dfordivam and @adetokunbo 👏 


@adetokunbo : note that I gave quite a few pointers wrt the usage of triggers, task outcomes, and retries. Please have a look, as you are going to work on a trigger as well this week.

In general, when implementing a trigger it's always worth to check its logging behavior carefully and think about whether these logs are "right" for the work the trigger is doing. Filtering to the trigger name in `lnav` gets that job done quite well.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
dfordivam: I had earlier considered that, but there was an issue around `getRootHash(round)` fetch for doing the `lookupConfirmationByActionWithOffset` in `retrieveTasks` / `isStaleTask`.

But you are right that while other SVs are doing vote, this SVs trigger should filter in `retrieveTasks`.
To keep both these APIs DB fetch only (ie no `getRootHash`) we have a couple of options.
1. Fetch all confirmations submitted by this SV, and do in-memory filter by `calculateRewardsCid`
2. Do DB filter by `calculateRewardsCid`

I think 1 is straightforward as we would typically not have many active confirmations. wdyt? @meiersi-da

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/AmuletConfigUtil.scala
dfordivam: This was an existing bug in `castVotes` API, it was only used in 1 SV configuration so nobody hit it.
The first vote request created by SV1 has `trackingCid` `None`, and subsequent vote requests set it to the first `voteRequestCid`.
I even confirmed that an existing test fails if we convert it to 4 SVs, and requires this fix.
("round management with scheduled config change of doubled tickDuration") in SvTimeBasedRoundMgmtIntegrationTest

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/AmuletConfigUtil.scala
dfordivam: This assertion was at the wrong place, as vote requests would only go away when majority does voting.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
dfordivam: We will have a similar issue in `SummarizingMiningRoundTrigger` also, as it also filters using `lookupConfirmationByActionWithOffset`. 

https://github.com/canton-network/splice/blob/428bec5c55fc0638c44c8652b486e0a64f8782be/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/SummarizingMiningRoundTrigger.scala#L66-L80

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
dfordivam: Ok, I have added a TODO linking to https://github.com/canton-network/splice/issues/5624

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
dfordivam: I have replaced this with 'mining_round' which has index. And in some ways 'mining_round' is more appropriate for the CalculateRewardsV2 and ProcessRewardsV2. Also bumped the store version in this PR, as I forgot to do that in the earlier PR where I added the ingestion. (#4724)

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
dfordivam: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
dfordivam: Ok, fixed this by returning `None`, The scan connection situation is a bit complicated actually  as we will not be able to likely make use of existing abstractions being used in sv-app. My idea is to remove this ephemeral connection later when BFT support is added, such that both are provided through a well designed abstraction. So I have just added a TODO for now.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
dfordivam: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
dfordivam: Added TODO

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/AmuletConfigUtil.scala
meiersi-da: Thx for the fix and the explanation.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
meiersi-da: Yes, 1 is what I was thinking about. We don't do any filtering based on whether Scan has data available. That we do as you do in the code now: throw a retryable exception.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
dfordivam: Replaced fresh scan connection with a long lived one, as discussed yesterday.

dfordivam:
@meiersi-da This PR is now ready for another review. All comments have been addressed, and/or TODOs added.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -0,0 +1,232 @@
...
+
+  override def completeTask(
+      task: Task
+  )(implicit tc: TraceContext): Future[TaskOutcome] = {
+    val round = task.calculateRewards.payload.round.number
+    getRootHash(round).flatMap {
+      case None =>
+        throw Status.FAILED_PRECONDITION
+          .withDescription(
+            s"waiting for scan to compute root hash for CalculateRewardsV2 round $round, will retry"
```
meiersi-da:
```suggestion
            s"Scan has not yet computed the root hash for CalculateRewardsV2 round $round."
```
That's the precondition that is actually violated.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -0,0 +1,232 @@
...
+        val action = startProcessingRewardsAction(
+          task.calculateRewards.contractId,
+          rootHash,
+        )
+        for {
+          queryResult <- store.lookupConfirmationByActionWithOffset(svParty, action)
+          taskOutcome <- queryResult match {
+            case QueryResult(_, Some(_)) =>
+              Future.successful(
+                TaskSuccess(
```
meiersi-da:
```suggestion
                TaskNoop(
```

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
```diff
@@ -0,0 +1,186 @@
...
+    val batchHash = task.payload.batchHash.value
+    val batchF = fetchBatch(round, batchHash)
+    val dsoRulesF = store.getDsoRules()
+    for {
+      batch <- batchF
+      dsoRules <- dsoRulesF
+      damlBatch = convertBatch(batch)
+      choiceArg = new ProcessRewardsV2_ProcessBatch(
+        damlBatch,
+        new DamlSet(java.util.Collections.emptyMap()),
```
meiersi-da, line 85: Missing TODO item.

meiersi-da: To check the vetting state.

meiersi-da (approved):
Nice work! Thanks a lot!

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
dfordivam: Right, added a todo linking to the issue

PR #5576, opened 2026-05-18
Scan: Reward accounting root-hash and activity-totals return three values

dfordivam:
Modified the following endpoints to return "undetermined" and "cannot-compute" values, in addition to the result.
- /v0/reward-accounting-process/rounds/{round_number}/activity-totals
- /v0/reward-accounting-process/rounds/{round_number}/root-hash

As discussed last week with Simon and Robert; these are the criterion for the return value selection.
- undetermined
  returned if scan does not see any CalculateRewardsContract with a round number <= the queried round
or if the root hash (and associated data) is not found in the store
- cannot-compute	
   if scan does not have a complete set of activity records for the round

But in the code the "undetermined" is actually a catch-all case, because even if the `CalculateRewardsV2` exist for this round, and the scan is yet to compute the totals, "undetermined" would still work since the sv-app would retry.

As these are `internal` endpoints, marked as "subject to change", the existing /v0 endpoints have been modified, and PR has no release-notes mention. (original addition of these endpoints also did not have release-notes entry)

```
  /v0/reward-accounting-process/rounds/{round_number}/root-hash:
    get:
      tags: [internal, scan]
      x-jvm-package: scan
      operationId: "getRewardAccountingRootHash"
      description: |
        SV node internal API (CIP-0104, subject to change).
        Returns the root hash computed for the specified round.

```

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/openapi/scan.yaml
+++ b/apps/scan/src/main/openapi/scan.yaml
@@ -1901,8 +1901,6 @@ paths:
             application/json:
               schema:
                 "$ref": "#/components/schemas/GetRewardAccountingActivityTotalsResponse"
-        "404":
-          $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/404"
 
   /v0/reward-accounting-process/rounds/{round_number}/root-hash:
     get:
@@ -1926,8 +1924,6 @@ paths:
             application/json:
               schema:
                 "$ref": "#/components/schemas/GetRewardAccountingRootHashResponse"
-        "404":
-          $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/404"
 
   /v0/reward-accounting-process/rounds/{round_number}/batches/{batch_hash}:
     get:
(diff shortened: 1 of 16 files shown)


dfordivam:
@meiersi-da This is a small change we discussed last week. I have originally done this as part of my [integ test PR](https://github.com/canton-network/splice/pull/5552), but took this out in this separate PR on `main`. 

The change was in fact necessary to get the CI green by avoiding unnecessary error logging from the `CalculateRewardsTrigger` when it just needed to retry the fetch.

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -1926,8 +1924,6 @@ paths:
             application/json:
               schema:
                 "$ref": "#/components/schemas/GetRewardAccountingRootHashResponse"
-        "404":
-          $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/404"
 
   /v0/reward-accounting-process/rounds/{round_number}/batches/{batch_hash}:
```
meiersi-da:
```suggestion
  /v0/internal/reward-accounting-process/rounds/{round_number}/batches/{batch_hash}:
```
perhaps in this PR or in separate one. Just move all three APIs to use an '/internal/' so it's very clear that they should not be used by external scan users.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -740,6 +742,18 @@ class BftScanConnection(
...
+      ec: ExecutionContext,
+      tc: TraceContext,
+  ): Future[GetRewardAccountingRootHashResponse] =
+    bftCall(_.getRewardAccountingRootHash(roundNumber))
+
+  override def getRewardAccountingBatch(roundNumber: Long, batchHash: String)(implicit
+      ec: ExecutionContext,
+      tc: TraceContext,
+  ): Future[Option[GetRewardAccountingBatchResponse]] =
+    bftCall(_.getRewardAccountingBatch(roundNumber, batchHash))
```
meiersi-da, line 756: need to check whether doing BFT calls for this endpoint using the standard code path works as expected, as deviations in the responses are expected. I suspect we might be logging mismatches too aggressively.

Nothing to do right now, but something to be aware of.

meiersi-da (approved):
Thanks a lot.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
dfordivam: Yes, at the moment this is just to compile the code. When the triggers actually use BFT client, we will have to circle back to these concerns.

apps/scan/src/main/openapi/scan.yaml
dfordivam: Done in this PR

PR #5582, opened 2026-05-18
Public Sequencer: Add automation to SV app to look for ValidatorLicenseRequest and create ValidatorLicense

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/5285

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvSvAutomationService.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvSvAutomationService.scala
@@ -74,6 +74,18 @@ class SvSvAutomationService(
     )
   )
 
+  registerTrigger(
+    new ValidatorLicenseRequestTrigger(
+      triggerContext,
+      svStore,
+      dsoStore,
+      connection(
+        SpliceLedgerConnectionPriority.High
+      ),
+      participantAdminConnection,
+    )
+  )
+
   // notice the absence of UpdateHistory: the history for the sv party is not needed as we don't foresee ever adding TxLog for it
 
   registerTrigger(
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvSvAutomationService.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvSvAutomationService.scala
@@ -80,7 +80,7 @@ class SvSvAutomationService(
       svStore,
       dsoStore,
       connection(
-        SpliceLedgerConnectionPriority.High
+        SpliceLedgerConnectionPriority.Medium
       ),
       participantAdminConnection,
     )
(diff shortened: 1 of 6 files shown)


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvSvAutomationService.scala
```diff
@@ -74,6 +74,18 @@ class SvSvAutomationService(
     )
   )
 
+  registerTrigger(
+    new ValidatorLicenseRequestTrigger(
+      triggerContext,
+      svStore,
+      dsoStore,
+      connection(
+        SpliceLedgerConnectionPriority.High
```
moritzkiefer-da: high seems off, this probably should be medium

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/ValidatorLicenseRequestTrigger.scala
```diff
@@ -0,0 +1,167 @@
...
+      ]
+  )(implicit tc: TraceContext): Future[TaskOutcome] = {
+    val payload = task.payload
+    val svParty = svStore.key.svParty
+
+    logger.info(
+      s"Running ValidatorLicenseRequestTrigger for ${payload.validator} ValidatorLicenseRequest"
+    )
+
+    if (payload.sponsor != svParty.toProtoPrimitive) {
```
moritzkiefer-da: is this check doing anything? You already filter on store ingestion don't you?
i'm fine keeping it as an extra safeguard but then let's add a comment

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/ValidatorLicenseRequestTrigger.scala
```diff
@@ -0,0 +1,167 @@
...
+                .map(_.flatten)
+
+              _ <-
+                if (permissionResults.nonEmpty) {
+                  Future.unit
+                } else {
+                  Future.failed(
+                    Status.NOT_FOUND
+                      .withDescription(
+                        s"ParticipantSynchronizerPermission not yet found for any participant of $validatorParty. Checked participants: ${participantIds
```
moritzkiefer-da: don't we want to check that it's found for all nodes rather than any? It doesn't make a whole lot of sense to have a party hosted on 3 participants where only 1 is permissioned

moritzkiefer-da (approved):
thanks

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvSvAutomationService.scala
pasindutennage-da: Solved

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/ValidatorLicenseRequestTrigger.scala
pasindutennage-da: Thanks, added

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/ValidatorLicenseRequestTrigger.scala
pasindutennage-da: Thanks, fixed

PR #5593, opened 2026-05-19
Fix sequencer flow control in istio

moritzkiefer-da:
fixes https://github.com/DACH-NY/canton-network-internal/issues/4901

This just applies the config we already manually applied and tested. There is almost certainly further config tweaking we could do here but for now this definitely works far better than the default.

Note that the initial stream window size is quite a bit smaller than the default of 16mb. The connection window size is larger than the default of 24mb though. This is different from what https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge recommends but after searching around a bit it seems like the values there likely don't work well with grpc streams so this is a tradeoff to make sure we don't get performance regressions while still improving the situation.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/infra/src/istio.ts
+++ b/cluster/pulumi/infra/src/istio.ts
@@ -811,6 +811,100 @@ function configureSequencerHighPerformanceGrpcDestinationRule(
   });
 }
 
+// Istio proxies lots of client connections over relatively few connections. If one of the client connections gets stuck
+// (e.g. because the client died) buffers will fill up and eventually istio will stop sending connection-level window updates
+// to the sequencer and trigger netty flow control. This surfaces as requests that send back response headers but then nothing else until the client times out.
+// To mitigate that we set the connection window size meaningfully higher than the stream window size. To fully mitigate it it likely needs to be
+// connection window size >= stream window size * maxConcurrentStreams but that would increase istio memory significantly so for now the values are usually just high enough that it
+// doesn't happen in practice but not enough to fully prevent it.
+// We also enable http2 pings to ensure that connections get closed when clients go away and the istio buffers are cleared again.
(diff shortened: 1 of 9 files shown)


canton-network-da:
# [backport] Reminder #
- [x] release-line-0.6.4
- [x] release-line-0.6.3
- [x] release-line-0.6.2
- [ ] release-line-0.6.1

cluster/pulumi/infra/src/istio.ts
```diff
@@ -811,6 +811,100 @@ function configureSequencerHighPerformanceGrpcDestinationRule(
...
+              },
+            },
+          },
+        },
+        {
+          // istio -> upstream (aka sequencer)
+          applyTo: 'CLUSTER',
+          match: {
+            cluster: {
+              portNumber: 5008,
```
nicu-da, line 873: Do we also need to touch on the dabft port (that one is different right?)

nicu-da (approved):
Thank you, this is some torturous tuning. 
Should we also run a mediator/participant catch-up to see if it somehow gets negatively impacted by the changes?

cluster/pulumi/infra/src/istio.ts
moritzkiefer-da: > Do we also need to touch on the dabft port (that one is different right?)

maybe but that is also less of an issue as it has far less clients. So I'd defer that for now.

moritzkiefer-da:
>  Should we also run a mediator/participant catch-up to see if it somehow gets negatively impacted by the changes?

Good idea, let me merge it first given that it's already applied anyway and then run one (or trick the rotation into running one).

moritzkiefer-da:
makes zero difference in catchup, the start is when we had the config set, after the gap it was removed. it catches up at the same speed

image

PR #5600, opened 2026-05-19
Support bft-custom in docker-compose

### LINKED ISSUE #5560: support bft-custom in docker-compose
In https://github.com/canton-network/splice/pull/3651 we added it for k8s, but never extended that to docker-compose afaict, at least not in the start.sh script. Let's do that, as people (unfortunately) use docker-compose in prod.

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/5560


cluster/compose/validator/start.sh
```diff
@@ -35,6 +35,10 @@ function usage() {
   echo "  -i <identities_dump>: restore identities from a dump file. Requires a new participant identifier to be provided."
   echo "  -w: Wait for the validator to be fully up and running before returning."
   echo "  -l: Connect participant and validator also to docker network compose-sv_splice-sv-public, to use an SV deployed locally on docker compose"
+  echo "  -B: Enable bft-custom mode for scan and sequencer connections."
+  echo "  -u <comma_separated_urls>: Comma-separated list of Scan URLs for bft-custom mode (no spaces)."
+  echo "  -S <comma_separated_sv_names>: Comma-separated list of SV names for bft-custom mode."
+  echo "  -T <threshold>: Consensus threshold integer for bft-custom mode."
   echo "      Also implies -s, -c and -C to be the defaults for such a deployment."
```
isegall-da: This belongs with the `-l`, no? Seems like you added the new ones one line too early

cluster/compose/validator/start.sh
```diff
@@ -142,6 +162,19 @@ if [ "${local_compose_sv}" -eq 1 ]; then
   fi
 fi
 
+if [ $bft_custom -eq 1 ]; then
```
isegall-da, line 171: I think we should also enforce that -b and -B are not used together, that would just lead to a confusing config.

isegall-da: or maybe get rid of trust-single altogether, and adjust our tests to use custom-bft. I was anyway going to say that we should test bft-custom in the integration test.

pasindutennage-da: @isegall-da @moritzkiefer-da 

Isn't it the case that some people still use trust-single?

If yes, then I am tempted to keep both, however, adjust the test for bft-custom. WDYT?

moritzkiefer-da: Yeah I wouldn't remove it. Seems like unnecessary breakage. We could probably make it syntactic sugar for bft-custom with only a single entry though if it isn't already?

isegall-da: Fair. Supporting the flag as a syntactic sugar seems right.

cluster/compose/validator/start.sh
pasindutennage-da: Thanks, fixed

cluster/compose/validator/start.sh
pasindutennage-da: @moritzkiefer-da @isegall-da 

Added the -B -b mutual exclusive check

Modified DockerComposeFullNetworkFrontendIntegrationTest to test bft-custom setup

In a tech-debt issue, I will remove the trust-single all together.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DockerComposeFullNetworkFrontendIntegrationTest.scala
```diff
@@ -130,7 +130,7 @@ class DockerComposeFullNetworkFrontendIntegrationTest
   ): A = {
     try {
       val command =
-        (Seq("build-tools/splice-compose.sh", "start_network", "-w") ++ startFlags).asJava
+        (Seq("build-tools/splice-compose.sh", "start_network", "-w", "-B") ++ startFlags).asJava
```
isegall-da, line 133: Is there a way to somehow query the validator and confirm that it is indeed configured to do bft-custom?
Or maybe do something as ugly as checking in the logs that it is?
Just wonder how we can test that this is actually working..

pasindutennage-da: @isegall-da @moritzkiefer-da 

In this test, we are treating the validator as a closed-box: its started asynchronously in docker, and hence we don't get the luxary we have in the regular integration tests where we can do validator.config.getScanConfig.

In my understanding, loggerFactory.assertEventuallyLogsSeq doesn't work with this setup to check logs inside the test.

In my local run, I see the following logs, which confirm the correct scan config operation.

WDYT?

```
timestamp":"2026-05-20T18:22:35.022Z",with config = {scan-client {
amulet-rules-cache-time-to-live="10 minutes"
scans-refresh-interval="10 minutes"
seed-urls=["http://scan:5012\"
sv-names=[sv
threshold=1
type=bft-custom
use-last-known-connections-for-initialization=true

timestamp":"2026-05-20T18:23:11.564Z", Successfully connected to 1 scan(s), meeting the threshold of  1 logger_name":"o.l.s.s.a.a.c.BftScanConnection$ConfigurationProvidedScansBft:validator=validator_backend",
```

moritzkiefer-da: I think I wouldn't bother. Let's just confirm manually that it works (which you have already done).

cluster/compose/validator/compose-bft-custom.yaml
```diff
@@ -0,0 +1,13 @@
+services:
+  validator:
+    environment:
+      - ADDITIONAL_CONFIG_BFT_SCAN=
+      - |
+        ADDITIONAL_CONFIG_BFT_CUSTOM=
+          canton.validator-apps.validator_backend.scan-client.type = "bft-custom"
+          canton.validator-apps.validator_backend.scan-client.seed-urls = ${BFT_CUSTOM_SEED_URLS}
+          canton.validator-apps.validator_backend.scan-client.sv-names = ${BFT_CUSTOM_SV_NAMES}
```
moritzkiefer-da, line 9: For the helm chart we seem to support configuring different names and thresholds for scan and sequencer connections. What's the argument for doing it differently in docker compose?

moritzkiefer-da (approved):
Thanks! Probably worth mentioning in the docs as well

cluster/compose/validator/compose-bft-custom.yaml
pasindutennage-da: @moritzkiefer-da @isegall-da 

Actually it is just for simplicity: in helm one can provide sv1-2 for trusted scans and then sv3-4 for trusted sequeners.

To achieve 2 different set of trusted svs, we just need more flags in the input, but i don't see practical usecases for that - a validator trusting sv for scan and not for sequencer seems rare. 

Same goes for threshold - can easily extend for more, but it will just add confusion, IMO.

TLDR: Just for simplicity and practicality.

moritzkiefer-da: I don't love the fact that we're back to being inconsistent between helm and compose which is what this issue started with.

But I also don't love the UX of 30 different 1-letter flags. So probably you're right and the pragmatic option is indeed to keep it as you have it atm.

PR #5606, opened 2026-05-19
Increase timeout and exponential backoff retry

jagathweerasinghe-da:
[static]
Potential fix for: https://github.com/DACH-NY/cn-test-failures/issues/8462


OriolMunoz-da (approved):
worth a try i guess

Issue #5499, opened 2026-05-12
Minimal wallet tx parsing in wallet for transfers v2


OriolMunoz-da:

OriolMunoz-da:
fixed in https://github.com/canton-network/splice/pull/5613

PR #5613, opened 2026-05-19
Add minimal wallet tx parsing in wallet for transfers v2

OriolMunoz-da:
As discussed, targetting [meiersi/ts2/v1-support-for-test-token2](https://github.com/canton-network/splice/tree/meiersi/ts2/v1-support-for-test-token2) to fix v2 allocations with the v1 parser (prev: tx-kind=transfer) 

Fixes https://github.com/canton-network/splice/issues/5499

Differences with the v1 parser:
- `label` is just whatever now, but since we're including the transferlegs maybe we're almost there already? (quote: "I think there's no need to do rich parsing. I think if we can show the raw holding changes we're actually already good.")
- `transferInstruction` is always null, what should we do there in v2?
- These updates are different (v2 vs v1):
``` 
update-13 => 2 events (release lock, send to bob) VS only one with both operations
update-24 => 2 events (create unlocked&archive locked, something that looks like a merge?) vs all in the same event
update-29 => root create of dummy that's just a holdingv1, excluded in v2
update-30 => root create of dummy that's just a holdingv1, excluded in v2
```

Used this AI-written script to get the updates that were different, if you want to use it (adjust file paths as needed):
```
#!/usr/bin/env python3

import json
from pathlib import Path
from typing import Any

LEFT_FILE = Path(
    "/home/oriolmunoz/workspace/splice-ts-v2/token-standard/cli/__tests__/actual/txs-v2.json"
)
RIGHT_FILE = Path(
    "/home/oriolmunoz/workspace/splice-ts-v2/token-standard/cli/__tests__/expected/txs.json"
)


def normalize(value: Any, *, in_events: bool = False) -> Any:
    """Recursively normalize JSON values for comparison."""
    if isinstance(value, dict):
        normalized: dict[str, Any] = {}
        for key in sorted(value.keys()):
            # Ignore selected keys only inside the events subtree.
            if in_events and key in {"label", "transferInstruction"}:
                continue
            child_in_events = in_events or key == "events"
            normalized[key] = normalize(value[key], in_events=child_in_events)
        return normalized

    if isinstance(value, list):
        return [normalize(item, in_events=in_events) for item in value]

    return value


def load_json(path: Path) -> Any:
    with path.open("r", encoding="utf-8") as f:
        return json.load(f)


def tx_index(value: Any) -> dict[str, Any]:
    txs = value.get("transactions", []) if isinstance(value, dict) else []
    return {
        tx["updateId"]: tx
        for tx in txs
        if isinstance(tx, dict) and isinstance(tx.get("updateId"), str)
    }


def main() -> int:
    left_json = normalize(load_json(LEFT_FILE))
    right_json = normalize(load_json(RIGHT_FILE))

    if left_json == right_json:
        print("Equivalent")
        return 0

    print("Different")
    left_txs = tx_index(left_json)
    right_txs = tx_index(right_json)

    if left_txs and right_txs:
        all_ids = sorted(set(left_txs.keys()) | set(right_txs.keys()))
        different_ids = [
            update_id
            for update_id in all_ids
            if left_txs.get(update_id) != right_txs.get(update_id)
        ]
        if different_ids:
            print("Different updateId values:")
            for update_id in different_ids:
                print(update_id)
        else:
            print("No per-transaction difference found by updateId.")

    return 1


if __name__ == "__main__":
    raise SystemExit(main())
```


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/plugins/TokenStandardCliSanityCheckPlugin.scala
OriolMunoz-da, line 110: strict mode is already included

token-standard/cli/src/apis/ledger-api-utils.ts
```diff
@@ -438,3 +438,71 @@ async function promiseWithTimeout<T>(
     }
   }
 }
+
+export async function getEventsOfContract(
```
OriolMunoz-da, line 443: everything here is moved from v1

token-standard/cli/src/commands/listHoldingTransactions.ts
```diff
@@ -50,7 +52,12 @@ export async function listHoldingTransactions(
...
     }
-    const result = await toPrettyTransactions(updates, partyId, ledgerClient);
+    const result = await toPrettyTransactions(
+      updates,
+      partyId,
+      opts.standardVersion,
+      ledgerClient,
+    );
     if (opts.strict) {
       validateStrict(result, opts.strictIgnore || []);
```
OriolMunoz-da, line 62: strict validation runs in both cases

token-standard/cli/src/txparse/parserv1.ts
```diff
@@ -348,6 +351,12 @@ export class TransactionParser {
         return await this.buildBasic(exercisedEvent, "Unlock", null);
       case "expire-dust":
         return await this.buildBasic(exercisedEvent, "ExpireDust", null);
+      case "settle-V2-allocation":
```
OriolMunoz-da, line 354: necessary for V2 test to not fart. This builds it with the same logic as v1, meaning that it assigns all creates and archives as holding changes under this event

token-standard/cli/src/txparse/parserv2.ts
```diff
@@ -0,0 +1,457 @@
...
+            cachedHoldings,
+          );
+          holdingChanges.push({ event: parsed, nodeId: nodeId });
+        }
+      } else {
+        throw new Error(`Impossible event: ${JSON.stringify(event)}`);
+      }
+    }
+
+    const accountedForHoldings = new Set(
```
OriolMunoz-da, line 111: In our 1:1 doc we have:
```
Sanity check should ensure that all tokens creates and archives are explained with the right account
- The set of all input holdings is equal to the set of all archived holdings
- The set of all output holdings is equal to the set of all created holdings
```

correct me if I'm wrong: I think this is handled here, as thanks to the the `--strict` flag any unaccounted create/archive (meaning: not explained by an EventLog_HoldingsChange) will be output and the strict check will fail

token-standard/cli/src/txparse/summary.ts
OriolMunoz-da, line 1: moved from v1

token-standard/cli/src/txparse/types.ts
```diff
@@ -77,8 +81,16 @@ export type Label =
   | Mint
   | Unlock
   | ExpireDust
-  | UnknownAction;
+  | UnknownAction
+  | SettleV2Allocation
+  | V2Label;
 type UnknownAction = RawArchive | RawCreate;
+export interface V2Label {
```
OriolMunoz-da, line 89: see PR description & let's decide what to put in here

token-standard/cli/src/constants.ts
```diff
@@ -43,9 +49,8 @@ export const TransferInstructionInterface = buildInterfaceId(
   "TransferInstruction",
 );
 
-// TODO (#563): include allocations
```
OriolMunoz-da, line 46: leftover

token-standard/cli/src/token-standard-cli.ts
```diff
@@ -25,7 +25,7 @@ export function createProgram(): Command {
       .description("List the holdings of a party")
       .argument("partyId", "The party for which to list the holdings"),
   ).action((partyId, opts) =>
-    listContractsByInterface(HoldingInterface, partyId, opts),
+    listContractsByInterface(HoldingInterfaceV1, partyId, opts),
```
OriolMunoz-da, line 28: we have some other utils (list holding, transfer instructions, accept...) that we didn't discuss, should we add v2 support? (they're pretty much the same, so no at a lot of value for what's meant to be as an example-CLI)

meiersi-da:
> As discussed, targetting [meiersi/ts2/v1-support-for-test-token2](https://github.com/canton-network/splice/tree/meiersi/ts2/v1-support-for-test-token2) to fix v2 allocations with the v1 parser (prev: tx-kind=transfer)
> 
> Fixes #5499
> 
> Differences with the v1 parser:
> 
> * `label` is just whatever now, but since we're including the transferlegs maybe we're almost there already? (quote: "I think there's no need to do rich parsing. I think if we can show the raw holding changes we're actually already good.")
> * `transferInstruction` is always null, what should we do there in v2?
> * These updates are different (v2 vs v1):
> 
> ```
> update-13 => 2 events (release lock, send to bob) VS only one with both operations
> update-24 => 2 events (create unlocked&archive locked, something that looks like a merge?) vs all in the same event
> update-29 => root create of dummy that's just a holdingv1, excluded in v2
> update-30 => root create of dummy that's just a holdingv1, excluded in v2
> ```

Thanks for the summary. I've looked at both update-13 and update-24, and they are as expected. The reason why they are two events now is that the `EventLog_HoldingsChange` events are generated as part of the base Amulet operations (`Amulet_Unlock` and executeTransfer). `TransferInstruction_Accept` and the other token standard choices are built on top of these operations. Thus removing the double-events would require to archive the locked amulet directly, which is a change I didn't want to embark on at this point of time. I'll create a tech-debt issue to do that once things have settled down some more. 

As for the `DummyHolding` that's expected. Let's leave it as is.

token-standard/cli/src/txparse/types.ts
meiersi-da: add:
- admin
- account

both of these are critical to check by downstream consumers; e.g., not checking admin would allow spoofing transfer events.

token-standard/cli/src/token-standard-cli.ts
meiersi-da: nothing to validate there.

docs/src/app_dev/token_standard/index.rst
```diff
@@ -281,7 +281,7 @@ and continue by passing the offset of the last transaction from the previous res
 Parsing the history
 ^^^^^^^^^^^^^^^^^^^
 
-Example code: `the parser here <https://github.com/canton-network/splice/blob/main/token-standard/cli/src/txparse/parser.ts>`_.
+Example code: `the parser here <https://github.com/canton-network/splice/blob/main/token-standard/cli/src/txparse/parserv1.ts>`_.
```
meiersi-da, line 284: please link both the v1 and v2 parsers.

meiersi-da (commented):
Thanks a lot. Looks great.

I reviewed the core of it. Please ping me again when the PR is ready to merge.

docs/src/app_dev/token_standard/index.rst
```diff
@@ -323,6 +323,109 @@ In each Token Standard exercise node, one can find:
...
 
+Parsing the history (V2)
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Example code: `the parser here <https://github.com/canton-network/splice/blob/main/token-standard/cli/src/txparse/parserv2.ts>`_.
+
+As of Token Standard V2, the new ``EventLog`` interface simplifies a lot the implementation of a parser.
+It has the choice ``EventLog_HoldingsChange`` that contains all input and output holdings as arguments,
+and all necessary metadata to be able to construct a transaction history.
+For example, one exercise of this choice could look like:
```
meiersi-da:
```suggestion
As of Token Standard V2, the new ``EventLog`` interface significantly simplifies the implementation of a tx parser.
It has the choice ``EventLog_HoldingsChange`` whose arguments contain all input and output holdings, all transfers that happened, together with meatadata
about the change.
For example, one exercise of this choice could look like:
```

docs/src/app_dev/token_standard/index.rst
```diff
@@ -323,6 +323,109 @@ In each Token Standard exercise node, one can find:
...
+              "otherside" : {
+                "id" : "",
+                "owner" : "aliceValidator::normalized",
+                "provider" : null
+              },
+              "side" : "SenderSide",
+              "transferLegId" : "leg0"
+            },
+            {
+              "amount" : "49.3200000000",
```
meiersi-da: The example is a bit confusing, as it repeats the very same leg twice. Can you pick an allocation settlement example with decent metadata one the transfer leg?

meiersi-da (approved):
Excellent! :shipit:

OriolMunoz-da:
pushing with [static] to fix the docs as CI already passed:
image

PR #5048, opened 2026-04-15
fix: change default log levels and document how to configure them (#721)

baktybekb:
## Summary

Addresses #721 — SVs report high log aggregation costs due to verbose default log levels.

### Changes to defaults

| Setting | Location | Before | After |
|---------|----------|--------|-------|
| `LOG_LEVEL_STDOUT` | Helm template | `DEBUG` | `INFO` |
| `LOG_API_REQUEST_CANTON` | Helm template | `DEBUG` | `INFO` |
| `LOG_LEVEL_STDOUT` | logback.xml | `WARN` | `INFO` |
| `LOG_LEVEL_API_REQUEST` | logback.xml | `DEBUG` | `INFO` |
| `LOG_LEVEL_API_REQUEST` | Docker Compose (SV) | `DEBUG` | `INFO` |

The Helm `LOG_LEVEL_STDOUT=DEBUG` default was the main cost driver for SV operators running in Kubernetes, as all DEBUG-level output was captured by log collectors (CloudWatch, Datadog, etc.).

### Documentation

Added a "Configuring log levels" section to `docs/src/deployment/troubleshooting.rst`:
- Table of all log level environment variables with their defaults
- CLI flags (`--log-level-canton`, `--verbose`, `--debug`)
- Docker Compose and Kubernetes (Helm) configuration examples
- Disclaimer that non-default log levels may limit debugging support
- Fixed incorrect statement that the default log level was Debug

### Files changed
- `cluster/helm/splice-util-lib/templates/_helpers.tpl` — Helm defaults
- `scripts/canton-logback.xml` — Logback defaults
- `cluster/compose/sv/compose.yaml` — Docker Compose defaults
- `docs/src/deployment/troubleshooting.rst` — Documentation

Fixes #721

PR #5648, opened 2026-05-21
Lower default Canton log level to INFO and document overrides

samsondav:
## Summary

Fixes #721.

The splice-node container and the SV / scan / validator Helm charts defaulted Canton classes (`com.daml`, `com.digitalasset`) to `DEBUG`, which SVs have flagged as expensive for log aggregation. The shared `scripts/canton-logback.xml` further conflated the Canton Network application logger (`org.lfdecentralizedtrust.splice`) onto the same `LOG_LEVEL_CANTON` variable, so it was not possible to express the issue's requested config — Canton at `INFO`, CN at `DEBUG` — without a logback change.

This PR makes both halves of that split land in the defaults.

Changes:

- `scripts/canton-logback.xml`: the `org.lfdecentralizedtrust.splice` logger now reads from a separate `LOG_LEVEL_CN` env var, defaulting to `DEBUG`. `com.daml` / `com.digitalasset` stay on `LOG_LEVEL_CANTON` (with the existing `INFO` fallback in this file).
- `cluster/images/common/entrypoint.sh`: `LOG_LEVEL_CANTON` default `DEBUG` → `INFO`; `LOG_LEVEL_STDOUT` default `DEBUG` unchanged.
- `cluster/helm/splice-util-lib/templates/_helpers.tpl`: emit `LOG_LEVEL_CN` alongside `LOG_LEVEL_CANTON` / `LOG_LEVEL_STDOUT`, value coming from a chart-side `logLevelCn` (default `DEBUG`).
- `cluster/helm/splice-{sv-node,scan,validator}/values-template.yaml`: `logLevel: DEBUG` → `INFO`. The helper already defaults this to `INFO` when omitted; this aligns the chart-bundled defaults with that.
- `docs/src/deployment/troubleshooting.rst`: rewrite the stale Debug-by-default note. Document the three env vars (`LOG_LEVEL_CANTON`, `LOG_LEVEL_CN`, `LOG_LEVEL_STDOUT`) and corresponding chart values (`logLevel`, `logLevelCn`, `logLevelStdout`). Include the disclaimer that debugging support is best-effort under non-default levels.
- `docs/src/faq.rst`: the `TrafficStateController` traffic-cost troubleshooting step previously assumed `DEBUG` was on for Canton classes by default. Update step 1 to instruct setting `LOG_LEVEL_CANTON=DEBUG` on the participant container before reproducing.

Effective container defaults after this PR:

| Class group | Logger root | Env var | Default |
|---|---|---|---|
| Canton | `com.daml`, `com.digitalasset` | `LOG_LEVEL_CANTON` | `INFO` |
| Canton Network app | `org.lfdecentralizedtrust.splice` | `LOG_LEVEL_CN` | `DEBUG` |
| Stdout sink filter | (appender) | `LOG_LEVEL_STDOUT` | `DEBUG` |

Out of scope (per the maintainer comment on the issue):

- General documentation of arbitrary Canton config overrides via `additionalEnvVars` / `ADDITIONAL_CONFIG_` — left for a separate issue if maintainers want it.
- Docker-compose paths (`cluster/compose/{sv,validator}/compose.yaml`) — already default to `INFO` via `LOG_LEVEL:-INFO`. They don't currently set `LOG_LEVEL_CN`, so the logback default `DEBUG` applies, matching the new helm defaults.
- Multi-validator pulumi path — already defaults to `INFO` via `cluster/pulumi/multi-validator/src/config.ts`; with the logback change, splice loggers pick up the `LOG_LEVEL_CN` default `DEBUG` from the same file.
- `splice-global-domain` chart — already inherits the helper's `INFO` default (no `logLevel:` in `values-template.yaml`); the helper now also injects `LOG_LEVEL_CN`. The existing `splice-global-domain/tests/sequencer_test.yaml` tests use `contains` assertions on `LOG_LEVEL_CANTON` and `LOG_LEVEL_STDOUT` and are unaffected.
- `splice-cometbft` chart — uses cometbft's own `logLevel: info` semantic (separate subsystem).

Note: the `docs/src/` Sphinx site was just deprecated (#5644) in favor of `docs.canton.network`. The notes have been updated in the existing location since that is where today's operators land; happy to mirror the override paragraph into cf-docs in a follow-up if maintainers prefer.

## Test plan

- [x] `shellcheck cluster/images/common/entrypoint.sh` — clean (only pre-existing SC1091 infos for runtime-mounted `tools.sh` / `pre-bootstrap.sh`).
- [x] Reviewed the helper at `cluster/helm/splice-util-lib/templates/_helpers.tpl` and the consumer charts (`splice-{sv-node,scan,validator,participant,splitwell-app,global-domain}/templates/*.yaml`) — all rely on the helper for `LOG_LEVEL_*` injection, so the new `LOG_LEVEL_CN` reaches every pod.
- [x] Verified `cluster/helm/splice-global-domain/tests/sequencer_test.yaml` uses `contains` assertions (not exact-list), so the added env var does not break the existing log-level wiring tests.
- [x] Verified `cluster/expected/multi-validator/expected.json` snapshots use the multi-validator pulumi path (`cluster/pulumi/multi-validator/src/multiValidator.ts`), not `splice-util-lib.log-level`, so the new env var does not affect those snapshots.
- [ ] CI: helm chart tests — let CI confirm.
- [ ] CI: docs build — let CI confirm the reST renders cleanly.

### DIFF
--- a/cluster/helm/splice-scan/values-template.yaml
+++ b/cluster/helm/splice-scan/values-template.yaml
@@ -64,7 +64,7 @@ spliceInstanceNames:
 service:
   ui:
     port: 80
-logLevel: DEBUG
+logLevel: INFO
 
 # bulk storage configuration (still in dev, please do not enable in prod)
 # bulk-storage:
--- a/cluster/helm/splice-sv-node/values-template.yaml
+++ b/cluster/helm/splice-sv-node/values-template.yaml
@@ -68,7 +68,7 @@ spliceInstanceNames:
 service:
   ui:
     port: 80
-logLevel: DEBUG
+logLevel: INFO
 
 # should stakater reloader annotation be added (note: reloader needs to be installed separately)
 enableReloader: true
--- a/cluster/helm/splice-util-lib/templates/_helpers.tpl
+++ b/cluster/helm/splice-util-lib/templates/_helpers.tpl
@@ -203,6 +203,8 @@ spec:
   value: {{ .logLevel | default "INFO" }}
 - name: LOG_API_REQUEST_CANTON
   value: {{ .apiRequestLogLevel | default "DEBUG" }}
+- name: LOG_LEVEL_CN
+  value: {{ .logLevelCn | default "DEBUG" }}
 - name: LOG_LEVEL_STDOUT
   value: {{ .logLevelStdout | default "DEBUG" }}
 - name: LOG_IMMEDIATE_FLUSH
--- a/cluster/helm/splice-validator/values-template.yaml
+++ b/cluster/helm/splice-validator/values-template.yaml
@@ -85,7 +85,7 @@ validatorWebUi:
 
 # set to true to disable auth (this is highly insecure)
 disableAuth: false
-logLevel: DEBUG
+logLevel: INFO
 
 # should stakater reloader annotation be added (note: reloader needs to be installed separately)
 enableReloader: true
--- a/cluster/images/common/entrypoint.sh
+++ b/cluster/images/common/entrypoint.sh
@@ -16,7 +16,7 @@ if [[ ! " ${*} " =~ " --console " ]]; then
     ARGS+=( daemon --no-tty )
 fi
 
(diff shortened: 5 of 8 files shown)


martinflorian-da:
FYI in case you missed this discussion: https://github.com/canton-network/splice/pull/5048#pullrequestreview-4170729875

samsondav:
Thanks for the pointer to #5048's review — agreed that the issue is stale and that touching defaults isn't the right call. Closing this in favor of picking up a `now`-labeled ticket next.

PR #5649, opened 2026-05-21
Fix the target before eventual ensure

jagathweerasinghe-da:
Potential fix for: https://github.com/DACH-NY/cn-test-failures/issues/8563


jagathweerasinghe-da:
@moritzkiefer-da Not much confidence, but is this a potential fix for https://github.com/DACH-NY/cn-test-failures/issues/8563?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanWithGradualStartsTimeBasedIntegrationTest.scala
```diff
@@ -88,11 +88,20 @@ class ScanWithGradualStartsTimeBasedIntegrationTest
           }
         }
         clue("Ensure ValidatorLivenessActivityRecord are received") {
+          val targetRound =
```
moritzkiefer-da, line 91: Can you explain why you think this solves the issue? or maybe taking a step back what is the concrete behavior you observed when it fails?

PR #5652, opened 2026-05-21
Resolve binaries in parent shell before tmux send-keys

### LINKED ISSUE #458: Make `start-canton.sh` and `start-frontends.sh` robust to `tmux` having been launched elsewhere
[ This issue was auto-migrated from DA's internal repo (DACH-NY/canton-network-node#19801). Original author: @martinflorian-da ]

## What is this about?

Context: https://github.com/DACH-NY/canton-network-node/pull/19787#discussion_r2125941247

If you start tmux in nix environment 1, and then run `start-x.sh` in nix environment 2, then it will use the binaries from environment 1 which is very confusing.

Ideas for fixing:

- do a `which command` and execute the result in the tmux instead of just `command`
- do a `nix develop path:nix` in every new tmux pane, like what was done in `start-frontends.sh` prior to it being removed in https://github.com/DACH-NY/canton-network-node/pull/19787 ; note that on the nix upgrade attempt there this made things slower to the point of actually failing to start on CI

## How important is this and why?

@moritzkiefer-da is unhappy without this and to be fair it seems easy to fix, while the issue seems like it can easily lead to difficult debugging sessions

samsondav:
Fixes #458.

## Problem

If tmux is launched from one nix env and `start-canton.sh` or `start-frontends.sh` is later re-run from another nix env, the panes pick up the binary from the tmux server's inherited PATH (the first env) rather than the active env. The failure is silent and produces confusing debugging sessions, as flagged by @moritzkiefer-da on DACH-NY/canton-network-node#19787.

## Fix

Resolve `canton`, `toxiproxy-server`, and `npm` via `command -v` in the parent shell and substitute the absolute path into the tmux command string. The pane invokes the binary directly regardless of the tmux server's PATH.

If the binary is missing on PATH, fail fast in the parent shell with a clear message before any tmux session is created, instead of failing silently inside a pane.

This matches the first idea from the issue body (resolve via `which`, execute the result). The second idea (`nix develop path:nix` in every pane) is not used here because the issue notes a prior attempt made CI startup measurably slower.

## Scope

- `start-canton.sh`: `$CANTON` (default `canton`, overridable via `-c`) and `toxiproxy-server`.
- `start-frontends.sh`: `npm` (used by `start_frontend`, `start_test`, and `wait_for_workspace_build` panes).

Out of scope: `tee` and other coreutils — these were not the reported failure mode and are typically identical across nix envs. `sbt` and `jq`/`jsonnet` are invoked from the parent shell, not from a tmux pane, so they already pick up the active env's PATH.

## Behavior

- Default invocation (single nix env): no behavior change. `command -v canton` returns the same absolute path the pane would have resolved through PATH.
- Mixed nix envs (the bug): pane runs the binary the user's current shell sees, not the tmux server's first-env binary.
- Missing binary: script exits with `start-canton.sh: 'canton' not found on PATH` (or similar) before any tmux session is created.
- `-c <custom-canton>` with an absolute path: passes through unchanged via `command -v`.

### DIFF
--- a/start-canton.sh
+++ b/start-canton.sh
@@ -107,6 +107,16 @@ do
     shift
 done
 
+# Resolve binaries in the parent shell so the panes execute the binary from
+# the active environment's PATH. Without this, if tmux is launched from one
+# nix env and start-canton.sh is then re-run from another, the panes pick up
+# the binary from the tmux server's inherited PATH (the first env), which is
+# silent and confusing. See https://github.com/canton-network/splice/issues/458.
+# Assumes the binaries are real executables on PATH; aliases and shell
+# functions would resolve to strings that are not usable as a tmux command.
+canton_input=$CANTON
+CANTON=$(command -v "$canton_input") || { echo "start-canton.sh: '$canton_input' not found on PATH" >&2; exit 1; }
+
 tmux_session="canton"
 tmux_window=0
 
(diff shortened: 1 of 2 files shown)


martinflorian-da:
Thanks for your contribution but I think I'm going to refer to our https://github.com/DACH-NY/ai-tools/blob/main/AI_POLICY.md again and close this PR.

Mostly:

- because this looks like unreviewed AI output (see https://github.com/DACH-NY/ai-tools/blob/main/AI_POLICY.md#guard-reviewers-time) I don't trust that anyone actually confirmed this does the right things; so I'd need to this testing myself, which is about as much effort as redoing the work from scratch
- https://github.com/DACH-NY/ai-tools/blob/main/AI_POLICY.md#a-note-for-new-contributors

PR #5668, opened 2026-05-22
enable hotreload for logback configuration

### LINKED ISSUE #5622: Read logback from configmap and enable hot reload
Currently we can't really change log level without a restart which means if a restart already fixes your issue that doesn't help.

Canton has an rpc endpoint but it's just the global log level so you can't use it to change e.g. netty logging.

If the logback is in a config map and has logback hot reload enabled this works much nicer.

mblaze-da:
Fixes https://github.com/canton-network/splice/issues/5622

Cluster test: https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/66732

Please LMK if this somehow doesn't address your needs for editing the logging config.

### DIFF
--- a/cluster/helm/splice-global-domain/templates/mediator.yaml
+++ b/cluster/helm/splice-global-domain/templates/mediator.yaml
@@ -202,7 +202,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 data:
   logback.xml: |
-{{ .Files.Get "files/canton-logback.xml" | indent 4 }}
+{{ .Files.Get "files/logback.xml" | indent 4 }}
 ---
 {{- if .Values.enablePostgresMetrics }}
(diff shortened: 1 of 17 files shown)


cluster/helm/splice-global-domain/templates/mediator.yaml
```diff
@@ -100,8 +100,13 @@ spec:
           failureThreshold: 3
           timeoutSeconds: 10
         {{- end }}
-        {{- if .Values.pvc }}
         volumeMounts:
+          - name: logback-config
+            # Must mount the whole configmap as a directory instead of shadowing /app/logback.xml with subPath as the latter
```
moritzkiefer-da, line 105: til thx for documenting it

moritzkiefer-da (approved):
Nice thanks! I assume you tested that the hot reload works?

mblaze-da:
> Nice thanks! I assume you tested that the hot reload works?

Yes, I changed root log level to DEBUG and observed a ton of DEBUG logs shortly after.

mblaze-da:
@moritzkiefer-da there is no point in backporting it, right? AFAIU helm charts are only released when cutting a release.

moritzkiefer-da:
yep

PR #5677, opened 2026-05-26
Fix SV reward weight persistence in Vote History

### LINKED ISSUE #5092: Update SV Reward Weight proposal don't persist base weight
The `DsoRules_UpdateSvRewardWeight` contract doesn't persist a `baseRewardWeight`, unlike the DSO Rules and Amulet Rules config-change proposals. As a result, we have no way to display the "before" value after the proposal has been completed and the UI falls back to the current SV weight in DSO Rules, which by that point already equals the new value, so the diff renders as `Weight X -> X`. See: https://github.com/hyperledger-labs/splice/blob/4cf5ff84ddca2cf534e64db49e1d798bc2056f4e/daml/splice-dso-governance/daml/Splice/DsoRules.daml#L972-L975

Inflight proposal:

Image

Vote History:

Image

pawelperek-da:
Fixes #5092 

Set DSO Rules -> Set SV Weight:
image

Set SV Weight -> Set SV Weight (created on main):
image

Set SV Weight -> Set SV Weight (created on branch):
image

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V069__sv_reward_weight_history.sql
```diff
@@ -0,0 +1,17 @@
...
+alter table scan_txlog_store
+    add column sv_onboarding_effective_at bigint;
+
+alter table scan_txlog_store
+    add column vote_sv_party text;
+
+-- Supporting indices are created concurrently by SqlIndexInitializationTrigger
+-- to avoid blocking ingestion on large scan_txlog_store tables.
+
+-- Backfill vote_sv_party for vote-result rows ingested before this migration.
```
pawelperek-da: Backfill query plan on CILR:
```
Update on scan_txlog_store (cost=0.12..35934.08 rows=0 width=0) (actual time=0.053..0.053 rows=0 loops=1)
--
Buffers: shared hit=45
-> Index Scan using scan_txlog_store_sid_en_vot on scan_txlog_store (cost=0.12..35934.08 rows=1 width=38) (actual time=0.052..0.052 rows=0 loops=1)
Filter: ((vote_sv_party IS NULL) AND (entry_type = 'vot'::text) AND (vote_action_name = 'SRARC_UpdateSvRewardWeight'::text))
Rows Removed by Filter: 44
Buffers: shared hit=45
Planning:
Buffers: shared hit=257
Planning Time: 4.099 ms
Execution Time: 0.128 ms
```

pawelperek-da: No longer valid

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
```diff
@@ -848,6 +849,49 @@ class DbScanStore(
...
+      DsoRules.COMPANION,
+      updateHistory,
+      CantonTimestamp.assertFromInstant(before),
+    ).map(
+      _.flatMap(dsoRules =>
+        Option(dsoRules.payload.svs.get(svParty.toProtoPrimitive)).map(_.svRewardWeight.toLong)
+      )
+    )
+
+  private def lookupContractByRecordTimeBefore[C, TCId <: ContractId[?], T](
```
pawelperek-da: This is an adapted version of `lookupContractByRecordTime` below, just ordered by `row_id DESC` instead of `ASC`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
```diff
@@ -848,6 +849,49 @@ class DbScanStore(
...
+      tc: TraceContext,
+  ): Future[Option[Contract[TCId, T]]] = {
+    val pqn @ PackageQualifiedName(packageName, QualifiedName(moduleName, entityName)) =
+      companionClass.packageQualifiedName(companion)
+    for {
+      row <- storage
+        .querySingle(
+          selectFromUpdateCreatesTableResult(
+            updateHistory.historyId,
+            where = sql"""template_id_module_name = ${lengthLimited(moduleName)}
```
pawelperek-da: Query plan on CILR:
```
Limit (cost=0.70..1.42 rows=1 width=1275) (actual time=2.230..2.232 rows=1 loops=1)
  Buffers: shared hit=2 read=4
  I/O Timings: read=2.111
  -> Index Scan Backward using updt_hist_crea_hi_tidmn_tiden_pn_rid on update_history_creates (cost=0.70..10.80 rows=14
  width=1275) (actual time=2.228..2.229 rows=1 loops=1)
  Index Cond: ((history_id = 1) AND (template_id_module_name = 'Splice.DsoRules'::text) AND (template_id_entity_name =
  'DsoRules'::text) AND (package_name = 'splice-dso-governance'::text))
  Filter: (record_time < ((EXTRACT(epoch FROM now()) * '1000000'::numeric))::bigint)
  Buffers: shared hit=2 read=4
  I/O Timings: read=2.111
  Planning:
  Buffers: shared hit=267 read=47
```

ray-roestenburg-da:
As discussed, maybe nice to expose a lookupContractByRecordTime API which can be used for the same purpose with some arguments

apps/sv/frontend/src/routes/voteRequestDetails.tsx
```diff
@@ -74,7 +74,8 @@ export const VoteRequestDetails: React.FC = () => {
 
   const action = amuletOrDsoAction.tag as SupportedActionTag;
   const actionName = actionTagToTitle(amuletName)[action];
-  const createdAt = voteRequest ? dayjs(voteRequest.createdAt).format(dateTimeFormatISO) : '';
+  const proposalTime = voteRequest?.createdAt ?? voteResult?.completedAt;
```
rautenrieth-da, line 77: How does the user know if the UI displays the request or result time?

apps/sv/frontend/src/components/governance/ProposalDetailsContent.tsx
```diff
@@ -653,13 +656,20 @@ interface UpdateSvRewardWeightSectionProps {
...
 const UpdateSvRewardWeightSection = ({
   svToUpdate,
   currentWeight,
   weightChange,
+  before,
 }: UpdateSvRewardWeightSectionProps) => {
+  const { data: priorWeight } = useSvRewardWeightBefore(svToUpdate, before);
+
+  const beforeValue =
+    priorWeight != null ? formatBasisPoints(priorWeight.toString()) : currentWeight;
```
rautenrieth-da: AFAICT, before the query completes, the UI will display `currentWeight`. As soon as it completes, it will jump to `priorWeight`. Isn't this confusing? Should it not display a placeholder while the query is loading?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
rautenrieth-da: Ordering by `row_id` does not necessarily give you results ordered by record time. Because of backfilling, rows are not inserted in record time order.

The result is that lookupContractByRecordTimeBefore will return a random contract created before `before`, not the latest one.

rautenrieth-da (commented):
If I understand correctly, our problem is that `DsoRules_UpdateSvRewardWeight` doesn't store the old value. Is this the only vote affected? Wouldn't it be more future proof to add a `lookupDsoRulesBefore` instead of a `lookupSvRewardWeightBefore`, and then let the UI extract the weight from the rules?

apps/sv/frontend/src/routes/voteRequestDetails.tsx
pawelperek-da: UI always displays result time when the vote passed and request time otherwise, it's an established pattern

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -255,6 +255,14 @@ object SqlIndexInitializationTrigger {
           where entry_type = 'vot'
         """,
       ),
+    IndexAction
+      .Create(
+        indexName = "updt_hist_crea_hi_tidmn_tiden_pn_rt_rid",
+        createAction = sqlu"""
+          create index concurrently if not exists updt_hist_crea_hi_tidmn_tiden_pn_rt_rid
+          on update_history_creates (history_id, template_id_module_name, template_id_entity_name, package_name, record_time, row_id)
```
ray-roestenburg-da: Do you really need this index, it seems very expensive.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
```diff
@@ -739,29 +739,34 @@ class DbScanStore(
...
+    // Literal comparison/sort order since Postgres complains when trying to bind it to a parameter.
+    val (compareRecordTime, orderLimit, spanLabel) = sortOrder match {
+      case SortOrder.Ascending =>
+        (sql" > ", sql"order by record_time asc, row_id asc limit 1", "lookup")
+      case SortOrder.Descending =>
+        (sql" < ", sql"order by record_time desc, row_id desc limit 1", "lookupBefore")
+    }
     for {
       row <- storage
         .querySingle(
```
pawelperek-da, line 764: Query plans on CILR:
1. Cold cache
```
Limit (cost=9.15..9.15 rows=1 width=1275) (actual time=500.644..500.646 rows=1 loops=1)
--
Buffers: shared hit=8 read=653
I/O Timings: read=489.509
-> Sort (cost=9.15..9.15 rows=2 width=1275) (actual time=500.643..500.644 rows=1 loops=1)
Sort Key: record_time DESC, row_id DESC
Sort Method: top-N heapsort Memory: 28kB
Buffers: shared hit=8 read=653
I/O Timings: read=489.509
-> Index Scan using updt_hist_crea_hi_tidmn_tiden_pn_rid on update_history_creates (cost=0.70..9.14 rows=2 width=1275) (actual time=1.359..499.612 rows=186 loops=1)
Index Cond: ((history_id = 1) AND (template_id_module_name = 'Splice.DsoRules'::text) AND (template_id_entity_name = 'DsoRules'::text) AND (package_name = 'splice-dso-governance'::text))
Filter: (record_time < '1735689600000000'::bigint)
Rows Removed by Filter: 457
Buffers: shared hit=5 read=653
I/O Timings: read=489.509
Planning:
Buffers: shared hit=320
Planning Time: 6.643 ms
Execution Time: 500.729 ms
```

2. Warm cache:
```
Limit (cost=9.15..9.15 rows=1 width=1275) (actual time=1.993..1.994 rows=1 loops=1)
--
Buffers: shared hit=661
-> Sort (cost=9.15..9.15 rows=2 width=1275) (actual time=1.992..1.993 rows=1 loops=1)
Sort Key: record_time DESC, row_id DESC
Sort Method: top-N heapsort Memory: 28kB
Buffers: shared hit=661
-> Index Scan using updt_hist_crea_hi_tidmn_tiden_pn_rid on update_history_creates (cost=0.70..9.14 rows=2 width=1275) (actual time=0.051..1.892 rows=186 loops=1)
Index Cond: ((history_id = 1) AND (template_id_module_name = 'Splice.DsoRules'::text) AND (template_id_entity_name = 'DsoRules'::text) AND (package_name = 'splice-dso-governance'::text))
Filter: (record_time < '1735689600000000'::bigint)
Rows Removed by Filter: 457
Buffers: shared hit=658
Planning:
Buffers: shared hit=320
Planning Time: 5.870 ms
Execution Time: 2.063 ms
```

ray-roestenburg-da: It's not as efficient as it could be, but better than dropping existing index and creating a new one for all creates that includes the `record_time`, it's a query only from SV app in certain cases, so I think it's good enough

docs/src/release_notes_upcoming.rst
```diff
@@ -114,3 +114,5 @@
         is listed in the ``ignored-amulet-versions`` configuration and prior to the minimal supported version.
         Parties from skipped batches are added to an in-memory ignore list. This allows sv-1 to better handle
         participants that have not yet vetted the latest amulet package and avoid repeated failures.
+
+      - Fixed "Set SV Weight" base weight display in Vote History
```
ray-roestenburg-da:
```suggestion
      - Fixed "Set SV Weight" base weight display in Vote History.
```

ray-roestenburg-da (approved):
Thanks!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
OriolMunoz-da: > Index Cond: ((history_id = 1) AND (template_id_module_name = 'Splice.DsoRules'::text) AND (template_id_entity_name = 'DsoRules'::text) AND (package_name = 'splice-dso-governance'::text))

filtering by DsoRules is highly selective (at least compared to amulet). Pawel has confirmed this is not in a hot path, so I don't have any big objection with this query.

caveat that @ray-roestenburg-da can answer better: how does this work with pruning? For a record time enough in the past, results might be pruned.

ray-roestenburg-da: Thanks @OriolMunoz-da. For Pruning votes are treated differently, there will be a separate VotesStore synced between Svs.

OriolMunoz-da: so will we then inline the older DSORules that we're looking for in this query? (avoiding querying update_history_creates if not necessary is the ideal)

moritzkiefer-da: > Thanks @OriolMunoz-da. For Pruning votes are treated differently, there will be a separate VotesStore synced between Svs.

but this queries the update history which we will prune, shouldn't it query the vote txlog instead?

ray-roestenburg-da: Very good point, yes. The votesstore will have to contain this. The design will have to consider any query to update_history like this, I'll keep it in mind.

ray-roestenburg-da: The votes store is basically the vote txlog. (and we'll have to find a way for the vote txlog to have this information)

moritzkiefer-da: Can we make this a prerequisite for this PR? Introducing changes that will make pruning harder seems like a bad idea given that we'll hit a very hard deadline for pruning. So I'd rather push more work on this PR to get it right from the start.

ray-roestenburg-da: It does raise a good question though, how that will actually work in the new design, I'll add it there 👍

ray-roestenburg-da: I agree, let's make sure we do this right the first time @pawelperek-da.

ray-roestenburg-da (changes requested):
Needs change, a txlog entry needs to be added for dso rules.

PR #5680, opened 2026-05-26
0.6.5-snapshot.20260526.2931.0.ve28a6722 in token standard v2 devnet readme

OriolMunoz-da:
https://token-std-v2-dev.global.canton.network.digitalasset.com/version


meiersi-da (approved):
ty

PR #5681, opened 2026-05-26
retire round aggregates from Scan

isegall-da:
I really tried cutting this into smaller steps that one can reasonably review, but no matter which side I started from, I ended up having to rip it all out.
Please LMK if you think I missed something.

Part of https://github.com/canton-network/splice/issues/4520

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
```diff
@@ -169,55 +163,12 @@ class DbScanStore(
   ] = new DbScanTxLogStoreConfig(loggerFactory)
 
   override protected def closeAsync(): Seq[AsyncOrSyncCloseable] = {
-    implicit def traceContext: TraceContext = TraceContext.empty
     Seq(
-      AsyncCloseable(
-        "db_scan_store",
```
isegall-da, line 175: This is called "db_scan_store" so I was a bit worried about removing this, but AFAICT, that's just a bad name, and it's actually just the aggregator.

ray-roestenburg-da: there were no syncclosables here before, so at that point it was the closeable for the enclosing class

ray-roestenburg-da:
you can probably add a flyway migration to drop the `round_totals` and `round_party_totals` tables (I did not see it in the PR)

ray-roestenburg-da:
SpliceDbTest still refers to round_* tables

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -1450,49 +1379,6 @@ paths:
...
-        "200":
-          description: ok
-          content:
-            application/json:
-              schema:
-                $ref: "#/components/schemas/GetRewardsCollectedResponse"
-        "404":
-          $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/404"
-
   /v0/activities:
```
ray-roestenburg-da, line 1382: Should we also just delete these? they were only for the Scan UI.

ray-roestenburg-da (approved):
Thanks! two comments to consider re: drop tables and updating the SpliceDbTest to no longer reset those tables. and of course why not also just get rid of activities.

apps/scan/src/main/openapi/scan.yaml
isegall-da: yes, coming in a separate PR. Tried to still not bloat this beyond necessary.

isegall-da:
Thanks @ray-roestenburg-da. I think I won't actually delete the actual data just yet, maybe wait a bit to see it doesn't blow up in some strange way in prod (it's not huge amounts of data), but at least the test I don't see a reason not to. I'll track deleting the data in the issue.

PR #5545, opened 2026-05-14
Remove offline_access scope

pawelperek-da:
Fixes #5608 

Examples of 60s long access token:
- SV:

https://github.com/user-attachments/assets/ee754eaa-5c09-4728-b002-980526dd2f88

- Wallet:

https://github.com/user-attachments/assets/f7c4cbad-6f91-44b8-baf9-cb85fbe8beb7

- ANS:

https://github.com/user-attachments/assets/8234322e-f327-4e7f-bb18-7c77c50173cf

- Splitwell:

https://github.com/user-attachments/assets/e2412aa8-d611-4bc6-a754-2a7a1ca92332

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a logical synchronizer upgrade test is required (from canton-3.5), comment `/lsu_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/canton-network/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/canton-network/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #5683, opened 2026-05-26
Restore refresh-token silent renew in frontend OIDC auth

### LINKED ISSUE #5682: Starting from 0.6.5 all UIs require re-login after access token expires
**Issue**

**Keycloak**

- Session Idle: 30 minutes  
- Session Max: 10 hours  
- Offline Session Max: none  

**Suggested solution**

samsondav:
PR #5545 set `automaticSilentRenew={false}` on the `react-oidc-context` provider and removed `offline_access` from the scope request. With both in place the UIs no longer use any refresh token, so users get bounced to the IdP every time the access token expires, which is every 5 minutes on a default Keycloak deployment.

`oidc-client-ts` uses the `refresh_token` grant for silent renewal when the stored `User` has a refresh token. That works against regular session-bound refresh tokens and does not require `offline_access`.

## Changes

- Re-enable `automaticSilentRenew` on `OidcAuthProvider`.
- Add **required** `enable_offline_scope: boolean` to the rs256 `AuthConfig`. Each deployment picks per-IdP:
  - **Auth0**: `true`. Auth0 only issues a refresh token when `offline_access` is requested.
  - **Keycloak**: `false`. Keycloak issues a session-bound refresh token without `offline_access`. Requesting it on Keycloak produces a long-lived offline refresh token, which is a larger XSS surface in a browser-based SPA.
- Wire `enable_offline_scope` through the four shipped web-ui image templates (`cluster/images/{sv,ans,splitwell,wallet}-web-ui/`) via a new `SPLICE_APP_UI_AUTH_ENABLE_OFFLINE_SCOPE` env var. The three images that support `unsafe` mode (sv, ans, wallet) strip the field with `sed` before envsubst on hs256-unsafe builds, since the strict schema would otherwise reject it.
- Update `apps/app/src/test/resources/frontend-config.jsonnet` to set `enable_offline_scope: true` (integration tests target Auth0).
- Keep the 401 → `fireAuthExpired` fallback. It is still the correct path when renewal fails (session-idle / session-max reached upstream, refresh token revoked).

## Breaking schema change

`enable_offline_scope` has no default. Every existing rs256 `config.js` will fail to parse at app load with a zod error pointing at the missing field. This is deliberate, and matches what Pawel and Stas agreed in the Slack thread on this PR: a silent default produces a broken Auth0 deployment, so the Auth0-vs-Keycloak decision should be a configuration-time error rather than a runtime expiry.

Operators upgrading from 0.6.4 or earlier (which globally requested `offline_access`) likely want `enable_offline_scope: true` to preserve the prior scope. Operators on 0.6.5 should pick per their IdP.

**Operators deploying the shipped images MUST set `SPLICE_APP_UI_AUTH_ENABLE_OFFLINE_SCOPE` to `true` or `false`.** If unset, `envsubst` substitutes empty string, producing `enable_offline_scope: ,` in the generated `config.js`, which fails the browser's JS parse at app load.

## Testing

New unit tests under `apps/common/frontend/src/__tests__/auth/`:

- `AuthProvider.test.tsx`: asserts `OidcAuthProvider` receives `automaticSilentRenew: true`. Regression guard for the exact bug.
- `authSchema.test.ts`: covers mandatory `enable_offline_scope` (true / false / missing-rejected / string-rejected / numeric-rejected), strict-unknown-key rejection, and rejection on hs256-unsafe configs.
- `oidcAuthToProviderProps.test.ts`: scope construction with and without the field, the `daml_ledger_api` no-openid carve-out, audience plumbing, and no-leak of the field into provider props.

Existing 401-fallback coverage in `BaseApiMiddleware.test.ts` and `LedgerApiClient.test.ts` is untouched.

## Known limitation

A silent-renew vs concurrent-401 race exists: if a token expires after a request is sent but before its response, the 401 fires `fireAuthExpired` even though `oidc-client-ts` may simultaneously be refreshing in the background. This was dormant before this PR (silent renew was disabled) and is now reachable. PR #5700 is an alternative design that closes this race by attempting a single-flight refresh before signing out on 401.

Fixes #5682


martinflorian-da:
This PR looks stale and content-wise qualifies for https://github.com/canton-network/splice/blob/main/AI_POLICY.md#guard-reviewers-time IMO.

PR #5685, opened 2026-05-27
Traffic-Based App Reward: Support bootstrapping with TBA enabled

### LINKED ISSUE #5624: Support bootstrapping network with traffic-based rewards enabled
**Note:** This one focuses only on adding a single integration test that shows that a network can be bootstrapped with TBAR enabled using a four SV topology.

We need add support for computing round 0 activity totals
Ref [this discussion](https://github.com/canton-network/splice/pull/5552#discussion_r3265723328) about the need for this support for TestNet reset and use of traffic-based rewards as default in integration tests.

adetokunbo:
Fixes #5624

Assumes the following to detect bootstrapping:
- the activity meta table is initially empty if bootstrapping
- so if there is any existing row, then it's not bootstrapping

#### Cluster Testing

#### PR Guidelines
- [] Include a screenshot for frontend-related PRs - see [README](https://github.com/canton-network/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
@@ -114,7 +114,10 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
       // oldest=5: rounds 5,6,7 open. R8 will have
       // both dryRunVersion and mintingVersion set.
       clue("vote to enable dryRunVersion + mintingVersion") {
-        changeRewardConfig(enableDryRun = true, enableMinting = true)
+        changeRewardConfig(
+          enableDryRun = true,
+          rewardVersion = RewardVersion.REWARDVERSION_TRAFFICBASEDAPPREWARDS,
+        )
       }
 
       val svBackends = Seq(sv1Backend, sv2Backend, sv3Backend, sv4Backend)
@@ -409,14 +412,13 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
 
   private def changeRewardConfig(
       enableDryRun: Boolean,
-      enableMinting: Boolean = false,
+      rewardVersion: RewardVersion = RewardVersion.REWARDVERSION_FEATUREDAPPMARKERS,
   )(implicit env: SpliceTestConsoleEnvironment): Unit = {
     val amuletRules = sv1Backend.getDsoInfo().amuletRules
     val existing = AmuletConfigSchedule(amuletRules).getConfigAsOf(env.environment.clock.now)
     val rc = existing.rewardConfig.get()
     val newRc = new RewardConfig(
-      if (enableMinting) RewardVersion.REWARDVERSION_TRAFFICBASEDAPPREWARDS
-      else RewardVersion.REWARDVERSION_FEATUREDAPPMARKERS,
+      rewardVersion,
       if (enableDryRun) Optional.of(RewardVersion.REWARDVERSION_TRAFFICBASEDAPPREWARDS)
       else Optional.empty[RewardVersion](),
       rc.batchSize,
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
@@ -67,6 +67,7 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
       .addConfigTransform((_, config) =>
         ConfigTransforms.withRewardConfig(
           InitialRewardConfig(
+            mintingVersion = "RewardVersion_FeaturedAppMarkers",
             dryRunVersion = None,
             appRewardCouponThreshold = BigDecimal("0"),
           )
@@ -415,7 +416,7 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
     val rc = existing.rewardConfig.get()
     val newRc = new RewardConfig(
       if (enableMinting) RewardVersion.REWARDVERSION_TRAFFICBASEDAPPREWARDS
-      else rc.mintingVersion,
+      else RewardVersion.REWARDVERSION_FEATUREDAPPMARKERS,
       if (enableDryRun) Optional.of(RewardVersion.REWARDVERSION_TRAFFICBASEDAPPREWARDS)
       else Optional.empty[RewardVersion](),
       rc.batchSize,
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
@@ -272,6 +272,27 @@ abstract class TrafficBasedRewardsTimeBasedIntegrationTestBase
                   s"CalculateRewardsV2 should exist for round $round"
               }
             }
+            // Test the archiveDryRunRewardAccountingContracts admin API
+            // by archiving a few early rounds. The remaining rounds are
+            // consumed by triggers naturally.
+            if (dryRunEnabled) {
+              clue("Archive dry-run CalculateRewardsV2 for rounds 0..2 via sv admin API") {
+                sv1Backend.archiveDryRunRewardAccountingContracts((0L to 2L).toSeq)
+              }
+              clue("CalculateRewardsV2 contracts for rounds 0..2 are gone") {
+                eventually() {
+                  val remaining = sv1Backend.appState.dsoStore
+                    .listCalculateRewardsV2()
+                    .futureValue
+                    .map(_.payload.round.number)
+                    .toSet
+                  (0L to 2L).foreach { round =>
+                    remaining should not contain round withClue
+                      s"CalculateRewardsV2 for round $round should be archived"
+                  }
+                }
+              }
+            }
           }
 
           (id0, id1, id3, id4, id5, id6, id7, aliceCreateId, svExpireId)
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
@@ -272,49 +272,6 @@ abstract class TrafficBasedRewardsTimeBasedIntegrationTestBase
                   s"CalculateRewardsV2 should exist for round $round"
               }
             }
-            // For rounds 0..5 scan will not be able calculate activity totals
-            // and root hashes, so archiving them here keep the
-            // CalculateRewardsTrigger's logs clean, as they expect each round
-            // with CalculateRewardsV2 to have a root hash.
-            if (dryRunEnabled) {
-              clue("Archive dry-run CalculateRewardsV2 for rounds 0..5 via sv admin API") {
-                sv1Backend.archiveDryRunRewardAccountingContracts((0L to 5L).toSeq)
-              }
-            } else {
-              // TODO (#5624): add support for bootstrapping
-              // Bootstrapping a network with mintingVersion set to trafficBasedAppRewards
-              // is in principle not supported, as the round 0 will never have
-              // activity totals/root-hash calculated, and its CalculateRewardsV2 cannot be processed.
-              // So the only way to handle this in test is via direct archive.
-              clue("Archive CalculateRewardsV2 for rounds 0..5 directly as dso") {
-                val cids = sv1Backend.appState.dsoStore
-                  .listCalculateRewardsV2()
-                  .futureValue
-                  .filter(c => c.payload.round.number >= 0L && c.payload.round.number <= 5L)
-                  .map(_.contractId)
-                if (cids.nonEmpty) {
-                  sv1Backend.participantClientWithAdminToken.ledger_api_extensions.commands
-                    .submitJava(
-                      userId = sv1Backend.config.ledgerApiUser,
-                      actAs = Seq(dsoParty),
-                      commands = cids.flatMap(_.exerciseArchive().commands.asScala.toSeq),
-                    )
-                }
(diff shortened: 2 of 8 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
```diff
@@ -272,49 +272,6 @@ abstract class TrafficBasedRewardsTimeBasedIntegrationTestBase
                   s"CalculateRewardsV2 should exist for round $round"
               }
             }
-            // For rounds 0..5 scan will not be able calculate activity totals
-            // and root hashes, so archiving them here keep the
-            // CalculateRewardsTrigger's logs clean, as they expect each round
-            // with CalculateRewardsV2 to have a root hash.
-            if (dryRunEnabled) {
-              clue("Archive dry-run CalculateRewardsV2 for rounds 0..5 via sv admin API") {
-                sv1Backend.archiveDryRunRewardAccountingContracts((0L to 5L).toSeq)
```
dfordivam, line 281: As we do need to test the `archiveDryRunRewardAccountingContracts` API, could we keep that for at least a couple of rounds (like 1~3), and confirm contracts for those rounds are gone. The rest of rounds could be consumed via triggers.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
```diff
@@ -172,7 +173,13 @@ class DbAppActivityRecordStore(
               )
       """.as[Option[Long]].headOption.map(_.flatten),
       "appActivity.earliestRoundWithCompleteAppActivity",
-    )
+    ).map {
+      // The first SV has complete history from genesis, so round 0 is always
+      // the earliest complete round. The query still runs to confirm ingestion
+      // has started (returns None if no meta row or no activity records).
+      case Some(_) if isFirstSv => Some(0L)
```
dfordivam: This means the first SV will never be able to update activity meta?
We should check `earliest_ingested_round` is 0 and only then return 0. Else fallback to the Some value.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbAppActivityRecordStoreTest.scala
```diff
@@ -471,6 +471,121 @@ class DbAppActivityRecordStoreTest
...
+        baseTs = CantonTimestamp.now()
+        _ <- store.insertActivityRecordMeta(1, 0, baseTs.toMicros, 0L)
+        _ <- insertRecordsForRounds(store, historyId, baseTs, ("round-0", 0L))
+        result <- store.earliestRoundWithCompleteAppActivity()
+      } yield {
+        result shouldBe None
+      }
+    }
+
+    "return round 0 on first SV even when ingestion started after genesis" in {
```
dfordivam: what is the intention behind this?
We have to handle the ingestion "started" after genesis gracefully right?

dfordivam (changes requested):
See comments, the logic needs to allow the first SV to bump meta userVersion + us to bump codeVersion

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
adetokunbo: Fixed

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbAppActivityRecordStoreTest.scala
adetokunbo: I've updated the descriptions; PTAL

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
adetokunbo: I tried that initially, but `earliest_ingested_round` value as 3 reflecting where the FAP grant was as added, so that check did not work.  That's why I've adopted this approach that does not depend on its value.  As you have pointed out, as originally implemented this was incorrect, as it blocked bumping the meta versions on firstSV; the latest commit fixes this; PTAL

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
```diff
@@ -172,7 +173,18 @@ class DbAppActivityRecordStore(
               )
       """.as[Option[Long]].headOption.map(_.flatten),
       "appActivity.earliestRoundWithCompleteAppActivity",
-    )
+    ).flatMap {
+      case Some(n) if isFirstSv =>
+        // First SV: round 0 unless a version bump occurred.
+        runQuerySingle(
+          hasPriorIngestionVersionDBIO.map(Some(_)),
```
dfordivam: This isn't compatible with current mainnet, which may not have a prior ingestion version

adetokunbo: Good point.  That actually makes the 'read-path only' approach adopted so far unworkable.  
I've updated the change to write 0 as the value of `earliest_ingested_round` on network bootstrap, and updated the `read-path` logic to reflect this.

PTAL

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
```diff
@@ -447,11 +465,15 @@ class DbAppActivityRecordStore(
               case None =>
                 DBIO.successful(NotReady: EnsureResult)
               case Some((firstRecordTimeMicros, earliestRound)) =>
+                // First SV on fresh network: earliest complete round is
+                // always 0, so override the given value.
+                val adjustedEarliestRound =
```
dfordivam, line 462: Some how I still feel that actual evidence of bootstrapping, ie round 0 is open, no verdicts ingested, etc is a better way to handle this.
Otherwise the correctness depends upon the state of DB.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
```diff
@@ -221,7 +227,19 @@ class DbAppActivityRecordStore(
...
+            if (isFirstSv)
+              sql"""select count(*) = 1 and min(earliest_ingested_round) = 0
+                    from #${Tables.activityRecordMeta}
+                    where history_id = $historyId
+              """.as[Boolean].head
+            else DBIO.successful(false)
+          hasPrev <-
+            if (isFirstSvFreshNetwork)
+              // First SV on fresh network: no prior-round proof needed.
+              DBIO.successful(true)
```
dfordivam: Wouldn't this break if the first SV does actual meta-version increment (assuming they also have `earliest_ingested_round` == 0 row)

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbAppActivityRecordStoreTest.scala
```diff
@@ -471,6 +471,133 @@ class DbAppActivityRecordStoreTest
...
+        } yield {
+          result.value shouldBe 11L
+        }
+      }
+    }
+  }
+
+  "assertCompleteActivity" should {
+
+    "on first SV" should {
```
dfordivam: Lets cover scenario where first SV does meta-version bump

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
dfordivam: ah nevermind, I missed count = 1.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
adetokunbo: Understood; though the upside of depending on that state of the db is that it's unaffected by unexpected errors that might interfere with the other evidence

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/DbAppActivityRecordStoreTest.scala
adetokunbo: Fixed

dfordivam (commented):
Lets get more review from Robert.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
dfordivam: Lets change the assertion here also

dfordivam: Fix comment

dfordivam: fix assertion, confirm all contracts for rounds < 11 are gone

dfordivam: Lets use `simpleTopology4SvsWithSimTime` now, as we no longer need the `actAs = Seq(dsoParty` block

dfordivam:
> Assumes the following to detect bootstrapping:

>    the activity meta table is initially empty if bootstrapping
>   so if there is any existing row, then it's not bootstrapping

Another assumption worth mentioning, the networks which are not bootstrapping must already have meta-table row, when this change lands.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
adetokunbo: Fixed

adetokunbo: Fixed

adetokunbo: Fixed, but only on reverting the environment back 1SV
With the 4SV environment, the contracts did not get archived; an quick investigation was because the non-first-SV scans could not serve root hashes for the early rounds

adetokunbo: Implemented then reverted this because it caused one of the changed assertions to fail, as explained above.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -239,7 +239,7 @@ object SvOnboardingConfig {
 }
 
 final case class InitialRewardConfig(
-    mintingVersion: String = "RewardVersion_FeaturedAppMarkers",
+    mintingVersion: String = "RewardVersion_TrafficBasedAppRewards",
```
dfordivam: I doubt that this is actually doing anything.
Notice `initialRewardConfig: Option[InitialRewardConfig] = None,` 
So lets make sure we test actually turning the rewardConfig ON

adetokunbo: Removed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsTimeBasedIntegrationTest.scala
adetokunbo: Now  added back since #5768,  #5888 got added to main, now the test succeed

adetokunbo:
@meiersi-da PTAL.  So far, this PR meets its goal the TBA integration tests pass using a 4SV topology; please do an initial review pass with feedback about the implementation approach

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
meiersi-da: I think the current solution is fine. The blast radius from this being off is minimal: it will never happen on MainNet, and in the worst case the app activity records on a reset TestNet or DevNet are off or stuck for Round 0.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
meiersi-da:
```suggestion
      rewardVersion: RewardVersion = RewardVersion.REWARDVERSION_FEATUREDAPPMARKERS,
```

 nit: less boolean blindness, https://existentialtype.wordpress.com/2011/03/15/boolean-blindness/

meiersi-da (approved):
Thanks. Seems like @dfordivam has some outstanding comment/concern. I'm good with this implementation, if he is.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
adetokunbo: Fixed

dfordivam (approved):
Approving to remove my "request changes" state and make PR mergeable

PR #5697, opened 2026-05-27
Support iterated settlement in Splice Amulet Wallet

OriolMunoz-da:
Fixes https://github.com/canton-network/splice/issues/5498


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AllocationsFrontendIntegrationTest.scala
```diff
@@ -384,6 +386,235 @@ class AllocationsFrontendIntegrationTest
...
+
+            // Check the committed checkbox
+            inside(find(id("create-allocation-committed"))) { case Some(element) =>
+              element.underlying.click()
+            }
+
+            // Set next iteration funding: key = amulet instrument id, value = 100
+            eventuallyClickOn(id("create-allocation-next-iteration-funding-add-meta"))
+            textField("create-allocation-next-iteration-funding-meta-key-0").underlying
+              .sendKeys(amuletInstrumentKey)
```
OriolMunoz-da: does this make sense, or should it be hardcoded to be a (single or several?) optional amount field always hardcoded to Amulet?

apps/wallet/frontend/src/components/MetaDisplay.tsx
OriolMunoz-da, line 1: moved to TextMap.tsx below

apps/wallet/frontend/src/components/TextMap.tsx
```diff
@@ -0,0 +1,90 @@
...
+                {value}
+              </Typography>
+            </Stack>
+          );
+        })}
+    </Stack>
+  );
+};
+
+export const TextMapEditor: React.FC<{
```
OriolMunoz-da, line 31: already existed in main as MetaEditor, was unused in v2-upcoming so it got deleted, restored it here

apps/wallet/frontend/src/components/CreateAllocation.tsx
```diff
@@ -348,7 +397,9 @@ function validatedForm(
       amount: leg.amount,
     });
   }
-  if (validLegSides.length === 0) return null;
+  // You must specify either transfger legs, or the next iteration funding
+  if (validLegSides.length === 0 && Object.keys(partial.next_iteration_funding).length === 0)
+    return null;
```
OriolMunoz-da: Image

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AllocationsFrontendIntegrationTest.scala
meiersi-da: a single optional amount field hardcoded to Amulet seems like the right choice. 

I'd probably do a checkbox for " [ ] allow iterated settlement ", which enables a field "funding amount: [ ... amount ... ]" and set the amount to zero by default.

apps/wallet/frontend/src/components/CreateAllocation.tsx
```diff
@@ -251,6 +254,46 @@ const CreateAllocation: React.FC = () => {
...
+                setAllocation({
+                  ...allocation,
+                  meta,
+                })
+              }
+              idPrefix="create-allocation"
+            />
+            <Typography variant="body2">Next iteration funding (optional)</Typography>
+            <TextMapEditor
+              meta={allocation.next_iteration_funding}
```
meiersi-da: Do I see it correctly that the user must know that "Amulet" is the right key to use for setting next iteration funding? 

Seems just about OK, but not great.

meiersi-da (approved):
Thanks. I feel a bit uneasy about the user having to know about the magic "Amulet" string for setting next iteration funding. If you have a cheap solution to avoid that then that would be welcome.

OriolMunoz-da:
fixed version with 
> a single optional amount field hardcoded to Amulet seems like the right choice.

image

meiersi-da:
> fixed version with
> 
> > a single optional amount field hardcoded to Amulet seems like the right choice.
> 
> image

Very nice! Thank you!

meiersi-da:
Small correction. `committed` is not dependent on "Iterated settlement". You can move that above that section. The same holds for the metadata, which is just general metadata for the amulet allocation as a whole.

PR #5700, opened 2026-05-27
Restore refresh-token silent renew with renew-or-signout on 401

### LINKED ISSUE #5682: Starting from 0.6.5 all UIs require re-login after access token expires

samsondav:
Alternative to #5683 — same bug, different design. Posting both so maintainers can pick.

PR #5545 set `automaticSilentRenew={false}` on the `react-oidc-context` provider and removed `offline_access` from the scope request. Either change on its own would have been fine; together they leave the UIs with no refresh-token mechanism at all, so every access-token expiry forces a full re-login (5 minutes on a default Keycloak deployment).

This PR restores token renewal for both Keycloak and Auth0 without re-enabling the library's timer-based silent renew (which races the new 401 handler under refresh-token rotation, see "Why a single renewal path" below).

## What changes

**1. `renewOrSignout` in `UserProvider`** (`UserContext.tsx`). On `splice:auth-expired` (fired from `BaseApiMiddleware` / `LedgerApiContext` on any 401), try one `auth.signinSilent()` before signing the user out. Single-flighted via a `renewInFlight` ref so concurrent 401s collapse to one grant. On success, react-query cache invalidation is deferred until the refreshed access token has propagated into `userAccessToken` state — a separate effect watches that state and fires `queryClient.invalidateQueries()` so the deferred invalidation doesn't race the state update.

**2. Library's `automaticSilentRenew` stays `false`** (matches what #5545 set). With Splice owning the renewal path, there is no need for the library's proactive timer — and re-enabling it would race the 401 handler (next section).

**3. Optional `enable_offline_scope: boolean` on the rs256 `AuthConfig`** for Auth0 deployments where the IdP only returns a refresh token when `offline_access` is in the scope. Default false: Keycloak issues a session-bound refresh token without it, and on Keycloak `offline_access` produces a long-lived offline refresh token (larger XSS surface in a browser-based SPA).

## Why a single renewal path

An earlier revision of this PR re-enabled `automaticSilentRenew={true}` alongside `renewOrSignout`, on the theory that the proactive timer would cover normal expiry and the 401 handler would cover edge cases (suspended tabs, clock skew). Adversarial review surfaced a race: the library timer fires `signinSilent` at exp - 60s using the stored refresh token; a near-simultaneous user request 401s and the handler fires its own `signinSilent`. With IdPs that rotate refresh tokens (Auth0 by default, Keycloak when configured), one of the two concurrent grants is rejected — the refresh token was already exchanged. The handler's catch block then signs the user out even though the other renewal succeeded.

Dropping `automaticSilentRenew` to `false` and routing all renewal through the single-flighted handler eliminates the race. Trade: one extra 401 round-trip per token expiry per active session (the user's first request after expiry takes a renewal + retry instead of using a pre-refreshed token). That cost is bounded and the net UX is far better than the forced re-login this PR replaces.

## How this compares to #5683

#5683 takes a different approach: keep `automaticSilentRenew={true}` and make `enable_offline_scope` a mandatory operator-set boolean. That forces the Auth0-vs-Keycloak decision at configuration time but leaves two cases open:

- Auth0 without `offline_access` opt-in: no refresh token, silent renew has nothing to use, user is logged out at access-token TTL.
- The same silent-renew vs concurrent-401 race documented above.

This PR keeps the schema field optional (no breaking config change) and uses the `renewOrSignout` design to handle both cases at runtime. Pick whichever shape fits.

## Testing

New tests under `apps/common/frontend/src/__tests__/auth/`:

- `UserContextRenewOrSignout.test.tsx`: signinSilent called when a refresh token is present; `invalidateQueries` fires after token state propagates (deferred path); removeUser called when no refresh token; failure path falls through to signout; `invalidateQueries` not called on the signout path; single-flight under concurrent expired events; listener removed on unmount.
- `AuthProvider.test.tsx`: asserts `automaticSilentRenew=false` (the design intent).
- `authSchema.test.ts`: parse behaviour of optional `enable_offline_scope`.
- `oidcAuthToProviderProps.test.ts`: scope construction with and without the opt-in.

Existing 401-fallback coverage in `BaseApiMiddleware.test.ts` and `LedgerApiClient.test.ts` is untouched — the handler still fires `splice:auth-expired` on 401, only the handler's response changes.

`apps/sv/frontend/src/__tests__/auth/redirect.test.tsx` is updated to wrap `UserProvider` in `QueryClientProvider` (now required because `UserProvider` calls `useQueryClient`). `apps/sv/frontend/package.json` pins `@tanstack/react-query: 5.90.20` directly to match the other apps and prevent singleton-resolution drift.

## References

- Refresh-token path: `oidc-client-ts.js:2533` (`if (user?.refresh_token) return this._useRefreshToken(state)`).
- Iframe fallback gate: `oidc-client-ts.js:2540` (throws "No silent_redirect_uri configured" when no refresh token and no `silent_redirect_uri` — Splice never sets `silent_redirect_uri`, so the iframe path is unreachable).
- Auth0 docs require `offline_access` to receive a refresh token: https://auth0.com/docs/secure/tokens/refresh-tokens/get-refresh-tokens

Fixes #5682

### DIFF
--- a/apps/common/frontend/src/__tests__/auth/AuthProvider.test.tsx
+++ b/apps/common/frontend/src/__tests__/auth/AuthProvider.test.tsx
@@ -28,7 +28,7 @@ const rs256Config: AuthConfig = {
 };
 
 describe('AuthProvider', () => {
-  test('passes automaticSilentRenew=true to the OIDC provider', () => {
+  test('disables the library timer-based silent renew (UserProvider handles it on 401 instead)', () => {
     mockOidcAuthProvider.mockClear();
     render(
       <AuthProvider authConf={rs256Config}>
@@ -37,18 +37,7 @@ describe('AuthProvider', () => {
     );
     expect(mockOidcAuthProvider).toHaveBeenCalled();
     const props = mockOidcAuthProvider.mock.calls[0][0] as Record<string, unknown>;
-    expect(props.automaticSilentRenew).toBe(true);
-  });
-
-  test('does not pass silent_redirect_uri (iframe fallback intentionally disabled)', () => {
-    mockOidcAuthProvider.mockClear();
-    render(
-      <AuthProvider authConf={rs256Config}>
-        <div />
-      </AuthProvider>
-    );
-    const props = mockOidcAuthProvider.mock.calls[0][0] as Record<string, unknown>;
-    expect(props.silent_redirect_uri).toBeUndefined();
+    expect(props.automaticSilentRenew).toBe(false);
   });
 
   test('hs256-unsafe config short-circuits without rendering the OIDC provider', () => {
--- a/apps/common/frontend/src/__tests__/auth/AuthProvider.test.tsx
+++ b/apps/common/frontend/src/__tests__/auth/AuthProvider.test.tsx
@@ -1,0 +1,69 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { render } from '@testing-library/react';
+import React from 'react';
+import { describe, expect, test, vi } from 'vitest';
+
+const mockOidcAuthProvider = vi.fn(({ children }: React.PropsWithChildren) => <>{children}</>);
+
+vi.mock('react-oidc-context', () => ({
+  AuthProvider: (props: React.PropsWithChildren<Record<string, unknown>>) =>
+    mockOidcAuthProvider(props),
+}));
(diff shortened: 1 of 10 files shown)


martinflorian-da:
This PR looks stale and content-wise qualifies for https://github.com/canton-network/splice/blob/main/AI_POLICY.md#guard-reviewers-time IMO.

PR #5701, opened 2026-05-27
Upload splice-util-token-standard-wallet by default on all validator nodes

OriolMunoz-da:
Fixes https://github.com/canton-network/splice/issues/4661


apps/dar-resources-generator/src/main/scala/org/lfdecentralizedtrust/splice/darutils/DarResourcesGenerator.scala
```diff
@@ -40,6 +40,7 @@ object DarResourcesGenerator {
     "splice-api-token-allocation-instruction-v2" -> "1.0.0",
     "splice-token-test-trading-app" -> "1.0.0",
     "splice-token-test-trading-app-v2" -> "1.0.0",
+    "splice-util-token-standard-wallet" -> "1.1.0",
```
OriolMunoz-da, line 46: btw this is the "minimum supported version" afaict, and it's unclear to me why we have 3 different versions for this

meiersi-da: what do you mean by three different versions?

meiersi-da: token standard wallet needed a bump as I added a new template

meiersi-da (approved):
Thanks. Please also add a comment to the release notes stub for the token standard v2 explaining that this is now deployed by default.

apps/dar-resources-generator/src/main/scala/org/lfdecentralizedtrust/splice/darutils/DarResourcesGenerator.scala
OriolMunoz-da: my bad, I had thought this was introduced as part of v2, but I see now that it existed since before that

PR #5721, opened 2026-05-28
Fix splice docs links TOKEN_STANDARD_V2_DEVNET.md

OriolMunoz-da:
After the org change they're broken

credit to Phillip for the report 

[static]


meiersi-da (approved):
ty

PR #5727, opened 2026-05-28
Release Notes for 0.6.6

OriolMunoz-da:


moritzkiefer-da (approved):
thx, there may be some Canton changes worth mentioning but their release notes seem like a mess so let's not do that and move ahead

PR #5728, opened 2026-05-28
Bump version to 0.6.7

OriolMunoz-da:


moritzkiefer-da (approved):
newer is always better

PR #5730, opened 2026-05-28
Use real Scan server in OpenAPI specs

danielporterda:
## Summary
- replace the Scan OpenAPI placeholder server with the requested public Scan target

Upstream fix for https://github.com/canton-network/cf-docs/issues/649. The generated cf-docs curl for `/v0/scans` should become:

```bash
curl https://scan.sv-1.global.canton.network.sync.global/api/scan/v0/scans
```

### DIFF
--- a/apps/scan/src/main/openapi/scan-stream-server.yaml
+++ b/apps/scan/src/main/openapi/scan-stream-server.yaml
@@ -19,7 +19,7 @@ tags:
     description: |
       Still under active development, highly unstable. Do not use in production.
 servers:
-  - url: https://example.com/api/scan
+  - url: https://scan.sv-1.global.canton.network.sync.global/api/scan
 paths:
 
   # Note for developers working on this openAPI spec: we generate only server-side code from this file,
@@ -50,4 +50,3 @@ paths:
                   x-scala-type: org.apache.pekko.http.scaladsl.model.ResponseEntity
         "404":
           $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/404"
-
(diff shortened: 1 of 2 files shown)


martinflorian-da:
I'm not entirely sure that this is the right place to fix this; leaving it to @isegall-da to decide.

danielporterda:
Closing this approach. The fix should live in canton-network/cf-docs as a transform applied when retrieved OpenAPI specs are materialized, rather than changing the upstream splice specs.

isegall-da:
@martinflorian-da why not? Why is example.com better?

martinflorian-da:
> @martinflorian-da why not? Why is example.com better?

@isegall-da 

At some point splice was supposed to be network-agnostic; I guess this is not the goal anymore but also do we want to link (only) to MainNet specifically?

Also by making it look like an actual URL, we kind of promise that this URL will continue working. But will we notice and remember to fix it if it stops working?

Not saying that these are good reasons, but that's what I meant with "not sure".

isegall-da:
@danielporterda 

> do we want to link (only) to MainNet specifically?

This is definitely a good point regardless of where the URL is injected. That URL is mainnet-specific.

isegall-da:
The maintenance question I'm slightly less worried about, don't see CF's url changing any time soon...

danielporterda:
@isegall-da the openapi spec allows specifying multiple hosts, so i guess i'll include devnet and testnet as well.

i'd prefer to do it in this repo as that's one fewer moving part

isegall-da:
> specifying multiple hosts

And then how are those rendered in the generated pages?

PR #5732, opened 2026-05-28
Add stable cf-docs snippet tags

danielporterda:
## Summary

Adds stable `CF_DOCS_SPLICE_SNIPPET_*` RST comment tags around the Splice documentation ranges consumed by cf-docs external snippet generation. This lets cf-docs extract snippets by source tags instead of brittle line numbers.

Six implicit literal blocks are converted to explicit `code-block` directives so the tag comments can wrap the exact snippet body cleanly without changing the rendered intent. I also fixed one adjacent inline-literal spacing typo caught by `sphinx-lint`.

Confirmed that existing docs generation still works.


isegall-da:
(see slack, I don't understand what we're trying to do. These snippets are no longer maintained in Splice, so why do we care about tagging them??)

martinflorian-da:
@danielporterda  This PR and the discussion around it look stale to me, so going to close. Please LMK if you think I'm in error.

PR #5737, opened 2026-05-29
Add mainnet history dump user GCP SA

jagathweerasinghe-da:
[static]

Configured by: https://github.com/DACH-NY/canton-network-internal/pull/5116


cluster/pulumi/gha/src/config.ts
```diff
@@ -45,3 +56,6 @@ console.error(
 );
 
 export const ghaConfig = fullConfig.gha;
+
+export const isSpliceCluster =
```
jagathweerasinghe-da: No need to guard for splice

cluster/pulumi/gha/src/mainnetHistoryDumpsUser.ts
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
jagathweerasinghe-da, line 1: GH secret in PUlumi?

cluster/pulumi/gha/src/config.ts
```diff
@@ -28,6 +30,15 @@ const GhaConfigSchema = z.object({
         })
       )
       .default([]),
+    mainnetHistoryDumpsUser: z
+      .object({
+        bucket: z.string().min(1),
```
isegall-da, line 33: What's the min(1)? I don't recall seeing us doing that elsewhere. It's preventing an empty string basically?

cluster/pulumi/gha/src/config.ts
```diff
@@ -28,6 +30,15 @@ const GhaConfigSchema = z.object({
...
+        wifProjectNumber: z.string().min(1),
+        wifPoolId: z.string().min(1),
+        // GitHub repos (full "org/name") allowed to impersonate the SA via WIF.
+        githubRepositories: z.array(z.string().min(1)).min(1),
+      })
+      .optional(),
```
isegall-da, line 39: Consider not even making this optional. We deploy gha only in one place, and we want to use it there.

isegall-da (approved):
Nice, thank you

cluster/pulumi/gha/src/config.ts
```diff
@@ -28,6 +30,15 @@ const GhaConfigSchema = z.object({
         })
       )
       .default([]),
+    mainnetHistoryDumpsUser: z
+      .object({
+        bucket: z.string().min(1),
+        wifProjectNumber: z.string().min(1),
```
isegall-da, line 34: Can/should this be an int instead of string?

cluster/pulumi/gha/src/mainnetHistoryDumpsUser.ts
jagathweerasinghe-da: As discussed, we do this in a later PR.

cluster/pulumi/gha/src/config.ts
jagathweerasinghe-da: Yes, making sure empty strings are not passed to GCP APIs.

cluster/pulumi/gha/src/config.ts
jagathweerasinghe-da: Done

cluster/pulumi/gha/src/config.ts
jagathweerasinghe-da: Done. Until the conifg.yml is merged from network-internals, this will end up in an error if another 'pulumi up' executed.

cluster/pulumi/gha/src/config.ts
jagathweerasinghe-da: project number is eventually used in a string for IAM role config; string is the way to go.

cluster/pulumi/gha/src/config.ts
isegall-da: Yeah, I thought of that too. Initially I thought it's fine, and you just merge your other PR quickly. But given that that might be a bit tricky to get fully working, with permissions etc., maybe it's better for now to keep it optional so that we can always easy not apply it yet? And then at a later stage once everything is applied, drop the optional.

PR #5747, opened 2026-05-29
Auto-ignore parties vetting unsupported and ignored amulet versions

### LINKED ISSUE #5725: Auto-ignore parties having broken amulet version

julientinguely-da:
fixes https://github.com/canton-network/splice/issues/5725

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/IgnoredPartiesStore.scala
```diff
@@ -0,0 +1,23 @@
...
+import java.util.concurrent.ConcurrentHashMap
+import scala.jdk.CollectionConverters.*
+
+class IgnoredPartiesStore(initialParties: Set[PartyId]) {
+
+  private val parties: ConcurrentHashMap.KeySetView[PartyId, java.lang.Boolean] = {
+    val set = ConcurrentHashMap.newKeySet[PartyId]()
+    initialParties.foreach(set.add)
+    set
+  }
```
moritzkiefer-da, line 17:
```suggestion
  private val parties: ConcurrentHashMap.KeySetView[PartyId, java.lang.Boolean] = ConcurrentHashMap.newKeySet[PartyId]()
  initialParties.foreach(parties.add)
```

moritzkiefer-da: probably better to use an `addAll`

cluster/configs/shared/base-sv.yaml
```diff
@@ -22,6 +22,10 @@ participant:
       memory: '18Gi'
   additionalEnvVars:
     - !include(./participant_pruning.yaml)
+svApp:
+  brokenAmuletVersions:
```
moritzkiefer-da: rather than hardcoding that here could we use anything < minimum version as the default and then add overwrites in addition to that?

cluster/pulumi/common-sv/src/singleSvConfig.ts
```diff
@@ -102,6 +102,8 @@ const SvAppConfigSchema = z
     // Map of package name -> list of versions to explicitly unvet
     additionalPackagesToUnvet: z.record(z.string(), z.array(z.string())).optional(),
     resources: K8sResourceSchema,
+    // List of amulet package versions considered broken (expiry will skip and ignore parties)
```
moritzkiefer-da: broken sounds a bit too strong, maybe "ignored"?

cluster/pulumi/common-sv/src/svApp.ts
```diff
@@ -92,10 +92,20 @@ export function valuesForSvApp(
           },
         ]
       : [];
+  const brokenAmuletVersionsConfig =
+    config.svApp?.brokenAmuletVersions && config.svApp.brokenAmuletVersions.length > 0
+      ? [
+          {
+            name: 'ADDITIONAL_CONFIG_BROKEN_AMULET_VERSIONS',
```
moritzkiefer-da: not quite convinced this is worth it over just setting it through `additionalEnvVars`

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpiredAmuletAllocationTrigger.scala
```diff
@@ -48,14 +49,30 @@ class ExpiredAmuletAllocationTrigger(
...
 
   private val store = svTaskContext.dsoStore
 
   override def completeTaskAsDsoDelegate(task: Task, controller: String)(implicit
       tc: TraceContext
-  ): Future[TaskOutcome] = {
+  ): Future[TaskOutcome] =
+    completeWithBrokenVersionCheck(
+      task.work.vettedVersion.toString,
+      task.work.expiredContracts
```
moritzkiefer-da: can we share the logic with the one in line 42 (minus the dso party)? seems like we really want the parties we use for vetting to be in sync with the parties we ignore

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/BrokenAmuletVersionGuard.scala
```diff
@@ -0,0 +1,37 @@
...
+      fallback: => Future[TaskOutcome]
+  ): Future[TaskOutcome] = {
+    if (
+      svConfig.brokenAmuletVersions.contains(vettedVersion) &&
+      svConfig.parameters.enabledFeatures.ignorePartyIdWithBrokenAmulet
+    ) {
+      ignoredPartiesStore.addAll(expiredOwners)
+      Future.successful(
+        TaskSuccess(
+          s"Skipped batch with broken version $vettedVersion: added ${expiredOwners.size} parties to ignore list"
```
moritzkiefer-da: probably want to log the actually ignored parties

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/AutomationConfig.scala
```diff
@@ -101,11 +101,9 @@ case class AutomationConfig(
       * before we consider the trigger to be in an unhealthy state.
       */
     futureCompletionGracePeriod: PositiveFiniteDuration = PositiveFiniteDuration.ofSeconds(1L),
+    ignoredUnavailablePartyIds: Set[PartyId] = Set.empty,
```
moritzkiefer-da: Given that it's not actually unavailable but vetting issues I would just call it "ignoredPartyIds".

it's also a breaking change, I think that's ok given that it only affects SVs but we should clearly mention it in release notes.

moritzkiefer-da (commented):
Thanks! This looks promising, mentioned most of my suggestions in comments. At a high-level I would:

1. Avoid the term `broken`. That sounds more like a critical bug than just "too old and can't downgrade"
2. Try to share the logic for determining the parties
3. Ignore versions < minimum version by default

cluster/configs/shared/base-sv.yaml
julientinguely-da: > anything < minimum version as the default

oups, added that along in the AutomationConfig

julientinguely-da: note the minimal version is 0.1.15

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/BrokenAmuletVersionGuard.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  protected def completeWithIgnoredAmuletVersionCheck(
+      vettedVersion: String,
+      expiredOwners: Set[PartyId],
+  )(
+      fallback: => Future[TaskOutcome]
+  )(implicit tc: TraceContext): Future[TaskOutcome] = {
+    if (
+      svConfig.allIgnoredAmuletVersions.contains(vettedVersion) &&
+      svConfig.parameters.enabledFeatures.ignorePartyIdWithBrokenAmulet
```
moritzkiefer-da: nit: still using broken here

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/IgnoredAmuletVersionGuard.scala
```diff
@@ -0,0 +1,43 @@
...
+    if (
+      svConfig.allIgnoredAmuletVersions.contains(vettedVersion) &&
+      svConfig.parameters.enabledFeatures.ignorePartyIdWithBrokenAmulet
+    ) {
+      logger.info(
+        s"Batch resolved to ignored version $vettedVersion, adding ${expiredOwners.size} parties to the ignore list: $expiredOwners"
+      )
+      ignoredPartiesStore.addAll(expiredOwners)
+      Future.successful(
+        TaskSuccess(
```
moritzkiefer-da, line 29: this already gets logged at info, if you include the list of parties you can just drop the log line above

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpiredAmuletAllocationTrigger.scala
```diff
@@ -48,20 +49,32 @@ class ExpiredAmuletAllocationTrigger(
...
+    with BrokenAmuletVersionGuard {
 
   private val store = svTaskContext.dsoStore
 
   override def completeTaskAsDsoDelegate(task: Task, controller: String)(implicit
       tc: TraceContext
   ): Future[TaskOutcome] = {
-
-    val allParties = task.work.expiredContracts.flatMap { contract =>
+    val informees = task.work.expiredContracts.flatMap { contract =>
```
moritzkiefer-da, line 60: we still need the dso when querying vetting state, we just never should ignore it. Might also be an option to filter it out in the IgnoredPartiesStore

cluster/configs/shared/base-sv.yaml
```diff
@@ -22,6 +22,11 @@ participant:
       memory: '18Gi'
   additionalEnvVars:
     - !include(./participant_pruning.yaml)
+svApp:
+  additionalEnvVars:
+    - name: ADDITIONAL_CONFIG_IGNORED_AMULET_VERSIONS
+      value: |
+        canton.sv-apps.sv.ignored-amulet-versions = ["0.1.15", "0.1.16"]
```
moritzkiefer-da: why those two versions? Don't those work and the only reason we are currently getting issues is because we didn't pass in preferred packages but once your change here lands we also have your fix for that so there is never a good reason to have this here?

moritzkiefer-da (approved):
thanks!

cluster/configs/shared/base-sv.yaml
julientinguely-da: you're right!

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpiredAmuletAllocationTrigger.scala
julientinguely-da: yes, the dso is added in the private method below

PR #5762, opened 2026-06-01
remove the migration id from the sv and validator app and make it optional in scan 

nicu-da:
part of https://github.com/canton-network/splice/issues/5425

changed to read the migration id from the acs stores instead of history tables 

### sv app:

- removed the config
- during onboarding we read the id from the sponsor SV

### Scan

- kept the config and use it only during initial onboarding
- we check afterwards that we have the same id as the other scans
- we might wanna consider removing it all together in the future but it would require scan to know about the sv app of the sponsor to retrieve it 

### Validators

- removed the config 
- during onboarding do a bft read from scan
- the id is still used for the participant database name, a bit weary if doing anything there as most likely people will just upgrade without reading the release notes 

will follow up with an update of the docs as well 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


canton-network-da:
# [backport] Reminder #
- [ ] release-line-0.6.6
- [ ] release-line-0.6.5
- [ ] release-line-0.6.4
- [ ] release-line-0.6.3

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
```diff
@@ -108,6 +110,25 @@ class JoiningNodeInitializer(
     )
   )
 
+  /** Reads the current migration id from the SV's own scan. Used as a fallback when the
+    * local update history does not yet know about any migration id.
+    */
+  private def migrationIdFromScan(): Future[Long] =
+    ScanConnection
+      .singleUncached(
+        ScanAppClientConfig(NetworkAppClientConfig(config.scan.internalUrl)),
```
nicu-da: this doesn't make sense actually, need to read it from the other scans :sad:

nicu-da:
will merge it in the next release just for safety

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbAppStore.scala
```diff
@@ -109,3 +112,32 @@ abstract class DbAppStore(
...
+  def getHighestKnownMigrationId(
+      storage: DbStorage,
+      acsTableName: String,
+  )(implicit
+      ec: ExecutionContext,
+      closeContext: CloseContext,
+      tc: TraceContext,
+  ): Future[Option[Long]] = {
+    val queryResult = storage.query(
+      sql"""select max(migration_id) from #$acsTableName""".as[Option[Long]],
```
moritzkiefer-da: What's the query plan for that? Not sure we have the right index. I also don't understand why you need to query the ACS table instead of last_ingested_offsets.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
```diff
@@ -60,8 +60,10 @@ case class ScanAppBackendConfig(
     isFirstSv: Boolean = false,
     miningRoundsCacheTimeToLiveOverride: Option[NonNegativeFiniteDuration] = None,
     enableForcedAcsSnapshots: Boolean = false,
-    // TODO(DACH-NY/canton-network-node#9731): get migration id from sponsor sv / scan instead of configuring here
-    domainMigrationId: Long = 0L,
+    // The migration id is normally read from the DB (the highest known migration id in the
+    // update history). It only needs to be configured to bootstrap a node that does not yet
+    // have any migration id in its DB (e.g. the network-founding scan or a freshly joining scan).
```
moritzkiefer-da, line 65: why does a freshly joining scan not read it from the sponsor?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
```diff
@@ -247,6 +250,16 @@ class BftScanConnectionTest
       } yield dsoPartyId should be(partyIdA)
     }
 
+    "return the agreed migration id when all agree" in {
```
moritzkiefer-da: can we change it to threshold agree to make sure we don't end up introducing a single point of failure for everyone

moritzkiefer-da: just asking for a test change, the code seems fine

moritzkiefer-da (approved):
thanks!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
nicu-da: we don't have any connection from scan to the sponsor yet and decided to avoid adding it here. 
though from a UX perspective I think you're right and adding the sponsor url here is easier than dealing with having to add the config

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbAppStore.scala
nicu-da: Not that familiar with the stores, simplified it 

```
sv_sv_1=> EXPLAIN (ANALYZE, BUFFERS, VERBOSE)
SELECT max(migration_id) FROM store_last_ingested_offsets;
                                                                                    QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=3.77..3.78 rows=1 width=8) (actual time=0.237..0.239 rows=1 loops=1)
   Output: max(migration_id)
   Buffers: shared hit=11
   ->  Index Only Scan using store_last_ingested_offsets_pkey on sv_sv_1.store_last_ingested_offsets  (cost=0.14..3.73 rows=17 width=8) (actual time=0.192..0.227 rows=17 loops=1)
         Output: store_id, migration_id
         Heap Fetches: 14
         Buffers: shared hit=11
 Query Identifier: 8294497768896216590
 Planning:
   Buffers: shared hit=98
 Planning Time: 0.605 ms
 Execution Time: 0.394 ms
(12 rows)

```

moritzkiefer-da: looks better :+1:

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/config/ScanAppConfig.scala
nicu-da: will actually remove it but in a different PR as there's a bunch of changes required for that and this PR is kind of too fat already

PR #5768, opened 2026-06-02
SvApp: Do BFT read in CalculateRewardsTrigger, ProcessRewardsTrigger

### LINKED ISSUE #5623: SvApp: Do BFT read in CalculateRewardsTrigger, ProcessRewardsTrigger

dfordivam:
Fixes #5623 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
@@ -19,7 +19,6 @@ import org.lfdecentralizedtrust.splice.codegen.java.splice.amuletrules.AmuletRul
 import org.lfdecentralizedtrust.splice.codegen.java.splice.dsorules.actionrequiringconfirmation.ARC_AmuletRules
 import org.lfdecentralizedtrust.splice.codegen.java.splice.dsorules.amuletrules_actionrequiringconfirmation.CRARC_StartProcessingRewardsV2
 import org.lfdecentralizedtrust.splice.config.ConfigTransforms
-import org.lfdecentralizedtrust.splice.config.RateLimitersConfig
 import org.lfdecentralizedtrust.splice.http.v0.definitions
 import definitions.GetRewardAccountingBatchResponse
 import definitions.GetRewardAccountingRootHashResponse
@@ -38,7 +37,6 @@ import org.lfdecentralizedtrust.splice.sv.config.InitialRewardConfig
 import org.lfdecentralizedtrust.splice.util.{
   AmuletConfigSchedule,
   AmuletConfigUtil,
-  SpliceRateLimitConfig,
   TimeTestUtil,
   TriggerTestUtil,
   WalletTestUtil,
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
@@ -276,7 +276,7 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
             sv2Db
               .update_(
                 sqlu"update app_activity_record_meta set earliest_ingested_round = $round",
-                "test.raiseEarliestIngestedRound",
+                "test.increaseAppActivityMeta_EarliestIngestedRound",
               )
               .futureValueUS
           },
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
@@ -1,17 +1,25 @@
 package org.lfdecentralizedtrust.splice.integration.tests
 
 import com.digitalasset.canton.HasExecutionContext
+import com.digitalasset.canton.config.NonNegativeDuration
 import com.digitalasset.canton.data.CantonTimestamp
+import com.digitalasset.canton.lifecycle.CloseContext
+import com.digitalasset.canton.resource.DbStorage
 import com.digitalasset.canton.topology.PartyId
 import java.time.Duration
 import java.util.Optional
+import org.lfdecentralizedtrust.splice.codegen.java.splice.amulet.cryptohash.Hash
 import org.lfdecentralizedtrust.splice.codegen.java.splice.amuletconfig.{
   AmuletConfig,
   RewardConfig,
   RewardVersion,
   USD,
 }
+import org.lfdecentralizedtrust.splice.codegen.java.splice.amuletrules.AmuletRules_StartProcessingRewardsV2
+import org.lfdecentralizedtrust.splice.codegen.java.splice.dsorules.actionrequiringconfirmation.ARC_AmuletRules
+import org.lfdecentralizedtrust.splice.codegen.java.splice.dsorules.amuletrules_actionrequiringconfirmation.CRARC_StartProcessingRewardsV2
 import org.lfdecentralizedtrust.splice.config.ConfigTransforms
+import org.lfdecentralizedtrust.splice.config.RateLimitersConfig
 import org.lfdecentralizedtrust.splice.http.v0.definitions
 import definitions.GetRewardAccountingBatchResponse
 import definitions.GetRewardAccountingRootHashResponse
@@ -24,14 +32,20 @@ import org.lfdecentralizedtrust.splice.sv.automation.confirmation.{
   CalculateRewardsDryRunTrigger,
   CalculateRewardsTrigger,
 }
+import org.lfdecentralizedtrust.splice.sv.automation.delegatebased.ProcessRewardsTrigger
+import org.lfdecentralizedtrust.splice.scan.automation.RewardComputationTrigger
 import org.lfdecentralizedtrust.splice.sv.config.InitialRewardConfig
 import org.lfdecentralizedtrust.splice.util.{
   AmuletConfigSchedule,
   AmuletConfigUtil,
+  SpliceRateLimitConfig,
   TimeTestUtil,
   TriggerTestUtil,
   WalletTestUtil,
 }
+
+import scala.concurrent.duration.DurationInt
+import slick.jdbc.canton.ActionBasedSQLInterpolation.Implicits.actionBasedSQLInterpolationCanton
 
 // This test focuses on the SV app side triggers testing
 // - Turning on/off of dry-run and minting-version in rewardConfig
@@ -181,6 +195,146 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
           ) should not be empty
         }
       }
+
+      confirmBftRead(bobParty)
+  }
+
+  // Here we confirm that sv2 can do BFT read of root-hash and batch from sv1 and sv4 only
+  // And the rewards processing works even when sv3 is offline.
+  private def confirmBftRead(
+      bobParty: PartyId
+  )(implicit env: SpliceTestConsoleEnvironment): Unit = {
+    clue("Stop sv3 to confirm we can perform bft read from sv1, sv4 only") {
+      sv3Backend.stop()
+      sv3ScanBackend.stop()
+    }
+
+    // sv2 will process all ProcessRewardsV2
+    val otherProcessRewardsTriggers =
+      Seq(sv1Backend, sv4Backend).map(_.dsoDelegateBasedAutomation.trigger[ProcessRewardsTrigger])
+
+    otherProcessRewardsTriggers.foreach(_.pause().futureValue)
+
+    try {
+      val sv2CalculateRewards = sv2Backend.dsoAutomation.trigger[CalculateRewardsTrigger]
+
+      // Pausing this ensures that the root-hash is not calculated while we advance round
+      val sv2RewardComputation = sv2ScanBackend.automation.trigger[RewardComputationTrigger]
+
+      // Here we ensure that SV2 has done ingestion of app-activity for the round just closed
+      // But then its AppActivityRecordMetaT is bumped so that it cannot compute the
+      // root-hash for the round.
+      val (calculateRewardsCid, round) = setTriggersWithin(
+        triggersToPauseAtStart = Seq(sv2CalculateRewards, sv2RewardComputation)
+      ) {
+        val round = oldestOpenRound
+        doTransfer(bobParty)
+        advanceRoundsToNextRoundOpening
+        doTransfer(bobParty)
+
+        val (calculateRewardsCid, rootHash) =
+          clue(
+            s"Round $round just closed: its CalculateRewardsV2 exists and sv1 serves root-hash"
+          ) {
+            eventually() {
+              val calc = sv1Backend.appState.dsoStore
+                .listCalculateRewardsV2()
+                .futureValue
+                .filterNot(_.payload.dryRun)
+                .find(_.payload.round.number == round)
+                .value
+              val rootHash = inside(sv1ScanBackend.getRewardAccountingRootHash(round)) {
+                case GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashOk(h) =>
+                  h.rootHash
+              }
+              (calc.contractId, rootHash)
+            }
+          }
+
+        clue(s"Only sv1 and sv4 confirm round $round, so it is not yet processed") {
+          eventually() {
+            val startProcessingAction = new ARC_AmuletRules(
+              new CRARC_StartProcessingRewardsV2(
+                new AmuletRules_StartProcessingRewardsV2(calculateRewardsCid, new Hash(rootHash))
+              )
+            )
+            sv1Backend.appState.dsoStore
+              .listConfirmations(startProcessingAction)
+              .futureValue should have size 2
+          }
+        }
+
+        // This is trying to simulate AppActivityRecordMetaT's userVersion bump
+        // albeit in a direct way, to avoid restart of scan app, etc.
+        actAndCheck(
+          s"Reset sv2's earliest-ingested round to $round", {
+            val sv2Db = sv2ScanBackend.appState.storage match {
+              case db: DbStorage => db
+              case other => fail(s"Expected DbStorage")
+            }
+            implicit val closeContext: CloseContext = CloseContext(sv2Db)
+            sv2Db
+              .update_(
+                sqlu"update app_activity_record_meta set earliest_ingested_round = $round",
+                "test.raiseEarliestIngestedRound",
+              )
+              .futureValueUS
+          },
+        )(
+          s"sv2's own scan now answers CannotProvide for round $round",
+          _ =>
+            sv2ScanBackend.getRewardAccountingRootHash(round) shouldBe
+              a[GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashCannotProvide],
+        )
+
+        (calculateRewardsCid, round)
+      }
+
+      // setTriggersWithin has resumed sv2's CalculateRewardsTrigger. sv3 is stopped and sv2's own
+      // scan CannotProvide, so the deciding 3rd confirmation can only come from sv2 via bft read.
+      clue(s"sv2's own scan still answers CannotProvide for round $round") {
+        sv2ScanBackend.getRewardAccountingRootHash(round) shouldBe
+          a[GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashCannotProvide]
+      }
+
+      clue(
+        s"sv2 reads round $round from the sv1 and sv4, and supplies the 3rd confirmation vote"
+      ) {
+        eventually() {
+          sv1Backend.appState.dsoStore
+            .listCalculateRewardsV2()
+            .futureValue
+            .map(_.contractId) should not contain calculateRewardsCid
+        }
+      }
+
+      clue(
+        s"sv2 does processing of ProcessRewardsV2"
+      ) {
+        eventually() {
+          sv1Backend.appState.dsoStore
+            .listRewardCouponsV2()
+            .futureValue
+            .map(_.payload.round.number) should contain(round)
+          sv1Backend.appState.dsoStore
+            .listProcessRewardsV2()
+            .futureValue
+            .map(_.payload.round.number) should not contain round
+        }
+      }
+    } finally {
+      otherProcessRewardsTriggers.foreach(_.resume())
+      clue("Restart sv3") {
+        sv3ScanBackend.start()
+        sv3Backend.start()
+        sv3Backend.waitForInitialization(
+          timeout = NonNegativeDuration.tryFromDuration(120.seconds)
+        )
+        sv3ScanBackend.waitForInitialization(
+          timeout = NonNegativeDuration.tryFromDuration(120.seconds)
+        )
+      }
+    }
   }
 
   private def doTransfer(
@@ -208,6 +362,11 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
       case GetRewardAccountingBatchResponse.members.RewardAccountingBatchOfMintingAllowances(b) =>
         b.mintingAllowances.toSeq
     }
+
+  private def oldestOpenRound(implicit env: SpliceTestConsoleEnvironment): Long = {
+    val (openRounds, _) = sv1ScanBackend.getOpenAndIssuingMiningRounds()
+    openRounds.map(_.contract.payload.round.number.toLong).min
+  }
 
   private def assertOldestOpenRound(
       expected: Long
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
@@ -883,14 +883,14 @@ class BftScanConnection(
         )
   }
 
-  /** Here we query all scans in parallel and resolve to the first response that has the
-    * batch because the contents of batch are verifiable using the hash.
+  /** The batch contents are verifiable via the hash, so BFT agreement across scans is not
+    * required: we query a single random scan and use its response if it has the batch.
     */
   override def getRewardAccountingBatch(roundNumber: Long, batchHash: String)(implicit
       ec: ExecutionContext,
       tc: TraceContext,
   ): Future[Option[GetRewardAccountingBatchResponse]] = {
-    val callConfig = BftCallConfig.default(scanList.scanConnections).copy(targetSuccess = 1)
+    val callConfig = BftCallConfig.randomSingleCall(scanList.scanConnections)
     if (!callConfig.enoughAvailableScans) Future.successful(None)
     else
       bftCall[GetRewardAccountingBatchResponse](
@@ -1062,6 +1062,12 @@ object BftScanConnection {
           )
       }
     }
+
+    // For targetSuccess==1, by setting the requestsToDo to 1, bftCall would
+    // select a random scan during retries, thereby preventing a single malicious
+    // scan to prevent reaching concesus by doing a failed response faster than others.
+    def randomSingleCall(connections: ScanConnections): BftCallConfig =
+      default(connections).copy(requestsToDo = 1, targetSuccess = 1)
 
     def forAvailableData(
         connections: ScanConnections,
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
@@ -848,20 +848,18 @@ class BftScanConnection(
     val callConfig = BftCallConfig.default(scanList.scanConnections)
     if (!callConfig.enoughAvailableScans) Future.successful(undetermined)
     else
-      BftScanConnection
-        .executeCall[String, SingleScanConnection](
-          call = scan =>
-            scan.getRewardAccountingRootHash(roundNumber).flatMap {
-              case GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashOk(ok) =>
-                Future.successful(ok.rootHash)
-              case _: GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashUndetermined |
-                  _: GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashCannotProvide =>
-                Future.failed(BftScanConnection.IgnoreResponse(scan.url))
-            },
-          requestFrom = callConfig.connections,
-          nTargetSuccess = callConfig.targetSuccess,
-          logger = logger,
-        )
+      bftCall[String](
+        call = scan =>
+          scan.getRewardAccountingRootHash(roundNumber).flatMap {
+            case GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashOk(ok) =>
+              Future.successful(ok.rootHash)
+            case _: GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashUndetermined |
+                _: GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashCannotProvide =>
+              Future.failed(BftScanConnection.IgnoreResponse(scan.url))
+          },
+        endpoint = "getRewardAccountingRootHash",
+        callConfig = callConfig,
+      )
         .transform(tryRootHash =>
           Success(
             tryRootHash.toOption.fold(undetermined)(rootHash =>
@@ -884,20 +882,18 @@ class BftScanConnection(
       ec: ExecutionContext,
       tc: TraceContext,
   ): Future[Option[GetRewardAccountingBatchResponse]] = {
-    val callConfig = BftCallConfig.default(scanList.scanConnections)
+    val callConfig = BftCallConfig.default(scanList.scanConnections).copy(targetSuccess = 1)
     if (!callConfig.enoughAvailableScans) Future.successful(None)
     else
-      BftScanConnection
-        .executeCall[GetRewardAccountingBatchResponse, SingleScanConnection](
-          call = scan =>
-            scan.getRewardAccountingBatch(roundNumber, batchHash).flatMap {
-              case Some(batch) => Future.successful(batch)
-              case None => Future.failed(BftScanConnection.IgnoreResponse(scan.url))
-            },
-          requestFrom = callConfig.connections,
-          nTargetSuccess = 1,
-          logger = logger,
-        )
+      bftCall[GetRewardAccountingBatchResponse](
+        call = scan =>
+          scan.getRewardAccountingBatch(roundNumber, batchHash).flatMap {
+            case Some(batch) => Future.successful(batch)
+            case None => Future.failed(BftScanConnection.IgnoreResponse(scan.url))
+          },
+        endpoint = "getRewardAccountingBatch",
+        callConfig = callConfig,
+      )
         .transform(tryBatch => Success(tryBatch.toOption))
   }
 }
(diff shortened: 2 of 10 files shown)


dfordivam:
@meiersi-da This draft PR is almost complete and I would be happy to get some early feedback while I resolve the failure in the integration test. Will discuss more on call today

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -817,17 +821,71 @@ class BftScanConnection(
...
       ec: ExecutionContext,
       tc: TraceContext,
   ): Future[Option[GetRewardAccountingBatchResponse]] =
-    bftCall(_.getRewardAccountingBatch(roundNumber, batchHash), "getRewardAccountingBatch")
+    Future
+      .find(
+        scanList.scanConnections.open.toList
+          .map(_.getRewardAccountingBatch(roundNumber, batchHash))
+      )(_.isDefined)
+      .map(_.flatten)
```
meiersi-da: Won't this fail if any of the inner future's fails?

Why not reuse `executeCall` and use a threshold of `1`? That would also reuse potential call cancellation logic, and exception logging behavior.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -817,17 +821,71 @@ class BftScanConnection(
...
+    else
+      BftScanConnection
+        .executeCall[String, SingleScanConnection](
+          call = scan =>
+            scan.getRewardAccountingRootHash(roundNumber).map {
+              case GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashOk(ok) =>
+                ok.rootHash
+              case _: GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashUndetermined |
+                  _: GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashCannotProvide =>
+                throw new RuntimeException(
```
meiersi-da: I'm always wary about raising `RuntimeException`, as that's considered an internal error by most of our code. Are the logs OK for this?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
```diff
@@ -903,4 +953,141 @@ class BftScanConnectionTest
...
+  "BftScanConnection.getRewardAccountingRootHash" should {
+
+    // n=4 scans -> default BFT threshold requiredNumScanThreshold(4) = f+1 = 2.
+    "reaches consensus when f+1 scans agree on the same hash" in {
+      val round = 42L
+      val connections = getMockedConnections(n = 4)
+      makeMockReturnRootHashOk(connections(0), round, "aabb")
+      makeMockReturnRootHashOk(connections(1), round, "aabb")
+      makeMockReturnRootHashUndetermined(connections(2), round)
+      makeMockReturnRootHashUndetermined(connections(3), round)
```
meiersi-da: Test at least one of them failing with an exception, e.g., because it's Scan server is not reachable.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
```diff
@@ -903,4 +953,141 @@ class BftScanConnectionTest
...
+      for {
+        resp <- bft.getRewardAccountingRootHash(round)
+      } yield inside(resp) {
+        case GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashOk(ok) =>
+          ok.rootHash should be("aabb")
+          ok.roundNumber should be(round)
+      }
+    }
+
+    "returns Undetermined when no quorum agrees on a hash" in {
```
meiersi-da, line 1071: We'd like to see a warning logged if there is disagreement on the actual hash, as that hints at an operational problem (or in the most extreme case, a compromised SV node).

Review the logs and add assertions here that they are as expected.

meiersi-da: We probably also want to see one info level log, when we are fetching our root hash from the other SVs, as that is a security relevant event.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
```diff
@@ -903,4 +953,141 @@ class BftScanConnectionTest
...
+
+  "BftScanConnection.getRewardAccountingBatch" should {
+
+    "returns the batch from the first scan that has it" in {
+      val round = 7L
+      val hash = "abcdabcd"
+      val connections = getMockedConnections(n = 3)
+      makeMockReturnBatch(connections(0), round, hash, None)
+      makeMockReturnBatch(connections(1), round, hash, Some(rewardAccountingBatchResponse))
+      makeMockReturnBatch(connections(2), round, hash, None)
```
meiersi-da: as above: test with mocks that fail with a retryable exception.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -151,16 +152,24 @@ abstract class CalculateRewardsTriggerBase(
     }
 
   private def getRootHash(round: Long)(implicit tc: TraceContext): Future[Option[Hash]] =
-    scanConnectionF.flatMap(_.getRewardAccountingRootHash(round)).map {
+    scanConnectionF.flatMap(_.getRewardAccountingRootHash(round)).flatMap {
```
meiersi-da: consider using `for` comprehension for better readability of these nested flatMaps

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -151,16 +152,24 @@ abstract class CalculateRewardsTriggerBase(
...
-          s"Scan cannot provide root hash for round $round"
+        logger.info(
+          s"Our scan cannot provide the root-hash for round $round, doing BFT read."
         )
+        bftScanConnectionF.flatMap(_.getRewardAccountingRootHash(round)).map {
+          case GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashOk(ok) =>
+            Some(new Hash(ok.rootHash))
+          case _ =>
+            logger.warn(
+              s"Could not obtain root-hash for round $round via BFT read."
```
meiersi-da: I'm a bit worried that this warning becomes quite noisy when larger parts of the system are down. I'm wondering whether it wouldn't be better to only warn if too many retries for the data failed.

I believe we should get this behavior for free thanks to using a task based trigger.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -151,16 +152,24 @@ abstract class CalculateRewardsTriggerBase(
     }
 
   private def getRootHash(round: Long)(implicit tc: TraceContext): Future[Option[Hash]] =
```
meiersi-da:
```suggestion
  private def getRootHash(round: Long)(implicit tc: TraceContext): Future[Hash] =
```

With this type, you can use retryable exceptions to report the reason for why the hash could not be fetched. No need to report `None` here only to convert it to an exception in `completeTask`.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -151,16 +152,24 @@ abstract class CalculateRewardsTriggerBase(
...
+        Future.successful(Some(new Hash(ok.rootHash)))
       case GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashUndetermined(_) =>
-        None
+        Future.successful(None)
       case GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashCannotProvide(_) =>
-        // TODO (#5623) replace with BFT read
-        throw new RuntimeException(
-          s"Scan cannot provide root hash for round $round"
+        logger.info(
+          s"Our scan cannot provide the root-hash for round $round, doing BFT read."
```
meiersi-da: I'd suggest to not log this on attempting the read, but when succeeding with the BFT read. This way we log the message only once.

We can log the message here as well, but that should then be at DEBUG level to avoid log spam due to retries.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
```diff
@@ -139,20 +140,29 @@ private[delegatebased] abstract class ProcessRewardsTriggerBase(
...
         )
+        bftScanConnectionF
+          .flatMap(_.getRewardAccountingBatch(round, batchHash))
+          .map {
+            case Some(response) => response
+            case None =>
+              throw new RuntimeException(
+                s"Batch could not be obtained via BFT read for round $round with hash $batchHash"
+              )
+          }
```
meiersi-da: Same logging and exception selection concerns apply as highlighted in the above comments. Please adjust the changes to this trigger as well.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/SvDsoAutomationService.scala
```diff
@@ -121,6 +137,13 @@ class SvDsoAutomationService(
         case scala.util.Failure(_) => scala.util.Success(())
       },
       NonNegativeDuration.tryFromDuration(timeouts.shutdownNetwork.duration),
+    ) :+ AsyncCloseable(
+      "bft-scan-connection",
+      bftScanConnectionF.transform {
+        case scala.util.Success(c) => scala.util.Try(c.close())
+        case scala.util.Failure(_) => scala.util.Success(())
+      },
+      NonNegativeDuration.tryFromDuration(timeouts.shutdownNetwork.duration),
```
meiersi-da:
```suggestion
      timeouts.shutdownNetwork,
```

usually this kind of conversion is not required... AFAIR

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
```diff
@@ -181,6 +195,146 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
...
+          s"Reset sv2's earliest-ingested round to $round", {
+            val sv2Db = sv2ScanBackend.appState.storage match {
+              case db: DbStorage => db
+              case other => fail(s"Expected DbStorage")
+            }
+            implicit val closeContext: CloseContext = CloseContext(sv2Db)
+            sv2Db
+              .update_(
+                sqlu"update app_activity_record_meta set earliest_ingested_round = $round",
+                "test.raiseEarliestIngestedRound",
```
meiersi-da:
```suggestion
                "test.increaseAppActivityMeta_EarliestIngestedRound",
```

meiersi-da: nit

meiersi-da (commented):
Thanks. I agree with the direction.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
dfordivam: The way `executeCall` has been written currently, for threshold of 1 it can form "consensus" even on a non `BftScanConnection.SuccessfulResponse` value. So if one scan is offline, its http error would be considered as a consensus.
And it is actually already causing an issue even for `getRewardAccountingRootHash`, because in the integration test the f == 0, and threshold ends up being 1.

This will require a somewhat redesign of `executeCall`, for both of these APIs.

dfordivam: @meiersi-da I have opened #5770 with a relatively benign fix, such that existing logic remains untouched. I will rebase this PR on #5770 and use `executeCall` here.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
dfordivam: Ok makes sense, and also this makes my integ test more robust, as I am making sv3 offline, and among [sv1, sv3, sv4], two will be picked randomly, and even with bft retries there is a chance of not reaching consensus in one try and causing warning.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -817,17 +823,76 @@ class BftScanConnection(
   ): Future[NonNegativeInt] =
     bftCall(_.getActivePhysicalSynchronizerSerial(), "getActivePhysicalSynchronizerSerial")
 
+  /** This is special because in addition to 'Ok' we can receive
+    * 'Undetermined' - This might indicate that scan is yet to process root hash for this round
+    * 'CannotProvide' - Indicates that scan does not have required app-activity data to provide a response
+    *
+    * So simple equality comparison on responses is not possible, and we treat
+    * the two non-Ok responses as a "no response" by throwing an exception so
```
dfordivam: @dfordivam fix comment

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -964,7 +1033,7 @@ object BftScanConnection {
           val f = connections.f
           BftCallConfig(
             connections = connections.open,
-            requestsToDo = 2 * f + 1,
+            requestsToDo = if (f == 0) connections.open.size else 2 * f + 1,
```
dfordivam, line 1064: Note: this is necessary because in our case even with 4 nodes, we end up with n == 3, and f == 0. So the integ test wont work if only one scan is fetched. Also I think this was wrong, since `2*f+1` should be valid for `f > 0` only.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1754,6 +1837,11 @@ object BftScanConnection {
     }
   }
 
+  // url makes each abstention a distinct key, so abstentions never group into a deciding quorum
+  final case class IgnoreResponse(url: Uri)
```
dfordivam, line 1876: We are still making use of `RuntimeException` but this makes the intention clearer. Also this does not cause any problems in logs

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
dfordivam: Added warn on disagreement. Info is added in the trigger itself.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
dfordivam: Added testing of retrying

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -150,19 +147,36 @@ abstract class CalculateRewardsTriggerBase(
...
+
+    def bftReadRootHash: Future[Hash] = {
+      for {
+        bftScan <- bftScanConnectionF
+        response <- bftScan.getRewardAccountingRootHash(round)
+      } yield response match {
+        case RewardAccountingRootHashOk(ok) =>
+          logger.info(s"Obtained the root-hash for round $round via BFT read.")
+          new Hash(ok.rootHash)
+        case _ => rootHashUnavailable("could not obtain root-hash via BFT read.")
```
dfordivam, line 164: Note the absence of warning. The trigger will retry and throw error/warning with its existing mechanism.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
dfordivam: Ok, only doing logging after succeeding

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
dfordivam: Added test for failure response

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
```diff
@@ -138,21 +140,39 @@ private[delegatebased] abstract class ProcessRewardsTriggerBase(
...
-        )
-    }
+  ): Future[GetRewardAccountingBatchResponse] = {
+    def bftReadBatch: Future[GetRewardAccountingBatchResponse] =
+      for {
+        bftScan <- bftScanConnectionF
+        response <- bftScan.getRewardAccountingBatch(round, batchHash)
+      } yield response match {
+        case Some(batch) =>
+          logger.info(s"Obtained the batch for round $round via BFT read.")
```
meiersi-da: How difficult would it be to log SVs from which we computed consensus? That would be helpful to know in case of a security incident.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -817,17 +823,76 @@ class BftScanConnection(
...
+                  roundNumber = roundNumber,
+                  rootHash = rootHash,
+                )
+              )
+            )
+          )
+        )
+  }
 
+  /** Here we query all scans in parallel and resolve to the first response that has the
```
meiersi-da: I suspect that using the first response causes an availability problem, if a malicious SV answers much more quickly than the other ones. 

What we actually want is to just ask one random SV that should have the data; and have the trigger's own retry loop take care of picking another SV if that fails.

meiersi-da (commented):
Partial review. Submitted to get access to the comment.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
meiersi-da, line 1064:
```suggestion
            // Play it safe wrt availability in case we have no fault tolerance.
            requestsToDo = if (f == 0) connections.open.size else 2 * f + 1,
```

meiersi-da: ack.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1343,6 +1412,24 @@ object BftScanConnection {
           )
       } yield domainScans.map(scanInfo => DsoScan(scanInfo.publicUrl, scanInfo.svName))
     }
+
+    def getPeerScansFromDsoRules(store: DsoRulesStore, ownSvParty: String)(implicit
```
meiersi-da:
```suggestion
    def getPeerScansFromDsoRules(store: DsoRulesStore, ownSvParty: PartyId)(implicit
```

we try to convert as late as possible to avoid passing in the wrong strings.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1343,6 +1412,24 @@ object BftScanConnection {
...
+        tc: TraceContext,
+        ec: ExecutionContext,
+    ): Future[Seq[DsoScan]] =
+      store.getDsoRulesWithSvNodeStates().map { rulesAndStates =>
+        rulesAndStates.svNodeStates.values
+          .filterNot(_.payload.sv == ownSvParty)
```
meiersi-da:
```suggestion
          .filterNot(_.payload.sv == ownSvParty.toProtoPrimitive)
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
meiersi-da: Thx

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
dfordivam: Ok, its simple I can make config `requestFrom = Random.shuffle(callConfig.connections).take(1)`

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
dfordivam: Not that difficult. The `logDisagreements` could be passed a bool to log consensus too. Perhaps make `disagreementLogLevel` part of a struct and pass it to `logDisagreements`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -817,17 +823,76 @@ class BftScanConnection(
...
+  /** Here we query all scans in parallel and resolve to the first response that has the
+    * batch because the contents of batch are verifiable using the hash.
+    */
   override def getRewardAccountingBatch(roundNumber: Long, batchHash: String)(implicit
       ec: ExecutionContext,
       tc: TraceContext,
-  ): Future[Option[GetRewardAccountingBatchResponse]] =
-    bftCall(_.getRewardAccountingBatch(roundNumber, batchHash), "getRewardAccountingBatch")
+  ): Future[Option[GetRewardAccountingBatchResponse]] = {
+    val callConfig = BftCallConfig.default(scanList.scanConnections).copy(targetSuccess = 1)
```
meiersi-da: If we set `requestsToDo` == 1, then we should get the behavior of just picking a random connection or not?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -150,19 +147,36 @@ abstract class CalculateRewardsTriggerBase(
...
-        None
-      case GetRewardAccountingRootHashResponse.members.RewardAccountingRootHashCannotProvide(_) =>
-        // TODO (#5623) replace with BFT read
-        throw new RuntimeException(
-          s"Scan cannot provide root hash for round $round"
-        )
+  private def getRootHash(round: Long)(implicit tc: TraceContext): Future[Hash] = {
+    def rootHashUnavailable(reason: String): Nothing =
+      throw Status.FAILED_PRECONDITION
+        .withDescription(s"For round: $round $reason")
```
meiersi-da:
```suggestion
        .withDescription(s"For round $round: $reason")
```

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -150,19 +147,36 @@ abstract class CalculateRewardsTriggerBase(
...
+        .asRuntimeException()
+
+    def bftReadRootHash: Future[Hash] = {
+      for {
+        bftScan <- bftScanConnectionF
+        response <- bftScan.getRewardAccountingRootHash(round)
+      } yield response match {
+        case RewardAccountingRootHashOk(ok) =>
+          logger.info(s"Obtained the root-hash for round $round via BFT read.")
```
meiersi-da, line 162: Same here: log whose responses were used to form consensus for debugging purposes. Probably using the Scan URLs is a good idea.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -150,19 +147,36 @@ abstract class CalculateRewardsTriggerBase(
...
+      }
     }
 
+    for {
+      ownScan <- scanConnectionF
+      response <- ownScan.getRewardAccountingRootHash(round)
+      rootHash <- response match {
+        case RewardAccountingRootHashOk(ok) => Future.successful(new Hash(ok.rootHash))
+        case RewardAccountingRootHashUndetermined(_) =>
+          rootHashUnavailable("scan has not yet computed the root hash.")
```
meiersi-da:
```suggestion
          rootHashUnavailable("our own Scan has not yet computed the root hash.")
```

meiersi-da (approved):
Nice work! Looks good modulo some small tweaks 👏

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
meiersi-da: I'd prefer to not have to correlate two log messages (and pay for their ingestion). What about returning the Scan URLs from which consensus was reached with the response?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
dfordivam: ah right, this is the correct way `bftCall` does not even take `requestFrom` as input

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
meiersi-da: why not set `requestsToDo = 1`?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
rautenrieth-da: What happens if you add `requestsToDo = 1` to this? (btw, instead of the copy I would create a new helper, something like `BftCallConfig.randomSingleCall`).

`bftCall` has an internal retry that should pick a new connection for each retry attempt. Would this properly keep trying until we get one successful response?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
dfordivam: Changing the return type would be a much bigger change. and even go against the current design of ScanConnections. I don't see any clean way of achieving this.
We could add a TODO in day-2 concerns to do this. It would probably make sense to write a custom class in sv-app to do the BFT read and not even use BftScanConnection.

meiersi-da: OK. Let's create a Day 2 concern ticket.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
```diff
@@ -967,4 +1017,166 @@ class BftScanConnectionTest
...
+      val connections = getMockedConnections(n = 3)
+      connections.foreach(makeMockReturnBatch(_, round, hash, None))
+      val bft = getBft(connections)
+
+      for {
+        resp <- bft.getRewardAccountingBatch(round, hash)
+      } yield resp should be(None)
+    }
+
+    "on multiple retries, we can obtain batch even in prescense of failing scans" in {
```
dfordivam, line 1163: @meiersi-da @rautenrieth-da Note the new behaviour of `getRewardAccountingBatch` with `randomSingleCall`

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
dfordivam: Logging this in `BftScanConnection` intenally. As discussed logging here would be a Day 2 concern.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
dfordivam: Added #5844

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
```diff
@@ -138,21 +140,39 @@ private[delegatebased] abstract class ProcessRewardsTriggerBase(
...
+          batch
+        case None =>
+          throw Status.FAILED_PRECONDITION
+            .withDescription(
+              s"Batch could not be obtained via BFT read for round $round with hash $batchHash"
```
meiersi-da:
```suggestion
              s"Failed to obtain batch for round $round with hash $batchHash via BFT read."
```

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
meiersi-da:
```suggestion
          logger.info(s"Obtained batch for round $round with hash $batchHash via BFT read.")
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -922,14 +994,29 @@ object BftScanConnection {
...
     keyToGroupResponses(consensusResponse).foreach { consensusResponseKey =>
-      responses.remove(consensusResponseKey)
-      responses.forEach { (disagreeingResponse, scanUrls) =>
-        logger.info(
-          s"Scans $scanUrls disagreed with the Consensus $consensusResponse and instead returned $disagreeingResponse"
+      val agreeingScanUrls = responses.remove(consensusResponseKey)
+      consensusLogConfig.agreementLogLevel.foreach { level =>
+        LoggerUtil.logAtLevel(
+          level,
+          s"Scans $agreeingScanUrls agreed on the Consensus $consensusResponse",
```
meiersi-da: the response could be rather large or not? I'm wary of printing that. It should not be required provided the TraceContext is passed through correctly. This way the trace-id can be used to correlate the agreement and the task completed in the trigger.
```suggestion
          s"Reached consensus from:\n${agreeingScanUrls.mkString("\n")",
```

so the URLs are easy to read in logs.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -922,14 +994,29 @@ object BftScanConnection {
...
       }
+      responses.forEach { (disagreeingResponse, scanUrls) =>
+        val shouldLog = disagreeingResponse match {
+          case _: SuccessfulResponse[?] => true
+          case _ => !consensusLogConfig.onlyLogDisagreementsInSuccessResponse
+        }
+        if (shouldLog) {
+          LoggerUtil.logAtLevel(
+            consensusLogConfig.disagreementLogLevel,
+            s"Scans $scanUrls disagreed with the Consensus $consensusResponse and instead returned $disagreeingResponse",
```
meiersi-da: add line-wrapping and consider pretty-printing via `show`, so we get proper ellipsis. Generally make sure that we don't log payload of unbounded size.

meiersi-da: consider `show` optional, as this might not be used with large enough responses. Wrapping the log line is still useful; e.g.,
```
The following Scan URLs disagreed with consensus:
  url1
  url2 
consensus response: resp
disagreeing response: resp'
```

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
```diff
@@ -967,4 +1017,166 @@ class BftScanConnectionTest
...
+      makeMockReturnRootHashUndetermined(connections(2), round)
+      makeMockFail(connections(3), notFoundFailure)
+      val bft = getBft(connections)
+
+      // With n=4, we query only two connections randomly, and even with
+      // retries it can sometimes fail. This eventually is here to avoid flakyness
+      //
+      // The agreement is logged shortly after the call resolves on
+      // consensus. The trailing delay keeps the suppression
+      // window open long enough to capture that log.
```
meiersi-da: You should be able to use: https://github.com/canton-network/splice/blob/aea1ac435279621d6ff7b6ff84b390969f5b20c6/canton/community/testing/src/main/scala/com/digitalasset/canton/BaseTest.scala#L782-L799

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
```diff
@@ -967,4 +1017,166 @@ class BftScanConnectionTest
...
+      val connections = getMockedConnections(n = 4)
+      makeMockReturnRootHashOk(connections(0), round, "aabb")
+      makeMockReturnRootHashOk(connections(1), round, "aabb")
+      makeMockReturnRootHashOk(connections(2), round, "ccdd")
+      makeMockReturnRootHashOk(connections(3), round, "ccdd")
+      val bft = getBft(connections)
+
+      // The disagreement is logged shortly after the call resolves on
+      // consensus. The trailing delay keeps the suppression
+      // window open long enough to capture that log.
```
meiersi-da: same here: https://github.com/canton-network/splice/blob/aea1ac435279621d6ff7b6ff84b390969f5b20c6/canton/community/testing/src/main/scala/com/digitalasset/canton/BaseTest.scala#L782-L799

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnectionTest.scala
meiersi-da: looks good.

meiersi-da (approved):
Very nice work! Thanks a lot!

PR #5769, opened 2026-06-02
Undo SvDsoStore version bump

rautenrieth-da:
This PR undoes the version bump of SvDsoStore from https://github.com/canton-network/splice/pull/5552.

Bumping the store descriptor would lead to significant downtime of the sv app, and thus to the sv node missing out on rewards. The version bump is not required as the new contracts ingested in the filter were not created before the filter was created.

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
@@ -92,6 +92,12 @@ class DbSvDsoStore(
       interfaceViewsTableNameOpt = None,
       // Any change in the store descriptor will lead to previously deployed applications
       // forgetting all persisted data once they upgrade to the new version.
+      // WARNING: Reinitializing the acs store is a very expensive operation, as it currently fetches the full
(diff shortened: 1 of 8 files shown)


meiersi-da (approved):
Thank you!

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
```diff
@@ -93,7 +93,7 @@ class DbSvDsoStore(
       // Any change in the store descriptor will lead to previously deployed applications
       // forgetting all persisted data once they upgrade to the new version.
       acsStoreDescriptor = StoreDescriptor(
-        version = 3,
+        version = 2,
```
meiersi-da, line 102:
```suggestion
        version = 2, // Note for next bump: SKIP version 3, as it was already used in the past, and there at least on CILR data for it might already exist. 
        // See https://github.com/canton-network/splice/pull/5769
```

rautenrieth-da: Will attempt to clean up the database first. If it doesn't work, I'll add the note in a separate PR.

PR #5770, opened 2026-06-02
Bft read: ignore consensus on failure response from single scan

dfordivam:
The `BftScanConnection.executeCall` can currently form a consensus on failure responses, like the `StreamTcpException` if the scan is temporarily unavailable.

This is particularly problematic if `nTargetSuccess` == 1, as the failure from an offline scan would happen quicker than the response received from online scans, and the `executeCall` thus would always ignore the success responses.

Note: the  `BftScanConnection.executeCall` can still form a consensus on  `StreamTcpException` for `nTargetSuccess` > 1, if a number of scans are offline.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
@@ -858,7 +858,13 @@ object BftScanConnection {
               (_, scans) => scan.url :: Option(scans).getOrElse(List.empty),
             )
 
-          if (agreements.size == nTargetSuccess) { // consensus has been reached
+          // Only a quorum of matching successful responses should decide the call; failures
+          // fall through to ConsensusNotReached.
+          val isSuccessfulResponse = key match {
+            case _: SuccessfulResponse[?] => true
+            case _ => false
+          }
+          if (isSuccessfulResponse && agreements.size == nTargetSuccess) { // consensus has been reached
             finalResponse.tryComplete(response): Unit
           }
 
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
@@ -858,13 +858,13 @@ object BftScanConnection {
               (_, scans) => scan.url :: Option(scans).getOrElse(List.empty),
             )
 
-          // Only a quorum of matching successful responses should decide the call; failures
-          // fall through to ConsensusNotReached.
-          val isSuccessfulResponse = key match {
+          // In the special case of nTargetSuccess == 1, consider the response only if it is a success
+          // Otherwise a single HTTP error or network failure would prevent reading the responses from others
+          val considerResponseForQuorum = key match {
             case _: SuccessfulResponse[?] => true
-            case _ => false
+            case _ => !(nTargetSuccess == 1 && requestFrom.size != 1)
           }
-          if (isSuccessfulResponse && agreements.size == nTargetSuccess) { // consensus has been reached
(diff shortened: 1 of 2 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -858,7 +858,13 @@ object BftScanConnection {
               (_, scans) => scan.url :: Option(scans).getOrElse(List.empty),
             )
 
-          if (agreements.size == nTargetSuccess) { // consensus has been reached
+          // In the special case of nTargetSuccess == 1, consider the response only if it is a success
+          // Otherwise a single HTTP error or network failure would prevent reading the responses from others
+          val considerResponseForQuorum = key match {
+            case _: SuccessfulResponse[?] => true
```
meiersi-da: consider all Http responses to be successes. So that only the other exceptions are ignored.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -858,7 +858,13 @@ object BftScanConnection {
               (_, scans) => scan.url :: Option(scans).getOrElse(List.empty),
             )
 
-          if (agreements.size == nTargetSuccess) { // consensus has been reached
+          // In the special case of nTargetSuccess == 1, consider the response only if it is a success
```
meiersi-da: add a `TODO(tech-debt): this problem also exists for `nTargetSuccess > 1`, but we don't handle it yet, as we didn't need that case to work. sorry.

meiersi-da (approved):
ty

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
dfordivam: Actually the issue won't happen for nTargetSuccess > 1, for the `StreamTcpException` kind of exceptions, as these exceptions don't have equality and they will not form consensus.

And I recalled this just now as I was writing the issue description for the tech-debt issue. (during yesterday's call this thing didn't click)

So I don't add the comment with a TODO right? @meiersi-da

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
dfordivam: fixed

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
dfordivam: I did add a test and mention that we still have the tech-debt of scan endpoints not able to specify valid failure responses.

```
    // Unlike a transport exception, an http failure still counts towards the quorum.
    // Although this behaviour is mostly a result of the tech-debt of the
    // inability for the scan endpoints to specify what responses are expected (like 404)
    "but let a single http failure decide the call when n == 2" in {
      val connections = getMockedConnections(n = 2)
      makeMockFail(connections.head, notFoundFailure)
      val delayedSuccess =
        org.apache.pekko.pattern.after(200.millis, actorSystem.scheduler)(
          Future.successful(partyIdA)
        )
      connections.tail.foreach(c => when(c.getDsoPartyId()).thenReturn(delayedSuccess))

      for {
        failure <- BftScanConnection
          .executeCall(call, connections, nTargetSuccess = 1, logger)
          .failed
      } yield failure should be(notFoundFailure)
    }

```

meiersi-da: ty. sounds good.

PR #5772, opened 2026-06-02
implement physical synchronizer config overrides

### LINKED ISSUE #5524: Make sequencer configs configurable per migration id
For things like roll forward LSU we really want to set different versions.

mblaze-da:
Fixes https://github.com/canton-network/splice/issues/5524

@moritzkiefer-da please let me know if this satisfies the requirements you had in mind when creating the issue.

I'm marking this as draft as I asked one question in a comment.


cluster/pulumi/common-sv/src/singleSvConfig.ts
```diff
@@ -188,7 +174,30 @@ const SingleSvConfigSchema = z
     periodicSnapshots: z.object({ topology: TopologySnapshotSchema.optional() }).optional(),
     versionOverride: CnChartVersionSchema.optional(),
   })
-  .strict();
+  .strict()
+  .transform(({ physicalSynchronizerOverrides, sequencer, mediator, ...svConfig }) => {
+    // Implicit config merging is confusing but for now I don't really see a better way of introducting these overrides.
```
moritzkiefer-da, line 181: so we're basically reintroducing the issues that made us move away from `default`? What's the issue with actually listing it under each migration id and sharing the config through includes like we do across SVs now?

moritzkiefer-da: don't love it but the alternative of different configs also sucks so probably this is the best option given that we expect the per physical synchronizer id overwrites to be rare (mainly around roll forward lsus)

moritzkiefer-da (commented):
thx! lgtm, also interested in @nicu-da's input

cluster/pulumi/common-sv/src/singleSvConfig.ts
nicu-da: I'm trying to understand which configs we are trying to override per migration. The version is already overridable right?

moritzkiefer-da: for roll forward lsus we need so set the time configs in the sequencer additionalEnvVars

nicu-da (approved):
LGTM, not a lot of ideas on how to improve this tbh

cluster/pulumi/common-sv/src/singleSvConfig.ts
mblaze-da: The main reason I decided against going with creating a common physical synchronizer config and `!include`ing it for each PSID was due to additional modifications it would incur in synchronizerMigration config editing automation. I just thought it would be a bit fragile to assume include paths in those scripts.

cluster/pulumi/sv-canton/src/decentralizedSynchronizerNode.ts
```diff
@@ -169,7 +171,10 @@ abstract class InStackDecentralizedSynchronizerNode
             },
           },
           livenessProbeInitialDelaySeconds: domainLivenessProbeInitialDelaySeconds,
-          additionalJvmOptions: getAdditionalJvmOptions(svConfig.sequencer?.additionalJvmOptions),
+          // These are used both for sequencer and mediator. This is probably wrong, isn't it?
```
mblaze-da: @nicu-da @moritzkiefer-da do we want to do something about this now?

moritzkiefer-da: that was already the case before though no? If so definitely don't change it in this PR but it seems maybe worth creating a tech debt for

nicu-da: Might be worth checking if we actually set that anywhere. a bit weird the way the chart is built as well. maybe just create a tech debt

nicu-da (commented):
it might be worth also adding an example to the mock config to see how it would work out

PR #5774, opened 2026-06-02
Fix session storage clear flake

jagathweerasinghe-da:
[ci]

Fix for: https://github.com/DACH-NY/cn-test-failures/issues/8586


ray-roestenburg-da (approved):

stephencompall-DA:
Why would removing the retry on the idempotent length read potentially fix the flake?

stephencompall-DA (approved):
Is the idea something like "you clear, browser adds something else to session storage, check then happens"? 👍

jagathweerasinghe-da:
> Is the idea something like "you clear, browser adds something else to session storage, check then happens"? 👍

Exactly, otherwise there is no reason why we should get a non-zero value intermittently.

PR #5792, opened 2026-06-03
Make isV2Allocation safer

OriolMunoz-da:
Before, if the field didn't exist on either v1 or v2 it wouldn't cause a compiler error. This should cause one.


meiersi-da (approved):
ty

PR #5793, opened 2026-06-03
Set template id filters in acs queries

### LINKED ISSUE #4862: Specify template filter in ACS query
We don't specify a template filter which makes some sense for tx history where we want to see full trees.

But we inherited that behavior for the ACS queries on store (re)initialization.

A good example where that caused issues is the new traffic based reward stores. Those should query a handful of templates but instead queried the full ACS which caused a bunch of issues.

moritzkiefer-da:
fixes #4862

[ci]

Not sure how to test it before merging, I did confirm in the logs that it seems to set things correctly.

I also checked with the indexer team that these queries are expected to be fast even if you pass lots of templates (like we do for example for the dso stores). The real test ofc will be sv runbook reonboarding on cilr.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AutoIgnoreUnresponsivePartiesIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AutoIgnoreUnresponsivePartiesIntegrationTest.scala
@@ -170,6 +170,8 @@ class AutoIgnoreUnresponsivePartiesIntegrationTest
       )
 
       clue("Disconnect alice's participant from the synchronizer") {
+        // stop to avoid log noise.
+        aliceValidatorBackend.stop()
         aliceValidatorBackend.participantClient.synchronizers.disconnect_all()
         aliceValidatorBackend.participantClient.synchronizers.is_connected(
           synchronizerId
@@ -192,9 +194,12 @@ class AutoIgnoreUnresponsivePartiesIntegrationTest
           sv1Backend.dsoDelegateBasedAutomation.expiredAmuletIgnoredPartiesStore.getAll should contain(
             aliceParty
           )
-          aliceWalletClient.list().amulets should have length numAmulets.toLong
-          aliceWalletClient.list().lockedAmulets should have length numAmulets.toLong
         },
       )
+
+      // reconnect or other tests might get unhappy, in particular `withNoVettedPackages` gets confused if the nodes is disconnected.
+      clue("Reconnect alice's participant") {
+        aliceValidatorBackend.participantClient.synchronizers.reconnect_all()
+      }
   }
 }
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
@@ -22,6 +22,7 @@ import com.digitalasset.canton.lifecycle.FutureUnlessShutdown
 import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
 import com.digitalasset.canton.logging.{NamedLoggerFactory, NamedLogging}
 import com.digitalasset.canton.protocol.LocalRejectError.ConsistencyRejections.InactiveContracts
+import com.digitalasset.canton.time.Clock
 import com.digitalasset.canton.topology.admin.grpc.TopologyStoreId
 import com.digitalasset.canton.topology.{Namespace, PartyId, SynchronizerId, UniqueIdentifier}
 import com.digitalasset.canton.tracing.TraceContext
@@ -157,10 +158,18 @@ class BaseLedgerConnection(
       filter: IngestionFilter,
       offset: Long,
       restartSettings: RestartSettings,
+      clock: Clock,
+      packageVersionSupport: PackageVersionSupport,
   )(implicit
       tc: TraceContext
   ): Source[BaseLedgerConnection.ActiveContractsItem, NotUsed] =
-    activeContracts(filter.toEventFormat, offset, restartSettings)
+    Source
+      .futureSource {
+        filter
+          .toAcsEventFormat(packageVersionSupport, clock)
+          .map(activeContracts(_, offset, restartSettings))
+      }
+      .mapMaterializedValue(_ => NotUsed)
 
   def getContract(
       contractId: ContractId[?],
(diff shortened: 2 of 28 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AutoIgnoreUnresponsivePartiesIntegrationTest.scala
```diff
@@ -192,9 +192,12 @@ class AutoIgnoreUnresponsivePartiesIntegrationTest
           sv1Backend.dsoDelegateBasedAutomation.expiredAmuletIgnoredPartiesStore.getAll should contain(
             aliceParty
           )
-          aliceWalletClient.list().amulets should have length numAmulets.toLong
```
moritzkiefer-da, line 195: this assertion did nothing, we already disconnect before so even if the contract gets archived we wouldn't notice. But it means I can't stop it which makes stuff annoying.

Not actually related to this PR at all but it did flake here.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AutoIgnoreUnresponsivePartiesIntegrationTest.scala
```diff
@@ -192,9 +194,12 @@ class AutoIgnoreUnresponsivePartiesIntegrationTest
...
-          aliceWalletClient.list().lockedAmulets should have length numAmulets.toLong
         },
       )
+
+      // reconnect or other tests might get unhappy, in particular `withNoVettedPackages` gets confused if the nodes is disconnected.
+      clue("Reconnect alice's participant") {
```
moritzkiefer-da, line 201: not really related to this PR but I did trigger a flake due to this.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/ScanApp.scala
OriolMunoz-da: do you want to keep this one around? it's shadowing the one above (and if you do want to keep it, maybe document why)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
```diff
@@ -524,19 +540,71 @@ object MultiDomainAcsStore extends StoreErrors {
     */
   final case class IngestionFilter(
       primaryParty: PartyId,
+      includeTemplates: Seq[
+        (
+            PackageQualifiedName,
+            (PackageVersionSupport, CantonTimestamp) => (
+                TraceContext
+            ) => Future[PackageVersionSupport.FeatureSupport],
```
OriolMunoz-da, line 542: nit: can we introduce a type alias for VersionGuard? it gets pretty unwieldy

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
```diff
@@ -157,10 +158,18 @@ class BaseLedgerConnection(
...
+      clock: Clock,
+      packageVersionSupport: PackageVersionSupport,
   )(implicit
       tc: TraceContext
   ): Source[BaseLedgerConnection.ActiveContractsItem, NotUsed] =
-    activeContracts(filter.toEventFormat, offset, restartSettings)
+    Source
+      .futureSource {
+        filter
+          .toAcsEventFormat(packageVersionSupport, clock)
```
OriolMunoz-da, line 169: This is only called once.
The reason this works for ACS is that given the offset from which you fetch ACS a package is either supported and there's (probably) contracts from it, or it isn't supported and there's no contracts from said package in the ACS

In the updates function you commented:
```
// The way we parse transaction trees currently requires no filtering.
    // **We may want to consider filtering for stores that just use the ACS**
    // and don't care about tx trees but the extra overhead has not been an issue so far.
    def toUpdatesEventFormat: EventFormat = toEventFormat(
```
We'd have to change the format on the fly if we wanted that, which is non-trivial

(nothing to do here, just pointing out that filtering in updates even without txlog is not so straightfwd afaict)

OriolMunoz-da (approved):
thanks!

> Not sure how to test it before merging, I did confirm in the logs that it seems to set things correctly.

You can add a StoreTest with a fake "PackageVersionSupport" and check that a contract in a filter that is not yet supported according to the PackageVersionSupport is not ingested.
This way we have at least _some_ test for the versionGuard stuff.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
moritzkiefer-da: > (nothing to do here, just pointing out that filtering in updates even without txlog is not so straightfwd afaict)

yep agreed, let me fold that into the comment

moritzkiefer-da:
> You can add a StoreTest with a fake "PackageVersionSupport" and check that a contract in a filter that is not yet supported according to the PackageVersionSupport is not ingested. This way we have at least some test for the versionGuard stuff.

Good idea, I mainly meant test performance (wrote this before I even realized I need the version guards) but definitely makes sense to add a store test. will do before merging.

OriolMunoz-da:
> Good idea, I mainly meant test performance

You should be able to run performance ingestion tests on the PR [via GHA](https://github.com/canton-network/splice/actions/workflows/performance_tests.yml), or let them be run daily on main and check the dashboard the day after (we might even get alerts for it, I forgot).

moritzkiefer-da:
> You should be able to run performance ingestion tests on the PR [via GHA](https://github.com/canton-network/splice/actions/workflows/performance_tests.yml), or let them be run daily on main and check the dashboard the day after (we might even get alerts for it, I forgot).

But I care mainly about the participant performance here not our store performance. Our stores will just see far less data so if the participant can serve it faster it's obviously faster (e.g. for the traffic store reingestion we have ~100 total contracts for an ACS of > 1 million). Do we test that in the performance ingestion tests? I thought those directly hit our stores and don't involve the participant?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala
moritzkiefer-da: added

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
moritzkiefer-da: done

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/ScanApp.scala
moritzkiefer-da: good catch, removed

OriolMunoz-da:
they don't test participant no, I thought you wanted to measure any changes by calling the packageversionsupport, mb

PR #5807, opened 2026-06-03
feat: port backports from splice-shared-gha

krzysztofczyz-da:
Also: remove one unneeded action `.github/actions/setup/action.yml`

Note: using a separate dir to make it easy to see that this a separate component (also easier to manage).


mblaze-da:
@krzysztofczyz-da, I wonder if we shouldn't just move stuff from `gha-scripts` into `.github` to keep the directory structure simple.

scripts/todo/src/checkTodos.sc
```diff
@@ -312,7 +312,9 @@ val todoStyleExcludePrefixes =
     "experiments/",
     "CONTRIBUTING.md",
     "token-standard/dependencies",
-    ".github/actions/tests/split_tests/dist/"
+    ".github/actions/tests/split_tests/dist/",
+    "gha-scripts/actions/parse_backport_comments/",
+    "gha-scripts/actions/backport_reminder"
```
mblaze-da, line 317:
```suggestion
    "gha-scripts/actions/**/*.js"
```
would be more correct but I guess we can't use globs here.

mblaze-da (approved):
I'd consider moving to `.github` now if there are no issues with that. Thanks!

krzysztofczyz-da:
> I'd consider moving to `.github` now if there are no issues with that. Thanks!

I think this setup is a bit easier to maintain as it is easier to spot what this thing does, where the code lies, etc.
I'm happy to reconsider if this proves to be problematic in the future

mblaze-da:
> I think this setup is a bit easier to maintain as it is easier to spot what this thing does, where the code lies, etc. I'm happy to reconsider if this proves to be problematic in the future

I don't think it will cause any particular, technical issues. Just that it's a bit confusing why we split off just this part of our GHA scripts while the rest stays in `.github`.

Issue #5135, opened 2026-04-21
Integrate session signing keys


martinflorian-da:
- [x] Wait for it to be ready (s.a. https://github.com/DACH-NY/canton/issues/30003): integrate/test/enable session signing keys by default (for better performance + lower KMS costs) :: upstream says "ready for testing"!
- [x] regroup; notes
  - docs added here: https://github.com/DACH-NY/canton/pull/31838/
- [x] Enable on those integration tests that use KMS
- [x] Enable on CILR for testing
  - [x] validator1
  - [x] DA-1
  - [x] sv-13
- [x] figure out https://github.com/DACH-NY/canton-network-internal/issues/5758
- [x] Add pulumi/Helm logic to add session signing key configs automatically whenever we set `kms:`
- [x] Add to our docs
- [x] Enable on DevNet
  - [x] validator1
  - [x] DA-1
  - [x] sv runbook
- [x] Enable on TestNet DA-1
  - [x] validator1
  - [x] DA-1
- [x] internal cleanup (look for TODO)
- [ ] Enable on MainNet DA-1
  - [x] Review effect on KMS usage and costs; share with upstream
- [ ] (testing / parameter tuning)
  - [ ] Review effect on KMS usage and costs; share with upstream

martinflorian-da:
<img width="1802" height="672" alt="Image" src="https://github.com/user-attachments/assets/305c0f26-5481-42e0-a1fe-a5cd303ef2b0" />

Enabled on MainNet, finally 🎉  (not signatures rate dropping waaaay down; the decryptions rate increasing is most likely just due to the participant restart and will calm down again in a few minutes)

FYI @JoaoSa-da 

I'm actually not aware of anything TODO w.r.t.  "(testing / parameter tuning)", so going to close this issue now 🎉

If we notice something we'll yell, and if you think of something to check please yell as well! 🙏

thibault-da:
NIce! @martinflorian-da Is that both on participant and sequencer node?

martinflorian-da:
@thibault-da Only participant! We have no KMS on sequencers yet.

PR #5835, opened 2026-06-04
Enable session signing keys on KMS integration tests

martinflorian-da:
Part of #5135

~Holding off merge until PV35 is the default, so we don't need to hard-code the PV in that test.~ :heavy_check_mark: 

[ci]


moritzkiefer-da:
> Holding off merge until PV35 is the default, so we don't need to hard-code the PV in that test.

Mostly recording this so it doesn't get lost: The problem is not that PV 35 is not the default, it is in integration tests. What will be a problem is the daily compat tests againt PV 34. So you either need to exclude it from those or wait until they are gone (after mainnet is PV 35)

martinflorian-da:
> Mostly recording this so it doesn't get lost: The problem is not that PV 35 is not the default, it is in integration tests.

Not if we do `.clearConfigTransforms()`, which is what we do here.

> What will be a problem is the daily compat tests againt PV 34. So you either need to exclude it from those or wait until they are gone (after mainnet is PV 35)

Ah yes, I missed that...

martinflorian-da:
ftr the test was green: https://github.com/canton-network/splice/actions/runs/26960289065/job/79548660320?pr=5835

moritzkiefer-da (approved):
thx

PR #5839, opened 2026-06-04
Add extra logging and metrics around ACS ingestion

OriolMunoz-da:
[Context](https://github.com/DACH-NY/cn-test-failures/issues/8718)
(but it's anyway nice to have those extra metrics)


moritzkiefer-da (approved):
thx

PR #5840, opened 2026-06-04
Test V2 settlement of a V1 Amulet allocation

OriolMunoz-da:
Fixes https://github.com/canton-network/splice/issues/4949


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardAllocationHandler.scala
```diff
@@ -135,12 +135,16 @@ class HttpTokenStandardAllocationHandler(
           body.excludeDebugFields.getOrElse(false)
         )
         externalPartyAmuletRules <- store.getExternalPartyAmuletRules()
-        // TODO (#4949): this only lists V2 allocations, but it should also do V1
-        allocations <- contractFetcher.lookupContractsById(
```
meiersi-da, line 139: The same problem of dealing with both v1 and v2 allocation holds for v2 withdraw and v2 cancel of an allocation created using the v1 factory. Or do you have that tested and handled already?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardAllocationHandler.scala
```diff
@@ -135,12 +135,16 @@ class HttpTokenStandardAllocationHandler(
...
           amuletallocationv2.AmuletAllocationV2.COMPANION
         )(settleBatch.allocations.asScala.toSeq.map(_.allocationCid))
+        allocationsV1 <- contractFetcher.lookupContractsById(
+          amuletallocationv1.AmuletAllocation.COMPANION
+        )(settleBatch.allocations.asScala.toSeq.map(_.allocationCid))
+        lockedAmuletCids = allocationsV2.flatMap(_.payload.lockedAmulet.toScala) ++
+          allocationsV1.map(_.payload.lockedAmulet)
         lockedAmulets <- contractFetcher.lookupContractsById(LockedAmulet.COMPANION)(
-          allocations.flatMap(_.payload.lockedAmulet.toScala)
+          lockedAmuletCids
```
meiersi-da: there are probably duplicates here or not? does that matter?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardV2AllocationIntegrationTest.scala
```diff
@@ -108,42 +110,74 @@ class TokenStandardV2AllocationIntegrationTest
         settleBatch
       )
 
+    val bobMissingAllocationArgs = new allocationinstructionv2.AllocationFactory_Allocate(
```
meiersi-da, line 116:
```suggestion
    // Arguments to create Bob's allocation to receive Alice's incoming transfer.
    // The allocation for it will be created by `OTCTrade_Settle`, but we need them
    // here as well to fetch the choice context that we need to pass in.
    val bobMissingAllocationArgs = new allocationinstructionv2.AllocationFactory_Allocate(
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardV2AllocationIntegrationTest.scala
```diff
@@ -249,17 +283,8 @@ class TokenStandardV2AllocationIntegrationTest
...
-    val venuePartyHint = s"venue-party-${Random.nextInt()}"
-    val venueParty = splitwellValidatorBackend.onboardUser(
-      splitwellWalletClient.config.ledgerApiUser,
-      Some(
-        PartyId.tryFromProtoPrimitive(
-          s"$venuePartyHint::${splitwellValidatorBackend.participantClient.id.namespace.toProtoPrimitive}"
-        )
-      ),
-    )
+    // We use the same venue party for bob so we can setup the TradeSettlementAgreement
```
meiersi-da:
```suggestion
    // We use the same venue party for bob so we can setup the TradeSettlementAgreement using a submit with `actAs = Seq(venueParty, bobParty)`.
```

meiersi-da (approved):
Nice work! Thanks.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardAllocationHandler.scala
OriolMunoz-da: there's a cancel test failing atm with a missing allocation context, as bob creates the v1 alloc and we were only fetching v2

meiersi-da: OK. Do we also have a test for `Withdraw`?

OriolMunoz-da: nope, adding it already :)

PR #5851, opened 2026-06-08
Merge latest main onto token-standard-v2-upcoming

OriolMunoz-da:
Another Monday another merge

# ORIOL: DO NOT SQUASH MERGE


meiersi-da (approved):
ty

PR #5854, opened 2026-06-08
Support overriding full image names in Helm

pasindutennage-da:
Fix https://github.com/DACH-NY/canton-network-internal/issues/5608


cluster/helm/splice-global-domain/templates/mediator.yaml
```diff
@@ -46,7 +46,7 @@ spec:
         fsGroup: 1001
       containers:
       - name: mediator
-        image: "{{ .Values.imageRepo }}/canton-mediator:{{ .Chart.AppVersion }}{{ ((.Values.imageDigests).canton_mediator) }}"
+        image: {{ .Values.imageRepo }}/{{ .Values.mediatorImageName | default "canton-mediator" }}:{{ .Chart.AppVersion }}{{ ((.Values.imageDigests).canton_mediator) }}
```
nicu-da: why use the default in the templates and also in the default values? I suggest we just set them in the values files.

cluster/helm/splice-global-domain/templates/mediator.yaml
```diff
@@ -113,7 +113,7 @@ spec:
       restartPolicy: Always
       initContainers:
         - name: pg-init
-          image: postgres:14
+          image: {{ .Values.mediator.persistence.initImage | default "postgres:14" }}
```
nicu-da: can we actually try to use `imageName` in most places? if we have just one central deployment in the chart  have that be a top level prop.
here we could have `mediator.imageName`, `mediator.initImageName`

cluster/helm/splice-global-domain/templates/mediator.yaml
pasindutennage-da: Fixed all, except for `splice-global-domain/sequencer.yaml` `.Values.sequencer.driver.persistence.initImage` which is not set by default, but only when client needs.

cluster/helm/splice-global-domain/templates/mediator.yaml
pasindutennage-da: FIxed

cluster/helm/cn-docs/values.schema.json
```diff
@@ -7,6 +7,22 @@
   "properties": {
     "networkName": {
       "type": "string"
+    },
+    "docs": {
```
nicu-da, line 11: nit: we should enforce that the image name is required here right?

nicu-da (approved):
Thx!

cluster/helm/cn-docs/values.schema.json
pasindutennage-da: Fixed in all places

pasindutennage-da:
Cluster Test: https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/69618

PR #5862, opened 2026-06-08
Avoid connecting to all scans when bft-custom mode is enabled

### LINKED ISSUE #5861: Avoid connecting to all scans when bft-custom mode is enabled
## Issue

```
scan-client {
  type = "bft-custom"
  sv-names = ["DA1", "DA2"]
  seed-urls = [
    "ssed-1",
    "seed-2"
  ]
}
```

## Logs

Instead of connecting to just 2 svs, client discovers all scan nodes and attempts to connect to all of them.

## Analysis

This is indeed a bug in the code. It is a race condition.

1.  Codeu ses bootstrapWithSeedNodes (BftScanConnecion.scala line 1470), do build a temoporary AllDsoScansBft with just the 2 seed-urls.

2.  Using the temoporary AllDsoScansBft, we fetch all the available scans and connect to only scans in svNames provided by user. (BftScanConnecion.scala line 1611)

3.  And then we create the ConfigurationProvidedScansBft

Now what goes wrong: in step 2 above, we opted this approach to use a temporary AllDsoScansBft, for the sake of reusability across AllDsoScansBft and ConfigurationProvidedScansBft initalization. However,  in (BftScanConnecion.scala line 141), we instantly trigger a period refreshAction action on temoporary AllDsoScansBft. When this temoporary AllDsoScansBft is refreshed, it try connecting to all scans! And hence the race condition!

As the solution, we can disable the refreshAction only when we are in the middle of making the ConfigurationProvidedScansBft.

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/5861

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
@@ -130,6 +130,7 @@ class BftScanConnection(
     val retryProvider: RetryProvider,
     val loggerFactory: NamedLoggerFactory,
     val connectionMetrics: Option[ScanConnectionMetrics] = None,
+    val disableBackgroundRefresh: Boolean = false,
 )(implicit protected val ec: ExecutionContextExecutor, protected val mat: Materializer)
     extends FlagCloseableAsync
     with NamedLogging
@@ -141,7 +142,7 @@ class BftScanConnection(
   private val refreshAction: Option[PeriodicAction] = scanList match {
     case _: BftScanConnection.TrustSingle =>
       None
-    case bft: BftScanConnection.Bft =>
+    case bft: BftScanConnection.Bft if !disableBackgroundRefresh =>
       Some(
         new PeriodicAction(
           clock,
@@ -164,6 +165,7 @@ class BftScanConnection(
           )
         })
       )
+    case _ => None
   }
 
   override def listVoteRequests()(implicit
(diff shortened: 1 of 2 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -141,7 +142,7 @@ class BftScanConnection(
   private val refreshAction: Option[PeriodicAction] = scanList match {
     case _: BftScanConnection.TrustSingle =>
       None
-    case bft: BftScanConnection.Bft =>
+    case bft: BftScanConnection.Bft if !disableBackgroundRefresh =>
```
isegall-da, line 149: Can we add a comment here on when we choose to disable this and why?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1605,11 +1609,14 @@ object BftScanConnection {
             if (ts.useLastKnownConnectionsForInitialization) { persistScanUrlsCallback }
             else { _ => Future.unit },
             connectionMetrics,
+            disableBackgroundRefresh = true,
           )
 
           // Use the temporary connection to get a consensus on the full list of scans
           allScans <- Bft.getScansInDsoRules(tempBftConnection)
 
+          _ = tempBftConnection.close()
```
nicu-da: lets do a use with resources, using `Using`

nicu-da (approved):
Any test we can add?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: Added

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
pasindutennage-da: since `allScans <- Bft.getScansInDsoRules(tempBftConnection)` only returns a future, we cannot sync close the scan connection. I used `.andThen(close())`

nicu-da (approved):

PR #5866, opened 2026-06-09
Traffic-Based App Rewards: add/update triggers to perform reward sharing

### LINKED ISSUE #5787: Wallet app: Add RewardSharingTrigger
Provides
- `RewardSharingTrigger` and any necessary unittests
- An integration test, likely a  `XXXTimeBasedIntegrationTest` 
- Also adds reward sharing to `MintingDelegationCollectRewardsTrigger`

adetokunbo:
Fixes #5787

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


adetokunbo:
@meiersi-da PTAL and let me know if the main changes are heading in the right direction

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
```diff
@@ -0,0 +1,145 @@
...
+    for {
+      allCoupons <- store.multiDomainAcsStore
+        .listContracts(RewardCouponV2.COMPANION)
+    } yield {
+      val assigned = allCoupons
+        .filter { rw =>
+          rw.payload.provider == endUserParty.toProtoPrimitive &&
+          rw.payload.beneficiary.isEmpty
+        }
+        .flatMap(_.toAssignedContract)
```
meiersi-da: let's push the filters down into the DB query by introducing a dedicated query function in the (External)UserWalletStore. Otherwise this trigger can get stuck due to the page limit returning a limited number of contracts and all of them getting filtered out. I'd suggest you return the results sorted by `expiresAt`, so that you always see the 100 oldest ones.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,137 @@
...
+        .lookupEndUserPartyWallet(aliceValidatorParty)
+        .valueOrFail("Expected alice to have a wallet")
+        .store
+        .multiDomainAcsStore
+        .listContracts(RewardCouponV2.COMPANION)
+        .futureValue
+
+    clue("Create unassigned RewardCouponV2 for alice's validator") {
+      val openRound = sv1ScanBackend.getLatestOpenMiningRound(env.environment.clock.now)
+      sv1Backend.participantClientWithAdminToken.ledger_api_extensions.commands.submitJava(
```
meiersi-da: didn't you create a utility for this already in one of the TestUtil classes?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,137 @@
...
+    clue("Unassigned coupon exists before TTL threshold") {
+      eventually() {
+        listV2Coupons().filter(_.payload.beneficiary.isEmpty) should have size 1
+      }
+    }
+
+    // Advance past the sharing threshold (36h - 30h = 6h)
+    advanceTime(Duration.ofHours(7))
+
+    clue("Unassigned coupon is consumed and assigned coupons created") {
```
meiersi-da: What about creating two coupons to see that the batching works as expected?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,137 @@
...
+
+        byBeneficiary should contain key bobValidatorParty.toProtoPrimitive withClue "bob should receive a coupon"
+        byBeneficiary should contain key aliceValidatorParty.toProtoPrimitive withClue "alice should receive remainder"
+
+        byBeneficiary(bobValidatorParty.toProtoPrimitive) shouldBe BigDecimal(
+          4.0
+        ) withClue "bob gets 40%"
+        byBeneficiary(aliceValidatorParty.toProtoPrimitive) shouldBe BigDecimal(
+          6.0
+        ) withClue "alice gets 60%"
```
meiersi-da: This check races with the minting triggers. I'd suggest to disable them to be on the safe side.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletRewardsTimeBasedIntegrationTest.scala
```diff
@@ -143,52 +151,71 @@ class WalletRewardsTimeBasedIntegrationTest
...
-      val prevBalance = bobValidatorWalletClient.balance().unlockedQty
-
-      // Bob's validator collects rewards
-      // it takes 3 ticks for the IssuingMiningRound 1 to be created and open.
-      advanceRoundsToNextRoundOpening
-      advanceRoundsToNextRoundOpening
-      advanceRoundsToNextRoundOpening
-      advanceTimeForRewardAutomationToRunForCurrentRound
+      // Pause alice's reward collection so the sharing trigger's assigned
+      // coupon is not minted before we can observe it.
```
meiersi-da: I'd suggest to pause bob's minting trigger so we can observe the assigned coupon there, while we should be able to observe for alice that her percentage of the reward coupon is minted. It's more interesting to keep alice's minting trigger enabled, as that would also allow us to see the bug when the minting trigger would assign the unshared coupon.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,137 @@
...
+      }
+    }
+  }
```
meiersi-da: This test overlaps quite a bit with the WalletRewardsTimeBasedIntegrationTest. Consider what the delineation is and avoid duplication, if possible.

It might well make sense to have this dedicated test here to simplify reasoning about test coverage and debugging test failures. However, we do want to avoid duplicated tests, as they cost in terms of test execution time and maintenance overhead.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
```diff
@@ -378,4 +395,139 @@ class MintingDelegationCollectRewardsTrigger(
...
   }
+
+  private def shouldShareNow(
+      coupons: Seq[Contract[RewardCouponV2.ContractId, RewardCouponV2]],
+      config: RewardSharingConfig,
+  ): Boolean = {
+    val now = context.clock.now.toInstant
+    val minTtl = config.minTtlAfterSharing.asJava
+    coupons.exists { c =>
+      Duration.between(now, c.payload.expiresAt).compareTo(minTtl) <= 0
```
meiersi-da: consider using
```
!c.payload.expiresAt.isAfter(now.plus(minTtl))
```

I also wonder whether we couldn't stay in Scala land a bit more so we get access to `<=`

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
```diff
@@ -152,72 +159,74 @@ class MintingDelegationCollectRewardsTrigger(
...
-        .submit(
-          actAs = Seq(delegateParty),
-          readAs = Seq(delegateParty, externalParty),
-          delegation.contractId.exerciseMintingDelegation_Mint(inputs.asJava, paymentContext),
+      val mintCouponsData = filteredCouponsData.copy(rewardCouponsV2 = assignedV2)
+      // Assign beneficiaries to unassigned V2 coupons when either:
+      // - the TTL threshold is reached (coupon is close to expiry), or
+      // - amulets need merging, in which case we include the assignment
+      //   in the same transaction to save traffic cost rather than
+      //   waiting for the TTL threshold and submitting separately.
```
meiersi-da: Why don't we also assign the coupons if we need to merge the other coupons? We're anyways building a transaction, so it makes sense to batch the reward sharing part as well.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
```diff
@@ -378,4 +395,139 @@ class MintingDelegationCollectRewardsTrigger(
...
+      maxNumInputs: Int,
+      validatorRightOpt: Option[Contract[ValidatorRight.ContractId, ValidatorRight]],
+  )(implicit tc: TraceContext): Future[Boolean] = {
+    val newBeneficiaries = config.allDamlBeneficiaries(externalParty).map { case (party, pct) =>
+      new RewardBeneficiary(party.toProtoPrimitive, pct)
+    }
+
+    // headOption is safe: shouldAssign checks nonEmpty before calling
+    unassignedV2.headOption match {
+      case None => Future.successful(false)
```
meiersi-da: this case should never happen, should it? consider raising an internal error if it happens nevertheless

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
```diff
@@ -378,4 +395,139 @@ class MintingDelegationCollectRewardsTrigger(
...
+              paymentContext,
+            ),
+          )
+          .withDisclosedContracts(contractsToDisclose)
+          .noDedup
+          .yieldUnit()
+          .map { _ =>
+            logger.debug(
+              s"Assigned ${unassignedV2.size} V2 coupons to ${newBeneficiaries.size} beneficiaries " +
+                s"and minted rewards for delegation ${delegation.contractId}"
```
meiersi-da: The debug log is quite different here compared to `performMint`. Are we missing information above? Consider using `show" ... "` interpolation, and defining a `Pretty` instance for `CouponsData` to share it more easily.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
```diff
@@ -378,4 +395,139 @@ class MintingDelegationCollectRewardsTrigger(
...
+  )(implicit tc: TraceContext): Future[Boolean] = {
+    val inputs = buildTransferInputs(couponsData, amuletsToMerge, maxNumInputs)
+    val transferContext =
+      buildTransferContext(openRound, openIssuingRounds, couponsData, validatorRightOpt)
+    val paymentContext = new PaymentTransferContext(amuletRules.contractId, transferContext)
+    val contractsToDisclose = spliceLedgerConnection.disclosedContracts(
+      amuletRules,
+      openRound,
+    ) addAll openIssuingRounds
+    val delegateParty = PartyId.tryFromProtoPrimitive(delegation.payload.delegate)
```
meiersi-da: It would be nice to share more of this code with `performAssignAndMint`. It seems to me that passing in the `paymentContext` would reduce the duplication quite a bit. Potentially also passing in the `inputs` would help as well.

meiersi-da: and actually also the contracts to disclose; and perhaps some other pieces.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
```diff
@@ -152,72 +159,74 @@ class MintingDelegationCollectRewardsTrigger(
       ]],
       validatorRightOpt: Option[Contract[ValidatorRight.ContractId, ValidatorRight]],
   )(implicit tc: TraceContext): Future[Boolean] = {
```
meiersi-da, line 156: Where are you testing that the new code in this trigger work as expected? Seems like the test has not yet been adapted.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,137 @@
...
+import org.lfdecentralizedtrust.splice.wallet.config.{
+  AppRewardBeneficiaryConfig,
+  RewardSharingConfig,
+}
+
+import java.time.Duration
+import java.util.Optional
+import scala.jdk.CollectionConverters.*
+
+class RewardSharingTimeBasedIntegrationTest
```
meiersi-da: You'll need to guard this test so it only runs with `SpliceAmulet_0_1_19` as that is the one introducing the new Daml models. Actually, this problem likely already holds for the tests merged to `main`. See https://github.com/canton-network/splice/blob/main/DEVELOPMENT.md#daml-version-guards-in-integration-tests for context.

meiersi-da (commented):
Thanks.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
meiersi-da: I've kicked off a build for `main` here: https://github.com/canton-network/splice/actions/runs/27219264312

Let's see what it says.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
adetokunbo: Done.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
adetokunbo: Done

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
adetokunbo: Fixed

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
adetokunbo: Done, CouponsData is now Pretty

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
adetokunbo: Fixed by introducing some case classes to carry parameters, PTAL

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
adetokunbo: I added these guards.  IUC, these tests don't fail on main now, but without the tags; they will if someone runs a compat test workflow that targets a release branch that does not include 0.1.19

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
adetokunbo: Yes - I just needed to update it with another default parameter.  Fixed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
adetokunbo: Done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
adetokunbo: Fixed.  The test has been reorganized to address this and some other comments.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletRewardsTimeBasedIntegrationTest.scala
adetokunbo: Thank you, I've changed the test so that operates like this.  PTAL

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
adetokunbo: The two tests now have distinct responsibilities, explained in the class comments: 
 - RewardSharingTest — assignment correctness in isolation 
   -  freezes minting to inspect assignment details.
 - WalletRewardsTest — sharing + minting interaction end-to-end
   - lets minting run (for alice) to verify the triggers interact correctly, and only pauses bob to observe the intermediate state.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
adetokunbo: I think this is already happening, but the comment and method branch structure make this hard to see.
I've reorganized `performMintIfNeeded` so this should be clearer.  Please let me know if I'm missing something.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
```diff
@@ -116,6 +116,24 @@ trait TransferInputStore extends AppStore with LimitHelpers {
...
+    for {
+      coupons <- multiDomainAcsStore.listContracts(RewardCouponV2.COMPANION)
+    } yield applyLimit(
+      "listUnassignedRewardCouponsV2",
+      limit,
+      coupons
+        .filter(_.payload.beneficiary.isEmpty)
+        .sortBy(_.payload.expiresAt),
+    )
+
```
meiersi-da: why can't you use the new functions that you introduced in the DB stores here as well?

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
adetokunbo: Sorry, I've just added these.  I mentioned this in the PR description, but I should have repeated in the comment asking for a review.
I added a new test [utility behavior](https://github.com/canton-network/splice/pull/5866/changes/1611befd15e06087e5a897ee5b4e9abb0adde8d8) for this, as I could not find an existing way of obtaining the external party ID at configuration time; please advise if a different approach ought to be used

adetokunbo:
@meiersi-da I am currently working on the [recent comment you added](https://github.com/canton-network/splice/pull/5866#discussion_r3386061833), but have addressed all prior comments.  I think have a good idea about how to resolve the comment; since there are many changes, do you mind beginning your next review pass while I'm working on it ?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
adetokunbo: I plan to add more parameters to and rename  `listUnassignedRewardCouponsV2`, then use it here.  Does that sound reasonable?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ExternallySignedPartyTestUtil.scala
```diff
@@ -300,4 +306,130 @@ trait ExternallySignedPartyTestUtil extends TestCommon {
...
+
+  case class PreGeneratedParty(
+      partyId: PartyId,
+      keyPairBytes: ByteString,
+  )
+
+  /** Onboard an external party using a pre-generated key from [[preGenerateExternalParty]].
+    * Imports the key pair into the participant, then follows the standard onboarding flow.
+    */
+  def onboardExternalParty(
```
meiersi-da, line 402: This is essentially a full copy of the existing `onboard` function. Can you share the code, so we don't duplicate the maintenance effort?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ExternallySignedPartyTestUtil.scala
```diff
@@ -300,4 +306,130 @@ trait ExternallySignedPartyTestUtil extends TestCommon {
...
+  /** Pre-generate an Ed25519 key pair and compute the external party's ID
+    * without a running participant. This is needed because external party IDs
+    * depend on cryptographic key fingerprints that are normally only known at
+    * runtime, but some test configs (e.g. `rewardSharingConfigByParty`) require
+    * the party ID at config transform time — before participants start.
+    *
+    * Use the returned [[PreGeneratedParty]] in config transforms to reference
+    * the party ID, then call [[onboardExternalParty]] with it at test time.
+    */
+  def preGenerateExternalParty(partyHint: String): PreGeneratedParty = {
```
meiersi-da, line 342: Where does this code come from? i.e., can we share this with the source to avoid it going stale?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
meiersi-da: Hmm... which kind of problem would caught only by the RewardSharingTest? The reward split can also be determined based on the amounts, can it not? I'm also asking as you are not doing this kind of "split-testing" for the WalletMintingDelegation trigger.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        // Advance past the sharing threshold (36h - 30h = 6h)
+        advanceTime(Duration.ofHours(7))
+      }
+
+      // Sharing trigger resumes — both coupons are past TTL, so the
+      // trigger batches them in a single AssignBeneficiaries call.
+      clue("Unassigned coupons are consumed and assigned coupons created") {
+        eventually(40.seconds) {
```
meiersi-da: Why do we need such a long time for this? The action should be trivial.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
```diff
@@ -40,6 +48,9 @@ class WalletMintingDelegationTimeBasedIntegrationTest
     with ExternallySignedPartyTestUtil {
 
   private val DefaultAmuletMergeLimit = 10
+  // Pre-generate key pairs so external party IDs are known at config time
+  private val sharingBeneficiary = preGenerateExternalParty("sharing_beneficiary")
```
meiersi-da:
```suggestion
  private val sharingAppProvider = preGenerateExternalParty("sharing_app_provider")
```

meiersi-da: I assume that's what you mean by beneficiary, is it not?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
```diff
@@ -626,6 +655,109 @@ class WalletMintingDelegationTimeBasedIntegrationTest
...
+            (beneficiaryParty.party, unassignedAmount1, None),
+            (beneficiaryParty.party, unassignedAmount2, None),
+            (beneficiaryParty.party, assignedAmount, Some(beneficiaryParty.party)),
+          ),
+          round = Some(issuingRound.round),
+        )
+      }
+
+      advanceRoundsToNextRoundOpening
+      advanceTimeForRewardAutomationToRunForCurrentRound
```
meiersi-da: why do you need these? the minting of reward coupon's V2 should work independently of rounds

meiersi-da: the only reason to advance time would be to ensure the TTL on the coupon is small enough.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletRewardsTimeBasedIntegrationTest.scala
```diff
@@ -126,71 +141,104 @@ class WalletRewardsTimeBasedIntegrationTest
...
-          .lookupEndUserPartyWallet(aliceValidatorParty)
-          .valueOrFail("Expected alice to have a wallet")
-        val unassigned = aliceWallet.store.multiDomainAcsStore
-          .listContracts(RewardCouponV2.COMPANION)
-          .futureValue
-          .filter(c =>
-            c.payload.provider == aliceValidatorParty.toProtoPrimitive &&
-              c.payload.beneficiary.isEmpty
+      // Give bob's trigger time to mint alice's shared coupon
+      advanceTimeForRewardAutomationToRunForCurrentRound
```
meiersi-da: Do we need this for minting reward coupons v2? I would expect not.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/UserWalletStore.scala
```diff
@@ -682,7 +682,11 @@ object UserWalletStore {
             (tc) => pkgVersionSupport.supportsTrafficBasedAppRewards(Seq(key.endUserParty), now)(tc)
           },
         )(co =>
-          UserWalletAcsStoreRowData(co, None, rewardCouponRound = Some(co.payload.round.number))
+          UserWalletAcsStoreRowData(
+            co,
+            contractExpiresAt = Some(Timestamp.assertFromInstant(co.payload.expiresAt)),
+            rewardCouponRound = Some(co.payload.round.number),
+          )
```
meiersi-da, line 689: Changing row data post-hoc without a store bump requires justification. Seems OK here, as the relevant contracts cycle at least every 36 hours on all environments, and on most environments no contracts will exist at all.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
meiersi-da: yes, but what's the difference between what you compute in this definition above and the code you implemented in DbExternalPartyWalletStore:
```
override def listUnassignedRewardCouponsV2(
      limit: Limit = defaultLimit
  )(implicit tc: TraceContext): Future[Seq[
    ContractWithState[amuletCodegen.RewardCouponV2.ContractId, amuletCodegen.RewardCouponV2]
  ]] =
    waitUntilAcsIngested {
      for {
        result <- storage.query(
          selectFromAcsTableWithState(
            WalletTables.externalPartyAcsTableName,
            acsStoreId,
            domainMigrationId,
            amuletCodegen.RewardCouponV2.COMPANION,
            additionalWhere = sql"and acs.create_arguments->>'beneficiary' is null",
            orderLimit = sql"order by acs.contract_expires_at asc limit ${sqlLimit(limit)}",
          ),
          "listUnassignedRewardCouponsV2",
        )
      } yield result.map(contractWithStateFromRow(amuletCodegen.RewardCouponV2.COMPANION)(_))
    }
``` ?

File: apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala [hunk 22]
```diff
@@ -138,86 +149,117 @@ class MintingDelegationCollectRewardsTrigger(
...
+          .exerciseMintingDelegation_Mint(submission.inputs, submission.paymentContext),
       )
+      .withDisclosedContracts(submission.contractsToDisclose)
+      .noDedup
+      .yieldUnit()
+      .map { _ =>
+        logger.debug(
+          show"Minted ${submission.couponsData}" +
+            show" and merged ${submission.amuletsToMerge.size} amulets" +
+            show" for delegation ${PrettyContractId(submission.delegation)}"
```
meiersi-da:
```suggestion
          show"Minted ${submission.couponsData} and merged ${submission.amuletsToMerge.size} amulets for delegation ${PrettyContractId(submission.delegation)}"
```
here and below to avoid repeated concatenation and allow sharing of the context for show, e.g., to determine ellipsis at the whole log message level

meiersi-da (commented):
I'm out of time for today, and in particular did not manage to re-review the large logic change in the MintingDelegationCollectRewardsTrigger. Will continue tomorrow.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ExternallySignedPartyTestUtil.scala
adetokunbo: Extracted `submitTopologyAndOnboard`

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ExternallySignedPartyTestUtil.scala
adetokunbo: Added a comment explaining the source: It replicates behavior in canton that can't be called directly

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
adetokunbo: Removed.  It was an artefact that should have been removed, sorry.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
adetokunbo: yes, fixed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
adetokunbo: Removed.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/UserWalletStore.scala
adetokunbo: Agreed

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
adetokunbo: Fixed

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
adetokunbo: They sorted differently (and returned slightly different types, which was less important). I've unified them into a single `listRewardCouponsV2` method with filter and sortOrder parameters, with just DB-level implementations.

adetokunbo:
@meiersi-da all comments are now addressed, except the decision about whether to the merge the `RewardSharingTimeBasedIntegration` test into the other one.  Am happy to do that if preferred.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletRewardsTimeBasedIntegrationTest.scala
adetokunbo: Not for V2 specifically, but for non-V2 reward collection in the same transaction

meiersi-da: But there are no such coupons in this test, so we could remove this, can't we?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
```diff
@@ -116,36 +116,50 @@ trait TransferInputStore extends AppStore with LimitHelpers {
...
+sealed trait RewardCouponV2SortOrder {
+  def ordering: Ordering[ContractWithState[RewardCouponV2.ContractId, RewardCouponV2]]
+}
+
+object RewardCouponV2SortOrder {
+  case object ByExpiresAtAsc extends RewardCouponV2SortOrder {
+    def ordering: Ordering[ContractWithState[RewardCouponV2.ContractId, RewardCouponV2]] =
+      Ordering.by(_.payload.expiresAt)
+  }
+  case object ByRoundAscAmountDesc extends RewardCouponV2SortOrder {
```
meiersi-da: It's not clear to me why we need this sort order.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/treasury/TreasuryService.scala
```diff
@@ -1088,16 +1092,21 @@ class TreasuryService(
...
   ): Future[(BigDecimal, Seq[(Round, BigDecimal, InputRewardCouponV2)])] =
     for {
-      rewardCouponV2Inputs <- userStore.listSortedMintableRewardCouponV2s(
-        includeUnassigned = mintUnassignedRewardCouponsV2,
-        PageLimit.tryCreate(maxNumInputs),
+      rewardCouponV2Results <- userStore.listRewardCouponsV2(
+        filter =
+          if (mintUnassignedRewardCouponsV2) RewardCouponV2Filter.All
+          else RewardCouponV2Filter.AssignedOnly,
+        sortOrder = RewardCouponV2SortOrder.ByRoundAscAmountDesc,
```
meiersi-da: I believe sorting by expiresAt would actually be the better option here. What we care about is to avoid missing the collection of a coupon. For the round-bound coupons that means acting in round order, as earlier rounds are gone more quickly.

For `RewardCouponV2` going by `expiresAt` seems like the right option.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
```diff
@@ -116,36 +116,50 @@ trait TransferInputStore extends AppStore with LimitHelpers {
...
-    * with no beneficiary.
+  /** Returns RewardCouponV2 contracts filtered by assignment status and
+    * sorted according to the specified order. Implemented in DB stores
+    * with SQL-level filtering to avoid page-limit issues.
     */
-  def listSortedMintableRewardCouponV2s(
-      includeUnassigned: Boolean,
+  def listRewardCouponsV2(
+      filter: RewardCouponV2Filter,
+      sortOrder: RewardCouponV2SortOrder,
```
meiersi-da:
```suggestion
      includeUnassigned : Bool,
      includeAssigned : Bool,
```

should suffice, and require less extra machinery.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
```diff
@@ -0,0 +1,137 @@
...
+  override protected def retrieveTasks()(implicit
+      tc: TraceContext
+  ): Future[Seq[Task]] =
+    for {
+      unassigned <- store.listRewardCouponsV2(
+        RewardCouponV2Filter.UnassignedOnly,
+        RewardCouponV2SortOrder.ByExpiresAtAsc,
+      )
+    } yield {
+      val assigned = unassigned.flatMap(_.toAssignedContract)
```
meiersi-da:
```suggestion
      val couponsToAssign = unassignedCoupons.flatMap(_.toAssignedContract)
```

otherwise it reads quite confusing, as we have two notions of "assignment", and the meaning of the variable name changes from "unassigned coupon contract with state" to "unassigned coupon contract assigned to synchronizer"

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
```diff
@@ -0,0 +1,137 @@
...
+      if (assigned.isEmpty || !shouldShareNow(assigned)) Seq.empty
+      else Seq(Task(assigned))
+    }
+
+  override protected def completeTask(
+      task: Task
+  )(implicit tc: TraceContext): Future[TaskOutcome] =
+    task.coupons.headOption match {
+      case None => Future.successful(TaskSuccess("No coupons to assign"))
```
meiersi-da: This is more like an invariant error.

Consider using `cats.data.NonEmptyList` to transport the information about non-emptiness between task retrieval and here. You can then just use `task.coupons.head` and `task.coupons.tail` directly.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
```diff
@@ -0,0 +1,137 @@
...
+    for {
+      unassigned <- store.listRewardCouponsV2(
+        RewardCouponV2Filter.UnassignedOnly,
+        RewardCouponV2SortOrder.ByExpiresAtAsc,
+      )
+    } yield {
+      val unassignedIds = unassigned.map(_.contractId).toSet
+      // Stale if none of the task's coupons are still unassigned
+      !task.coupons.exists(c => unassignedIds.contains(c.contract.contractId))
+    }
```
meiersi-da: This only just works because the sort order is consistent.

There's an easier option: use https://github.com/canton-network/splice/blob/1f9102e20830d1a841c8f5fe962df0d62bf69144/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/MultiDomainAcsStore.scala#L126-L129

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  override protected def extraMetricLabels = Seq("party" -> endUserParty.toString)
+
+  override protected def retrieveTasks()(implicit
+      tc: TraceContext
+  ): Future[Seq[Task]] =
+    for {
+      unassigned <- store.listRewardCouponsV2(
+        RewardCouponV2Filter.UnassignedOnly,
+        RewardCouponV2SortOrder.ByExpiresAtAsc,
```
meiersi-da:
```suggestion
        RewardCouponV2SortOrder.ByExpiresAtAsc,
        limit = HardLimit.tryCreate(context.config.parallelism * config.batchSize)
```

setting a configurable batch size (default 100) seems like the prudent option.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
```diff
@@ -0,0 +1,137 @@
...
+      )
+    } yield {
+      val assigned = unassigned.flatMap(_.toAssignedContract)
+      if (assigned.isEmpty || !shouldShareNow(assigned)) Seq.empty
+      else Seq(Task(assigned))
```
meiersi-da: split into batches

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
```diff
@@ -286,8 +341,9 @@ class MintingDelegationCollectRewardsTrigger(
         Some(issuingRoundsMap.keySet.map(_.number))
       )
       appRewardCouponsWithQuantity <- store.listSortedAppRewards(issuingRoundsMap)
-      rewardCouponsV2WithQuantity <- store.listSortedMintableRewardCouponV2s(
-        includeUnassigned = rewardSharingConfig.beneficiaries.isEmpty
+      rewardCouponsV2 <- store.listRewardCouponsV2(
+        RewardCouponV2Filter.All,
+        RewardCouponV2SortOrder.ByRoundAscAmountDesc,
```
meiersi-da:
```suggestion
        RewardCouponV2SortOrder.ByRoundAscAmountDesc,
        limit = HardLimit.tryCreate(rewardSharingConfig.batchSize)
```

to avoid sharing too many at once.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
```diff
@@ -138,86 +150,112 @@ class MintingDelegationCollectRewardsTrigger(
...
-              s"and merged ${amuletsToMerge.size} amulets for delegation ${delegation.contractId}"
-          )
-          true
+      case Some(primaryCoupon) =>
+        val newBeneficiaries = config.allDamlBeneficiaries(externalParty).map { case (party, pct) =>
+          new RewardBeneficiary(party.toProtoPrimitive, pct)
         }
-    } else {
-      Future.successful(false)
+        val additionalCoupons = unassignedV2.drop(1)
```
meiersi-da: if you pattern match on the list itself, then you get access to head and tail, and don't have to remember to drop(1). Seems like the better option.

meiersi-da (commented):
Thanks. I think we are close.

What would you do whether to keep the test or not? How much CI time does the test cost?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
adetokunbo: Removed; sorting is now always by expiresAt.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/treasury/TreasuryService.scala
adetokunbo: Done; listRewardCouponsV2 now sorts by expiresAt for all callers including TreasuryService.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/TransferInputStore.scala
adetokunbo: Done; replaced sealed traits with includeUnassigned: Boolean, includeAssigned: Boolean.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
adetokunbo: Done; renamed `assigned` → `couponsToAssign` and `unassigned` → `unassignedCoupons`.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
adetokunbo: Done; `Task.coupons` is now `NonEmptyList`, so `completeTask` uses `.head`/`.tail` directly with no impossible `None` case.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
adetokunbo: To confirm: should `batchSize` be part of the general automation config rather than per-party `RewardSharingConfig`?

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
adetokunbo: Done; pattern matching on `primaryCoupon :: additionalCoupons` instead of `headOption` + `drop(1)`.

meiersi-da: as pointed out on Slack: it should be part of the RewardSharingConfig with a default of 100

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
adetokunbo: Done; using `listAssignedContracts`.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
adetokunbo: Done; fetches `parallelism * batchSize` coupons, splits into tasks of `batchSize`.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
adetokunbo: Done as part of above.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
adetokunbo: Done; `batchSize` added to `RewardSharingConfig` (default 100), used as limit.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletRewardsTimeBasedIntegrationTest.scala
adetokunbo: Sorry, you're correct; I mistakenly thought this was needed here. Removed.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/RewardSharingTimeBasedIntegrationTest.scala
adetokunbo: The split amounts can be checked in WalletRewardsTest too. I've merged the batching and amount assertions into WalletRewardsTest and deleted RewardSharingTest.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/MintingDelegationCollectRewardsTrigger.scala
adetokunbo: Done

adetokunbo:
> Thanks. I think we are close.
> 
> What would you do whether to keep the test or not? How much CI time does the test cost?

@meiersi-da
It was relatively low (on my machine 1 min), but so was the other test.  I've merged them, on my machine this saved about 35s.
PTAL

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
```diff
@@ -626,6 +655,99 @@ class WalletMintingDelegationTimeBasedIntegrationTest
...
+        createRewardCouponsV2(
+          Seq(
+            (beneficiaryParty.party, unassignedAmount1, None),
+            (beneficiaryParty.party, unassignedAmount2, None),
+            (beneficiaryParty.party, assignedAmount, Some(beneficiaryParty.party)),
+          )
+        )
+      }
+
+      advanceTimeForRewardAutomationToRunForCurrentRound
```
meiersi-da: Seems like another left-over... remove and consider grepping the git diff to find them.

meiersi-da: or do we need that because of the TTL?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
```diff
@@ -626,6 +655,99 @@ class WalletMintingDelegationTimeBasedIntegrationTest
...
+      // V2 coupon, then resume — the trigger should batch-assign
+      // beneficiaries to both unassigned coupons and mint all in one transaction.
+      setTriggersWithin(triggersToPauseAtStart = Seq(externalPartyMintingDelegationTrigger)) {
+        createRewardCouponsV2(
+          Seq(
+            (beneficiaryParty.party, unassignedAmount1, None),
+            (beneficiaryParty.party, unassignedAmount2, None),
+            (beneficiaryParty.party, assignedAmount, Some(beneficiaryParty.party)),
+          )
+        )
```
meiersi-da, line 730: check here that the coupons are visible to ensure that the later check is not accidentally vacuous. Probably best to do using `actAndAcheck`.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletRewardsTimeBasedIntegrationTest.scala
```diff
@@ -126,71 +139,108 @@ class WalletRewardsTimeBasedIntegrationTest
...
+      val bobRewardTrigger = bobValidatorBackend
+        .userWalletAutomation(bobValidatorWalletClient.config.ledgerApiUser)
+        .futureValue
+        .trigger[CollectRewardsAndMergeAmuletsTrigger]
+
+      // Pause bob's minting trigger so we can observe his assigned
+      // (unminted) coupon from alice's sharing, while alice's triggers
+      // run freely (sharing + minting). V2 coupons must be created after
+      // pausing because they can be minted immediately.
+      var prevBobBalance = BigDecimal(0)
```
meiersi-da: why do you need a var for this?

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
meiersi-da: How will this work given when there are more than the page limit contracts? Also retrieving all contract data is much more expensive than checking whether some contract-ids are no longer present. 

Why can't you use `containsArchived`?

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/db/DbUserWalletStore.scala
```diff
@@ -203,6 +208,36 @@ class DbUserWalletStore(
...
+            domainMigrationId,
+            amuletCodegen.RewardCouponV2.COMPANION,
+            additionalWhere = whereClause,
+            orderLimit = sql"order by acs.contract_expires_at asc limit ${sqlLimit(limit)}",
+          ),
+          "listRewardCouponsV2",
+        )
+      } yield result
+        .map(contractWithStateFromRow(amuletCodegen.RewardCouponV2.COMPANION)(_))
+    }
```
meiersi-da, line 220: a full copy of exactly the same code seems problematic.

Share as a method on a sibling object to `WalletTables`, e.g. `WalletQueries` . It seems the only diff here is the table name.

meiersi-da (commented):
Almost done! Thanks.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/db/DbUserWalletStore.scala
adetokunbo: Done; added `queryRewardCouponsV2` to the existing `DbTransferInputQueries` trait, same pattern as `listSortedRewardCoupons`.

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
adetokunbo: Done; using `containsArchived` instead.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletRewardsTimeBasedIntegrationTest.scala
adetokunbo: Done; `setTriggersWithin` block now returns the balance as its last expression.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
adetokunbo: Done; using `actAndCheck` to confirm coupons are visible before resuming the trigger.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
adetokunbo: Removed.

adetokunbo: Not needed for TTL — removed.

adetokunbo:
> Almost done! Thanks.

@meiersi-da:  I've addressed all the outstanding comments, PTAL

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
```diff
@@ -0,0 +1,136 @@
...
+        TaskSuccess(
+          s"Shared ${task.coupons.size} RewardCouponV2 with ${newBeneficiaries.size} beneficiaries for $endUserParty"
+        )
+      }
+  }
+
+  override protected def isStaleTask(
+      task: Task
+  )(implicit tc: TraceContext): Future[Boolean] =
+    // Stale if all of the task's coupons have been archived
```
meiersi-da, line 110:
```suggestion
    // Stale if any of the task's coupons have been archived
```

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/store/db/DbUserWalletStore.scala
meiersi-da: Good find. Thanks.

meiersi-da (approved):
Yay! Well done; and just on time! :shipit:

apps/wallet/src/main/scala/org/lfdecentralizedtrust/splice/wallet/automation/RewardSharingTrigger.scala
meiersi-da: nit: this comment is off, is it not? Consider fixing it in the next PR you are working on.

adetokunbo: Certainly!

PR #5874, opened 2026-06-09
Fix SV reward weight persistence in Vote History

### LINKED ISSUE #5092: Update SV Reward Weight proposal don't persist base weight

pawelperek-da:
Fixes #5092

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
```diff
@@ -667,6 +669,49 @@ class DbScanStore(
     } yield ResultsPage(recentVoteResults, afterToken)
   }
 
+  override def lookupLatestSvRewardWeightChange(
```
pawelperek-da, line 672:
```
Limit (cost=1.74..1.75 rows=1 width=1057) (actual time=0.182..0.183 rows=0 loops=1)
--
Buffers: shared hit=55
-> Sort (cost=1.74..1.75 rows=1 width=1057) (actual time=0.181..0.181 rows=0 loops=1)
Sort Key: vote_effective_at DESC
Sort Method: quicksort Memory: 25kB
Buffers: shared hit=55
-> Index Scan using scan_txlog_store_sid_en_vot on scan_txlog_store (cost=0.12..1.73 rows=1 width=1057) (actual time=0.153..0.153 rows=0 loops=1)
Index Cond: (store_id = 1)
Filter: (vote_accepted AND (vote_action_name = 'SRARC_UpdateSvRewardWeight'::text) AND ((entry_data #>> '{result,request,action,value,dsoAction,value,svParty}'::text[]) = 'any-sv::1220abc'::text))
Rows Removed by Filter: 51
Buffers: shared hit=52
Planning:
Buffers: shared hit=282
Planning Time: 5.162 ms
Execution Time: 0.257 ms
```

ray-roestenburg-da (approved):
Thanks!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanStore.scala
moritzkiefer-da: Am I missing something here? Your `index cond` is only `store_id = 1` so you're not actually using the index meaningfully.

pawelperek-da: `scan_txlog_store_sid_en_vot` is doing heavy lifting here because it's a partial index with `WHERE entry_type = 'vot'`. That means we are still O(n log n) or O(n)'ing all votes, but we agreed a small performance concession for a query sitting in the cold path is acceptable vs creating an index that'd be used just for this query. If you disagree, please speak up

moritzkiefer-da: I see thx, sounds reasonable. Let's make sure to test the corresponding code paths in the SV UI on CILR after merging

PR #5877, opened 2026-06-09
Release Notes: Avoid connecting to all scans when bft-custom mode is  enabled

### LINKED ISSUE #5861: Avoid connecting to all scans when bft-custom mode is enabled

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/5861

### DIFF
--- a/docs/src/release_notes_upcoming.rst
+++ b/docs/src/release_notes_upcoming.rst
@@ -29,11 +29,6 @@
             automatically at start-up from its database. The value must be removed from both the helm chart and
             the docker-compose configuration.
 
-          - Fixed a bug where validators using the ``bft-custom`` scan client configuration
-            would incorrectly attempt to establish scan connections with all scan nodes during
-            the validator startup. The scan client now strictly confines all scan connections to
-            configured, trusted SV endpoints.
-
       .. Important::
 
           The migration id must still be kept for participant database naming for backwards compatibility (``persistance.databaseName`` helm value,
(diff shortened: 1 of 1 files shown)


docs/src/release_notes_upcoming.rst
```diff
@@ -29,6 +29,11 @@
             automatically at start-up from its database. The value must be removed from both the helm chart and
             the docker-compose configuration.
 
+          - Fixed a bug where validators using the ``bft-custom`` scan client configuration
```
nicu-da: Can you add this to the top under a new section call bugfixes (it's under the validator deployment section now)

pasindutennage-da: @nicu-da 

Thank you. Fixed

isegall-da (approved):
Thank you

nicu-da (approved):

PR #5880, opened 2026-06-09
refactor for bulk staging

isegall-da:
Part of #5884.
The idea is that the workflow that copies from staging from committed follows the same pattern, thus extends the same abstract base images, and only needs to implement the `process` stage. That PR became a bit too large, so I extracted only the refactoring to a separate PR. This one should be a pure refactor, to allow that reuse.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


isegall-da:
ping @ray-roestenburg-da @rautenrieth-da can I get a review please?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -3,68 +3,110 @@
...
     extends NamedLogging
     with Spanning {
 
-  private def getStartTimestamp(implicit
+  protected val description: String
+
+  /** The key in the key-value store where the timestamp of the latest processed snapshot is stored. This is
+    * used to resume processing from the correct point in case of restarts.
+    */
+  protected val kvStoreKey: String
```
ray-roestenburg-da: Prefer for these vals (kvStoreKey and processedTimestampMetric)  as arguments to the constructor for this abstract class. no initialization order surprises and makes it natural that you have to specify a stable value when extending the abstract base class.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -3,68 +3,110 @@
...
+
+  /** A metric that should be updated with the timestamp of the latest snapshot that was processed.
+    */
+  protected val processedTimestampMetric: MetricHandle.Gauge[CantonTimestamp]
+
+  /** This method should return the timestamp of the next snapshot available, after `last`, if any.
+    * The pipeline will  poll this method until it returns a new snapshot, and then process that snapshot
```
ray-roestenburg-da:
```suggestion
    * The pipeline will poll this method until it returns a new snapshot, and then process that snapshot
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -3,68 +3,110 @@
...
+    * before polling for the next one. It is ok for this method to return a snapshot that should not actually
```
ray-roestenburg-da, line 31: nit: don't know if this info changes anything for reader that wants to implement this. Maybe better to document this in one place, in the non-abstract method where the process occurs and both should.. and this method are called.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -3,68 +3,110 @@
...
+    * be processed, as the pipeline will call `shouldProcessSnapshotAt` to check if the snapshot should be
+    * processed or skipped (but will remember that it was skipped and update `last` for the future calls accordingly).
+    */
+  protected def getNextSnapshotTimestampAfter(
+      last: TimestampWithMigrationId
+  )(implicit tc: TraceContext): Future[Option[TimestampWithMigrationId]]
+
+  /** This method should return true if the snapshot at the given timestamp should be processed,
+    * or false if it should be skipped. This is used to skip snapshots that are not relevant for bulk storage
+    * (e.g. because they are in the DB, but are more frequent then the frequency we need for bulk storage).
```
ray-roestenburg-da: I can't think of a situation at this moment where we would want to skip a snapshot (the snapshots are created every 3 hours and it seems very useful to have all of them), what is the use case for this?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
```diff
@@ -3,68 +3,110 @@
...
+  ): Boolean
+
+  /** This method should return the main Flow that processes the snapshot at the given timestamp.
+    * It must emit back the same timestamp as its output once processing is complete.
+    */
+  protected def processSnapshotAt(ts: TimestampWithMigrationId)(implicit
+      tc: TraceContext
+  ): Flow[TimestampWithMigrationId, TimestampWithMigrationId, NotUsed]
+
+  protected[bulk] def readLatestProcessedSnapshotTimestamp(implicit
```
ray-roestenburg-da: you could put these methods about progress in a separate class that wraps the kvProvider store and just does that, and then you use it where you need it. Pass that class in via constructor and delegate to it.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
ray-roestenburg-da, line 136: you could take readLatestProcessedSnapshotTimestamp, and the other methods as args. Might be easy for unit testing.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -105,11 +116,32 @@ class UpdateHistoryBulkStorage(
...
-  private def getFirstSegment(implicit tc: TraceContext): Future[Option[UpdatesSegment]] =
-    kvProvider.getLatestUpdatesSegmentInBulkStorage().value.flatMap {
+  private def getFirstSegment(implicit tc: TraceContext): Future[Option[UpdatesSegment]] = {
+    readLatestProcessedSegment.flatMap {
       case None => getFirstSegmentFromGenesis
       case Some(after) => getNextSegment(Some(after))
     }
+  }
+
+  protected[bulk] def readLatestProcessedSegment(implicit
```
ray-roestenburg-da: this looks very similar to protected[bulk] def readLatestProcessedSnapshotTimestamp(implicit
in the other class, maybe these kind of helpers can go to it's own class where you simplify reading any type from the kv store.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -3,47 +3,58 @@
...
-    val s3Connection: S3BucketConnection,
-    val historyMetrics: HistoryMetrics,
     override val loggerFactory: NamedLoggerFactory,
 )(implicit actorSystem: ActorSystem, ec: ExecutionContext)
     extends NamedLogging
     with Spanning {
 
+  protected val description: String
+
+  /** The key in the key-value store where the timestamp of the latest updates segment is stored. This is
```
ray-roestenburg-da: same here prefer these as args to the constructor

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
ray-roestenburg-da: I think this is actually the thing you want to make reusable, and provide args to it, so any user can create a Source that switches out the important parts, I think you can do without the abstract class, which would be my preference.

ray-roestenburg-da: You can then have an apply or method that composes the Source fromthe methods / flows that you want to use, for staging or committed for instance, probably also easeier to unit test cause you can provide test versions of readLatestProcessedSnapshotTimestamp, shouldProcessSnapshotAt, etc, just a thought to consider.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorageStaging.scala
```diff
@@ -0,0 +1,139 @@
...
+import cats.data.OptionT
+import cats.implicits.*
+import com.daml.metrics.api.MetricHandle
+import com.digitalasset.canton.data.CantonTimestamp
+import io.grpc.Status
+import org.apache.pekko.NotUsed
+import org.apache.pekko.stream.scaladsl.Flow
+import org.lfdecentralizedtrust.splice.scan.store.bulk.AcsSnapshotBulkStorage.AcsSnapshotObjects
+
+class AcsSnapshotBulkStorageStaging(
```
ray-roestenburg-da: would be nice to have some docs here, what is staging for (or intended to be for soon).
I assume it is for a "workplace" or scratchpad where files are stored while everything is getting processed, and then at the end (atomically at best) files are moved to committed? so from outside API, this is purely a write side process?

I don't fully get why that staging place has an "is a" relationship with AcsSnapshotBulkStorage (why it is designed that way). In the comment on getAcsSnapshotAtOrBefore, that becomes confusing for me, that the staging returns data from committed.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanKeyValueProvider.scala
```diff
@@ -8,57 +8,41 @@ import com.digitalasset.canton.logging.{NamedLoggerFactory, NamedLogging}
...
-    latestUpdatesSegmentInBulkStorageKey,
-    segment,
-  )
-
-  final def getLatestUpdatesSegmentInBulkStorage()(implicit
-      tc: TraceContext,
-      ec: ExecutionContext,
-  ): OptionT[Future, UpdatesSegment] =
-    store.readValueAndLogOnDecodingFailure(latestUpdatesSegmentInBulkStorageKey)
+//  private val latestUpdatesSegmentInBulkStorageKey = "latest_updates_segment_in_bulk_storage"
```
ray-roestenburg-da: Probably want to remove commented code.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanKeyValueProvider.scala
```diff
@@ -8,57 +8,41 @@ import com.digitalasset.canton.logging.{NamedLoggerFactory, NamedLogging}
...
-  implicit val timestampCodec: Codec[CantonTimestamp] =
+  private implicit val timestampCodec: Codec[CantonTimestamp] =
     Codec
       .from[Long](implicitly, implicitly)
       .iemap(timestamp => CantonTimestamp.fromProtoPrimitive(timestamp).leftMap(_.message))(
         _.toProtoPrimitive
       )
   implicit val acsSnapshotTimestampMigrationCodec: Codec[TimestampWithMigrationId] =
     deriveCodec[TimestampWithMigrationId]
+//  implicit val acsSnapshotTimestampMigrationEncoder: Encoder[TimestampWithMigrationId] =
```
ray-roestenburg-da: curious why it was needed to remove these

ray-roestenburg-da (commented):
Some feedback on structure / design, thanks

ray-roestenburg-da (approved):
Please see my suggestions, happy to approve this to speed up things since it's a refactor.

isegall-da:
@ray-roestenburg-da is something like [this commit](https://github.com/canton-network/splice/pull/5880/commits/c16cf6f69bf799c15aad256785edf8c5854ac184) what you had in mind? (of course, it's very partial, doesn't even compile as I haven't changed the call site where we construct it to adjust, just wondering if it's in the direction you had envisioned)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/AcsSnapshotBulkStorage.scala
isegall-da: We already skip them today, this just refactors that. We store in bulk storage only one snapshot per day, not every 3 hours. Just to save on storage, as ACSs can always be reconstructed from the updates.

ray-roestenburg-da: ok thanks

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanKeyValueProvider.scala
isegall-da: Well, note that this only adds them commented out, so they never actually existed :) 
I added them in an attempt to fix some implicit mess, which didn't work, but then accidentally only commented them out instead of deleting them.

isegall-da:
@ray-roestenburg-da I restructure this as discussed, can you make another pass please?

ray-roestenburg-da (approved):
Thanks!

PR #5881, opened 2026-06-09
Make the migration id optional in docker compose

nicu-da:
[static]
Decided to fully remove it from the values files as we embed this in the docs but not 100% sure about it.

When cleaning up the docs and removing the migration id this was the only place left. If we were to keep it, it adds a lot of confusion for new validators that follow the docs, as they have to learn about the migration id simply to set the database name that can actually have any value.

The issue with removing it, is making sure we don't break existing validators when they update. Made it pretty clear in the release notes but not a lot of things we can do to protect against people not messing up the config, reading the release notes. 

The other issue is how we keep backwards compatibility for the docker setup, my plan is to make the migration id optional and if it's passed then include it in the database name, if not keep as is. 

Both of these backwards compatibility setups would need to also show up in the docs, but they can be ignored by new validators and by validators that renamed their db. 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


nicu-da:
@isegall-da any thoughts on this?

isegall-da:
I don't like that we require action from existing validators. I'd rather maintain full bwd compat such that those that completely ignore release notes and just blindly upgrade with their existing config just continue working.

Either that, or fail very early, on schema check, with a very precise message what exactly needs to be changed and how

nicu-da:
> I don't like that we require action from existing validators. I'd rather maintain full bwd compat such that those that completely ignore release notes and just blindly upgrade with their existing config just continue working.
> 
> Either that, or fail very early, on schema check, with a very precise message what exactly needs to be changed and how

So for helm my thoughts were to not set the databaseName value at all and make it required in the helm chart, with a message explaning the two possible value when not set. That way if people blindly upgrade the values files as well then it should fail and you have to explicitly set it. Imo blindly updating the example helm values files should be a no go anyway.

The other option would be to keep as is and just document in the onboarding docs that you should just set the databaseName to the value `participant`. But we would still need to mention the MIGRATION_ID substitution. 

For docker compose there's no other way other than making the migration id optional, if you set it then the db name is modified accordingly. And we document that in the onboarding docs recommending that new validators don't set it. 


And for release 1.0 we can just default to the non migration id by default.

nicu-da:
@isegall-da restricted the scope and will do the removal and switch to non migration id by default in the next 0.7/1.0 release

docs/src/release_notes_upcoming.rst
```diff
@@ -24,16 +24,21 @@
 
       - Validator
 
-          - The ``migration.id`` value was removed from the validator (validator app) helm chart and is no longer
-            supported for docker-compose deployments. The validator now resolves the synchronizer migration id
-            automatically at start-up from its database. The value must be removed from both the helm chart and
-            the docker-compose configuration.
+          - The ``migration.id`` value was removed from the validator (validator app) helm chart. For docker-compose
```
moritzkiefer-da, line 27: how about adding an explicit line like this

"if you deployed your validator before this change you _must_ keep this flag set to the existing value".

moritzkiefer-da (approved):
Thanks!

I wonder if we could add some extra safety by checking if there is a DB with a name starting with `participant` before creating a new one in some init script. Seems like that would catch most attempts where someone significantly misconfigures things.

nicu-da:
> Thanks!
> 
> I wonder if we could add some extra safety by checking if there is a DB with a name starting with `participant` before creating a new one in some init script. Seems like that would catch most attempts where someone significantly misconfigures things.

Could work for compose, but for helm we expose the db name directly as a helm value, so we might have validators running multiple participants within the same db, or using different names for the db.

moritzkiefer-da:
> Could work for compose, but for helm we expose the db name directly as a helm value, so we might have validators running multiple participants within the same db, or using different names for the db.

we do but the recommendation and example values are still `participant_MIGRATION_ID` iirc. So I think we can't make this a hard requirement but adding a check that you can disable if needed might still catch a lot of potential misconfigurations.

moritzkiefer-da:
but I guess there may also be some permission fallout if we cannot actually list DBs or something so maybe not worth it.

nicu-da:
> but I guess there may also be some permission fallout if we cannot actually list DBs or something so maybe not worth it.

we already do list them for compose. I guess if you have permissions to create dbs you should be able to list them as well

nicu-da:
merging this as is and opening another PR with the checks

nicu-da:
part of https://github.com/canton-network/splice/issues/5425

PR #5891, opened 2026-06-10
Add Scan proxy endpoints for V2 contexts

OriolMunoz-da:
Fixes https://github.com/canton-network/splice/issues/5326

It's all boilerplate


meiersi-da (approved):
Thank you!

PR #5896, opened 2026-06-10
Add a metric for active CalculateRewardsV2 contracts

rautenrieth-da:


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardProcessingMetrics.scala
```diff
@@ -33,4 +33,18 @@ class RewardProcessingMetrics(metricsFactory: LabeledMetricsFactory) {
...
+
+  val calculateRewardsContractCount: Gauge[Int] =
+    metricsFactory.gauge(
+      MetricInfo(
+        name = prefix :+ "calculate_rewards_v2" :+ "active_contracts",
+        summary = "The number of active CalculateRewardsV2 contracts",
+        Saturation,
+      ),
+      initial = -1,
+    )(MetricsContext.Empty)
```
meiersi-da: it sounds like we should distinguish between dry-run and not-dry run.

Would we use the same name of the metric for scan as well, or will that have a different name?

rautenrieth-da: Implemented both suggestions. Adding the `dryRun` label forced some refactoring, as gauges have to be created (and closed) exactly once for each label/value combination.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -246,4 +255,49 @@ object CalculateRewardsTriggerBase {
...
+      )(metricsContext)
+
+    val calculateRewardsRootHashBftReads: Meter =
+      metricsFactory.meter(
+        MetricInfo(
+          name = prefix :+ "calculate_rewards_v2" :+ "root_hash_bft_reads",
+          summary = "Count of BFT reads of the reward-accounting root-hash",
+          description =
+            "This metric counts the BFT reads of the reward-accounting root-hash performed by the CalculateRewardsV2 trigger, i.e., the cases where this SV's own Scan could not provide the root-hash and it had to be obtained via a BFT read against peer Scans. Labeled with dryRun.",
+          qualification = Traffic,
```
dfordivam: Unrelated to this refactor, but this sounds off. Perhaps `Errors` is better?

```
  /** Any metric proportional to the normal usage of the system */
  case object Traffic extends MetricQualification

  /** Any metric that counts or relates to errors */
  case object Errors extends MetricQualification
```

rautenrieth-da: Done for all of our bft read metrics.

meiersi-da (approved):
Thanks a lot!

Issue #5621, opened 2026-05-20
Tracking: SV-app + wallet support for enabling traffic-based rewards + static reward sharing


dfordivam:
This captures all the sv-app, validator-app, and wallet support necessary for enabling traffic-based rewards.

#### Tasks
  - [x] #5623
  - [x] #4747 

  - [x] #5869 
  - [x] #5849
     - totalFeaturedAppRewardCoupons: set to total of all app activity records converted to CC using the CC price of the OpenMiningRound. totalUnfeaturedAppRewardCoupons: set to zero
     - Do BFT read of totals similar to root-hash

  - [x] #5715
      - ProcessRewardsTrigger: determines beneficiaries with wrong vetting state
      - ExpireRewardCouponV2Trigger: time-based expiry trigger for RewardCouponV2
      - DsoStore: Add lookup api with random sampling
      - UnhideRewardCouponV2Trigger: polling parallel task trigger publishing reward coupons that are hidden from their beneficiaries,
        - Task creation: sample 100 at random -> find the ones with a working vetting state using PackageVersionSupport.supportsRewardAccountingV2 and make each of them its own task (or is vetting state also checked inside the task?)
        - Task processing: unhide all the beneficiaries RewardCouponV2 contracts in batches by calling DsoRules_UnhideRewardCouponsV2. complete once no new RewardCouponV2 batch remains
  - [x] #5868
  - [x] #5742
     - Any mismatch in votes should trigger a warning
  
  - [x] #6026
    - polling trigger in the SV app that computes and reports the following metrics
      - Number of active CalculateRewardsV2 and ProcessRewardsV2 contracts (move from where they are currently captured)
      - gauge for parties with hidden coupons: one gauge for the number of hidden coupons, labelled with party-id 

  - [x] #5277 
  - [x] #5781
     - Both direct minting and via minting delegation
  - [x] #5624
     - Add support for computing round 0 activity totals
  - [x] #5720
  - [x] #5460
  - [x] Docs update
        Specifically on the [new docs ](https://github.com/canton-network/cf-docs)
        - https://docs.canton.network/appdev/modules/m4-featured-app-activity-marker#featured-app-activity-markers
        - https://docs.canton.network/appdev/app-rewards

dfordivam:
All tracked issues are closed

PR #5910, opened 2026-06-11
Add counters for BFT reads in the TBAR triggers

### LINKED ISSUE #5868: Add counters for BFT reads in the TBAR triggers
one per kind of BFT read, but only if it is used

PumpkinSeed:
I tried to resolve this issue because I'm in the phase of understanding the Splice codebase, and I thought it is a good idea to contribute.

This adds two counters for the BFT reads in the TBAR triggers, one per kind of read:

- `splice.calculate_rewards_v2.root_hash_bft_reads` - the CalculateRewardsV2 trigger reading the root-hash from peer Scans
- `splice.process_rewards_v2.batch_bft_reads` - the ProcessRewardsV2 trigger reading the batch from peer Scans

Both are labeled with dryRun, same as the existing reward metrics.

Fixes #5868
Part of #5621


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -154,6 +154,9 @@ abstract class CalculateRewardsTriggerBase(
         .asRuntimeException()
 
     def bftReadRootHash: Future[Hash] = {
+      rewardMetrics.calculateRewardsRootHashBftReads.mark()(
```
rautenrieth-da: The dry run label is useful for all reward metrics. Can you please add the context to the constructor of `RewardMetrics`, similar to https://github.com/canton-network/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/StoreMetrics.scala#L22?

rautenrieth-da (commented):
Thank you for your contribution!

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
PumpkinSeed: Yes, checking it right now.

PumpkinSeed: Added the change in https://github.com/canton-network/splice/pull/5910/changes/aeae9914f3918e7d0c54f2d8a1c5c341b14a845a

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
```diff
@@ -107,9 +109,7 @@ abstract class CalculateRewardsTriggerBase(
               task.calculateRewards.payload.roundClosedAt,
               context.clock.now.toInstant,
             )
-            _ = rewardMetrics.calculateRewardsProcessingDelay.update(delay)(
-              MetricsContext.Empty.withExtraLabels("dryRun" -> isDryRun.toString)
-            )
+            _ = rewardMetrics.calculateRewardsProcessingDelay.update(delay)(MetricsContext.Empty)
```
rautenrieth-da:
```suggestion
            _ = rewardMetrics.calculateRewardsProcessingDelay.update(delay)
```

We usually omit this, as `MetricsContext.Empty` is the default value

PumpkinSeed: Fixed in this commit: https://github.com/canton-network/splice/pull/5910/changes/21ef7c8ac94e31b868258be4aef5df8581ceeff2

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/CalculateRewardsTrigger.scala
meiersi-da: @rautenrieth-da @PumpkinSeed : how are you thinking about the place of the counter? Right now it measures the number of attempts at doing a BFT read, while the log message is placed strategically to only log a message for each time a BFT read was used to deliver the value used by the trigger.

It seems the current way is slightly better. The metric as is will give an early heads up when BFT reads are being used, and the logs can be used to understand exactly which rounds where filled via a BFT read.

Wrote it up here to share my thinking, so you guys can check it.

PumpkinSeed: So the thought model, the counter increments on each fallback event, which happens only when the own Scan returns CannotProvide.

rautenrieth-da: I was thinking the same. If we write the metric late, we could miss cases where the trigger fails after attempting a BFT read, and I'd rather see those in the metric.

rautenrieth-da (approved):

rautenrieth-da:
@PumpkinSeed can you please rebase your PR onto latest main? CI is failing with an unrelated error that was fixed in https://github.com/canton-network/splice/commit/f02edca57a11d5035ff10f7bd2bf20cb399f606e

PumpkinSeed:
@rautenrieth-da I clicked the button, but I'm not sure it's the right way or whether I should do it manually.

PR #5927, opened 2026-06-11
Fix Vote History sorting

### LINKED ISSUE #5692: Fix sorting in Vote history's infinite scroll

pawelperek-da:
Fixes #5692

Before:
Screenshot 2026-06-12 at 13 21 48

After:
Screenshot 2026-06-12 at 13 32 16

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/openapi/common-internal.yaml
+++ b/apps/common/src/main/openapi/common-internal.yaml
@@ -343,6 +343,7 @@ components:
           type: integer
           description: |
             Cursor for pagination. When requesting the next page of results, pass the `next_page_token` from the previous response.
+            Results are ordered by ledger record time, descending.
 
     ListDsoRulesVoteResultsResponse:
       type: object
--- a/apps/common/src/main/openapi/common-internal.yaml
+++ b/apps/common/src/main/openapi/common-internal.yaml
@@ -343,7 +343,7 @@ components:
           type: integer
           description: |
             Cursor for pagination. When requesting the next page of results, pass the `next_page_token` from the previous response.
-            Results are ordered by ledger record time, descending.
+            Results are ordered by effective date (the accepted vote's effectiveAt, or the result's completedAt otherwise), descending.
 
     ListDsoRulesVoteResultsResponse:
       type: object
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
@@ -257,15 +257,13 @@ object SqlIndexInitializationTrigger {
       ),
     IndexAction
       .Create(
-        indexName = "scan_txlog_store_sid_rt_en_vot",
+        indexName = "scan_txlog_store_sid_effat_en_vot",
         createAction = sqlu"""
-          create index concurrently if not exists scan_txlog_store_sid_rt_en_vot
-          on scan_txlog_store (store_id, record_time desc, entry_number desc)
+          create index concurrently if not exists scan_txlog_store_sid_effat_en_vot
+          on scan_txlog_store (store_id, coalesce(vote_effective_at, entry_data->'result'->>'completedAt') desc, entry_number desc)
           where entry_type = 'vot'
         """,
       ),
-    // Superseded by scan_txlog_store_sid_rt_en_vot when vote results switched from
-    // entry_number to record time ordering.
     IndexAction.Drop(indexName = "scan_txlog_store_sid_en_vot"),
   )
 
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
@@ -226,7 +226,7 @@ object SqlIndexInitializationTrigger {
   }
 
   /** Indexes managed by this trigger class */
-  val defaultIndexActions: List[IndexAction] = List(
+  val defaultIndexActions: List[IndexAction] = List[IndexAction](
     IndexAction
       .Create(
         indexName = "updt_hist_crea_hi_mi_ci_import_updates",
@@ -248,15 +248,6 @@ object SqlIndexInitializationTrigger {
       ),
     IndexAction
       .Create(
-        indexName = "scan_txlog_store_sid_en_vot",
-        createAction = sqlu"""
-          create index concurrently if not exists scan_txlog_store_sid_en_vot
-          on scan_txlog_store (store_id, entry_number desc)
-          where entry_type = 'vot'
-        """,
-      ),
-    IndexAction
-      .Create(
         indexName = "dso_acs_store_sid_mid_pn_tid_rbio",
         createAction = sqlu"""
           create index concurrently if not exists dso_acs_store_sid_mid_pn_tid_rbio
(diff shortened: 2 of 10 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
```diff
@@ -248,13 +248,16 @@ object SqlIndexInitializationTrigger {
       ),
     IndexAction
       .Create(
-        indexName = "scan_txlog_store_sid_en_vot",
+        indexName = "scan_txlog_store_sid_rt_en_vot",
         createAction = sqlu"""
-          create index concurrently if not exists scan_txlog_store_sid_en_vot
-          on scan_txlog_store (store_id, entry_number desc)
+          create index concurrently if not exists scan_txlog_store_sid_rt_en_vot
```
pawelperek-da: @rautenrieth-da could you take a look if that's a correct way to remove unused indices? We didn't use `IndexAction.Drop` before but AFAIU that's the way

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/DbVotesStoreQueryBuilder.scala
```diff
@@ -42,7 +42,10 @@ trait DbVotesTxLogStoreQueryBuilder[TXE]
     TxLogQueries.SelectFromTxLogTableResult
   ], TxLogQueries.SelectFromTxLogTableResult, Effect.Read] = {
     val afterCondition = after match {
-      case Some(a) => Some(sql"""entry_number < $a""")
+      case Some(a) =>
+        Some(
+          sql"""(record_time, entry_number) < ((select record_time from #$txLogTableName where store_id = $txLogStoreId and entry_number = $a), $a)"""
```
ray-roestenburg-da: did you check the plane for this? the SQL query looks weird to me.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala
rautenrieth-da: The code looks good to me. Since this is the first time we're using it, we should check if it really works as intended. You won't trigger the code path in integration tests, you need to first deploy an old version of the app and then a version that has your PR merged. At the very least, check logs of CILR after the changes from this PR have been deployed there.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/DbVotesStoreQueryBuilder.scala
rautenrieth-da: +1 to commenting with the query plan. We just dropped the index on `(store_id, entry_number)`, how can it quickly find the record time from the entry number?

I would have expected that we do either:
- Sort lexicographically by `(record_time, entry_number)`, and include the record time in the API pagination token (along with the entry number)
- Argue why two vote requests can never have the same record time, sort by record time only, and paginate by record time (instead of entry number). If you store the record time as "micros since epoch", you don't even need to change the type of `pageToken` in the API.

ray-roestenburg-da: We discussed this exactly (out of band) and decided we should use the (record_time, entry_number) in page token

pawelperek-da: Subquery is efficient because `entry_number` is the primary key. I am not sure if we still need to filter by `store_id` in this case. Query plan:
```
Index Scan using scan_txlog_store_pkey on scan_txlog_store (cost=0.57..2.79 rows=1 width=32) (actual time=0.029..0.029 rows=1 loops=1)
--
Index Cond: (entry_number = 357263371)
Filter: (store_id = 1)
Planning Time: 3.699 ms
Execution Time: 0.099 ms
```

ray-roestenburg-da: that was not the total problem though, remember we talked about skipped entries

pawelperek-da:
Query plan:

```
   Limit  (cost=0.42..5.47 rows=20 width=142) (actual time=0.062..0.078 rows=20 loops=1)
     Output: store_id, entry_number, transaction_offset, domain_id, entry_type, entry_data, (COALESCE(vote_effective_at, ((entry_data ->
  'result'::text) ->> 'completedAt'::text)))
     Buffers: shared hit=20 read=2
     ->  Index Scan using scan_txlog_store_sid_effat_en_vot on public.scan_txlog_store  (cost=0.42..45714.83 rows=181141 width=142) (actual
           Index Cond: ((store_id = 1) AND (ROW(COALESCE(vote_effective_at, ((entry_data -> 'result') ->> 'completedAt')), entry_number) <
  ROW('2025-10-27T01:30:02.872Z', 745686)))
           Buffers: shared hit=20 read=2
   Planning:
     Buffers: shared hit=6 read=2 dirtied=2
   Planning Time: 0.183 ms 
   Execution Time: 0.248 ms
   ```

File: apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/DbVotesStoreQueryBuilder.scala [hunk 3]
```diff
@@ -41,8 +41,13 @@ trait DbVotesTxLogStoreQueryBuilder[TXE]
   ): SqlStreamingAction[Vector[
     TxLogQueries.SelectFromTxLogTableResult
   ], TxLogQueries.SelectFromTxLogTableResult, Effect.Read] = {
+    val effectiveAtSortKey =
+      s"coalesce($effectiveAtColumnName, entry_data->'result'->>'completedAt')"
```
rautenrieth-da: Why do we pass `effectiveAtColumnName` as an argument? We only have an index on the `vote_effective_at` column, which happens to be the actual parameter used by the only call site. That seems needlessly complicated, can't we just hardcode the column name here?

rautenrieth-da (approved):
Thanks!

apps/common/src/main/openapi/common-internal.yaml
```diff
@@ -343,6 +343,7 @@ components:
           type: integer
           description: |
             Cursor for pagination. When requesting the next page of results, pass the `next_page_token` from the previous response.
+            Results are ordered by ledger record time, descending.
```
ray-roestenburg-da: This is not correct, right? it should be effective time?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/ScanStoreTest.scala
```diff
@@ -479,6 +479,135 @@ abstract class ScanStoreTest
...
+            )
+          } yield {
+            page1.resultsInPage.map(_.request.requester) shouldBe
+              Seq(1, 2, 3).map(userParty(_).toProtoPrimitive)
+            page2.resultsInPage.map(_.request.requester) shouldBe
+              Seq(4).map(userParty(_).toProtoPrimitive)
+          }
+        }
+
+        "order by coalesce(effectiveAt, completedAt) across mixed precision, tie-broken by entry_number" in {
```
ray-roestenburg-da, line 542: thanks

ray-roestenburg-da (approved):
Thanks! just add the comment we discussed offline

apps/common/src/main/openapi/common-internal.yaml
ray-roestenburg-da: Sorry outdated, did not realize rebase

PR #5937, opened 2026-06-12
Re-enable automaticSilentRenew on OidcAuthProvider

### LINKED ISSUE #5682: Starting from 0.6.5 all UIs require re-login after access token expires

samsondav:
PR #5545 disabled `automaticSilentRenew` when removing the global `offline_access` scope. Keycloak still issues a session-bound refresh token without that scope, so silent renew can use the refresh_token grant and users are no longer forced back to the IdP on every short access-token expiry (default 5 minutes).

Fixes #5682

## Change

Re-enable `automaticSilentRenew` on `OidcAuthProvider`.

## Testing

`AuthProvider.test.tsx` mocks `react-oidc-context` and asserts `automaticSilentRenew: true` is passed through. Also covers the hs256-unsafe short-circuit path.

## Note

This is the minimal fix Pawel asked for in lieu of #5683, which adds per-IdP `enable_offline_scope` configuration and operator docs. Auth0 deployments that require `offline_access` to receive a refresh token may still need a follow-up if silent renew does not work there without that scope.

### DIFF
--- a/apps/common/frontend/src/__tests__/auth/AuthProvider.test.tsx
+++ b/apps/common/frontend/src/__tests__/auth/AuthProvider.test.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import { render } from '@testing-library/react';
 import React from 'react';
-import { describe, expect, test, vi } from 'vitest';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
 
 const mockOidcAuthProvider = vi.fn(({ children }: React.PropsWithChildren) => <>{children}</>);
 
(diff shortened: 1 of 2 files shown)


pawelperek-da (approved):
Thanks!

samsondav:
The `wall-clock-time (0)` CI failure here is an unrelated flaky test, not caused by this PR (which only touches the frontend `AuthProvider`).

`WalletIntegrationTest` › "Failure to complete TransferPreapproval creation should be handled correctly" failed because the validator HTTP server timed out under CI load (38s → 503) and logged a WARN "resulted in a timeout" entry, which broke the test's strict ERROR-level log assertion. Failed job: https://github.com/canton-network/splice/actions/runs/27450544466/job/81436663390

Fix for the flake itself: #5988. Re-running CI on this PR should clear it in the meantime.

PR #5964, opened 2026-06-15
Add expiry trigger for AmuletAllocationV2

OriolMunoz-da:
Fixes https://github.com/canton-network/splice/issues/4525


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardCliTestDataTimeBasedIntegrationTest.scala
OriolMunoz-da, line 1: I don't love that the (v1) expiry tests were here. Afaict it was probably done to avoid having a separate timebased test with extra startup time.
We could maybe rename `TokenStandardMetadataTimeBasedIntegrationTest` to simply `TokenStandardTimeBasedIntegrationTest` and move them there I guess

any thoughts @pasindutennage-da since you worked on the v1 expiry?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpiredAmuletAllocationV2Trigger.scala
```diff
@@ -0,0 +1,184 @@
...
+        splice.amuletallocationv2.AmuletAllocationV2,
+      ]
+    ]
+
+  private def allocationV2Stakeholders(allocation: splice.amuletallocationv2.AmuletAllocationV2) =
+    (Seq(
+      allocation.allocation.admin
+    ) ++ allocation.allocation.authorizer.owner.toScala.toList ++ allocation.settlement.executors.asScala)
+      .map(PartyId.tryFromProtoPrimitive)
+}
```
OriolMunoz-da, line 175: Image

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/db/DbSvDsoStore.scala
```diff
@@ -1152,6 +1152,52 @@ class DbSvDsoStore(
       }
     }
 
+  override def listExpiredAmuletAllocationsV2(
```
OriolMunoz-da, line 1155: query-plan:

 we already have:
```
create index dso_acs_store_sid_mid_pn_tid_exp
    on public.dso_acs_store (store_id, migration_id, package_name, template_id_qualified_name, contract_expires_at)
    where (contract_expires_at IS NOT NULL);
```

both this and the already-existing v1 will use that index and then filter by ignored parties from the results, as AmuletAllocation{V1,V2} should be highly selective

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardCliTestDataTimeBasedIntegrationTest.scala
meiersi-da: Hmm.. couldn't the reason also be to check that tx history parsing works as expected when doing expiry?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpiredAmuletAllocationV2Trigger.scala
```diff
@@ -0,0 +1,184 @@
...
+                  splice.api.token.allocationv2.Allocation.INTERFACE
+                ),
+                cancel,
+              )
+            }
+
+            res <-
+              if (cancellations.isEmpty) {
+                Future.successful(
+                  TaskSuccess("Nothing to expire")
```
meiersi-da:
```suggestion
                  TaskNoop("Nothing to expire")
```

meiersi-da: how can it be that there's nothing to expire? Almost seems like an invariant violation.

meiersi-da (approved):
Nice work! Thank you.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardCliTestDataTimeBasedIntegrationTest.scala
OriolMunoz-da: no, because it doesn't dump the txs for the CLI to use

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpiredAmuletAllocationV2Trigger.scala
OriolMunoz-da: indeed, this makes no sense, removing

PR #5970, opened 2026-06-15
Expose Scan:: registry/token: Change istio authorization policies to allow traffic to the registry endpoints

### LINKED ISSUE #5946: Expose Scan:: registry/token: Change istio authorization policies to allow traffic to the registry endpoints

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/5946

Cluster Test 2026-July-9 : https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/73456/workflows/f0e09452-7305-4492-8e3b-f8aa02c9c085/jobs/437568

## Testing

```
#!/bin/bash

echo "Testing Scan ACS Endpoint (HTTP GET)..."
for i in {1..15}; do
  curl -s -o /dev/null -w "Request $i: %{http_code}\n" \
  https://scan.sv-2.scratcha.network.canton.global/api/scan/v0/acs/party123::abcsf
done

echo -e "\nTesting Token Registry Endpoint (HTTP GET)..."
for i in {1..15}; do
  curl -s -o /dev/null -w "Request $i: %{http_code}\n" \
  https://scan.sv-2.scratcha.network.canton.global/registry/metadata/v1/info
done

```

```
Testing Scan ACS Endpoint (HTTP GET)...

Request 1: 200
Request 2: 200
Request 3: 200
Request 4: 200
Request 5: 200
Request 6: 200
Request 7: 200
Request 8: 200
Request 9: 200
Request 10: 200
Request 11: 429
Request 12: 429
Request 13: 429
Request 14: 429
Request 15: 429


Testing Token Registry Endpoint (HTTP GET)...
Request 1: 200
Request 2: 200
Request 3: 200
Request 4: 200
Request 5: 200
Request 6: 200
Request 7: 200
Request 8: 200
Request 9: 200
Request 10: 200
Request 11: 429
Request 12: 429
Request 13: 429
Request 14: 429
Request 15: 429
```


cluster/configs/shared/base.yaml
pasindutennage-da, line 222: when should this be changed?

cluster/configs/shared/base.yaml
```diff
@@ -207,7 +207,13 @@ cloudArmor:
       pathPrefix: /api/scan
       throttleAcrossAllEndpointsAllIps:
         withinIntervalSeconds: 60
-        maxRequestsBeforeHttp429: 0
+        maxRequestsBeforeHttp429: 0    # Keeps scan completely closed to the public
+    tokenRegistry:
+      hostname: scan.sv-2.scratcha.global.canton.network.digitalasset.com
```
pasindutennage-da: what should we set here / or should this be set in pulumi dynamically?

cluster/configs/shared/rate-limits/v0-acs.yaml
```diff
@@ -6,3 +6,45 @@ rateLimits:
     maxTokens: 10
     tokensPerFill: 5
     fillInterval: 60s
+  /registry/allocations/v1:
```
moritzkiefer-da: hm a bit unfortunate that we need to list all endpoints individually here. iirc we had some safeguards somewhere that ensures that we don't miss endpoints, can we extend them to cover this. In particular I'd like to avoid that we break this for token std v2.

cluster/configs/shared/base.yaml
moritzkiefer-da: probably a question for @stephencompall-DA

cluster/configs/shared/base.yaml
moritzkiefer-da: hardcoding definitely won't work, so either you get rid of the need to set it or you need to set it from pulumi

cluster/configs/shared/base.yaml
```diff
@@ -207,7 +207,13 @@ cloudArmor:
...
+      pathPrefix: /registry
+      throttleAcrossAllEndpointsAllIps:
+        withinIntervalSeconds: 60
+        maxRequestsBeforeHttp429: 200
```
moritzkiefer-da, line 235: what number do we currently get on the prod clusters? we need to make sure this is not lower than what people already expect.

cluster/configs/shared/rate-limits/v0-acs.yaml
moritzkiefer-da: ah I see you've done that below already

moritzkiefer-da: you're adding this to v0-acs.yaml. probably want a separate file instead, this is not about acs endpoints

cluster/deployment/scratchnetb/config.resolved.yaml
```diff
@@ -9,6 +9,12 @@ cloudArmor:
       throttleAcrossAllEndpointsAllIps:
         maxRequestsBeforeHttp429: 0
         withinIntervalSeconds: 60
+    tokenRegistry:
```
moritzkiefer-da, line 12: you're enabling a global cloud armor limit but cloud armor is disabled. I was expecting we try to solve this in istio for now, is that not an option?

moritzkiefer-da: try setting the limit low enough that you can easily trigger it during testing and see that you trigger it

cluster/configs/shared/base.yaml
stephencompall-DA: Setting this on is meaningless without L7 ALB gateway enabled, but there are some unsolved routing problems with that. Right now everything goes directly to istio.

So, not yet.

cluster/configs/shared/rate-limits/v0-acs.yaml
pasindutennage-da: Thanks.

pasindutennage-da: Fixed

cluster/configs/shared/base.yaml
pasindutennage-da: Thanks. Leaving it disabled

cluster/configs/shared/base.yaml
pasindutennage-da: Thanks, fixed

cluster/configs/shared/base.yaml
pasindutennage-da: `sum(rate(istio_requests_total{destination_app="scan-app"}[1m])) * 60` == on average 12k requests per minute

` /registry/.*` returns no data

Considering the differences between token and scan end points, would a maxRequestsBeforeHttp429=12k/4=3000 make sense (mostly a holistic value, but we can increase it gradually, while looking at usage?)

cluster/deployment/scratchnetb/config.resolved.yaml
pasindutennage-da: Hi @moritzkiefer-da  @nicu-da 

Sure, will check in scratchnet with low limits.

I have a question about the scope:: As of now, we allow an endpoint to have either a per-IP token bucket (clientIp: true) or a global shared bucket. These two types are mutually exclusive. https://github.com/canton-network/splice/issues/5651 asks for both global limits and per-IP limits, AFAICU. Since Cloud Armor isn't available, we can't achieve both without refactoring the Pulumi rate limit logic to map two separate descriptors to a single route/path.

For this PR, should I stick to clientIp: true for all token endpoints (and hence no global limits), or just have global limits? or should I first refactor rate limiting logic to support both options for given end point? (sounds like different PR)

cluster/configs/shared/base.yaml
pasindutennage-da: given that cloudarmour isn't working anyway, there is no point in tuning this value for now right?

cluster/deployment/scratchnetb/config.resolved.yaml
moritzkiefer-da: can you check with Stephen when we expect cloud armor to be enabled? If the answer is soon we can maybe wait for that, if not then we do need to refactor our istio limits.

pasindutennage-da: Discussed, decided to refactor istio

cluster/configs/shared/rate-limits/token-registry.yaml
```diff
@@ -0,0 +1,121 @@
+rateLimits:
+  /registry/allocations/v1:
+    name: registry-allocations
+    type: limited
+    maxTokens: 3000
```
pasindutennage-da: Actual limits to be discussed, and tuned

pasindutennage-da:
@martinflorian-da @moritzkiefer-da @stephencompall-DA 

This code is ready for review now. 

I used mostly holistic values for `maxTokens` and `tokensPerFill`. Let me know if you know better values (or whether we have metrics on those numbers?)

Thanks

cluster/configs/shared/rate-limits/token-registry.yaml
moritzkiefer-da: The limits you configured here seem extremely bursty. Afaict this means that if there is no traffic for 15min you can immediately burst 3000 requests initially. Do we have a good reason to expect that:
1. We can support this type of burst
2. We actually get these bursts?

I would default to the opposite: Set tokensPerFill to maxTokens or maybe half of that so you spread out traffic more evenly and avoid those burst, maybe 500requests/min or something

cluster/pulumi/infra/src/cloudArmor.ts
```diff
@@ -147,9 +147,11 @@ function addThrottleAndBanRules(
       // this makes the pulumi update cleaner if toggling just one service
       if (throttleAcrossAllEndpointsAllIps.maxRequestsBeforeHttp429 > 0) {
         const ruleName = `throttle-all-endpoints-all-ips-${confEntryHead}`;
-
+        const hostnameRegex = hostname
```
moritzkiefer-da, line 150: why are we changing anything in cloud armor?

cluster/pulumi/infra/src/istio.ts
```diff
@@ -759,6 +759,40 @@ function configurePublicInfo(ingressNs: k8s.core.v1.Namespace): k8s.apiextension
     : [];
 }
 
+function configurePublicTokenRegistry(
```
moritzkiefer-da, line 762: I think it might make sense to guard this behind a config setting in config.yaml, that way we can easily turn it off or on if we encounter issues.

moritzkiefer-da (approved):
Thank you, the changes look good to me. A few questions on testing:

1. Did you test that the whitelisting works correctly i.e. you can access the registry endpoints without vpn but not the other endpoints?
2. Your curl commands show that you have a rate limit on the endpoint for your own IP. But I don't see how they demonstrate that the cross-IP rate limit works. How have you validated that?

pasindutennage-da:
@moritzkiefer-da 

Please see the answers below

## Token public and scan vpn only checking

**With VPN**

a. `curl -i https://scan.sv-2.scratcha.network.canton.global/api/scan/v0/acs/party123::123`

```
HTTP/2 200 
traceparent: 00-50d5e9e498c3355498b0080df31ff2ad-5aeff99e76f12cbe-03
server: istio-envoy
date: Thu, 09 Jul 2026 09:28:31 GMT
content-type: application/json
content-length: 47
x-envoy-upstream-service-time: 29

{"acs_snapshot":"H4sIAAAAAAAA/wMAAAAAAAAAAAA="}
```

b. `curl -i https://scan.sv-2.scratcha.network.canton.global/registry/metadata/v1/info`

```
HTTP/2 200 
traceparent: 00-64400cf90dff680f9bb9b855c9af585d-bf6ea671de0ac777-03
server: istio-envoy
date: Thu, 09 Jul 2026 09:29:27 GMT
content-type: application/json
content-length: 138
x-envoy-upstream-service-time: 14

{"adminId":"DSO::1220e1abe2c71cd7c227d8f52a2def9eb4533818e74d20de68a11cef0bf6b705bd1a","supportedApis":{"splice-api-token-metadata-v1":1}}

```


**Without VPN**

a. `curl -i https://scan.sv-2.scratcha.network.canton.global/api/scan/v0/acs/party123::123`

```
HTTP/2 403 
content-length: 19
content-type: text/plain
date: Thu, 09 Jul 2026 09:30:44 GMT
server: istio-envoy

RBAC: access denied
```

b. `curl -i https://scan.sv-2.scratcha.network.canton.global/registry/metadata/v1/info`

```
HTTP/2 200 
traceparent: 00-7999d4ddd47e0b4f418f2bbc65dab86b-2cac37bb6b1688fc-03
server: istio-envoy
date: Thu, 09 Jul 2026 09:31:07 GMT
content-type: application/json
content-length: 138
x-envoy-upstream-service-time: 6

{"adminId":"DSO::1220e1abe2c71cd7c227d8f52a2def9eb4533818e74d20de68a11cef0bf6b705bd1a","supportedApis":{"splice-api-token-metadata-v1":1}}

```

## Verify rate limiting across many IPs to hit global limit

```
  /registry/metadata/v1/info:
    name: registry-metadata-info
    type: limited
    maxTokens: 30
    tokensPerFill: 1
    fillInterval: 60s
    perIpLimits:
      maxTokens: 10
      tokensPerFill: 5
      fillInterval: 60s
```

```
#!/bin/bash

ENDPOINT="https://scan.sv-2.scratcha.network.canton.global/registry/metadata/v1/info"


echo "Sending 10 requests from current IP..."

for i in $(seq 1 10); do
  curl -s -o /dev/null -w "Request $i: %{http_code}\n" $ENDPOINT
done

```

1. With IP-1

```
Sending 10 requests from current IP...
Request 1: 200
Request 2: 200
Request 3: 200
Request 4: 200
Request 5: 200
Request 6: 200
Request 7: 200
Request 8: 200
Request 9: 200
Request 10: 200
```

2. With IP-2

```
Sending 10 requests from current IP...
Request 1: 200
Request 2: 200
Request 3: 200
Request 4: 200
Request 5: 200
Request 6: 200
Request 7: 200
Request 8: 200
Request 9: 200
Request 10: 200

```

3. With IP-3

```
Sending 10 requests from current IP...
Request 1: 200
Request 2: 200
Request 3: 200
Request 4: 200
Request 5: 200
Request 6: 200
Request 7: 200
Request 8: 200
Request 9: 200
Request 10: 200
```

4. With IP-4

```
Sending 10 requests from current IP...
Request 1: 200
Request 2: 429
Request 3: 429
Request 4: 429
Request 5: 429
Request 6: 429
Request 7: 429
Request 8: 429
Request 9: 429
Request 10: 429

```

5. With IP-5

```
Sending 10 requests from current IP...
Request 1: 429
Request 2: 429
Request 3: 429
Request 4: 429
Request 5: 429
Request 6: 429
Request 7: 429
Request 8: 429
Request 9: 429
Request 10: 429
```

## Why cloud armour changes are needed

To generate WAF rules for Token Registry paths. Before, allowedPathsCondition hardcoded the '/api/scan/' prefix when computing the regex match pattern for Cloud Armor rules. This hardcoding fails for /registry/ prefix. And I fixed it. Additionally, I made hostname an optional parameter to allow Cloud Armor to fall back to the cluster DNS.  For now, because CloudArmour is disabled, we won't see the impact, but when Steven completes CloudArmour work, it will work for Token end points.

cluster/pulumi/infra/src/cloudArmor.ts
pasindutennage-da: Please see my comment below.

cluster/pulumi/infra/src/istio.ts
pasindutennage-da: Fixed.

PR #5975, opened 2026-06-16
SvApp: Add UnhideRewardCouponV2Trigger and ExpireRewardCouponV2Trigger

### LINKED ISSUE #5715: SvApp: Add UnhideRewardCouponV2Trigger and ExpireRewardCouponV2Trigger

dfordivam:
Fixes #5715 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V072__dso_acs_store_reward_beneficiary_is_observer.sql
```diff
@@ -0,0 +1,7 @@
+-- For storing RewardCouponV2's provider/beneficiary is observer
+alter table dso_acs_store
+  add column reward_beneficiary_is_observer boolean;
+
+create index dso_acs_store_sid_mid_pn_tid_rbio
```
meiersi-da: @rautenrieth-da : this index creation will result in downtime for the SV app won't it? One option to limit the downtime would be to make the partial index more partial and also constrain it on the `template-id`, so the corresponding index can be used during its creation.

Or would you rather recommend to use the delayed index creation infrastructure?

meiersi-da: note @dfordivam : downtime for the SV app is a problem as the SVs will may miss a reward coupon if they are down for too long.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpireRewardCouponV2Trigger.scala
```diff
@@ -0,0 +1,142 @@
...
+      ExpireRewardCouponV2Trigger.listExpiredVettedRewardCouponsV2(
+        svTaskContext,
+        context.loggerFactory.getTracedLogger(classOf[ExpireRewardCouponV2Trigger]),
+      ),
+      splice.amulet.RewardCouponV2.COMPANION,
+      svTaskContext.vettingLookupService,
+      PackageIdResolver.Package.SpliceAmulet,
+      payload => (payload.dso +: observerParties(payload)).map(PartyId.tryFromProtoPrimitive(_)),
+    )
+    with SvTaskBasedTrigger[Task] {
```
meiersi-da: Comparing to this recent addition of a trigger https://github.com/canton-network/splice/pull/5964/changes#diff-415ea762d792a54415fcb004b7ec7ebe6b3355217b7a089edde6e22b5bf2900eR48, you are not use the `IgnoredAmuletVersionGuard`, probably by intention.  

Any particular reason why that trigger's tooling is insufficient for this usecase?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpireRewardCouponV2Trigger.scala
```diff
@@ -0,0 +1,142 @@
...
+      }
+    } yield {
+      val (vetted, skipped) = groups.partition { case (_, supported) => supported }
+      val skippedCoupons = skipped.flatMap(_._1)
+      if (skippedCoupons.nonEmpty)
+        logger.debug(
+          s"Skipping ${skippedCoupons.size} contracts whose observers are not correctly vetted."
+        )
+      vetted.flatMap(_._1)
+    }
```
meiersi-da: @moritzkiefer-da @julientinguely-da : you've recently worked on improved support for unresponsive parties or parties with the wrong vetting state. Is there support code that @dfordivam could reuse here to implement this? Any concern with this code?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/UnhideRewardCouponV2Trigger.scala
```diff
@@ -0,0 +1,134 @@
...
+  ): Future[Boolean] = {
+    val now = context.clock.now
+    for {
+      coupons <- store.listNonObserverRewardCouponsV2ForProvider(
+        task.providerParty,
+        PageLimit.tryCreate(1),
+      )
+      support <- svTaskContext.packageVersionSupport
+        .supportsTrafficBasedAppRewards(Seq(task.providerParty), now)
+    } yield coupons.isEmpty || !support.supported
```
meiersi-da, line 67: This is essentially the same function as what's implemented above, just with a different page limit. Consider sharing a `listProvidersWithUnhidableCoupons(pageLimit: PageLimit)`

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/UnhideRewardCouponV2Trigger.scala
```diff
@@ -0,0 +1,134 @@
...
+          else
+            for {
+              _ <- unhideBatch(coupons, controller)
+              _ = logger.debug(
+                s"Unhid batch of ${coupons.size} reward coupons for ${task.providerParty}"
+              )
+              result <- loop(totalUnhidden + coupons.size)
+            } yield result
+      } yield result
+    loop(0).map(count => TaskSuccess(s"unhid $count reward coupons for ${task.providerParty}"))
```
meiersi-da: Do we need this loop here? Can't we just rely on the top-level retrieve-tasks to take care of the looping?

I'm a bit worried that this extra loop will make interpreting the trigger's activity unnecessarily confusing.

I suspect changing the `Task` definition to `Task(provider: Party, coupons: Seq[CouponCid])` would be required to make the staleness check after completion work out. That doesn't seem like a problem. It actually feels like a simplification in terms of reasoning about the trigger's behavior.

meiersi-da (commented):
Thanks for the good work. Not exactly a trivial change. I would value @rautenrieth-da review on this as well.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/UnhideRewardCouponV2Trigger.scala
dfordivam: I interpreted the design-doc description as a single task execution handling all coupons. But it could be a separate task.

> task completion
>   * unhide all the beneficiaries RewardCouponV2 contracts in batches by calling DsoRules_UnhideRewardCouponsV2
>   * complete once no new batch remains

Also In case of multiple SVs trying the unhide, there would be clash, so the staleness check will need to handle that. This probably needs some more thinking

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V072__dso_acs_store_reward_beneficiary_is_observer.sql
dfordivam: I have done investigations and confirmed that there is no way to make the index creation faster.
The index creation wont make use of other indices and will go through all the rows
I did some experiments locally and confirmed that all rows are in fact read when creating the index.

Even a delayed index creation will also scan all the rows, and it appears to be a heavy change for our needs, given that we know that we don't have any existing rows in DB to index.

So I looked into alternate ways to make this work. The simplest approach looks like reuse of existing indexes for our query.

Specifically we have the following index already, and the `sv_onboarding_token` is a very specific column.

    "dso_acs_store_sid_pn_tid_sot" btree (store_id, migration_id, package_name, template_id_qualified_name, sv_onboarding_token) WHERE sv_onboarding_token IS NOT NULL

We could make this column and index generic by rename to `aux_lookup_key`, and add the provider to this when `providerIsBeneficiary` is false. This would work just fine for our query as the shape of `dso_acs_store_sid_pn_tid_sot` is identical to our `reward_party` one.

The usage of `sv_onboarding_token` is minimal/straightforward in the code, only`lookupSvOnboardingRequestByTokenWithOffset` .

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/UnhideRewardCouponV2Trigger.scala
dfordivam: On further consideration, I think it makes sense to get rid of this loop because of the following reasoning.

The V2 daml was introduced in the 0.6.7,and the TBA would be enabled on 0.6.11 onwards, so most of validators would have vetted by time. The cause of the wrong vetting state would be mostly operator errors.

And I think that the average size of coupons to unhide will be lower than 100, which is around 16 hours(16*6=96)
as the 16 hours time window of wrong wetting state for the app providers appears to be on the high end.

If we assume that the average count is around/below 100, and the max is 36*6=216.
I think that we should just increase the default batch size to 220, such that we almost have a guarantee that all the coupons should be handled in a single trigger run.
This would avoid the complications around SVs race.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V072__dso_acs_store_reward_beneficiary_is_observer.sql
meiersi-da: Good proposal. I like it. @rautenrieth-da wdyt?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpireRewardCouponV2Trigger.scala
dfordivam: I missed the `IgnoredAmuletVersionGuard` and it looks necessary even for our trigger. Though it is somewhat orthogonal to the `supportsTrafficBasedAppRewards` based filtering, I think I can make integrate both together.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/UnhideRewardCouponV2Trigger.scala
dfordivam: There is a slight difference, the above does random sample, and this one confirms existing coupons of task's party.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpireRewardCouponV2Trigger.scala
dfordivam: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpireRewardCouponV2Trigger.scala
dfordivam: This code block is now gone, please take a look at refactored/simplified trigger.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/UnhideRewardCouponV2Trigger.scala
dfordivam: Removed the loop and bumped the default size to 220

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpireRewardCouponV2Trigger.scala
dfordivam: Hmm now on re-review I think the vetting based filtering is sufficient in our case right?
We need not do ignore for the parties specified in `IgnoredPartiesStore` as these appear to be for contracts with party as signatory, and are probably offline?
Or do we need to have an option to do ignore via the unresponsive parties list?

meiersi-da: > Hmm now on re-review I think the vetting based filtering is sufficient in our case right?
We need not do ignore for the parties specified in IgnoredPartiesStore as these appear to be for contracts with party as signatory, and are probably offline?
Or do we need to have an option to do ignore via the unresponsive parties list?

You are probably right that we don't need it. Assuming we got everything right. I'm OK with making the assumption and adding the more complex code after the fact.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
```diff
@@ -84,10 +86,10 @@ private[delegatebased] abstract class ProcessRewardsTriggerBase(
       batch <- batchF
       dsoRules <- dsoRulesF
       damlBatch = convertBatch(batch)
+      providersWithWrongVettingState <- determineProvidersWithWrongVettingState(batch)
       choiceArg = new ProcessRewardsV2_ProcessBatch(
         damlBatch,
-        // TODO (#5715) determine 'providersWithWrongVettingState'
-        new DamlSet(java.util.Collections.emptyMap()),
+        providersWithWrongVettingState,
```
meiersi-da, line 92: Please record the size of this set in a metric.

meiersi-da: And create an issue for @rautenrieth-da to add it to the dashboard, and alerts.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/UnhideRewardCouponV2Trigger.scala
meiersi-da: OK

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/UnhideRewardCouponV2Trigger.scala
```diff
@@ -0,0 +1,123 @@
...
+    val batchSize = svConfig.delegatelessAutomationUnhideRewardCouponV2BatchSize
+    for {
+      coupons <- store.listNonObserverRewardCouponsV2ForProvider(
+        task.providerParty,
+        PageLimit.tryCreate(batchSize),
+      )
+      result <-
+        if (coupons.isEmpty)
+          Future.successful(
+            TaskSuccess(s"no reward coupons to unhide for ${task.providerParty}")
```
meiersi-da:
```suggestion
            TaskSkipped(s"no reward coupons to unhide for ${task.providerParty}")
```

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V072__dso_acs_store_reward_beneficiary_is_observer.sql
meiersi-da: I just discussed the case with @rautenrieth-da , and we concluded that it seems better to avoid the confusion, and instead use the infrastructure that exists for lazy index creation in: https://github.com/canton-network/splice/blob/43bbf37fb3d7b91dd059aa98af193ed1a020a8ac/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/automation/SqlIndexInitializationTrigger.scala#L228-L242

It was purpose-built for this problem and should work fine for this use-case.

meiersi-da (approved):
Thanks. Good work!

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
dfordivam: I realized this will conflict with Robert's #5896, and it is a fairly small change which could be added on top of 5896 or in a separate PR after it merges. So I will defer adding in this PR.
Also confirming we should be using Gauge for this right?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExpireRewardCouponV2Trigger.scala
dfordivam: Removed the ignore parties related code

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V072__dso_acs_store_reward_beneficiary_is_observer.sql
dfordivam: Ok, now using this for index creation

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ProcessRewardsTrigger.scala
meiersi-da: yes, gauge is right.

PR #5976, opened 2026-06-16
add a workflow to automatically request and upgrade test on changes to Pulumi or Nix packages

mblaze-da:
Fixes https://github.com/DACH-NY/canton-network-internal/issues/4002

### DIFF
--- a/.github/workflows/pr_auto_upgrade_test.yml
+++ b/.github/workflows/pr_auto_upgrade_test.yml
@@ -1,0 +1,27 @@
+name: Automatically trigger an upgrade test on PRs
+on:
+  pull_request_target:
+    # This is only triggered on PR creation to avoid spamming the PR with comments on every update.
+    # It is up to the author to decided if subsequent changes require another upgrade test.
+    types: [ opened ]
+    paths:
+      - 'cluster/pulumi/**/package.json'
+      - 'nix/**'
+permissions:
+  contents: read
+  issues: read
+  pull-requests: write # To comment back on the PR
+jobs:
+  request_upgrade_test:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Add a comment to the PR to request an upgrade test
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          script : |
+            await github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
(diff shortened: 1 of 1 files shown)


moritzkiefer-da (approved):
nice thanks

.github/workflows/pr_auto_upgrade_test.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - name: Add a comment to the PR to request an upgrade test
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          script : |
+            await github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: "Requesting an /upgrade_test due to changes in Pulumi or Nix package versions."
```
isegall-da, line 26: Did you test this? I thought the comment needs to start with the /upgrade_test string, but may be wrong

.github/workflows/pr_auto_upgrade_test.yml
```diff
@@ -0,0 +1,27 @@
+name: Automatically trigger an upgrade test on PRs
+on:
+  pull_request_target:
```
isegall-da, line 3: Since this runs on free runners and does not require secrets, on: pull_request is also fine, and simpler

isegall-da:
@mblaze-da I know this is already merged, but please note some comments I left above

.github/workflows/pr_auto_upgrade_test.yml
mblaze-da: It uses the `contains` function which as far as I've checked searches the whole string.

isegall-da: OK thanks for checking

.github/workflows/pr_auto_upgrade_test.yml
mblaze-da: AFAIU the difference, `pull_request_target` will always run the workflow from main. Is this undesireable here? What would be simplified by changing to `pull_request`?

isegall-da: The question should be the opposite. The default should be `on: pull_request`, as that is guaranteed not to use Splice resources/secrets, as for forks it runs on the fork, not on Splice. We should use `on: pull_request_target` only when there's a need for it, as it moves things closer to Splice. Here there's no need for it AFAICT.

PR #5940, opened 2026-06-12
Change TransferPreapprovalProposal retry delays

julientinguely-da:
fixes https://github.com/DACH-NY/cn-test-failures/issues/8765

see [difference between full and equal jitters](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/#:~:text=Looking%20at%20the%20amount%20of%20client%20work%2C%20the%20number%20of%20calls%20is%20approximately%20the%20same%20for%20%E2%80%9CFull%E2%80%9D%20and%20%E2%80%9CEqual%E2%80%9D%20jitter%2C%20and%20higher%20for%20%E2%80%9CDecorrelated%E2%80%9D.%20Both%20cut%20down%20work%20substantially%20relative%20to%20both%20the%20no%2Djitter%20approaches.) introduced by this canton [bump](https://github.com/canton-network/splice/pull/5876/changes#diff-e13ac08be0cf01a470e6e5f870fb3e323f95aafdea1d10970351386f5f36b045) from 2 days ago. Nice catch Moritz!!

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If an upgrade test is required, comment `/upgrade_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a logical synchronizer upgrade test is required (from canton-3.5), comment `/lsu_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/canton-network/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/canton-network/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #5988, opened 2026-06-16
test(wallet): make TransferPreapproval failure test robust to validator timeout

samsondav:
Make the "Failure to complete TransferPreapproval creation should be handled correctly" wallet test robust to a transient validator HTTP timeout that flakes the log assertion under CI load.

### Flake

Observed on the `wall-clock-time (0)` job: https://github.com/canton-network/splice/actions/runs/27450544466 (job: https://github.com/canton-network/splice/actions/runs/27450544466/job/81436663390)

```
forEvery failed, because:
  at index 0, Incorrect log level WARN. Expected: ERROR.
  Message: Request to .../api/validator/v0/wallet/transfer-preapproval (POST) resulted in a timeout after 38 seconds.
  Remaining log entries:
  ## ERROR ... HttpCommandException: HTTP 503 Service Unavailable POST at '/api/validator/v0/wallet/transfer-preapproval' ... The server is taking too long to respond
```

The test pauses validator automation so `createTransferPreapproval` cannot complete. The intended outcome is an HTTP 429 once the server-side `RetryFor.ClientCalls` loop in `HttpWalletHandler` gives up (`ABORTED` -> `TooManyRequests`). Under load the validator's HTTP request timeout fired first: `HttpErrorHandler.timeoutHandler` logged a WARN "resulted in a timeout after 38 seconds" and returned 503 instead. The strict ordered assertion in `assertThrowsAndLogsCommandFailures` then matched the unexpected WARN entry at index 0 with `errorMessage`, which requires ERROR level, and failed.

### Fix

Assert the failure via `assertThrowsAndLogsUnorderedOptional[CommandFailure]`: a required ERROR entry matching either "429 Too Many Requests" or "503 Service Unavailable", plus an optional WARN entry matching the server-side "resulted in a timeout". This keeps the requirement that a genuine failure is surfaced and logged as ERROR, while tolerating the transient timeout WARN/503 on a slow host. Mirrors the existing pattern in `ScanIntegrationTest` and `SplitwellUpgradeIntegrationTest`.

Observed on PR #5937's CI run. The fix is by inspection; the integration test was not run locally.

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletIntegrationTest.scala
@@ -27,6 +27,7 @@ import org.lfdecentralizedtrust.splice.wallet.store.{
 import com.digitalasset.canton.config.RequireTypes.PositiveInt
 import com.digitalasset.canton.console.CommandFailure
 import com.digitalasset.canton.data.CantonTimestamp
+import com.digitalasset.canton.logging.SuppressingLogger.LogEntryOptionality
 import com.digitalasset.canton.logging.SuppressionRule
 import com.digitalasset.canton.topology.{PartyId, SynchronizerId}
 import com.digitalasset.canton.{HasExecutionContext, SynchronizerAlias}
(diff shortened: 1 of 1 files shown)


moritzkiefer-da:
I believe this is already fixed in https://github.com/canton-network/splice/pull/5940. have you still encountered this issue after rebasing to include that change?

samsondav:
Thanks Moritz — you're right. #5937 branched ~45 min before #5940 merged and wasn't rebased, so that run didn't include the retry-delay fix. I'll rebase onto main to pick up #5940 and confirm; assuming the flake is gone I'll close this in favor of #5940.

PR #5992, opened 2026-06-16
Limit number of legs and parties in Amulet allocations and their settlement on off-ledger APIs

OriolMunoz-da:
Fixes https://github.com/canton-network/splice/issues/5556


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardAllocationInstructionHandler.scala
```diff
@@ -89,6 +94,22 @@ class HttpTokenStandardAllocationInstructionHandler(
...
+        ) match {
+          case Success(allocate) =>
+            Future {
+              tokenStandardSettlementConfig.validateAllocate(allocate)
+            }
+          case Failure(err) =>
+            Future.failed(
+              io.grpc.Status.INVALID_ARGUMENT
+                .withDescription(
+                  s"Field `choiceArguments` does not contain a valid `${allocationinstructionv2.AllocationFactory.CHOICE_AllocationFactory_Allocate.name}`: $err"
```
meiersi-da:
```suggestion
                  s"Field `choiceArguments` does not contain a valid `${allocationinstructionv2.AllocationFactory.CHOICE_AllocationFactory_Allocate.name}` value: $err"
```

meiersi-da (approved):
Nice. Thanks.

PR #5993, opened 2026-06-16
[ci] use DbStorageMulti to prevent concurrent Splice app instances.

### LINKED ISSUE #4080: Guard against duplicate splice apps breaking the DB
Due to operational mistakes it's possible to (temporarily or maybe also less temporarily) end up with e.g. two validator app instances running. One example of that is if you run kubectl delete pod which can start the new pod before the old one is dead.

While some parts of our apps are probably resilient against this, e.g. UpdateIngestion due to sufficient unqiue constraints other parts probably are not (e.g. the kv stores) We also don't really test for this so it's easy to end up in a situation where we break it even for parts that do work right now.

Canton uses DB locks in the participant to guard against this. We probably want to copy the same approach for our own apps.

ray-roestenburg-da:
Fixes https://github.com/canton-network/splice/issues/4080

#### Cluster Testing
- [x] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceStorageFactory.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceStorageFactory.scala
@@ -7,7 +7,12 @@ import com.digitalasset.canton.concurrent.FutureSupervisor
 import com.digitalasset.canton.config.{ReplicationConfig, StorageConfig}
 import com.digitalasset.canton.lifecycle.FutureUnlessShutdown
 import com.digitalasset.canton.logging.NamedLoggerFactory
-import com.digitalasset.canton.resource.{DbLockCounter, StorageFactory, StorageMultiFactory}
+import com.digitalasset.canton.resource.{
+  DbLockCounter,
+  StorageFactory,
+  StorageMultiFactory,
+  StorageSingleFactory,
+}
 import com.digitalasset.canton.tracing.TraceContext
 
 import java.util.concurrent.atomic.AtomicReference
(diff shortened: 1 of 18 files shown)


ray-roestenburg-da:
cluster_test succeeded.

ray-roestenburg-da:
Running perf test https://github.com/canton-network/splice/actions/runs/27814391925

ray-roestenburg-da:
Performance test succeeded, see grafana, no alerts (ran around 8:30 UTC)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceStorageFactory.scala
```diff
@@ -0,0 +1,99 @@
...
+      onPassive: () => Unit,
+  ): StorageFactory =
+    if (!instanceLockEnabled) new StorageSingleFactory(storage)
+    else
+      new StorageMultiFactory(
+        config = storage,
+        exitOnFatalFailures = exitOnFatalFailures,
+        replicationConfig = ReplicationConfig.withDefaultO(storage, None),
+        onActive = _ => FutureUnlessShutdown.unit,
+        onPassive = logger => {
```
moritzkiefer-da, line 85: what exactly happens in this case? Does it crash the node? Does it fail readiness probes? Does it fail liveness probes? If it doesn't crash will it recover again?

moritzkiefer-da (approved):
thanks! have you done some manual testing of trying to spin up two apps against the same db and see that it works and it recovers if you stop the other? the unit tests look good but a bit hard to see if everything is wired up correctly so the manual test seems valuable.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceStorageFactory.scala
ray-roestenburg-da: yes, on passive it crashes / exits. If another one has the lock, this node is passive and crashes.

ray-roestenburg-da: if it gets a transient connection loss it will try to connect again and only after a timeout waiting to reconnect it crashes / exits. (timeout / delay is configured with `storageConfig.failedToFatalDelay`). during the transient loss it will fail readyness, liveness succeeds (which allows attempts to reconnect within the window).

ray-roestenburg-da:
I'll do a manual test 👍

ray-roestenburg-da:
Tested locally with 1 sv, 1 scan, and a lock from a test, looks good.

PR #6007, opened 2026-06-17
Allow null ACS Commitment Reconciliation value in Set DSO Rules

### LINKED ISSUE #6003: Allow `null` value in ACS Commitment Reconciliation Interval
Front-end needs to stop mapping ACS Commitment Reconciliation in Set DSO Rules to "" when the value is null:

Image

pawelperek-da:
Fixes #6003 

Can vote `null`:
Screenshot 2026-06-17 at 10 58 48

Voting on another fields within the `decentralizedSynchronizer` map:
Screenshot 2026-06-17 at 12 07 50

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/frontend/src/utils/buildDsoRulesConfigFromChanges.ts
+++ b/apps/sv/frontend/src/utils/buildDsoRulesConfigFromChanges.ts
@@ -34,12 +34,14 @@ export function buildDsoRulesConfigFromChanges(dsoConfigChanges: ConfigChange[])
 
   for (let i = 1; i <= synchronizerCount; i++) {
     const key = getValue(`decentralizedSynchronizer${i}`);
+    const acsCommitmentReconciliationInterval = getValue(
+      `decentralizedSynchronizerAcsCommitmentReconciliationInterval${i}`
+    );
     const value = {
       state: getValue(`decentralizedSynchronizerState${i}`) as SynchronizerState,
       cometBftGenesisJson: getValue(`decentralizedSynchronizerCometBftGenesisJson${i}`),
-      acsCommitmentReconciliationInterval: getValue(
-        `decentralizedSynchronizerAcsCommitmentReconciliationInterval${i}`
-      ),
+      acsCommitmentReconciliationInterval:
+        acsCommitmentReconciliationInterval === '' ? null : acsCommitmentReconciliationInterval,
     };
     synchronizers = synchronizers.set(key, value);
   }
--- a/apps/sv/frontend/src/utils/buildDsoRulesConfigFromChanges.ts
+++ b/apps/sv/frontend/src/utils/buildDsoRulesConfigFromChanges.ts
@@ -18,13 +18,15 @@ export function buildDsoRulesConfigFromChanges(dsoConfigChanges: ConfigChange[])
   const changeMap = new Map<string, string>();
 
   dsoConfigChanges.forEach(change => {
-    changeMap.set(change.fieldName, change.newValue.toString());
+    changeMap.set(change.fieldName, change.newValue);
   });
 
-  const getValue = (fieldName: string, fallbackValue: string = '') => {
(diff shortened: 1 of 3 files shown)


apps/sv/frontend/src/utils/buildDsoRulesConfigFromChanges.ts
```diff
@@ -34,12 +34,14 @@ export function buildDsoRulesConfigFromChanges(dsoConfigChanges: ConfigChange[])
...
+      `decentralizedSynchronizerAcsCommitmentReconciliationInterval${i}`
+    );
     const value = {
       state: getValue(`decentralizedSynchronizerState${i}`) as SynchronizerState,
       cometBftGenesisJson: getValue(`decentralizedSynchronizerCometBftGenesisJson${i}`),
-      acsCommitmentReconciliationInterval: getValue(
-        `decentralizedSynchronizerAcsCommitmentReconciliationInterval${i}`
-      ),
+      acsCommitmentReconciliationInterval:
+        acsCommitmentReconciliationInterval === '' ? null : acsCommitmentReconciliationInterval,
```
pawelperek-da: We use similar nullification logic in set amulet rules, I'm not sure why it was missed here

moritzkiefer-da: Any chance of putting that logic into getValue instead? This seems like a thing we'll just keep forgetting and we break it again.

pawelperek-da: Sounds good but I don't think all fields are supposed to be nullable?

pawelperek-da: I don't think we can easily query nullability information from codegen

moritzkiefer-da: but we must have some place where we convert null => empty string to even get to an empty string as that's not what is stored in the Daml state. So surely we must also be able to preserve that information when converting back?

pawelperek-da: I made a mockup and it works well in all but one case:
1. FE receives `"acsCommitmentReconciliation": 1800` from the ledger
2. I cleanup the `acsCommitmentReconciliation` field
3. Now FE don't know wether it should serialise the field to `""` or `null`

We still need some sort of compile-time tagging to know whether value is nullable (thus serialised to `null`) or non-nullable (thus serialised to `""`). 
Come to think of it, I hope we do not have any fields for which both `null` and `""` are correct values, because right now UI has no way of expressing both

moritzkiefer-da: hm unfortunate but I don't see a great way around, it doesn't look like the codegen exposes any runtime type information we could check sadly, we only get the encode/decode methods. I guess you could do something horrible like trying to encode null and catch the exception, if it succeeds the field is nullable, if it fails use an empty string. But that feels pretty gross.

Another option might be to make it a mandatory argument to getValue whether a field should be nullable or not. That doesn't mean you can't get it wrong but at least you're forced to make an explicit choice when you add a field so it's less likely we accidentally miss this somewhere.

>  Come to think of it, I hope we do not have any fields for which both null and "" are correct values, because right now UI has no way of expressing both

It's possible in theory, in Daml those two are definitely different. But I think we don't have such a case yet or if we do the two are at least equivalent.

pawelperek-da: I actually kinda like the idea with getValue having mandatory nullability argument

moritzkiefer-da (approved):
thanks! the nullability argument looks like a pretty reasonable option

PR #6008, opened 2026-06-17
Don't include amulet rules and newest open round in token standard contexts

OriolMunoz-da:
# DO NOT MERGE UNTIL TBAR CONFIG HITS MAINNET

Fixes https://github.com/canton-network/splice/issues/4950

Minimnum init versions obtained from https://github.com/canton-network/splice/pull/3487/changes#diff-d2555057df0d1d7e4b946dee2ceeabeac77c2d51c94f0912e02c6b119bd7ad12R10 (24h submission PR, which removes the requirement for these contexts)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpTokenStandardAllocationInstructionHandler.scala
```diff
@@ -48,16 +48,6 @@ class HttpTokenStandardAllocationInstructionHandler(
       for {
         externalPartyAmuletRules <- store.getExternalPartyAmuletRules()
         amuletRules <- store.getAmuletRules()
-        newestOpenRound <- store
```
moritzkiefer-da, line 51: @OriolMunoz-da I was just wondering about this as I was looking at some transactions, this will also break BFT reads won't it? So once we do actually proceed with this pr we actually need to tie it to a switch over time?

moritzkiefer-da: I believe with https://github.com/canton-network/splice/pull/6000/changes#diff-369890d9086e525a35678ce3e29e062fbef90310f5c96bec940bb578b43efd99 you are unblocked on the tbar stuff? i think dry run mode is sufficient?

OriolMunoz-da: > this will also break BFT reads won't it? 

yeah

> So once we do actually proceed with this pr we actually need to tie it to a switch over time?

We only have a problem if <f+1 scans have upgraded AND <f+1 are in the old version AND the rest are unavailable, otherwise there will be consensus one way or the other (and log noise from scans disagreeing with consensus), no?

moritzkiefer-da: > We only have a problem if <f+1 scans have upgraded AND <f+1 are in the old version AND the rest are unavailable, otherwise there will be consensus one way or the other (and log noise from scans disagreeing with consensus), no?

you query 2f+1 scans and you expect f+1 to give you the same response. If everyone upgrades and noone is down this does indeed just work. But sadly people being down is not a theoretical issue, last time we tried this kind of reasoning it resulted in an incident because one (or more I don't recall) scans were down so out of the 2f+1 scans you queried one gave you nothing (or an outdated response, don't fully recall) and the others had f for one answer and f for the other answer. So I think we do need to bite the bullet and handle it properly sadly.

OriolMunoz-da: > last time we tried this kind of reasoning it resulted in an incident

you've convinced me :smile: 

this will require then adding a switch over check and a vote for it?

moritzkiefer-da: > this will require then adding a switch over check and a vote for it?

yep

meiersi-da: @OriolMunoz-da : thanks for starting to work on this. Given the complication and the pressure on scan pruning, I think this can wait to after scan pruning is done. wdyt?

OriolMunoz-da: wasn't starting on it now, the PR exists since end of tsv2. I've been indeed prio'ing scan pruning

PR #6009, opened 2026-06-17
Adjust scan mismatch thresholds

pasindutennage-da:
Fix https://github.com/DACH-NY/canton-network-internal/issues/5700


canton-network-da:
# [backport] Reminder #
- [ ] release-line-0.6.8
- [ ] release-line-0.6.7
- [ ] release-line-0.6.6
- [ ] release-line-0.6.5

nicu-da (approved):
Thx!

cluster/pulumi/observability/grafana-alerting/scan_connection_disagreement_alerts.yaml
```diff
@@ -133,7 +133,7 @@ groups:
         annotations:
           __dashboardUid__: cnndx7p
           __panelId__: "6"
-          description: A scan connection returned a successful (2xx) response that disagreed with the BFT consensus result in the last 10 minutes. A successful but divergent response indicates the scan node is serving incorrect or stale data and should be investigated.
+          description: A scan connection returned more than 5 successful (2xx) responses that disagreed with the BFT consensus result in the last 10 minutes. A successful but divergent response indicates the scan node is serving incorrect or stale data and should be investigated.
```
martinflorian-da, line 136: Hm the reasonable threshold depends on how often we query scan, which could vary based on network and activity? I honestly wonder if this absolute value-based alert buys us so much compared to the first one.

@nicu-da WDYT? Can we just remove that one?

nicu-da: It might make sense to convert it to percentage as well. 
This alert is only on success calls, so ideally we want to have it be more sensitive than the other one.

pasindutennage-da: @nicu-da @martinflorian-da Should I change it now to be a % rather than abs value? or ok to merge?

rautenrieth-da: What situation are we trying to detect with these alerts? A single disagreement on the successful result of `/v0/events/{update_id}` is already a rather serious issue, a disagreement on the result of `/v0/open-and-issuing-mining-rounds` is perfectly normal behavior.

So not only do the alerts depend on network activity, but also on how we do scan BFT calls from our apps. When we start doing many BFT calls on an eventually consistent endpoint, we will have to bump the threshold so high that we would miss disagreements on always consistent endpoints.

rautenrieth-da: Could we perhaps include a whitelist of endpoints that should always be consistent, and filter the `bft_per_connection_consensus` metric by the `request` label? Then we could set the alert threshold to zero. Or in general have separate alerts by request, so that it's easier to silence those that are known to be noisy.

nicu-da: > So not only do the alerts depend on network activity, but also on how we do scan BFT calls from our apps. When we start doing many BFT calls on an eventually consistent endpoint, we will have to bump the threshold so high that we would miss disagreements on always consistent endpoints.

if it's percentage based then it shouldn't matter the volume of calls right? 

> Could we perhaps include a whitelist of endpoints that should always be consistent, and filter the bft_per_connection_consensus metric by the request label? Then we could set the alert threshold to zero. Or in general have separate alerts by request, so that it's easier to silence those that are known to be noisy.

I think it would make sense to have ignores for specific requests, but honestly i would first wait to see real data from devnet. If we have scan connections that are constantly lagging and are out of sync we should actually pursue those SVs to fix their scans.

Issue #5884, opened 2026-06-09
support for data-staging of bulk storage


isegall-da:
Goals:
- Confirm BFT agreement between SVs before committing to objects in bulk storage, thus detecting divergence early and making it easier to recover from it
- Avoid granting the app overwrite/delete permissions on committed objects, thus preventing accidents

Idea: Use two buckets: A staging bucket where the app has full read/write/delete perms, and a committed bucket with only write-once and read. Write objects to staging first, move them to committed once confirmed BFT agreement.

Implementation plan: 
- [x] Disable bulk storage on CILR & DevNet where we enabled it already (only we, not other SVs), and delete the data (to not have to deal with bwd compat)
- [x] #6162
- [x] #6332
- [x] #6495
- [x] #6960
- [ ] Enable on all SVs on CILR to test BFT
- [ ] ~After confirmed to work, disable BFT checks and reduce to being enabled on only 1 or 2 SVs~ check costs and consider if we can keep it on on all SVs long term
- [x] #7021
- [ ] Document and advise enabling on all SVs on DevNet

PR #6018, opened 2026-06-17
support a staging bucket in bulk storage

### LINKED ISSUE #6162: Introduce the basic mechanism for the dual-bucket approach, including moving objects, but without BFT checks still

isegall-da:
Fixes #6162 
(See #5884 for context)

~Could benefit from some more unit testing of the update history part, but this PR is a behemoth as it is (sorry, couldn't break it in a very sensible way), so I'm leaving that as a followup PR (and hence not competing #6126 here)~ on closer look, the two pipelines share 99% of the code now, and I don't see significant coverage gaps actually.

Note that the changes are not bwd compatible, so require disabling bulk storage and deleting the data on CILR&DevNet where we enabled it (tracked in #5884)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


isegall-da:
@rautenrieth-da @ray-roestenburg-da I know this PR is a bit of a monster (sorry..), but can I get a review still please? :pray:

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
```diff
@@ -11,12 +11,20 @@ import org.apache.pekko.Done
...
 import org.lfdecentralizedtrust.splice.environment.RetryProvider
 
 import scala.concurrent.{ExecutionContext, Future}
 
+trait RetryableService[S] {
+  def asRetryableService(
+      automationConfig: AutomationConfig,
+      backoffClock: Clock,
+      retryProvider: RetryProvider,
+  )(implicit tracer: Tracer): PekkoRetryingService[S]
```
ray-roestenburg-da, line 25: should this not return `RetryingService[S]`? or rename the trait to `PekkoRetryableService`

ray-roestenburg-da (approved):
Thanks! looks good, I only had one minor question

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
```diff
@@ -154,6 +155,44 @@ class S3BucketConnection(
...
+    )
+    val copyReq = CopyObjectRequest
+      .builder()
+      .destinationBucket(bucketName)
+      .destinationKey(key)
+      .sourceBucket(sourceBucket)
+      .sourceKey(key)
+      .build()
+
+    s3Client.copyObject(copyReq).asScala.map(_ => ())
```
ray-roestenburg-da, line 188: Is unwrapCompletionException also useful here? (and in deleteObject)

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/HasS3Mock.scala
```diff
@@ -75,29 +77,32 @@ trait HasS3Mock
...
+          .delete(
+            Delete.builder().objects(objs.asJava).build()
+          )
+          .build()
+        _ <-
+          if (list.contents().asScala.isEmpty) { Future.successful(()) }
+          else { client.deleteObjects(deleteObjRequest).asScala }
+      } yield {
+        ()
+      }).futureValue
```
ray-roestenburg-da: Looks like you are ignoring the future value?

ray-roestenburg-da: Oh wait scalatest, so this will just fail, never mind

ray-roestenburg-da: (maybe slightly better to map over the buckets and future sequence them and then get the futureValue on that)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/BulkStorageCommitFromStaging.scala
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  private def copyToCommitted: Flow[
+    (T, Seq[ObjectKeyAndChecksum]),
+    (T, Seq[ObjectKeyAndChecksum]),
+    NotUsed,
+  ] =
+    Flow[(T, Seq[ObjectKeyAndChecksum])]
+      .mapAsync(parallelism = 1) { case (ts, objs) =>
```
ray-roestenburg-da, line 92: nit: you could set a val in this class to 1 and pass it to all mapAsyncs here cause it's important that's always one. `val noParallelism = 1` or something

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/UpdateHistoryBulkStorage.scala
```diff
@@ -70,140 +75,90 @@ class UpdateHistoryBulkStorage(
...
-      Source
-        .tick(0.seconds, appConfig.updatesPollingInterval.underlying, ())
-        .mapAsync(1)(_ =>
-          if (updateHistory.isReady)
-            updateHistory.isHistoryBackfilled(currentMigrationId)
-          else Future.successful(false)
-        )
-        .filter(identity)
-        .take(1)
+//    // Wait for update history to initialize and for history backfilling to complete before starting bulk storage dumps
```
ray-roestenburg-da: nit: Prefer no commented code, rather remove it.

isegall-da: whoops, for sure, this is just a leftover

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/PekkoRetryingService.scala
isegall-da: renamed the trait, and also the method.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
isegall-da: Not sure what you mean, and in what way you think it would be useful.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/BulkStorageCommitFromStaging.scala
isegall-da: I actually am not sure if all these parallelism = 1 are actually necessary, or we can tolerate parallelism. Left a TODO to revisit that. Clearly 1 is safest, so leaving as 1 for now.

isegall-da:
(will merge on Monday, didn't yet merge the PR on cn-internal, and don't want to do that right before logging off for the long weekend)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
ray-roestenburg-da: you call a java async method, it seems you used that in the other async call so why not here?

isegall-da: I used it only in one place above, where an exception is intended, so I need the equivalent of a try-catch. That's not the case here.

PR #6019, opened 2026-06-17
Helm charts: Vault Injection 

pasindutennage-da:
Fix https://github.com/DACH-NY/canton-network-internal/issues/5714

### Overview

This PR introduces the `secretOverrides` configuration map across all component Helm charts, enabling operators to completely bypass Kubernetes `secretKeyRef`. It allows the injection of raw configuration strings that are resolved dynamically at runtime by external secret provider hooks.

### Target Audience

This feature is designed for operators who use secret management engines such as HashiCorp Vault.

### When to Use 

Use this option when your infrastructure relies on a mutating webhook injector that requires reading reference URIs or secret lookup paths.

### How to Use

1. Define the `secretOverrides` map under the target component in your custom `helm/values-template.yaml` file.
2. Supply the external secret engine lookup path or reference string.
3. Ensure your cluster has the corresponding mutating webhook, injector installed to intercept and translate these string tokens into credentials at runtime.

### What NOT to Do

Do not use this field to hardcode plaintext or cleartext passwords into your Git repositories
This field is explicitly designed for *references*, not raw data.


cluster/helm/splice-domain/tests/synchronizer_test.yaml
nicu-da, line 1: lets use synchronizer instead of domain for any new stuff we add (including renaming this)

nicu-da (commented):
can you add a short description on how is this expected to be used? when is the injection being done? 
Wondering if there's any way of preventing people from yoloing password directly into that

nicu-da (approved):
LGTM, though not ideal as we can't test it.

PR #6027, opened 2026-06-18
Refactor jdbc types tests

rautenrieth-da:
The main motivation is to have tests that show that we can handle empty arrays without an explicit type annotation, at least when using sql interpolation.

The previous tests also only used `GetResult` implicits for the individual tests; only the one test using `TestTable` actually wrote any data and used `SetParameter` implicits.


OriolMunoz-da (approved):
ty!

PR #6030, opened 2026-06-18
make auto upgrade test workflow directly request the test and refactor pr_cluster_test

mblaze-da:
@isegall-da I've addressed your comment about the trigger. I've also made it so it skips the comment and directly requests the test so it should now work @moritzkiefer-da.


moritzkiefer-da (approved):
thanks lgtm, have you tested this?

mblaze-da:
> thanks lgtm, have you tested this?

Not yet, I'm planning to test right after I merge. If there are any small issues I'll iron them out immediately. If it's something larger then I'll revert and do another iteration. Setting it up so that it's testable requires more work which I'd like to avoid.

PR #6043, opened 2026-06-18
Handle unavailable scan in ProcessRewardsTrigger

rautenrieth-da:
This PR changes ProcessRewardsTrigger to fall back to BFT reads also if scan is unavailable (i.e., if the request fails with an exception).

Previously, the fallback only handled the case where scan successfully answered that it has no data.


dfordivam (commented):
LGTM

meiersi-da (approved):
Thanks.

PR #6058, opened 2026-06-19
make auto upgrade test workflow directly request the test and refactor pr_cluster_test

mblaze-da:
The first commit contains the not-quite-working changes from https://github.com/canton-network/splice/pull/6030. The second and third commit, address the shortcomings, which are:
- incorrectly using workflow_call as an event type which doesn't exist - workflow calls inherit the parent event type
- messing up the formatting of the GH GraphQL query in `get_head`

### DIFF
--- a/.github/workflows/pr_auto_upgrade_test.yml
+++ b/.github/workflows/pr_auto_upgrade_test.yml
@@ -1,27 +1,21 @@
 name: Automatically trigger an upgrade test on PRs
 on:
-  pull_request_target:
+  pull_request:
     # This is only triggered on PR creation to avoid spamming the PR with comments on every update.
     # It is up to the author to decided if subsequent changes require another upgrade test.
     types: [ opened ]
     paths:
       - 'cluster/pulumi/**/package.json'
       - 'nix/**'
-permissions:
+permissions: # permissions have to match pr_cluster_test permissions as called workflows cannot escalate their own permissions
   contents: read
   issues: read
-  pull-requests: write # To comment back on the PR
+  pull-requests: write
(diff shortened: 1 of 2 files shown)


moritzkiefer-da (approved):
looks plausible but so did the prior version so let's make sure to test directly after merging

.github/workflows/pr_auto_upgrade_test.yml
```diff
@@ -1,27 +1,21 @@
...
-        with:
-          script : |
-            await github.rest.issues.createComment({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              issue_number: context.issue.number,
-              body: "Requesting an /upgrade_test due to changes in Pulumi or Nix package versions."
-            });
+    uses: ./.github/workflows/pr_cluster_test.yml
+    secrets: inherit
```
isegall-da, line 18: If you do need secrets now, then on: pull_request doesn't work any more

isegall-da: But why do you need secrets? I don't see that you use them

mblaze-da: I have to double check that.

mblaze-da: Yeah, so now that we directly trigger the CCI workflow it runs `google-github-actions/auth` which uses some secrets.

mblaze-da: So I'll revert to `pull_request_target`.

PR #6062, opened 2026-06-19
fix conditions in pr_cluster_test

mblaze-da:
This is a bit embarrasing but I choose to blame lack of a type system for this.


moritzkiefer-da (approved):
yaml is my favorite programming language

PR #6072, opened 2026-06-22
PackageWithDependencyIntegrationTest runs only for pv35 or greater

pasindutennage-da:
Fix https://github.com/DACH-NY/cn-test-failures/issues/8799

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
@@ -115,13 +115,11 @@ class PackageWithDependencyIntegrationTest extends AdditionalPackagesToUnvetInte
         vettedPackageIds should contain noElementsOf darsWithMissingDependency.map(_.packageId)
         vettedPackageIds should contain(missingDependency.packageId)
       }
-
       stopAllAsync(
         sv1Backend,
         sv1ValidatorBackend,
       ).futureValue
     }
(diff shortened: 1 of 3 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AdditionalPackagesToUnvetIntegrationTest.scala
```diff
@@ -103,41 +104,46 @@ class PackageWithDependencyIntegrationTest extends AdditionalPackagesToUnvetInte
...
-      eventually() {
-        val vettedPackageIds = getVettedPackageIds(
-          sv1Backend.appState.participantAdminConnection,
-          synchronizerId,
-        )
-        vettedPackageIds should contain noElementsOf darsWithMissingDependency.map(_.packageId)
-        vettedPackageIds should contain(missingDependency.packageId)
+    val currentPv = sv1Backend.config.localSynchronizerNodes.current.protocolVersion
+
+    if (currentPv >= ProtocolVersion.v35) {
```
moritzkiefer-da: that feels a bit hacky, can we add a scalatest tag instead and filter out all tests with the tag  in the pv compat test so it doesn't get run at all instead of running but doing nothing?

pasindutennage-da: @moritzkiefer-da 

I added a tag and excluded it in runtime. Please let me know if thats the right way to achieve it. Thanks

moritzkiefer-da (approved):
looks reasonable thx

PR #6073, opened 2026-06-22
helm, info, status: Check and report reachability sequencer + refactor

giner:
Possible status values:
- Sequencer: 0 (reachable and not lagging), 1 (lagging), 2 (unreachable), 3 (unreachable and lagging).

### DIFF
--- a/cluster/helm/splice-info/scripts/get-status.sh
+++ b/cluster/helm/splice-info/scripts/get-status.sh
@@ -11,14 +11,24 @@ SCAN_URL="${SCAN_URL:-http://scan-app:5012}"
 SV_THRESHOLD="${SV_THRESHOLD:-600}"
 MEDIATOR_THRESHOLD="${MEDIATOR_THRESHOLD:-900}"
 SCAN_THRESHOLD_ROUNDS="${SCAN_THRESHOLD_ROUNDS:-900}"
-SCAN_THRESHOLD_EVENT="${SCAN_THRESHOLD_EVENT:-300}"
+SCAN_THRESHOLD_EVENTS="${SCAN_THRESHOLD_EVENT:-300}"
 SEQUENCER_THRESHOLD="${SEQUENCER_THRESHOLD:-2520}" # 42 minutes, Sequencer acknowledgments are irregular, so we use a higher threshold here
 
-CURL_TIMEOUT="${CURL_TIMEOUT:-15}"
+PARALLELISM="${PARALLELISM:-8}"
 
+CURL_TIMEOUT="${CURL_TIMEOUT:-15}"
 TLS_SKIP_VERIFY="${TLS_SKIP_VERIFY:-false}"
+
 CURL_CMD=(curl -fs -m "$CURL_TIMEOUT")
-[[ $TLS_SKIP_VERIFY == true ]] && CURL_CMD+=(-k)
+GRPC_HEALTH_CMD=(grpc_health_code --max-time "$CURL_TIMEOUT")
+
+if [[ $TLS_SKIP_VERIFY == true ]]; then
+  CURL_CMD+=(-k)
+  GRPC_HEALTH_CMD+=(--insecure)
+fi
+
+CURL_CMD_JSON=$(jq -nc --args '$ARGS.positional' -- "${CURL_CMD[@]}")
+GRPC_HEALTH_CMD_JSON=$(jq -nc --args '$ARGS.positional' -- "${GRPC_HEALTH_CMD[@]}")
 
 prom2json() {
   P2J_VERSION="1.5.0"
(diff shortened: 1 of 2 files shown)


martinflorian-da:
> Deploy cluster test triggered for [Commit 01a3ced14e736c8d92fc1b7b9fb3d072cb36585b in ](https://github.com/canton-network/splice/pull/6073/files/01a3ced14e736c8d92fc1b7b9fb3d072cb36585b), please contact a Contributor to approve it in CircleCI: https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/71265

green: https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/71266/workflows/49d239f5-71f5-46d3-9c3b-0780a226216e

cluster/helm/splice-info/scripts/get-status.sh
```diff
@@ -42,6 +47,29 @@ prom2json() {
...
+  GRPCURL_ARCH="linux_x86_64"
+  GRPCURL_BIN="$HOME/.grpcurl-$GRPCURL_VERSION"
+  GRPCURL_URL="https://github.com/fullstorydev/grpcurl/releases/download/v$GRPCURL_VERSION/grpcurl_${GRPCURL_VERSION}_$GRPCURL_ARCH.tar.gz"
+  GRPCURL_EXPECTED_SHA="a926b62a85787ccf73ef8736b3ae554f1242e39d92bb8767a79d6dd23b11d1d5"
+
+  if [[ ! -f "$GRPCURL_BIN" ]]; then
+    GRPCURL_DIST=$(mktemp)
+    GRPCURL_TMPDIR=$(mktemp -d)
+
+    echo "Downloading grpcurl..." >&2
```
martinflorian-da: Can it happen that we do this multiple times in parallel if we run `grpcurl` in parallel? Can't we move that installation to some init step? I get that we don't want to change to image just for that... but that last-moment install seems a bit fishy.

martinflorian-da (approved):
I agree with Itai that this is hard to review. The downside of getting something wrong here is limited though IMO, so happy to rush this a bit to make the release cut tomorrow, and we can always fix things later on...

@giner I'll fix your merge conflict and set automerge now, so we get a little bit of more testing before the release cut.

cluster/helm/splice-info/scripts/get-status.sh
martinflorian-da: (happy for that to be a follow-up improvement, if you think it makes sense)

martinflorian-da (dismissed):
I actually need to retract my approval following some input from a watchful colleague :tm:. Downloading remote binaries in a bash script is fairly sketchy from a supply chain security standpoint; at the very least I'd like us to review this with less haste.

How robust is that sha check against future code changes to this script, for example?

Perhaps the clean way *is* to use a different image that has `grpcurl` baked in?

cluster/helm/splice-info/scripts/get-status.sh
giner: You are right, while it doesn't break anything here we do download in parallel. thank you for catching, I'll fix that.

giner:
> I agree with Itai that this is hard to review.

Let me try to refactor it a bit on the next iteration (remove some verbose bash 3 compatibility stuff, split into smaller easier to read, self described pieces). If it's still hard to read after that I'll look into rewriting it Python.

giner:
> Perhaps the clean way is to use a different image that has `grpcurl` baked in?

I couldn't find a good image containing all: `jq`, `curl`, `prom2json` and `grpcurl`. If downloading at runtime is no go we'll have to manage another image.

cluster/helm/splice-info/scripts/get-status.sh
giner: Fixed in https://github.com/canton-network/splice/pull/6073/changes/e18d260525e3900cf0494931bb17f77a3d33911f

moritzkiefer-da:
There are two somewhat orthogonal issues for me:

1. Downloading tools at runtime is a hard no for me here.
2. grpcurl itself is in a bit of a dodgy state. If you look through the issue tracker there are a bunch of issues for dependencies with known vulnerabilities without a release that fixes it and it gets flagged by most docker image scanners. That's why we just removed it from the base image of all the actual apps. I'm not necessarily saying that those vulnerabilities do apply the way we use it here but I don't even want to have to reason about this and explain this every time an auditor runs a scan.

I suspect in most SVs setups the info pod has direct access to things like participant admin APIs so it is highly security critical so we definitely should err on the side of being overly cautious.

If we can build an image containing https://github.com/grpc-ecosystem/grpc-health-probe to get the same info, that seems like a reasonable option. I don't really know any other alternative that is well maintained and doesn't get flagged. We could write some custom python program or similar but at that point, I'd also start to somewhat question whether the complexity is worth the ROI.

giner:
Thank you @moritzkiefer-da. Would it work if for now I just replace grpcurl with grpc-health-probe but keep downloading at runtime (we already do this for prom2json in the same script) and within the next few iteration I would refactor the script for more readability and add a docker image?

moritzkiefer-da:
Given that I don't see that we're in a particular rush to land this change I'd rather do it the other way around and switch to a base image containing the tools first and make the script more readable instead of adding more complexity and potentially risky downloads first.

giner:
Alright, converting draft for now

martinflorian-da:
@giner To be clear: It seems reasonable that we add a `splice-info` image of some sort for this. The base image we use for all splice images already contains grpc-health-probe which should also work here: https://github.com/canton-network/canton-base-images

About this PR: You could also shrink it to only scan monitoring, and follow up with the sequencer things in a follow up :shrug:

giner:
@martinflorian-da scan has been already monitored, in this PR I was only splitting the status into two distinct states. It's quite minor from monitoring point of view while having sequencer monitored is more substantial.

As for the base image, yes, it has `grpc-health-probe` but missing `curl`, `jq` and `prom2json`. I'm guessing we need a separate image for these?

giner:
The PR is ready for review.

Please let me know if more parts should be decomposed and named.

Here is what has been done:
- replaced `grpcurl` by a simple bash function `grpc_health_code` (so no more downloading of grpcurl)
- extracted concurrency to a separate function `run_parallel`
- applied simpler common patterns for processing data and falling back
- integrated changes from `main`

If everything looks fine I'll rebase, squash into a single commit and clean up the commit description.

giner:
Rebased on main and squashed

giner:
Rebased on the current main

isegall-da (approved):
I still think that the readability gain of doing this in python would have greatly outweighed the downsides, but I accept that you see things differently.

PR #6079, opened 2026-06-22
Vault Injection: _helpers.tpl  

pasindutennage-da:
Fix: https://github.com/DACH-NY/canton-network-internal/issues/5764

### DIFF
--- a/cluster/helm/splice-participant/tests/participant_test.yaml
+++ b/cluster/helm/splice-participant/tests/participant_test.yaml
@@ -341,6 +341,7 @@ tests:
     set:
       persistence.secretOverrides.postgresPassword: "vault:kv/data/participant/postgres-password"
       persistence.enablePgInitContainer: true
+      disableAuth: true
     documentSelector:
       path: kind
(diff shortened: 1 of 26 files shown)


cluster/helm/splice-participant/tests/participant_test.yaml
```diff
@@ -341,6 +341,7 @@ tests:
     set:
       persistence.secretOverrides.postgresPassword: "vault:kv/data/participant/postgres-password"
       persistence.enablePgInitContainer: true
+      disableAuth: true
```
moritzkiefer-da: why are we disabling auth? shouldn't the test actually work with auth enabled exactly to demonstrate that client id/secret and whatever else may be required is not set through secretKeyRef?

pasindutennage-da: @moritzkiefer-da Thanks. I fixed this.

moritzkiefer-da (approved):
thx looks plausible

PR #6092, opened 2026-06-23
Fixed ScanConnectionIntegrationTest flake

pasindutennage-da:
Fix https://github.com/DACH-NY/cn-test-failures/issues/8856

During eventually(40.seconds) sv1ScanBackend wait period, aliceValidator runs CO_BuyMemberTraffic, which updates the cache. Hence by the time the tap goes through, alice has the upto date cache, and hence no more further cache updates!

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanConnectionIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanConnectionIntegrationTest.scala
@@ -24,33 +24,39 @@ class ScanConnectionIntegrationTest
     // tap once so the AmuletRules are cached...
     aliceWalletClient.tap(5)
     val baseConfig = sv1ScanBackend.getAmuletRules().contract.payload.configSchedule.initialValue
-    clue("change the amulet config to invalidate the cache.") {
-      val newConfig =
-        mkUpdatedAmuletConfig(
-          sv1ScanBackend.getAmuletRules().contract,
-          tickDuration = defaultTickDuration,
-          maxNumInputs = 101,
-        )
-
-      setAmuletConfig(
-        Seq((Some(durationUntilOffboardingEffectivity), newConfig, baseConfig)),
-        durationUntilExpiration,
-      )
-    }
-    eventually(40.seconds) {
-      sv1ScanBackend
-        .getAmuletRules()
-        .contract
-        .payload
-        .configSchedule
-        .initialValue
-        .transferConfig
-        .maxNumInputs shouldBe 101
-    }
 
     loggerFactory.assertEventuallyLogsSeq(SuppressionRule.LevelAndAbove(Level.DEBUG))(
-      // tapping again..
-      aliceWalletClient.tap(5),
+      {
+        clue("change the amulet config to invalidate the cache.") {
+          val newConfig =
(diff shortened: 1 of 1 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanConnectionIntegrationTest.scala
```diff
@@ -24,33 +24,39 @@ class ScanConnectionIntegrationTest
...
-        .initialValue
-        .transferConfig
-        .maxNumInputs shouldBe 101
-    }
 
     loggerFactory.assertEventuallyLogsSeq(SuppressionRule.LevelAndAbove(Level.DEBUG))(
-      // tapping again..
-      aliceWalletClient.tap(5),
+      {
+        clue("change the amulet config to invalidate the cache.") {
```
moritzkiefer-da, line 30: Why is this guaranteed to be safe? You've moved it a bit later but why does this guarantee that the buy member traffic can not still interfere with things?

pasindutennage-da: ## Previous implementation 

Earlier the problem was in our assumption that until the 2nd tap() is made, the aliceValidator would not send any request that has AmuletRules. With this assumption, we checked logs to assert that aliceValidator updated its cache.

But, the problem with that assumption was: even before the second tap was made, aliceValidator had updated its cache. In this specific run it was a CO_BuyMemberTraffic that was run in the background. But in other runs, it can be something else (if there are more background tasks in validator app that refers to AmuletRules)


## Fix

My proposal to solve this: have the log assertion about cache invalidation in the same block as `setAmuletConfig`. So, we monitor logs for everything happened after setAmuletConfig, so whichever event triggers aliceValidator to send a request that has amulet rules contract, will trigger our expect log line.

> buy member traffic can not still interfere with things

Indeed it can interfere, but rather than destrucitve interference, it just helps log assertion pass -- like i said, we don't really care what triggered the scan cache update, we just care that it eventuall does.

moritzkiefer-da: can't you still end up with the traffic top-up triggering before you update the config which satisfies the log assertion so you're not actually checking that the cache gets invalidated after the update?

pasindutennage-da: t1: 1st tap in alice -- amulet rules contract cached at alice
t2: config-update that causes amulet rules archival and new generation
t3: 2nd tap

I am assuming that, in this perticular test, during (t1,t2) no other bakground task changes amulet rules. If that assmption is correct, then alice has no reason to update the cache until t2.

moritzkiefer-da (approved):

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/ScanConnectionIntegrationTest.scala
pasindutennage-da: but then the assertion text should be only for "alice-validator".... that is missing. is it what you are hinting at?

PR #6099, opened 2026-06-24
Add dryRun label to TBAR triggers

rautenrieth-da:
Previously we filtered by dry run in the dashboard by matching the trigger name, which is needlessly complicated.


dfordivam (commented):
Looks good to me! I added the label to the trigger I am working on, will open a PR shortly

dfordivam (approved):

PR #6100, opened 2026-06-24
SV app: Add RewardMetricsTrigger

### LINKED ISSUE #6026: SV-App Add RewardMetricsTrigger
Polling trigger in the SV app that computes and reports the following metrics

   * Number of active CalculateRewardsV2 and ProcessRewardsV2 contracts (move from where they are currently captured)
   * gauge for parties with hidden coupons: one gauge for the number of hidden coupons, labelled with party-id

dfordivam:
Fixes #6026 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
@@ -1,0 +1,141 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package org.lfdecentralizedtrust.splice.sv.automation
+
+import com.daml.metrics.api.MetricHandle.{Gauge, LabeledMetricsFactory}
+import com.daml.metrics.api.MetricQualification.Saturation
+import com.daml.metrics.api.{MetricInfo, MetricName, MetricsContext}
+import com.digitalasset.canton.lifecycle.{AsyncOrSyncCloseable, SyncCloseable}
+import com.digitalasset.canton.topology.PartyId
+import com.digitalasset.canton.tracing.TraceContext
+import io.opentelemetry.api.trace.Tracer
+import org.apache.pekko.stream.Materializer
+import org.lfdecentralizedtrust.splice.automation.{PollingTrigger, TriggerContext}
+import org.lfdecentralizedtrust.splice.environment.SpliceMetrics
+import org.lfdecentralizedtrust.splice.store.PageLimit
+import org.lfdecentralizedtrust.splice.sv.automation.RewardMetricsTrigger.RewardMetrics
+import org.lfdecentralizedtrust.splice.sv.store.SvDsoStore
+
+import scala.collection.concurrent.TrieMap
+import scala.concurrent.{blocking, ExecutionContext, Future}
+
+class RewardMetricsTrigger(
+    override protected val context: TriggerContext,
+    dsoStore: SvDsoStore,
+    hiddenCouponProvidersSampleSize: Int,
+)(implicit
+    override val ec: ExecutionContext,
+    override val tracer: Tracer,
+    val mat: Materializer,
+) extends PollingTrigger {
+
+  private val rewardMetrics = new RewardMetrics(context.metricsFactory)
+
+  override def performWorkIfAvailable()(implicit traceContext: TraceContext): Future[Boolean] =
+    for {
+      calculateRewards <- dsoStore.listCalculateRewardsV2()
+      _ = {
+        val (dryRun, minting) = calculateRewards.partition(_.payload.dryRun)
+        rewardMetrics.calculateRewardsActiveContractsDryRun.updateValue(dryRun.size)
+        rewardMetrics.calculateRewardsActiveContractsMinting.updateValue(minting.size)
+      }
+      processRewards <- dsoStore.listProcessRewardsV2()
+      _ = {
+        val (dryRun, minting) = processRewards.partition(_.payload.dryRun)
+        rewardMetrics.processRewardsActiveContractsDryRun.updateValue(dryRun.size)
+        rewardMetrics.processRewardsActiveContractsMinting.updateValue(minting.size)
+      }
+      providers <- dsoStore.listNonObserverRewardCouponsV2ProvidersSample(
+        PageLimit.tryCreate(hiddenCouponProvidersSampleSize)
+      )
+      _ <- Future.traverse(providers) { provider =>
+        dsoStore
+          .listNonObserverRewardCouponsV2ForProvider(provider, PageLimit.Max)
+          .map(coupons => rewardMetrics.updateHiddenCoupons(provider, coupons.size.toLong))
+      }
+      _ = rewardMetrics.pruneHiddenCouponGauges(providers)
+    } yield false
+
+  override def closeAsync(): Seq[AsyncOrSyncCloseable] =
+    super.closeAsync().appended(SyncCloseable("reward metrics", rewardMetrics.close()))
+}
+
+object RewardMetricsTrigger {
+
+  class RewardMetrics(metricsFactory: LabeledMetricsFactory) extends AutoCloseable {
+
+    private val prefix: MetricName = SpliceMetrics.MetricsPrefix
+
+    private def activeContractsGauge(
+        template: String,
+        templateName: String,
+        dryRun: Boolean,
+    ): Gauge[Int] =
+      metricsFactory.gauge(
+        MetricInfo(
+          name = prefix :+ template :+ "active_contracts",
+          summary =
+            s"The number of active $templateName contracts, as seen by the RewardMetricsTrigger",
+          qualification = Saturation,
+        ),
+        initial = -1,
+      )(MetricsContext.Empty.withExtraLabels("dryRun" -> dryRun.toString))
+
+    val calculateRewardsActiveContractsDryRun: Gauge[Int] =
+      activeContractsGauge("calculate_rewards_v2", "CalculateRewardsV2", dryRun = true)
+    val calculateRewardsActiveContractsMinting: Gauge[Int] =
+      activeContractsGauge("calculate_rewards_v2", "CalculateRewardsV2", dryRun = false)
+    val processRewardsActiveContractsDryRun: Gauge[Int] =
+      activeContractsGauge("process_rewards_v2", "ProcessRewardsV2", dryRun = true)
+    val processRewardsActiveContractsMinting: Gauge[Int] =
+      activeContractsGauge("process_rewards_v2", "ProcessRewardsV2", dryRun = false)
+
+    private val hiddenCouponGauges: TrieMap[PartyId, Gauge[Long]] = TrieMap.empty
+
+    @SuppressWarnings(Array("com.digitalasset.canton.RequireBlocking"))
+    private def getHiddenCouponGauge(provider: PartyId): Gauge[Long] =
+      hiddenCouponGauges.getOrElse(
+        provider,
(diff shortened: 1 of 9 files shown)


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
```diff
@@ -0,0 +1,141 @@
...
+    val processRewardsActiveContractsMinting: Gauge[Int] =
+      activeContractsGauge("process_rewards_v2", "ProcessRewardsV2", dryRun = false)
+
+    private val hiddenCouponGauges: TrieMap[PartyId, Gauge[Long]] = TrieMap.empty
+
+    @SuppressWarnings(Array("com.digitalasset.canton.RequireBlocking"))
+    private def getHiddenCouponGauge(provider: PartyId): Gauge[Long] =
+      hiddenCouponGauges.getOrElse(
+        provider,
+        // We must synchronize here to avoid allocating the metric for the same party multiple
```
dfordivam: This is based on the design of `AmuletPriceMetricsTrigger`

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
```diff
@@ -0,0 +1,141 @@
...
+    val mat: Materializer,
+) extends PollingTrigger {
+
+  private val rewardMetrics = new RewardMetrics(context.metricsFactory)
+
+  override def performWorkIfAvailable()(implicit traceContext: TraceContext): Future[Boolean] =
+    for {
+      calculateRewards <- dsoStore.listCalculateRewardsV2()
+      _ = {
+        val (dryRun, minting) = calculateRewards.partition(_.payload.dryRun)
```
rautenrieth-da, line 43: `listCalculateRewardsV2` returns at most 1000 entries, so the two `dryRun` and `minting` numbers may underestimate the actual count. We would get a log warning if that happens, and we have an alert that triggers on either metric being higher than 1, so we're fine. Consider adding a comment though.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
```diff
@@ -0,0 +1,141 @@
...
+        rewardMetrics.calculateRewardsActiveContractsDryRun.updateValue(dryRun.size)
+        rewardMetrics.calculateRewardsActiveContractsMinting.updateValue(minting.size)
+      }
+      processRewards <- dsoStore.listProcessRewardsV2()
+      _ = {
+        val (dryRun, minting) = processRewards.partition(_.payload.dryRun)
+        rewardMetrics.processRewardsActiveContractsDryRun.updateValue(dryRun.size)
+        rewardMetrics.processRewardsActiveContractsMinting.updateValue(minting.size)
+      }
+      providers <- dsoStore.listNonObserverRewardCouponsV2ProvidersSample(
```
rautenrieth-da: Do I understand this correctly, in each round we pick an independent random sample of providers that have active hidden coupons?

How do we expect that list of providers to evolve? If there are many such providers and new ones keep appearing, then this will generate a lot of metric series over time. And if we add a panel to grafana that shows a leaderboard of offending parties (like we have in [unresponsive_parties.json](https://github.com/canton-network/splice/blob/main/cluster/pulumi/observability/grafana-dashboards/canton/unresponsive_parties.json)), then the table content will change every 30sec due to the random sample.

If we expect this to not be an issue in practice, let's please add a corresponding comment in the code.

Apart from the above, I like this. Having a list of offending parties in grafana is very useful for monitoring.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
```diff
@@ -0,0 +1,141 @@
...
+        }
+      }
+    }
+
+    override def close(): Unit = {
+      calculateRewardsActiveContractsDryRun.close()
+      calculateRewardsActiveContractsMinting.close()
+      processRewardsActiveContractsDryRun.close()
+      processRewardsActiveContractsMinting.close()
+      hiddenCouponGauges.values.foreach(_.close())
```
rautenrieth-da: Similar here. This should not require a synchronization in practice, but make sure it's consistent with`pruneHiddenCouponGauges`.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
```diff
@@ -0,0 +1,141 @@
...
+      )
+
+    def updateHiddenCoupons(provider: PartyId, count: Long): Unit =
+      getHiddenCouponGauge(provider).updateValue(count)
+
+    @SuppressWarnings(Array("com.digitalasset.canton.RequireBlocking"))
+    def pruneHiddenCouponGauges(currentProviders: Seq[PartyId]): Unit = {
+      val toClose = hiddenCouponGauges.keySet.toSet -- currentProviders.toSet
+      hiddenCouponGauges.view.filterKeys(toClose.contains).foreach(_._2.close())
+      blocking {
```
rautenrieth-da: This is _probably_ not needed, as we should never call `pruneHiddenCouponGauges` concurrently with `close` or `getHiddenCouponGauge`. If you still want to keep it, then I'd include the 
```
hiddenCouponGauges.view.filterKeys(toClose.contains).foreach(_._2.close())
```
in the blocking block, as iterating over a TrieMap that is being modified has weak guarantees on what will actually happen.

Or have a look at [cleanupVoteGauges](https://github.com/canton-network/splice/blob/main/canton/community/synchronizer/src/main/scala/com/digitalasset/canton/synchronizer/metrics/BftOrderingMetrics.scala#L853).

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
```diff
@@ -0,0 +1,141 @@
...
+    private val hiddenCouponGauges: TrieMap[PartyId, Gauge[Long]] = TrieMap.empty
+
+    @SuppressWarnings(Array("com.digitalasset.canton.RequireBlocking"))
+    private def getHiddenCouponGauge(provider: PartyId): Gauge[Long] =
+      hiddenCouponGauges.getOrElse(
+        provider,
+        // We must synchronize here to avoid allocating the metric for the same party multiple
+        // times, which would lead to duplicate metric labels being reported by OpenTelemetry.
+        blocking {
+          synchronized {
```
rautenrieth-da: I think our current recommendation is to use a `Mutex` instead of `synchronized`, see https://github.com/DACH-NY/canton/pull/29939. See [here](https://github.com/DACH-NY/canton/blob/83132e48b142a285f27a55962ed77ae21f99fd8d/community/synchronizer/src/main/scala/com/digitalasset/canton/synchronizer/metrics/BftOrderingMetrics.scala#L856) for a mutex used in metrics.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
```diff
@@ -0,0 +1,141 @@
...
+        rewardMetrics.processRewardsActiveContractsMinting.updateValue(minting.size)
+      }
+      providers <- dsoStore.listNonObserverRewardCouponsV2ProvidersSample(
+        PageLimit.tryCreate(hiddenCouponProvidersSampleSize)
+      )
+      _ <- Future.traverse(providers) { provider =>
+        dsoStore
+          .listNonObserverRewardCouponsV2ForProvider(provider, PageLimit.Max)
+          .map(coupons => rewardMetrics.updateHiddenCoupons(provider, coupons.size.toLong))
+      }
```
meiersi-da, line 63: @dfordivam @rautenrieth-da : I'm a bit nervous about doing the counts and joins on the scala side. It's an anti-pattern from a scaling perspective and it does not seem to simplify the code.

Have you considered to add the two dedicated queries that this would require to the store directly and thus compute the full counts on the DB?

meiersi-da: As for the parties with the wrong vetting state I'd suggest to have the query only return the top 50 to conserve resources.

meiersi-da: structure query as:
* limit 10k on coupons
* group by provider order by count
* limit 50

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
dfordivam: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
dfordivam: I modified to use the lock + this pattern

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
dfordivam: Added a comment

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
dfordivam: Now the parties with max number of coupons is picked (after a 10k sample on coupons), so it should be relatively stable on 30s gapped retries.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
dfordivam: As discussed implemented the DB side query. Since we are only fetching the reward-party+count, it actually is an index-only query, and is extremely fast.

dfordivam:
@meiersi-da This is now ready for review

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/RewardMetricsTrigger.scala
```diff
@@ -0,0 +1,139 @@
...
+      // These listings are capped at the default page size (1000), so the dryRun/minting counts
+      // may underestimate if there are more than that. That's acceptable: we log a warning when
+      // the cap is hit, and an alert already fires when either count is above 1.
+      calculateRewards <- dsoStore.listCalculateRewardsV2()
+      _ = {
+        val (dryRun, minting) = calculateRewards.partition(_.payload.dryRun)
+        rewardMetrics.calculateRewardsActiveContractsDryRun.updateValue(dryRun.size)
+        rewardMetrics.calculateRewardsActiveContractsMinting.updateValue(minting.size)
+      }
+      processRewards <- dsoStore.listProcessRewardsV2()
```
meiersi-da, line 47: I originally expected that we would convert this query and the one above to a `count`-based one as well. However upon rereading the code it seems OK to not do so. The bit of extra work that we would save every 30s when the trigger runs is not worth the implementation effort.

meiersi-da (approved):
Thank you!

PR #6101, opened 2026-06-24
Helm Charts: service accounts

pasindutennage-da:
Fix https://github.com/DACH-NY/canton-network-internal/issues/5791

### DIFF
--- a/cluster/helm/cn-docs/values-template.yaml
+++ b/cluster/helm/cn-docs/values-template.yaml
@@ -23,3 +23,5 @@ enableReloader: true
 service:
   http:
     port: 80
+
+# serviceAccountName: ""
--- a/cluster/helm/cn-docs/values-template.yaml
+++ b/cluster/helm/cn-docs/values-template.yaml
@@ -24,4 +24,3 @@ service:
   http:
     port: 80
 
-# serviceAccountName: ""
--- a/cluster/helm/splice-global-domain/tests/mediator_test.yaml
+++ b/cluster/helm/splice-global-domain/tests/mediator_test.yaml
@@ -187,11 +187,11 @@ tests:
 
   - it: "injects custom serviceAccountName into Deployment when specified"
     set:
-      serviceAccountName: "custom-mediator-sa"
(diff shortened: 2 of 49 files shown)


cluster/helm/splice-global-domain/tests/mediator_test.yaml
```diff
@@ -185,3 +185,13 @@ tests:
       - notExists:
           path: spec.template.spec.containers[?(@.name=='mediator')].env[?(@.valueFrom.secretKeyRef)]
 
+  - it: "injects custom serviceAccountName into Deployment when specified"
+    set:
+      serviceAccountName: "custom-mediator-sa"
```
moritzkiefer-da: isn't this test somewhat misleading? This looks like you're setting a SA just for the mediator. But it seems like in reality because this is the same chart and a global option you actually set it for both mediator and sequencer to the same value? I don't mind that all that much for now I guess given that the request was for validators and those don't care about those charts anyway but it seems at least worth calling it custom-domain-sa or something like that then to make this explicit.

cluster/helm/cn-docs/values-template.yaml
```diff
@@ -23,3 +23,5 @@ enableReloader: true
 service:
   http:
     port: 80
+
+# serviceAccountName: ""
```
moritzkiefer-da: not sure we need this commented out here tbh

moritzkiefer-da: same for all the other charts

moritzkiefer-da (approved):
thx

cluster/helm/splice-global-domain/tests/mediator_test.yaml
pasindutennage-da: thanks. renamed to custom-domain-sa

cluster/helm/cn-docs/values-template.yaml
pasindutennage-da: fixed

PR #6108, opened 2026-06-25
add ACS snapshot size metric

### LINKED ISSUE #5398: Add a metric for the size of acs snapshots
The size is available at https://github.com/canton-network/splice/blob/bf5c94439a20e0c6dc711beda32531502f54c68e/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala#L639, it just needs to be exported in the result of `AcsSnapshotStore.saveIncrementalSnapshot()` and then the new metric updated in `AcsSnapshotTriggerBase.completeTask()`.

JYC11:
fixes: https://github.com/canton-network/splice/issues/5398

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
@@ -109,7 +109,10 @@ abstract class AcsSnapshotTriggerBase(
         )
         .map(size => {
           snapshotMetrics.latestRecordTimeSave.updateValue(snapshot.recordTime)
-          snapshotMetrics.snapshotSize.updateValue(size)
+          size match {
+            case Some(v) => snapshotMetrics.snapshotSize.updateValue(v)
+            case None => ()
+          }
           TaskSuccess(s"Saved incremental snapshot at ${snapshot.recordTime}")
         })
     case AcsSnapshotTriggerBase.DeleteIncrementalSnapshotTask(snapshot) =>
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
@@ -107,8 +107,9 @@ abstract class AcsSnapshotTriggerBase(
               nextSnapshotTargetRecordTime = nextAt,
             ),
         )
-        .map(_ => {
+        .map(size => {
           snapshotMetrics.latestRecordTimeSave.updateValue(snapshot.recordTime)
+          snapshotMetrics.snapshotSize.updateValue(size)
           TaskSuccess(s"Saved incremental snapshot at ${snapshot.recordTime}")
         })
     case AcsSnapshotTriggerBase.DeleteIncrementalSnapshotTask(snapshot) =>
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
@@ -111,7 +111,7 @@ abstract class AcsSnapshotTriggerBase(
           snapshotMetrics.latestRecordTimeSave.updateValue(snapshot.recordTime)
           size match {
             case Some(v) => snapshotMetrics.snapshotSize.updateValue(v)
-            case None => ()
+            case None => snapshotMetrics.snapshotSize.updateValue(-1)
           }
           TaskSuccess(s"Saved incremental snapshot at ${snapshot.recordTime}")
         })
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
@@ -467,20 +467,21 @@ class AcsSnapshotStore(
       table: IncrementalAcsSnapshotTable,
       action: DBIOAction[R, NoStream, E],
       expectedState: Option[IncrementalAcsSnapshot],
+      onSkip: R,
   )(implicit
       tc: TraceContext
-  ): DBIOAction[Unit, NoStream, Effect.Transactional & Effect.Read & E] = {
+  ): DBIOAction[R, NoStream, Effect.Transactional & Effect.Read & E] = {
     (for {
       actualState <- getIncrementalSnapshotAction(table)
       result <-
         if (actualState == expectedState) {
-          action.map(_ => ())
+          action
         } else {
           logger.info(
             s"Skipping action because actual state $actualState != expected state $expectedState. " +
               "In production, this is expected only during retries after transient network failures."
           )
-          DBIOAction.unit
+          DBIOAction.successful(onSkip)
         }
     } yield result).transactionally
   }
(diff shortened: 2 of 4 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -467,20 +467,21 @@ class AcsSnapshotStore(
       table: IncrementalAcsSnapshotTable,
       action: DBIOAction[R, NoStream, E],
       expectedState: Option[IncrementalAcsSnapshot],
+      onSkip: R,
   )(implicit
       tc: TraceContext
-  ): DBIOAction[Unit, NoStream, Effect.Transactional & Effect.Read & E] = {
+  ): DBIOAction[R, NoStream, Effect.Transactional & Effect.Read & E] = {
```
rautenrieth-da:
```suggestion
  ): DBIOAction[Option[R], NoStream, Effect.Transactional & Effect.Read & E] = {
```
then you can return `Some(_)` if the query is executed and `None` if not, without having to provide a `onSkip` value of type `R`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -693,14 +694,15 @@ class AcsSnapshotStore(
...
-      ()
+      copied_rows
     }
     storage.queryAndUpdate(
       withExclusiveSnapshotDataLock(
         withIncrementalSnapshotIdempotencyCheck(
           table,
           statement,
           Some(snapshot),
+          onSkip = 0,
```
rautenrieth-da: This means the this method will return 0 if query is retried (which should be very rare in practice), which in turn means the gauge will be updated with a value of 0. There's probably not much we can do - if the query is retried and skipped, then we simply won't have the count. I would however prefer if this method returned `None` if the count is unknown and if the trigger decided how to update the gauge in that case.

docs/src/release_notes_upcoming.rst
```diff
@@ -63,4 +63,7 @@
           ``CRARC_MiningRound_Archive``, ``CRARC_MiningRound_StartIssuing`` and
           ``CRARC_StartProcessingRewardsV2``
 
-        - The governance Vote History is now ordered by effective date rather than ingestion order
\ No newline at end of file
+        - The governance Vote History is now ordered by effective date rather than ingestion order
+
+      - Scan app
+        - add ACS snapshot size metric
```
rautenrieth-da:
```suggestion
        - Add a metric for the size of the most recent ACS snapshot
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
JYC11:
```
[error] /home/admin/Desktop/splice/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala:486:21: type mismatch;
[error]   slick.dbio.DBIOAction[
[error]     scala.Any|scala.Option[R]
[error]     ,
[error]     slick.dbio.NoStream
[error]     ,
[error]     E with slick.dbio.Effect.Read with slick.dbio.Effect.Transactional
[error]   ]
[error]     } yield result).transactionally
```
I initially went with the Option[R] approach but it caused this lint error which was why I didn't end up using it.

JYC11: I'll try to think of another way to do this

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -469,18 +473,18 @@ class AcsSnapshotStore(
...
   )(implicit
       tc: TraceContext
-  ): DBIOAction[Unit, NoStream, Effect.Transactional & Effect.Read & E] = {
+  ): DBIOAction[Option[R], NoStream, Effect.Transactional & Effect.Read & E] = {
     (for {
       actualState <- getIncrementalSnapshotAction(table)
       result <-
         if (actualState == expectedState) {
-          action.map(_ => ())
+          action.map(Some(_): Option[R])
```
rautenrieth-da:
```suggestion
          action.map(Option.apply)
```
`Option.apply` usually works better with type inference than `Some(_)`. Worst case you can use `Option.apply[R]` to specify the type.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -469,18 +473,18 @@ class AcsSnapshotStore(
...
         } else {
           logger.info(
             s"Skipping action because actual state $actualState != expected state $expectedState. " +
               "In production, this is expected only during retries after transient network failures."
           )
-          DBIOAction.unit
+          DBIOAction.successful(None: Option[R])
```
rautenrieth-da:
```suggestion
          DBIOAction.successful(Option.empty)
```

docs/src/release_notes_upcoming.rst
```diff
@@ -63,4 +63,7 @@
           ``CRARC_MiningRound_Archive``, ``CRARC_MiningRound_StartIssuing`` and
           ``CRARC_StartProcessingRewardsV2``
 
-        - The governance Vote History is now ordered by effective date rather than ingestion order
\ No newline at end of file
+        - The governance Vote History is now ordered by effective date rather than ingestion order
```
rautenrieth-da: You still have a merge conflict here - the `docs/src/release_notes_upcoming.rst` file is emptied on every release.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotTriggerBase.scala
```diff
@@ -107,8 +107,12 @@ abstract class AcsSnapshotTriggerBase(
               nextSnapshotTargetRecordTime = nextAt,
             ),
         )
-        .map(_ => {
+        .map(size => {
           snapshotMetrics.latestRecordTimeSave.updateValue(snapshot.recordTime)
+          size match {
+            case Some(v) => snapshotMetrics.snapshotSize.updateValue(v)
+            case None => ()
```
rautenrieth-da: This is slightly different behavior than before - now if we fail to determine the size of the snapshot, we keep the old value, which would be indistinguishable from a case where the new snapshot has exactly the same size as the previous one. I don't have a strong opinion, but consider writing the default ("unknown") value instead:
```suggestion
            case None => snapshotMetrics.snapshotSize.updateValue(-1)
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -53,6 +53,10 @@ class AcsSnapshotStore(
   override val profile: JdbcProfile = storage.profile.jdbc
   import profile.api.jdbcActionExtensionMethods
 
+  private implicit def rowsAlteredOfOption[A](implicit
```
rautenrieth-da:
```suggestion
  private implicit def rowsAlteredByIdempotencyCheck[A](implicit
```

to make it more explicit that it's for `withIncrementalSnapshotIdempotencyCheck` - it might not be correct for other actions that return an `Option`

rautenrieth-da (approved):
Thanks!

PR #6114, opened 2026-06-25
expose LSU information through Scan

### LINKED ISSUE #5855: expose lsu information through scan
should include freeze and upgrade time 

required by validators for automation

mblaze-da:
Fixes https://github.com/canton-network/splice/issues/5855

I was supposed to get an intro from @nicu-da for this issue but didn't manage to do so before his vacation. I inferred the current implementation from the existing code and I'm not very confident about that. I'm not sure if I should also add this endpoint to the Scan client. Should I add a test for it? I took inspiration from `getRollForwardLsu` and I couldn't find a test for it.

### DIFF
--- a/apps/scan/src/main/openapi/scan.yaml
+++ b/apps/scan/src/main/openapi/scan.yaml
@@ -172,6 +172,21 @@ paths:
               schema:
                 "$ref": "#/components/schemas/GetRollForwardLsuResponse"
 
+  /v0/lsu:
+    get:
+      tags: [internal, scan]
+      x-jvm-package: scan
+      operationId: "getLsu"
+      description: |
+        Retrieve information on the next LSU
+      responses:
+        "200":
+          description: ok
+          content:
+            application/json:
+              schema:
+                "$ref": "#/components/schemas/GetLsuResponse"
+
   /v0/active-synchronizer-serial:
     get:
       tags: [internal, scan]
@@ -2106,6 +2121,26 @@ components:
           type: string
         successorPhysicalSynchronizerId:
           type: string
+    GetLsuResponse:
+      type: object
+      properties:
+        lsu:
+          description: Info on the next LSU
+          "$ref": "#/components/schemas/Lsu"
+    Lsu:
+      type: object
+      required:
+        - freezeTime
+        - upgradeTime
+      properties:
+        freezeTime:
+          description: The time when topology freeze starts
+          type: string
+          format: date-time
+        upgradeTime:
+          description: The time at which to upgrade
+          type: string
+          format: date-time
     GetActivePhysicalSynchronizerSerialResponse:
       type: object
       required: ["serial"]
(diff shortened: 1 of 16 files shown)


apps/scan/src/main/openapi/scan.yaml
```diff
@@ -2106,6 +2121,26 @@ components:
...
+        lsu:
+          description: Info on the next LSU
+          "$ref": "#/components/schemas/Lsu"
+    Lsu:
+      type: object
+      required:
+        - freezeTime
+        - upgradeTime
+      properties:
+        freezeTime:
```
moritzkiefer-da:
```suggestion
        topologyFreezeTime:
```

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -2106,6 +2121,26 @@ components:
           type: string
         successorPhysicalSynchronizerId:
           type: string
+    GetLsuResponse:
+      type: object
+      properties:
+        lsu:
+          description: Info on the next LSU
+          "$ref": "#/components/schemas/Lsu"
+    Lsu:
```
moritzkiefer-da, line 2130: let's also include the successor serial in here.

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -172,6 +172,21 @@ paths:
               schema:
                 "$ref": "#/components/schemas/GetRollForwardLsuResponse"
 
+  /v0/lsu:
+    get:
+      tags: [internal, scan]
+      x-jvm-package: scan
+      operationId: "getLsu"
+      description: |
+        Retrieve information on the next LSU
```
moritzkiefer-da:
```suggestion
        Retrieve information on the next logical synchronizer upgrade (LSU)
```

moritzkiefer-da (commented):
looks reasonable but I think this deserves some tests, you can probably extend the existing lsu tests with some queries to this endpoint

cluster/configs/shared/rate-limits/unlimited.yaml
```diff
@@ -193,6 +193,10 @@ rateLimits:
     name: rollForwardLsu
     type: unlimited
```
pasindutennage-da, line 195: with unlimited, this will share the global bucket. Is that intended?

pasindutennage-da (commented):
Thank you

Looks good to me, but I will let @moritzkiefer-da approve this one because I don't have all the context about LSU.

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -2106,6 +2121,31 @@ components:
...
+      properties:
+        topologyFreezeTime:
+          description: The time when topology freeze starts
+          type: string
+          format: date-time
+        upgradeTime:
+          description: The time at which to upgrade
+          type: string
+          format: date-time
+        successorPhysicalSynchronizerSerial:
```
moritzkiefer-da: sorry I realize now I said serial but what I was trying to say is physical synchronizer id, so just include the full thing

moritzkiefer-da (approved):
thx

cluster/configs/shared/rate-limits/unlimited.yaml
mblaze-da: I used the same configuration as `rollForwardLsu`. Both are internal API so I thought it is fine. Am I missing something here @pasindutennage-da?

PR #6125, opened 2026-06-25
 Helm charts:: Security Contexts

pasindutennage-da:
Fix https://github.com/DACH-NY/canton-network-internal/issues/5825 

Cluster test https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/72827


moritzkiefer-da (commented):
The change itself looks fairly reasonable. But there is a very large number of configuration flags on security contexts and security contexts on various levels. They did mention this explicitly but their message sounds fairly ambigous and they may just have used this an example. Can we check with them what they expect?

martinflorian-da:
I think there are two parts here:

1. What are reasonable defaults that we would (eventually) like everyone to have.
2. What is a cheap way to allow arbitrary customizations.

For 1. we need some research; did we do that somewhere? 

For 2. I'm not sure if this PR goes in the right direction. Why not just allow overriding the `securityContext:` field completely? Then people can configure whatever they want, whenever they want.

cluster/helm/cn-docs/tests/docs_test.yaml
isegall-da, line 1: Can we use something other than docs? We know the docs are going away soon

pasindutennage-da:
@martinflorian-da 

> For 2. I'm not sure if this PR goes in the right direction. Why not just allow overriding the securityContext: field completely? Then people can configure whatever they want, whenever they want.

I agree with your point. We can let one override `securityContext`, however, we (I) should make sure that whatever they put inside that object is supported in our helm charts; else, it will break.

What I would do is; define `securityContext` in schema as an object with our selected subset (not just `allowPrivilegeEscalation`) of allowed options. Then in Helm helper, I will set only those supported options.

moritzkiefer-da:
> however, we (I) should make sure that whatever they put inside that object is supported in our helm charts; else, it will break.

I wouldn't try to enforce this in the schema. Trying to figure out all the various options that may be safe is a lost cause and some options may be safe in some usecase but not others. Just set the ones that are always valid by default and let people overwrite with whatever they want.

martinflorian-da:
Totally agree with @moritzkiefer-da ; for the defaults it's IMO fine enough if we ensure validity via our cluster testing. For everything else - up to the users to pick the right strings. If they want to override this it's reasonable to assume they have some idea what they are doing.

pasindutennage-da:
@martinflorian-da @moritzkiefer-da  

I will allow overriding the containerSecurityContext field, so users can put whatever they want. We will set a few defaults that won't break our deployments.

Full list of securitycontext: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.36/#securitycontext-v1-core

**Defaults**

`allowPrivilegeEscalation`: `false` (controls whether a process can gain more privileges than its parent

`privileged: false` (container does not run in privileged mode)

`runAsNonRoot: true` (ensures the container must run as a non-root user) -- cluster_test will say if this breaks some postgress pods

**Testing**

 I will update the tests to verify both the default state (no overrides) and the arbitrary overrides. Just extending my current 2 test per chart strcuture.

To define these default securitycontext, we can either put them in `values-template.yaml`, or directly in the Helm _helpers.tpl . Which one do you think is clearer?

moritzkiefer-da:
any reason not to set something like 
```
  seccompProfile:
    type: RuntimeDefault
```
and maybe switch to explicit capabilities whitelists?

it seems like `runAsUser`, `runAsGroup` and `fsGroup` are also set quite frequently from a quick search but I also can't come up with all that clear of a reason why it meaningfully improves things over `runAsNonRoot: false` so fine to skip for now.

> To define these default securitycontext, we can either put them in values-template.yaml, or directly in the Helm _helpers.tpl . Which one do you think is clearer?

I think I prefer values-templates. Then you also get automatic merging with user-supplied values.

Note that there is also both a pod and a container security context. I think we want to set both.

pasindutennage-da:
@moritzkiefer-da @martinflorian-da 

I made the changes for `scan helm chart`. To make sure that we are on the same page, can you please have a look and see if scan helm chart (only scan helm chart for now) supports all the security contexts?

The remaining work is to just duplicate this across all helm charts, and I will do it next.

cluster/helm/splice-scan/templates/scan.yaml
```diff
@@ -268,6 +276,10 @@ spec:
         {{- end }}
       initContainers:
         - name: {{ .Release.Name }}-init
+          {{- with .Values.containerSecurityContext }}
```
moritzkiefer-da: I wonder about the assumption that all containers have the same security context. seems like maybe you should be able to overwrite per container? or maybe it's not actually needed in practice. I don't really have any good intuition here. So maybe let's stick with what you have and we can add more configurability if we get requests.

cluster/helm/splice-scan/values-template.yaml
```diff
@@ -11,6 +11,17 @@ pod:
...
   labels: {}
 
+podSecurityContext:
+  runAsNonRoot: true
+  seccompProfile:
+    type: RuntimeDefault
+
+containerSecurityContext:
+  allowPrivilegeEscalation: false
+  capabilities:
```
moritzkiefer-da: have you tested this? I would expect that we start with drop: all but for things to work we actually need to add some back.

cluster/helm/splice-scan/values-template.yaml
```diff
@@ -11,6 +11,17 @@ pod:
   annotations: {}
   labels: {}
 
+podSecurityContext:
+  runAsNonRoot: true
+  seccompProfile:
+    type: RuntimeDefault
+
+containerSecurityContext:
```
moritzkiefer-da: maybe set privileged: false just for piece of mind. afaict it is actually the default but I had to convince myself that it really is and there seems to be little downside to setting it explicitly

moritzkiefer-da (commented):
thx, it looks good in configurability, the big question ofc is if it really works so definitely make sure to test this

cluster/helm/splice-scan/values-template.yaml
pasindutennage-da: @moritzkiefer-da @isegall-da 

What I have in mind for testing is:: in a scratchnet, deploy canton network, from the main branch (current working version). Then modify one chart at a time, and cncluster apply again, and see if it fails and fix. 

This is a very slow feedback testing, do you know any other way to test this? Alternatively, would local net help?

cluster/helm/splice-scan/values-template.yaml
pasindutennage-da: Fixed

cluster/helm/splice-scan/templates/scan.yaml
pasindutennage-da: Thanks. For this version, i will have a common config, and make sure it doesn't break. as new requests come in future i can make it more fine grained.

cluster/helm/splice-scan/values-template.yaml
isegall-da: Localnet is docker-compose based, not helm charts, so doesn't help you.

You can shortcut that feedback loop a bit by modifying helm values directly if that helps.

martinflorian-da:
> I think I prefer values-templates.

@moritzkiefer-da But do we have a way to avoid duplication there? This will be a lot of duplication that :100: we'll mess up at some point, and for a hard-to-miss security config that doesn't sound great.

> Then you also get automatic merging with user-supplied values.

If someone is competent enough to want to override our defaults surely they can manage the merging themself, i.e., copy the things from our defaults they actually like into whatever they want the final state to be. I can even see some people getting confused if their security context gets merged with something we set by default...

> Full list of securitycontext: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.36/#securitycontext-v1-core

@pasindutennage-da Did you just go through this list and decide individually which ones made sense for you? Is there some nice checklist or set of recommendations somewhere that we can just follow and link as a reference?

moritzkiefer-da:
> @moritzkiefer-da But do we have a way to avoid duplication there? This will be a lot of duplication that 💯 we'll mess up at some point, and for a hard-to-miss security config that doesn't sound great.

if we want to deduplicate, we alreday code generate the actual values files from values-template.yaml so factoring out some stuff seems pretty easy.

moritzkiefer-da:
> If someone is competent enough to want to override our defaults surely they can manage the merging themself, i.e., copy the things from our defaults they actually like into whatever they want the final state to be. I can even see some people getting confused if their security context gets merged with something we set by default...

that's a pretty poor argument imho, if you use helm you need to understand how default overwrites work and setting the default in values.yaml and then your overwrites get merged with them is standard practice.
imho we need a very strong argument to deviate from that and I don't see it.

cluster/helm/splice-domain/templates/domain.yaml
```diff
@@ -28,8 +28,20 @@ spec:
         {{- include "splice-util-lib.default-labels" (set . "app" .Release.Name) | nindent 8 }}
     spec:
       {{- include "splice-util-lib.service-account" .Values | nindent 6 }}
```
pasindutennage-da, line 30: I didn't check this chart in scratch ; i have not provided default values for this chart.

cluster/helm/splice-load-tester/templates/load-tester.yaml
```diff
@@ -26,8 +26,20 @@ spec:
         {{- include "splice-util-lib.default-labels" (set . "app" .Release.Name) | nindent 8 }}
     spec:
```
pasindutennage-da, line 27: I did not test this chart in scratch; and i did not set default context

cluster/helm/splice-party-allocator/templates/party-allocator.yaml
```diff
@@ -26,8 +26,20 @@ spec:
         {{- include "splice-util-lib.default-labels" (set . "app" .Release.Name) | nindent 8 }}
     spec:
```
pasindutennage-da, line 27: this chart is not tested in scratch; and i have not set default context

cluster/helm/splice-splitwell-app/templates/splitwell.yaml
```diff
@@ -26,8 +26,20 @@ spec:
         {{- include "splice-util-lib.default-labels" (set . "app" "splitwell-app") | nindent 8 }}
     spec:
       {{- include "splice-util-lib.service-account" .Values | nindent 6 }}
```
pasindutennage-da, line 28: i did not check this chart in scratch; i did not set default values

cluster/helm/splice-splitwell-web-ui/templates/splitwell-web-ui.yaml
```diff
@@ -23,8 +23,20 @@ spec:
         {{- include "splice-util-lib.default-labels" (set . "app" "splitwell-web-ui") | nindent 8 }}
     spec:
```
pasindutennage-da, line 24: i have not checked this chart in scratch; i have not set default values

pasindutennage-da:
@moritzkiefer-da

This PR is ready to be reviewed now. The approach i used

1. Started with the least privileges and then relaxed until  containers could run without error.
2. Tested  with `cncluster apply --sv1-only`
3. Avoided testing charts that we never use in production, but still included the contexts with empty defaults, just so that it is easy to reason about the charts at some later point in time.
4. Currently running cluster test and upgrade test. Links in the description.

cluster/helm/splice-party-allocator/values.schema.json
```diff
@@ -0,0 +1,17 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "securityContexts": {
+      "type": ["object", "null"],
+      "properties": {
+        "pod": {
+          "type": ["object", "null"]
+        },
+        "partyAllocator": {
```
moritzkiefer-da: should we put these under containers.containerName everywhere? that seems to make it a bit easier to understand what is going on

moritzkiefer-da (approved):
Thanks! The changes lgtm, I do think @martinflorian-da's concerns on duplication are valid so let's try to factor out the base profile that's restricted. It seems like putting a placeholder in there and resolving it in the translation from values-template.yaml -> values.yaml is probably the easiest option.

I would also suggest to not get this into this week's release but instead merge directly after the release so we get maximum testing time in case there is some fallout.

cluster/helm/splice-domain/templates/domain.yaml
martinflorian-da: I wonder if we should just remove that chart for now... (offtopic)

cluster/helm/splice-load-tester/templates/load-tester.yaml
martinflorian-da: Fair enough as we don't run this on MainNet, but at some point we should IMO harden that one too as it runs on DevNet. Let's add some TODO?

cluster/helm/splice-party-allocator/templates/party-allocator.yaml
martinflorian-da: Same as for load tester; let's maybe add a TODO to document that we do think it would be nice to harden these as well.

cluster/helm/splice-splitwell-app/templates/splitwell.yaml
martinflorian-da: That one I don't fully get though; deploying splitwell is not that big of a deal?

martinflorian-da: Happy to also add a TODO here, mainly curious why you decided to skip it.

cluster/helm/splice-splitwell-web-ui/templates/splitwell-web-ui.yaml
martinflorian-da: same comment as for the other splitwell component

docs/src/release_notes_upcoming.rst
```diff
@@ -113,3 +113,11 @@
         wallet             0.1.22
         walletPayments     0.1.21
         ================== =======
+
+  - Deployment
+
+    - Helm
+
+      - Added security contexts for Helm based deployments.
```
```suggestion
      - Added security contexts for all Helm-based deployments intended for production.
```

...or some better wording that avoids suggesting that we fixed all of them (we didn't). Also happy with a list (of charts we fixed or of the ones that remain unfixed).

cluster/helm/splice-participant/tests/participant_test.yaml
```diff
@@ -372,3 +372,78 @@ tests:
...
+      persistence:
+        enablePgInitContainer: true
+    documentSelector:
+      path: kind
+      value: Deployment
+    asserts:
+      - equal:
+          path: spec.template.spec.securityContext
+          value:
+            fsGroup: 1001
```
martinflorian-da, line 387: I think before you changes this included also

```
        runAsUser: 1001
        runAsGroup: 1001
```

So this is weaker now? I think that matters for the init container, that now doesn't inherit these anymore.

cluster/helm/splice-util-lib/security-profiles.yaml
```diff
@@ -0,0 +1,31 @@
...
+    runAsGroup: 1001
+    allowPrivilegeEscalation: false
+    privileged: false
+    capabilities:
+      drop:
+        - ALL
+
+  ui_security_context:
+    # runAsUser, runAsGroup, and runAsNonRoot are intentionally omitted
+    # to prevent entrypoint initialization script crashes (Exit Code 126)
```
martinflorian-da: That reads like "we got some error that we didn't understand and fixed it by reducing the security level"... can you give more context about the issue? Making it a TODO (pointing to a real issue on one of our security milestones) is also fine to me.

docs/src/release_notes_upcoming.rst
```diff
@@ -113,3 +113,11 @@
...
+        This improves the security of Kubernetes based deployments by enforcing the principle of least privilege across all pods and containers.
```
martinflorian-da: I'd also soften this to just "harden"... I don't think we can confidently claim "least possible privilege", and definitely not on "all pods and contains"

martinflorian-da (commented):
Thanks for pushing this forward guys!

I'm going to continue being annoying sadly: Even with deduplicating the security context boilerplate itself I'm still not super excited about the maintainability of this change...

- the tests are very duplicate and boilerplate-y
- we need to remember to add the right thing on new charts we add

I guess these concerns are largely consistent with the state of the rest of our Helm. But I think that at some point we need to look into more sophisticated scanning / static checks to better guard against regressions and "forgot to add context to new thing we added". 

@pasindutennage-da Did you happen to look into such tools already by any chance? So things like trivy, kyverno...

(TBC I'm thinking we should create a follow-up issue with the context we currently have; not that we should significantly change this PR.)

cluster/helm/splice-party-allocator/values.schema.json
pasindutennage-da: Fixed

cluster/helm/splice-participant/tests/participant_test.yaml
pasindutennage-da: I added them under container context, and checked that it deploys successfully.

For all charts, I checked the existing contexts, and made sure that my changes don't break any of them.

cluster/helm/splice-util-lib/security-profiles.yaml
pasindutennage-da: Thanks, made an issue, and added a todo.

cluster/helm/splice-participant/tests/participant_test.yaml
martinflorian-da: But I think what happens here is that the init container now runs as root? Which (if true; not 100% confident) not only sounds like worse security than we had before, but might actually break something for our partners (maybe @isegall-da remembers more details, but I think we had a nasty sequence of regression -fix - still doesn't work around the [PRs that added these](https://github.com/canton-network/splice/pull/2696/)).

martinflorian-da: Btw the same comment applies to sequencer and mediator, maybe also others (it's a big PR, didn't check all places 😇).

The tldr of my question is: Is there a good reason why init containers can't run as non-root (and overall be as hardened as the other containers)?

pasindutennage-da: @martinflorian-da You are correct. Init can `runAsNonRoot: true`

Sorry for missing this.

martinflorian-da: Do you mean the opposite of that you said or am I even more confused than I thought? :)

pasindutennage-da: Hi @martinflorian-da 

I was wrong, before, to assume that init containers **cannot** have `runAsNonRoot: true`. 

I realised that all init containers can indeed have `runAsNonRoot: true`, and i corrected (and verified it)

Sorry for der confusion.

cluster/helm/splice-participant/templates/participant.yaml
```diff
@@ -28,12 +28,16 @@ spec:
         {{- include "splice-util-lib.default-labels" (set . "app" .Release.Name) | nindent 8 }}
     spec:
       {{- include "splice-util-lib.service-account" .Values | nindent 6 }}
+      {{- with .Values.securityContexts.pod }}
```
martinflorian-da, line 32: Does this actually work if `.securityContexts` is `null`? Might be wroth testing... (also all the other place where you do similar chainings...)

cluster/helm/local.mk
```diff
@@ -56,9 +56,9 @@ cluster/helm/clean: $(foreach chart,$(app_charts),cluster/helm/$(chart)/helm-cle
 .PHONY: cluster/helm/test
 cluster/helm/test: cluster/helm/build $(foreach chart,$(app_charts),cluster/helm/$(chart)/helm-test)
 
-%/values.yaml: %/values-template.yaml
-  # We do not automatically run write-digests, as we do not want that for local dev, only for published artifacts
-	cp $< $@
+%/values.yaml: %/values-template.yaml cluster/helm/splice-util-lib/security-profiles.yaml
+    # We do not automatically run write-digests, as we do not want that for local dev, only for published artifacts
```
martinflorian-da: you might need to move that comment down now and perhaps add a new one describing you `yq` line

cluster/helm/splice-cometbft/values-template.yaml
```diff
@@ -111,3 +111,10 @@ mempool:
 # should stakater reloader annotation be added (note: reloader needs to be installed separately)
 enableReloader: true
 
+securityContexts:
+  pod:
+    security_context_profile: pod_security_context
+    fsGroup: 1001
```
martinflorian-da, line 117: TIL that `fsGroupChangePolicy: OnRootMismatch` and I think we might want to use it everywhere we set `fsGroup`. WDYT?

cluster/helm/splice-scan/tests/scan_test.yaml
```diff
@@ -637,3 +637,151 @@ tests:
...
+        equal:
+          path: spec.template.spec.securityContext
+          value:
+            seccompProfile:
+              type: RuntimeDefault
+      - documentIndex: 0
+        equal:
+          path: spec.template.spec.containers[0].securityContext
+          value:
+            runAsNonRoot: true
```
martinflorian-da, line 669: Should this also need fsGroup in case we want to write to `//persistent-data`, e.g., when taking a heap dump?

cluster/helm/splice-info/templates/deployment.yaml
martinflorian-da, line 1: Before we package this into a release, let's test that info still works after this PR (we don't have real tests for this and at this time of day I question my ability to reliably determine if this PR might be breaking that service).

martinflorian-da (approved):
Thanks a lot! I fully suspect that we'll find out about some things that will break after this, let's hope we find out quickly and before cutting the release. (I don't have a better strategy than "merge and find out" at this point.)

cluster/helm/splice-participant/templates/participant.yaml
pasindutennage-da: For each chart where I use `with .Values.securityContexts.`, it is mentioned in the respective `values-template.yaml`.

Since we ship values,  anyone using the chart will merge the newly introduced values automatically.

However, if someone only applies the template, without the defaults we ship, then it will fail. 

@martinflorian-da  @nicu-da  do you think this is a realistic assumption to make?

nicu-da: imo if we ship defaults in values-template.yaml it's normally fine. the question is if there's any reason for users to try to install while explicitly wanting to not set a security context and thus setting the value to null

cluster/helm/local.mk
pasindutennage-da: Fixed

cluster/helm/splice-cometbft/values-template.yaml
pasindutennage-da: Fixed

cluster/helm/splice-scan/tests/scan_test.yaml
pasindutennage-da: Fixed

cluster/helm/splice-participant/templates/participant.yaml
pasindutennage-da: Fixed

PR #6126, opened 2026-06-25
Fix informees in ExpiredAmuletAllocationTrigger

rautenrieth-da:
```
template AmuletAllocation
  with
    lockedAmulet : ContractId LockedAmulet -- ^ Locked amulet that holds the funds for the allocation
    allocation : AllocationSpecification
  where
    signatory allocationInstrumentAdmin allocation, allocation.transferLeg.sender
    observer allocation.settlement.executor
```

Receiver is not informee, and the content of that field is not enforced to contain a valid PartyId, so the trigger fails hard if someone enters nonsense in that field.


moritzkiefer-da (approved):
thx!

PR #6131, opened 2026-06-26
returns 400 on validator-faucets api when no validator_ids are sent

JYC11:
fixes: https://github.com/canton-network/splice/issues/1964

The SQL syntax crash does not seem to occur as an empty array produces `= ANY ('{}')` which is valid SQL. A test asserting store returns `Seq.empty` for an empty input was added as a regression test in case `inClause` semantics change.

This PR instead addresses the user-facing concern: calling `/v0/validators/validator-faucets` with no `validator_ids` (or a misspelled key like `validator=`) used to return 200 + empty list, masking the typo. It now returns HTTP 400 with a clear error.

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/http/HttpScanHandler.scala
@@ -2428,16 +2428,23 @@ class HttpScanHandler(
   ): Future[ScanResource.GetValidatorFaucetsByValidatorResponse] = {
     implicit val tc = extracted
     withSpan(s"$workflowId.getValidatorFaucetsByValidator") { _ => _ =>
-      store
-        .getValidatorLicenseByValidator(validators.map(v => PartyId.tryFromProtoPrimitive(v)))
-        .map(licenses =>
-          ScanResource.GetValidatorFaucetsByValidatorResponse.OK(
-            definitions
-              .GetValidatorFaucetsByValidatorResponse(
-                FaucetProcessor.process(licenses)
-              )
-          )
+      if (validators.isEmpty) {
+        Future.successful(
+          ScanResource.GetValidatorFaucetsByValidatorResponse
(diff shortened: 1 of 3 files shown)


isegall-da:
> fixes: #1964
> 
> The SQL syntax crash does not seem to occur as an empty array produces `= ANY ('{}')` which is valid SQL. A test asserting store returns `Seq.empty` for an empty input was added as a regression test in case `inClause` semantics change.
> 
> This PR instead addresses the user-facing concern: calling `/v0/validators/validator-faucets` with no `validator_ids` (or a misspelled key like `validator=`) used to return 200 + empty list, masking the typo. It now returns HTTP 400 with a clear error.

@JYC11 I'm confused by this statement:

> The SQL syntax crash does not seem to occur

The issue says that it does occur. Did you try reproducing the issue?

isegall-da:
Seems like you're right, and the sql error has already been fixed.
```
$ curl https://scan.sv-1.dev.global.canton.network.sync.global/api/scan/v0/validators/validator-faucets?validator_ids=foo::bar
{"validatorsReceivedFaucets":[]}
```

Given that the issue is almost a year old, that's not too surprising.
Your PR covers only a very specific case of the validator IDs list truly being empty, not the case where they're all invalid (as above). 
Having said that, given that liveness rewards have been removed from the network months ago, I don't think we want to invest in the UX of this endpoint TBH. I'd suggest we just close the issue and this PR.

JYC11:
> Seems like you're right, and the sql error has already been fixed.
> 
> ```
> $ curl https://scan.sv-1.dev.global.canton.network.sync.global/api/scan/v0/validators/validator-faucets?validator_ids=foo::bar
> {"validatorsReceivedFaucets":[]}
> ```
> 
> Given that the issue is almost a year old, that's not too surprising. Your PR covers only a very specific case of the validator IDs list truly being empty, not the case where they're all invalid (as above). Having said that, given that liveness rewards have been removed from the network months ago, I don't think we want to invest in the UX of this endpoint TBH. I'd suggest we just close the issue and this PR.

That's fine by me, I'll close the PR. Can you close the issue?

PR #6139, opened 2026-06-26
Always log how the app activity record store was initialized

rautenrieth-da:
Just one more log line during app startup, and can be helpful if you ever need to investigate the app activity store.

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
@@ -70,11 +70,7 @@ object DbAppActivityRecordStore {
       runningUser: Int,
       storedCode: Int,
       storedUser: Int,
-  ) extends MetaCheckResult {
-    def message: String =
-      s"Activity ingestion version downgrade detected: " +
-        s"running=($runningCode,$runningUser), stored=($storedCode,$storedUser)."
-  }
+  ) extends MetaCheckResult
 
   def checkMetaVersions(
       existing: Option[(Int, Int)],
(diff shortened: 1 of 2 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
```diff
@@ -355,9 +351,23 @@ class DbAppActivityRecordStore(
...
+        logger.error(
+          s"App activity ingestion version downgrade detected: " +
+            s"running=(${d.runningCode},${d.runningUser}), stored=(${d.storedCode},${d.storedUser}). " +
+            s"Make sure you did not accidentally remove or downgrade the 'activity-ingestion-user-version' field" +
+            s"in the scan app config. Shutting down to prevent data corruption."
+        )
         sys.exit(1)
-      case _ => ()
+      case Resume =>
+        logger.info(
```
dfordivam: Is it ok to do log at info level? note that this will be logged on every verdict batch ingestion

dfordivam (approved):
Thanks, Looks ok, the `Resume` branch logging may need adjustment

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
adetokunbo: Perhaps the logging could be moved inside `ensureMetaDBIO`; that way it could be called once at info level for any value?

adetokunbo (approved):

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
rautenrieth-da: Moved to `ensureMetaDBIO`.

PR #6164, opened 2026-06-30
helm, info: Make updater.sh more robust

giner:
- Clamp sleep time to 0 if updating takes longer than 60 seconds
- Handle TERM signal for near-instant pod deletion instead of waiting for the 30s grace period


isegall-da (approved):
tx

PR #6167, opened 2026-06-30
move SV specific code from canton-network to common-sv in preparation for migrating resources to the sv stack

mblaze-da:
I want to get this change in sperately and quickly as it will likely cause a lot of conflicts. In the end, after all the migrations are done I'll move it to the `sv` package. For now, this will help me avoid conflicts while I work on replicating those SV resources definitions in the `sv` stack.


cluster/pulumi/common-sv/src/bigQuery.ts
nicu-da, line 1: didn't look that deep into it but just wanted to make sure we don't also have things that get installed a single time here, for bigquery I would expect we have such things

nicu-da (approved):

cluster/pulumi/common-sv/src/bigQuery.ts
mblaze-da: Possible. Didn't look into bigquery stuff yet. Perhaps some of it will need migrating.

PR #6188, opened 2026-07-01
Drop receiver from ExpiredAmuletAllocation expiry informees

martinflorian-da:
Follow up to https://github.com/canton-network/splice/pull/6126


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AmuletAllocationExpiryTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,86 @@
...
+    with WalletTestUtil
+    with TimeTestUtil
+    with TriggerTestUtil {
+
+  override protected def runTokenStandardCliSanityCheck: Boolean = false
+
+  override def environmentDefinition: EnvironmentDefinition =
+    EnvironmentDefinition.simpleTopology1SvWithSimTime(this.getClass.getSimpleName)
+
+  "ExpiredAmuletAllocationTrigger" should {
```
moritzkiefer-da: A full test suite is quite expensive, time based ones even more so. why can we not extend the test we already have in apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TokenStandardCliTestDataTimeBasedIntegrationTest.scala to just set a borked receiver?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AmuletAllocationExpiryTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,86 @@
...
+
+    "expire a V1 allocation whose transfer-leg receiver is not a hosted party" in { implicit env =>
+      val aliceParty = onboardWalletUser(aliceWalletClient, aliceValidatorBackend)
+      aliceWalletClient.tap(1000)
+
+      // Well-formed but unhosted; the receiver is free-form data, not a stakeholder.
+      val unhostedReceiver = PartyId.tryFromProtoPrimitive(
```
moritzkiefer-da: that catches trying to use it as an observer but not a failing PartyId.tryFromProtoPrimitive, don't we want to test that as well? It seems like that is even a stronger test for the most part so we probably don't need both.

moritzkiefer-da (approved):
thx

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AmuletAllocationExpiryTimeBasedIntegrationTest.scala
martinflorian-da: The test just resembles the actual issue we had. Are you suggesting we try setting a party string that is not even well-formed?

Sounds reasonable but I'm not convinced I can remove the "well-formed but nonexistent party" check here; the failures could be on different code paths so why not keep both checks.

moritzkiefer-da: > The test just resembles the actual issue we had. Are you suggesting we try setting a party string that is not even well-formed?

yes that was the first issue we hit around this a week ago resulting in Robert's partial fix.

> Sounds reasonable but I'm not convinced I can remove the "well-formed but nonexistent party" check here; the failures could be on different code paths so why not keep both checks.

fair, ok with keeping both.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/AmuletAllocationExpiryTimeBasedIntegrationTest.scala
martinflorian-da: Well I expected that we'll need to set `runTokenStandardCliSanityCheck = false` for this test but I actually tested it now and it looks like the sanity checks pass just fine :angry: thanks for pushing back.

PR #6195, opened 2026-07-01
Fix party auto-ignore for ExpiredAmuletTrigger and ExpiredLockedAmuletTrigger

julientinguely-da:
see https://daholdings.slack.com/archives/C03J6NCPSE4/p1782892926662029

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


moritzkiefer-da (approved):
Thanks, fix looks reasonable to me.

Wondering if we can also extend the integration tests to cover this to make sure everything is wired up correctly but I'm struggling to come up with a great idea. I thought we could check that the ignored parties store gets updated but that's not the issue, the issue is that we read it wrong. So we would need to essentially need to test that we don't read a contract which is always a mess. I guess we could add a trace log for the parties we ignore and log assert on that but that also seems shitty.

So maybe let's just go with this and revisit if we break it again, surely it will just work now™.

julientinguely-da:
> So maybe let's just go with this and revisit if we break it again, surely it will just work now™.

That would make me really sad if it doesn't

julientinguely-da:
What if we enable the amulet expiry triggers on devnet on Friday to be sure that it'll work?

moritzkiefer-da:
sgtm

PR #6214, opened 2026-07-02
Add `helmForceUpdate` config option

krzysztofczyz-da:
Fixes #4940 

Added a `helmForceUpdate` config option (`true` by default).
Tested on scratchneta.

#### Test
1. Manually add an annotation to a deamonset (fluentbit)
2. run `cncluster pulumi cluster up`
3. The annotation disappears
4. readd the annotation
5. change the `helmForceUpdate` setting to `false`
6. run `cncluster pulumi cluster up`
7. the annotation stays on the resource

### DIFF
--- a/build-tools/cncluster
+++ b/build-tools/cncluster
@@ -2486,16 +2486,4 @@ if [ ! ${subcommand_whitelist[${SUBCOMMAND_NAME}]+_} ]; then
     _error  "Unknown subcommand: ${SUBCOMMAND_NAME}"
 fi
 
-declare -a NEW_ARGS=()
-for arg in "$@"; do
-    if [[ "$arg" == "--helm-force" ]]; then
-        _warning "You have passed --helm-force. This will force an update for all Pulumi Helm Releases, which is potentially destructive."
-        _prompt_to_confirm "Are you sure you want to proceed?"
-        export HELM_FORCE_UPDATE="true"
-    else
-        NEW_ARGS+=("$arg")
-    fi
-done
-set -- "${NEW_ARGS[@]}"
-
 "subcmd_${SUBCOMMAND_NAME}" "$@"
--- a/build-tools/cncluster
+++ b/build-tools/cncluster
@@ -2486,4 +2486,16 @@ if [ ! ${subcommand_whitelist[${SUBCOMMAND_NAME}]+_} ]; then
     _error  "Unknown subcommand: ${SUBCOMMAND_NAME}"
 fi
 
+declare -a NEW_ARGS=()
+for arg in "$@"; do
+    if [[ "$arg" == "--helm-force" ]]; then
+        _warning "You have passed --helm-force. This will force an update for all Pulumi Helm Releases, which is potentially destructive."
+        _prompt_to_confirm "Are you sure you want to proceed?"
+        export HELM_FORCE_UPDATE="true"
+    else
+        NEW_ARGS+=("$arg")
+    fi
+done
+set -- "${NEW_ARGS[@]}"
+
 "subcmd_${SUBCOMMAND_NAME}" "$@"
--- a/.envrc.vars
+++ b/.envrc.vars
@@ -70,6 +70,7 @@ export MULTIVALIDATOR_SIZE=0
 # This can over time lead to a very large number of secrets (which is how helm stores the release history) resulting in
 # unnecessary slowness when running helm commands.
 export HELM_MAX_HISTORY_SIZE=10
+export HELM_FORCE_UPDATE="false"
 export OIDC_AUTHORITY_LEDGER_API_AUDIENCE=https://canton.network.global
 export OIDC_AUTHORITY_VALIDATOR_AUDIENCE=https://canton.network.global
 export OIDC_AUTHORITY_SV_AUDIENCE=https://canton.network.global
--- a/cluster/expected/canton-network/expected.json
+++ b/cluster/expected/canton-network/expected.json
@@ -214,7 +214,7 @@
     "inputs": {
       "chart": "oci://ghcr.io/digital-asset/decentralized-canton-sync-dev/helm/cn-docs",
       "compat": "true",
-      "forceUpdate": false,
+      "forceUpdate": true,
       "maxHistory": 10,
       "name": "docs",
       "namespace": "docs",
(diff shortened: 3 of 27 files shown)


martinflorian-da:
Thanks!

> --force should not be used normally (as replace can nuke your apps' availablity, particulary damaging if done for a slow starting app)

Hmm so my understanding is that since we chain this after Pulumi, a Helm upgrade is (still) only triggered is something actually changed in the Pulumi state. And then you'd get a restart, true, but AFAICT we almost always get a restart when we change things even without force... I'd say let's just test it by merging and if it causes problems we can revert before it reaches MainNet. I.e., default the config thing to true.

> I assume most use cases will be applied from localhost.

So the main need is actually the long-running clusters (CILR, DevNet, TestNet, MainNet). These are managed by the operator, so AFAIU we'd need to deploy the operator once to enable this "feature"... and then it's enabled for all help operations.

> env var

We're trying to reduce the number of weird env vars; can we add this to the `config.yaml` schema instead?

And let's do a CI scratchnet test before merging, maybe an upgrade test as this actually changes many things that are already deployed.

build-tools/cncluster
martinflorian-da, line 1: I actually don't think we need that extra cncluster feature tbh; assuming you add this to `config.yaml`, the "edit `config.yaml` and rerun" is a quite established flow by now.

martinflorian-da:
I also suggest doing a basic "does it do what it should" test if you haven't already. If you're quick, you could just kubectl edit some deployment while the CI upgrade test is running (but before the upgrade step) to add some extra env var, and then confirm that the operator removed your manual changes again on the upgrade step.

krzysztofczyz-da:
> Hmm so my understanding is that since we chain this after Pulumi, a Helm upgrade is (still) only triggered is something actually changed in the Pulumi state. And then you'd get a restart, true, but AFAICT we almost always get a restart when we change things even without force...

Restart != replace - i.e. imagine you have a change which cannot get applied to a statefulset. a replace would kill it and create a new one (which may or may not work). A restart would roll out the change pod-by-pod.
Note that a replace would only happen if necessary (kinda rare tbh.).

Or to quote `helm upgrade --help`:

```
      --force                                      force resource updates through a replacement strategy
```

> So the main need is actually the long-running clusters (CILR, DevNet, TestNet, MainNet). These are managed by the operator, so AFAIU we'd need to deploy the operator once to enable this "feature"... and then it's enabled for all help operations.

This changes things - we won't need the cncluster thing then - removed 🫡 

> We're trying to reduce the number of weird env vars; can we add this to the config.yaml schema instead?

For sure, happily changed 😄 

> And let's do a CI scratchnet test before merging, maybe an upgrade test as this actually changes many things that are already deployed.

Lemme trigger that

> I also suggest doing a basic "does it do what it should" test if you haven't already. If you're quick, you could just kubectl edit some deployment while the CI upgrade test is running (but before the upgrade step) to add some extra env var, and then confirm that the operator removed your manual changes again on the upgrade step.

Done - added to description 👌

krzysztofczyz-da:
Upgrade unfortunately failed due to GCP quota 😿

martinflorian-da:
@krzysztofczyz-da AFAIU we should be in an OK place again w.r.t. scaling up scratchnet clusters... want to give this another try?

krzysztofczyz-da:
[Upgrade test](https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/75080/details?workflowId=81f89fc5-c492-4b6f-a421-0190d4a7bf10&job=2679e1e4-3188-40cd-9d80-b4c3b0507128&buildNumber=448165&jobType=build#step-114-) failed on:
```
[organization/canton-network/canton-network.scratcha] - error - error: 1 error occurred:
        * Helm release "sv-2/sv-app" failed to initialize completely. Use Helm CLI to investigate: failed to become available within allocated timeout. Error: Helm Release sv-2/sv-app: failed to replace object: PersistentVolumeClaim "sv-app-global-domain-migration-hd-pvc" is invalid: spec: Forbidden: spec is immutable after creation except resources.requests and volumeAttributesClassName for bound claims
    "storage": "20G"
   }
  },
- "VolumeName": "pvc-7a004e84-1788-464f-849b-aafc7ba8b431",
+ "VolumeName": "",
  "StorageClassName": "hyperdisk-standard-rwo",
  "VolumeMode": "Filesystem",
  "DataSource": null,
```
Which kinda makes sense I guess, Helm cannot re-create the PVC w/o dropping the PV (in our case **with data**) - which is probably not what we would want - force does not force the replace because the PVC is bound to a pod. This would _probably_ pass if we would be using PVC templates in STS instead of bare PVCs. 

So now the options seem to be @martinflorian-da :
- we can scrap the project (which I'd recommend as running kubectl replace/recreate on all resources is scary)
- we can adapt our resources to be compatible with force (move PVCs to PVC templates in STSes, etc.) - not fun and kinda scary as well
- add this option as as default=false and enable it on local if needed ( also okay in my book, but still needs to be handled with care)
- ???

martinflorian-da:
Thanks for all the digging here @krzysztofczyz-da !

> we can adapt our resources to be compatible with force (move PVCs to PVC templates in STSes, etc.) - not fun and kinda scary as well

Not worth the effort and risk short term, IMHO.

> add this option as as default=false and enable it on local if needed ( also okay in my book, but still needs to be handled with care)

Not really solving any problem? If I remember to manually flip this when needed I can also just manually fix my deployment, `helm uninstall` etc., no?

> ???

So actually the specific thing you hit here is a very fun one... this PVC is not really needed anymore and will be removed as part of https://github.com/canton-network/splice/issues/5641 .

The other scary PVC I can think of is CometBFT; that doesn't exist on CantonBFT deployments (which are becoming standard just these days).

However, there are sometimes also PVCs we use for persisting heap dumps and the like. Indeed we'd prefer these to not be flushed on random changes to values, and obviously we also don't want deployments to fail...

... and the alternative of defaulting to `force` only on the helm charts where it always works seems to add the complexity of not knowing where it is used and where not. It's all very sad, I'm out of reasonable ideas.

> we can scrap the project (which I'd recommend as running kubectl replace/recreate on all resources is scary)

Sounds like the most reasonable thing to do. cc @rautenrieth-da @moritzkiefer-da in case I'm missing something and there is still hope.

PR #6219, opened 2026-07-02
add more log ignores to address recent issues in resource-intensive tests

mblaze-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/9023
Fixes https://github.com/DACH-NY/cn-test-failures/issues/9029

I saw that we already ignore these logs in many other files so I thought that this might be the right thing to do here as well. Please LMK if I'm wrong.


martinflorian-da (approved):
Thanks! I think my OCD would have compelled me to restrict these more (to `mediator=sv4StandaloneMediator` and `UNAVAILABLE`) but this is definitely good enough.

PR #6238, opened 2026-07-06
helm, info, status: Increase threshold for sequencer acknowledgments

giner:
Increase the default threshold from 30 to 42 minutes to reduce false negatives (primarily on TestNet).

Here is a relevant graph for TestNet:
Screenshot from 2026-07-06 14-41-41


martinflorian-da (approved):
Thanks!

PR #6242, opened 2026-07-06
feat: bump checkout to v7 w/ allow-unsafe-pr-checkout arg

krzysztofczyz-da:


.github/workflows/build.daml_test.yml
```diff
@@ -15,11 +15,12 @@ jobs:
     steps:
 
       - name: Check out repository code
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```
martinflorian-da, line 18: Any reason not to bump **all** our current `uses: actions/checkout`?

martinflorian-da: offtopic: on internal we're currently on v4.2.2 🤡 

(keeping it consistent within the same repo still sounds better to me)

isegall-da: Agreed, let's move everything to v7

krzysztofczyz-da: Done

PR #6246, opened 2026-07-06
make apps node pools name unique in case multiple pools are configured

mblaze-da:


martinflorian-da (approved):
thx

cluster/pulumi/cluster/src/nodePools.ts
```diff
@@ -72,21 +72,26 @@ function installAppsNodePools(
...
 function hyperdiskNodePool(
+  index: number,
   cluster: string,
   config: GkeNodePoolConfig,
   location?: string
 ): gcp.container.NodePool {
-  return new gcp.container.NodePool('cn-apps-node-pool-hd', {
+  const name =
+    index === 0
+      ? 'cn-apps-node-pool-hd' // for backwards compat
```
martinflorian-da, line 92: I do wonder whether simply `pool`, `pool-1`, `pool-2` etc. might not be less surprising/confusing, but this is nit-level

PR #6247, opened 2026-07-06
Dso Governance POC

moritzkiefer-da:
Current limitations:

Current limitations:

1. The split between governance and operator votes is a placeholder.
2. The tests are a joke. Given that I think at this stage this is more
   about how we arrange stuff this hopefully doesn't detract from
   review too much.
3. There are some open questions on the migration.
4. We may want to move beneficiaries to a different place.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
+++ b/daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
@@ -376,7 +376,7 @@ executeAllDefinitiveVotes app = do
   now <- getTime
   requests <- query @VoteRequest app.dso
   forA_ requests $ \(requestCid, request) -> do
-    let activeSvs = Set.fromList (getSvRightOwners rules)
+    let activeSvs = if onLedgerSvRightOwners rules then Set.fromList (getSvRightOwners rules) else Set.fromList (map (.name) (Map.values rules.svs))
     let execute = request.voteBefore <= now ||
           (activeSvs == Set.fromList (Map.keys request.votes))
     rightOwnerCids <- querySvRightOwners app
--- a/daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
+++ b/daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
@@ -376,7 +376,7 @@ executeAllDefinitiveVotes app = do
   now <- getTime
   requests <- query @VoteRequest app.dso
   forA_ requests $ \(requestCid, request) -> do
-    let activeSvs = Set.fromList (map fst $ TextMap.toList (getSvRightOwners rules))
+    let activeSvs = Set.fromList (getSvRightOwners rules)
     let execute = request.voteBefore <= now ||
           (activeSvs == Set.fromList (Map.keys request.votes))
     rightOwnerCids <- querySvRightOwners app
@@ -495,7 +495,7 @@ checkSvRewardStates app = do
   rewardStates <- query @SvRewardState app.dso
   [(_, dsoRules)] <- query @DsoRules app.dso
   if onLedgerSvRightOwners dsoRules then
-    Set.fromList [state.svName | (_, state) <- rewardStates] === Set.fromList (map fst (TextMap.toList (getSvRightOwners dsoRules)))
+    Set.fromList [state.svName | (_, state) <- rewardStates] === Set.fromList (getSvRightOwners dsoRules)
   else
     Set.fromList [state.svName | (_, state) <- rewardStates] === Set.fromList (map (.name) (Map.values dsoRules.svs))
   
--- a/daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
+++ b/daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
@@ -34,7 +34,7 @@ import Splice.DsoBootstrap
 import Splice.DSO.AmuletPrice
 import Splice.DSO.DecentralizedSynchronizer
 import Splice.DSO.SvState
-import Splice.DSO.SvRightOwner
+import Splice.Amulet.SvRightOwner
 import Splice.Scripts.AnsRulesParameters
 
 -- | Multiplier for the SV weights, which are recorded in basis-points, i.e., 1/10000ths
@@ -399,7 +399,7 @@ lookupRightOwner app rightOwnerParty = do
 querySvRightOwners : AmuletApp -> Script (Optional (TextMap.TextMap (ContractId SvRightOwner)))
 querySvRightOwners app = do
   rightOwners <- query @SvRightOwner app.dso
-  pure $ if null rightOwners 
+  pure $ if null rightOwners
     then None
     else Some (TextMap.fromList [(ro.rightOwnerName, roCid) | (roCid, ro) <- rightOwners])
 
@@ -498,7 +498,7 @@ checkSvRewardStates app = do
     Set.fromList [state.svName | (_, state) <- rewardStates] === Set.fromList (getSvRightOwners dsoRules)
   else
     Set.fromList [state.svName | (_, state) <- rewardStates] === Set.fromList (map (.name) (Map.values dsoRules.svs))
-  
+
 
 -- | Check that there's exactly one status report per SV both off-boarded and onboarded.
 checkSvStatusReports : AmuletApp -> Script ()
--- a/daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
+++ b/daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
@@ -12,6 +12,7 @@ import qualified DA.Map as Map
 import DA.Optional (fromOptional)
 import qualified DA.Set as Set
 import qualified DA.Text as T
+import qualified DA.TextMap as TextMap
 import Daml.Script
 import DA.Time
 
@@ -33,34 +34,38 @@ import Splice.DsoBootstrap
 import Splice.DSO.AmuletPrice
 import Splice.DSO.DecentralizedSynchronizer
 import Splice.DSO.SvState
+import Splice.DSO.SvRightOwner
 import Splice.Scripts.AnsRulesParameters
-import Splice.Util
 
 -- | Multiplier for the SV weights, which are recorded in basis-points, i.e., 1/10000ths
 bpsMultiplier : Int
 bpsMultiplier = 10000
 
 initMainNet : Script (AmuletApp, Party, (Party, Party, Party, Party))
-initMainNet = initDecentralizedSynchronizer False None
+initMainNet = initDecentralizedSynchronizer False None True
+
+
+initMainNetNoOnLedgerSvRightOwners : Script (AmuletApp, Party, (Party, Party, Party, Party))
+initMainNetNoOnLedgerSvRightOwners = initDecentralizedSynchronizer False None False
 
 
 initMainNetWithAmuletPrice : Decimal -> Script (AmuletApp, Party, (Party, Party, Party, Party))
-initMainNetWithAmuletPrice amuletPrice = initDecentralizedSynchronizerWithAmuletPrice False 0 amuletPrice None
+initMainNetWithAmuletPrice amuletPrice = initDecentralizedSynchronizerWithAmuletPrice False 0 amuletPrice None True
 
 initDevNet : Script (AmuletApp, Party, (Party, Party, Party, Party))
-initDevNet = initDecentralizedSynchronizer True None
+initDevNet = initDecentralizedSynchronizer True None True
 
 initDevNetWithAmuletConfig : AmuletConfig USD -> Script (AmuletApp, Party, (Party, Party, Party, Party))
-initDevNetWithAmuletConfig amuletConfig = initDecentralizedSynchronizer True (Some amuletConfig)
+initDevNetWithAmuletConfig amuletConfig = initDecentralizedSynchronizer True (Some amuletConfig) True
 
-initDecentralizedSynchronizer : Bool -> Optional (AmuletConfig USD) -> Script (AmuletApp, Party, (Party, Party, Party, Party))
-initDecentralizedSynchronizer isDevNet optAmuletConfig = initDecentralizedSynchronizerWithAmuletPrice isDevNet 0 1.0 optAmuletConfig
+initDecentralizedSynchronizer : Bool -> Optional (AmuletConfig USD) -> Bool -> Script (AmuletApp, Party, (Party, Party, Party, Party))
+initDecentralizedSynchronizer isDevNet optAmuletConfig onLedgerSvRightOwners = initDecentralizedSynchronizerWithAmuletPrice isDevNet 0 1.0 optAmuletConfig onLedgerSvRightOwners
 
-initDecentralizedSynchronizerWithNonZeroRound : Bool -> Int -> Script (AmuletApp, Party, (Party, Party, Party, Party))
-initDecentralizedSynchronizerWithNonZeroRound isDevNet initialRound = initDecentralizedSynchronizerWithAmuletPrice isDevNet initialRound 1.0 None
+initDecentralizedSynchronizerWithNonZeroRound : Bool -> Int ->  Bool -> Script (AmuletApp, Party, (Party, Party, Party, Party))
+initDecentralizedSynchronizerWithNonZeroRound isDevNet initialRound onLedgerSvRightOwners = initDecentralizedSynchronizerWithAmuletPrice isDevNet initialRound 1.0 None onLedgerSvRightOwners
 
-initDecentralizedSynchronizerWithAmuletPrice : Bool -> Int -> Decimal -> Optional (AmuletConfig USD) -> Script (AmuletApp, Party, (Party, Party, Party, Party))
-initDecentralizedSynchronizerWithAmuletPrice isDevNet initialRound amuletPrice optAmuletConfig = do
+initDecentralizedSynchronizerWithAmuletPrice : Bool -> Int -> Decimal -> Optional (AmuletConfig USD) -> Bool -> Script (AmuletApp, Party, (Party, Party, Party, Party))
+initDecentralizedSynchronizerWithAmuletPrice isDevNet initialRound amuletPrice optAmuletConfig onLedgerSvRightOwners = do
   [sv1, sv2, sv3, sv4] <- forA ["sv1", "sv2", "sv3", "sv4"] allocateParty
 
   dso <- allocateParty "dso-party"
@@ -129,6 +134,8 @@ initDecentralizedSynchronizerWithAmuletPrice isDevNet initialRound amuletPrice o
       initialTrafficState = Map.empty
       isDevNet
       initialRound = Some initialRound
+      sv1RightOwnerName = if onLedgerSvRightOwners then Some "sv1" else None
+      sv1VoteWeight = if onLedgerSvRightOwners then Some 1 else None
     DsoBootstrap_Bootstrap
 
   dsoUserId <- validateUserId "dso-user"
@@ -136,32 +143,43 @@ initDecentralizedSynchronizerWithAmuletPrice isDevNet initialRound amuletPrice o
 
   -- add more sv nodes
   forA_ (zip [sv2, sv3, sv4] ["sv2", "sv3", "sv4"]) $ \(svParty, svName) -> do
+    let rewardWeight = 3 * bpsMultiplier -- Simulate the new nodes being Tier 2 SV nodes
     [(dsoRulesCid, _)] <- query @DsoRules dso
     submit (actAs sv1 <> readAs dso) $ exerciseCmd dsoRulesCid $ DsoRules_OnboardValidator with
       sponsor = sv1
       validator = svParty
       version = Some "0.1.0"
       contactPoint = Some (svName <> "@example.com")
-    (_, earliestOpenRound) <- head <$> getOpenRoundsSorted app
-    submit dso $ exerciseCmd dsoRulesCid $ DsoRules_AddSv with
+    (earliestOpenRoundCid, earliestOpenRound) <- head <$> getOpenRoundsSorted app
+    result <- submit dso $ exerciseCmd dsoRulesCid $ DsoRules_AddSv with
       newSvParty = svParty
       newSvName = svName
-      newSvRewardWeight = 3 * bpsMultiplier -- Simulate the new nodes being Tier 2 SV nodes
+      newSvRewardWeight = if onLedgerSvRightOwners then 0 else rewardWeight
       newSvParticipantId = svName <> "-participant-id"
       joinedAsOfRound = earliestOpenRound.round
-
+    when onLedgerSvRightOwners $  do
+      instructionCid <- submit dso $ exerciseCmd result.newDsoRules $ DsoRules_AddSvRightOwner with
+        rightOwnerName = svName
+        info = SvRightOwnerInfo with
+          rightOwnerParty = svParty
+          voteWeight = 1
+          rewardWeight = rewardWeight
+          rewardNodeOperatorName = svName
+          beneficiaries = []
+      void $ submit (actAs sv1 <> readAs dso) $ exerciseCmd result.newDsoRules DsoRules_ExecuteAddSvRightOwnerInstruction with
+        instructionCid
+        openMiningRoundCid = earliestOpenRoundCid
+        svOperator = sv1
   -- check that the SvStatusReports are present
   checkSvContractInvariants app
 
   pure (app, dso, (sv1, sv2, sv3, sv4))
 
 
-getSvInfoByParty : AmuletApp -> Party -> Script SvInfo
+getSvInfoByParty : AmuletApp -> Party -> Script (Text, SvInfo)
 getSvInfoByParty app sv = do
   [(_, rules)] <- query @DsoRules app.dso
-  case Map.lookup sv rules.svs of
-    None -> fail $ "Not a sv: " <> show sv
-    Some info -> pure info
+  getSvInfoByOperatorParty sv rules
 
 generateUnclaimedReward : AmuletApp -> AmuletUser -> Script ()
 generateUnclaimedReward app provider1 = do
@@ -181,13 +199,13 @@ generateUnclaimedReward app provider1 = do
 dsoDelegateSubmits : AmuletApp -> (ContractId DsoRules -> Commands a) -> Script a
 dsoDelegateSubmits app mkCommands = do
   [(dsoRulesCid, dsoRules)] <- query @DsoRules app.dso
-  let sv = head (Map.keys dsoRules.svs)
+  let sv = head (operatorParties dsoRules)
   submit (actAs sv <> readAs app.dso) $ mkCommands dsoRulesCid
 
 dsoDelegateSubmitsMustFail : AmuletApp -> (ContractId DsoRules -> Commands a) -> Script ()
 dsoDelegateSubmitsMustFail app mkCommands = do
   [(dsoRulesCid, dsoRules)] <- query @DsoRules app.dso
-  let sv = head (Map.keys dsoRules.svs)
+  let sv = head (operatorParties dsoRules)
   submitMustFail (actAs sv <> readAs app.dso) $ mkCommands dsoRulesCid
 
 svSubmits : AmuletApp -> Party -> (ContractId DsoRules -> Commands a) -> Script a
@@ -228,7 +246,7 @@ runNextIssuanceD app amuletPrice appActivityRoundTotal = do
             svRewardCouponCids = map fst svRewards
             optValidatorFaucetCouponCids = Some (map fst validatorFaucetCoupons)
             optValidatorLivenessActivityRecordCids = Some (map fst validatorLivenessActivityRecords)
-          sv = Some (head (Map.keys dsoRules.svs))
+          sv = Some (head (operatorParties dsoRules))
 
       -- the closed round
       confirmAWC_MiningRound_Archive app
@@ -255,7 +273,7 @@ runNextIssuanceD app amuletPrice appActivityRoundTotal = do
           middleRoundCid = r2Cid
           latestRoundCid = r3Cid
           amuletPriceVoteCids
-          sv = Some (head (Map.keys dsoRules.svs))
+          sv = Some (head (operatorParties dsoRules))
     let summarizingRoundCid = advanceResult.summarizingRound
 
     Some summarizingRound <- queryContractId app.dso summarizingRoundCid
@@ -322,7 +340,7 @@ confirmAWC_MiningRound_Archive app = do
 confirmAndExecutionAction : AmuletApp -> ActionRequiringConfirmation -> Script ()
 confirmAndExecutionAction app action = do
   [(dsoRulesCid, rules)] <- query @DsoRules app.dso
-  forA_ (Map.keys rules.svs) $ \sv -> do
+  forA_ (operatorParties rules) $ \sv -> do
     -- mallory does not act
     unless ("mallory" `T.isPrefixOf` partyToText sv) $ do
       submit (actAs sv <> readAs app.dso) $ exerciseCmd dsoRulesCid DsoRules_ConfirmAction with
@@ -335,37 +353,37 @@ executeAllConfirmedActions : AmuletApp -> Script ()
 executeAllConfirmedActions app = do
   [(amuletRulesCid, _)] <- query @AmuletRules app.dso
   [(dsoRulesCid, rules)] <- query @DsoRules app.dso
-  let s = summarizeDso rules
   -- query all confirmations visible to DSO party
   confirmations0 <- query @Confirmation app.dso
   -- execute the actions for which there are enough confirmations
   let executableConfirmations =
           filter (const True) -- (\cs -> s.requiredNumConfirmations <= length cs)
         $ groupEqualOn (\c -> c._2.action) confirmations0
+      submitter = head (operatorParties rules)
   forA_ executableConfirmations $ \cs -> do
-    submit (actAs s.dsoDelegate <> readAs app.dso) $ exerciseCmd dsoRulesCid DsoRules_ExecuteConfirmedAction with
+    submit (actAs submitter <> readAs app.dso) $ exerciseCmd dsoRulesCid DsoRules_ExecuteConfirmedAction with
       action = (head cs)._2.action
       amuletRulesCid = Some amuletRulesCid
       confirmationCids = map fst cs
-      sv = Some s.dsoDelegate
+      sv = Some submitter
 
 executeAllDefinitiveVotes : AmuletApp -> Script ()
 executeAllDefinitiveVotes app = do
   [(amuletRulesCid, _)] <- query @AmuletRules app.dso
   [(dsoRulesCid, rules)] <- query @DsoRules app.dso
-  let s = summarizeDso rules
+  let submitter = head (operatorParties rules)
   -- query all voting requests visible to DSO party
   now <- getTime
   requests <- query @VoteRequest app.dso
   forA_ requests $ \(requestCid, request) -> do
-    let activeSvs = Set.fromList $ map (.name) $ Map.values rules.svs
+    let activeSvs = Set.fromList (map fst $ TextMap.toList (getSvRightOwners rules))
     let execute = request.voteBefore <= now ||
           (activeSvs == Set.fromList (Map.keys request.votes))
     when execute $ do
-      void $ submit (actAs s.dsoDelegate <> readAs app.dso) $ exerciseCmd dsoRulesCid DsoRules_CloseVoteRequest with
+      void $ submit (actAs submitter <> readAs app.dso) $ exerciseCmd dsoRulesCid DsoRules_CloseVoteRequest with
         requestCid
         amuletRulesCid = Some amuletRulesCid
-        sv = Some s.dsoDelegate
+        sv = Some submitter
 
 initiateAndCastVote : AmuletApp -> [Party] -> Optional Time -> ActionRequiringConfirmation -> Script (ContractId VoteRequest)
 initiateAndCastVote _ [] _ _ = error "initiateVote: require at least one party"
@@ -447,27 +465,23 @@ checkSvRewardStates : AmuletApp -> Script ()
 checkSvRewardStates app = do
   rewardStates <- query @SvRewardState app.dso
   [(_, dsoRules)] <- query @DsoRules app.dso
-  forA_ (Map.values dsoRules.svs) $ \info ->
-    require "Each onboarded andSV name has one SvRewardState contract" $
-      any (\(_, rewardState) -> rewardState.svName == info.name) rewardStates
-  forA_ (Map.values dsoRules.offboardedSvs) $ \info ->
-    require "Each offboarded and SV name has one SvRewardState contract" $
-      any (\(_, rewardState) -> rewardState.svName == info.name) rewardStates
-  let offAndOnboardedSvs = Set.fromList (map (\info -> info.name) (Map.values dsoRules.svs) <> map (\info -> info.name) (Map.values dsoRules.offboardedSvs))
-  Set.fromList [state.svName | (_, state) <- rewardStates] === offAndOnboardedSvs
-  require "No duplicate SvRewardstate contracts" $
-    length rewardStates == Set.size offAndOnboardedSvs
+  if onLedgerSvRightOwners dsoRules then
+    Set.fromList [state.svName | (_, state) <- rewardStates] === Set.fromList (map fst (TextMap.toList (getSvRightOwners dsoRules)))
+  else
+    Set.fromList [state.svName | (_, state) <- rewardStates] === Set.fromList (map (.name) (Map.values dsoRules.svs))
+  
 
 -- | Check that there's exactly one status report per SV both off-boarded and onboarded.
 checkSvStatusReports : AmuletApp -> Script ()
 checkSvStatusReports app = do
   [(_, dsoRules)] <- query @DsoRules app.dso
-  let svParties = Map.keys dsoRules.svs
+  let svParties = operatorParties dsoRules
   let offboardedParties = Map.keys dsoRules.offboardedSvs
   let expectedSvParties = Set.fromList $ svParties ++ offboardedParties
   reports <- query @SvStatusReport app.dso
   let reportingParties = Set.fromList $ [ report.sv | (_, report) <- reports ]
   reportingParties === expectedSvParties
+  pure ()
 
 getSvNodeState : AmuletApp -> Party -> Script (ContractId SvNodeState, SvNodeState)
 getSvNodeState app sv = do
--- a/daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
+++ b/daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
@@ -379,25 +379,46 @@ executeAllDefinitiveVotes app = do
     let activeSvs = Set.fromList (map fst $ TextMap.toList (getSvRightOwners rules))
     let execute = request.voteBefore <= now ||
           (activeSvs == Set.fromList (Map.keys request.votes))
+    rightOwnerCids <- querySvRightOwners app
     when execute $ do
+      Some request <- queryContractId app.dso requestCid
       void $ submit (actAs submitter <> readAs app.dso) $ exerciseCmd dsoRulesCid DsoRules_CloseVoteRequest with
         requestCid
         amuletRulesCid = Some amuletRulesCid
         sv = Some submitter
+        rightOwnerCids = if voteType request.action == RightOwnerVote then rightOwnerCids else None
+
+lookupRightOwner : AmuletApp -> Party -> Script (Optional (ContractId SvRightOwner))
+lookupRightOwner app rightOwnerParty = do
+  rightOwners <- queryFilter @SvRightOwner app.dso (\ro -> ro.info.rightOwnerParty == rightOwnerParty)
+  case rightOwners of
+    [] -> pure None
+    [(roCid, _)] -> pure (Some roCid)
+    _ -> abort "lookupRightOwner: multiple SvRightOwner contracts found for the same party"
+
+querySvRightOwners : AmuletApp -> Script (Optional (TextMap.TextMap (ContractId SvRightOwner)))
+querySvRightOwners app = do
+  rightOwners <- query @SvRightOwner app.dso
+  pure $ if null rightOwners 
+    then None
+    else Some (TextMap.fromList [(ro.rightOwnerName, roCid) | (roCid, ro) <- rightOwners])
 
 initiateAndCastVote : AmuletApp -> [Party] -> Optional Time -> ActionRequiringConfirmation -> Script (ContractId VoteRequest)
 initiateAndCastVote _ [] _ _ = error "initiateVote: require at least one party"
 initiateAndCastVote app (initiator::others) targetEffectiveAt action = do
     [(dsoRulesCid, _)] <- query @DsoRules app.dso
+    rightOwnerCid <- lookupRightOwner app initiator
     requestCid <- submit (actAs initiator <> readAs app.dso) $ exerciseCmd dsoRulesCid DsoRules_RequestVote with
       requester = initiator
       action
       reason = Reason with url = ""; body = "let's get it done"
       targetEffectiveAt
       voteRequestTimeout = None
+      rightOwnerCid = if voteType action == RightOwnerVote then rightOwnerCid else None
     foldlA (castVote dsoRulesCid) requestCid.voteRequest others
   where
     castVote dsoRulesCid requestCid sv = do
+      rightOwnerCid <- lookupRightOwner app sv
       result <- submit (actAs sv <> readAs app.dso) $ exerciseCmd dsoRulesCid DsoRules_CastVote with
           requestCid
           vote = Vote with
@@ -405,22 +426,26 @@ initiateAndCastVote app (initiator::others) targetEffectiveAt action = do
             accept = True
             reason = Reason with url = ""; body = "✓"
             optCastAt = None
+          rightOwnerCid = if voteType action == RightOwnerVote then rightOwnerCid else None
       return result.voteRequest
 
 initiateAndAcceptVote : AmuletApp -> [Party] -> ActionRequiringConfirmation -> Script ()
 initiateAndAcceptVote _ [] _ = error "initiateAndAcceptVote: require at least one party"
 initiateAndAcceptVote app (initiator::others) action = do
     [(dsoRulesCid, _)] <- query @DsoRules app.dso
+    rightOwnerCid <- lookupRightOwner app initiator
     requestCid <- submit (actAs initiator <> readAs app.dso) $ exerciseCmd dsoRulesCid DsoRules_RequestVote with
       requester = initiator
       action
       reason = Reason with url = ""; body = "let's get it done"
       targetEffectiveAt = None
       voteRequestTimeout = None
+      rightOwnerCid = if voteType action == RightOwnerVote then rightOwnerCid else None
     foldlA (castVote dsoRulesCid) requestCid.voteRequest others
     executeAllDefinitiveVotes app
   where
     castVote dsoRulesCid requestCid sv = do
+      rightOwnerCid <- lookupRightOwner app sv
       result <- submit (actAs sv <> readAs app.dso) $ exerciseCmd dsoRulesCid DsoRules_CastVote with
           requestCid
           vote = Vote with
@@ -428,13 +453,16 @@ initiateAndAcceptVote app (initiator::others) action = do
             accept = True
             reason = Reason with url = ""; body = "✓"
             optCastAt = None
+          rightOwnerCid = if voteType action == RightOwnerVote then rightOwnerCid else None
       return result.voteRequest
 
 -- cast a vote on a request
 castVote : AmuletApp -> Party -> ContractId VoteRequest -> Bool -> Script (ContractId VoteRequest)
 castVote app sv requestCid vote = do
     [(dsoRulesCid, _)] <- query @DsoRules app.dso
-    result <- submit (actAs sv <> actAs app.dso) $ exerciseCmd dsoRulesCid DsoRules_CastVote with
+    rightOwnerCid <- lookupRightOwner app sv
+    Some request <- queryContractId app.dso requestCid
+    result <- submit (actAs sv <> readAs app.dso) $ exerciseCmd dsoRulesCid DsoRules_CastVote with
         requestCid = requestCid
         vote = Vote with
             sv = sv
@@ -443,6 +471,7 @@ castVote app sv requestCid vote = do
                 url = ""
                 body = ""
             optCastAt = None
+        rightOwnerCid = if voteType request.action == RightOwnerVote then rightOwnerCid else None
     return result.voteRequest
 
 -- | Update all currently desired prices to the target price
--- a/daml/splice-dso-governance-test/daml/Splice/Scripts/TestOnboarding.daml
+++ b/daml/splice-dso-governance-test/daml/Splice/Scripts/TestOnboarding.daml
@@ -5,7 +5,6 @@ module Splice.Scripts.TestOnboarding where
 
 import DA.Assert
 import DA.Foldable (forA_)
-import DA.Optional (fromSome)
 import qualified DA.Map as Map
 import Daml.Script
 import DA.Time
@@ -144,7 +143,7 @@ testSvOnboarding = do
           newSvName = "svX"
           newParticipantId = "svX-participant-id"
           reason = "because"
-          newSvRewardWeight = 19 -- we use 19 as a dummy reward weight for testing
+          newSvRewardWeight = 0 -- weight is not used for on-ledger sv right owners
   executeAllConfirmedActions app
 
   -- svX is now a confirmed SV
@@ -154,7 +153,7 @@ testSvOnboarding = do
     svName = "svX"
     svParticipantId = "svX-participant-id"
     reason = "because"
-    svRewardWeight = 19
+    svRewardWeight = 0 -- weight is not used for on-ledger sv right owners
     dso
     expiresAt = svOnboardingConfirmed.expiresAt
 
@@ -323,7 +322,7 @@ testSvOnboardingRequestArchiveWhenSvIsAlreadyAddedAsSV = do
   submit dso $ exerciseCmd dsoRulesCid $ DsoRules_AddSv with
     newSvParty = svX
     newSvName = "svX"
-    newSvRewardWeight = 19
+    newSvRewardWeight = 0 -- weight is not used for on-ledger sv right owners
     newSvParticipantId = "svX-participant-id"
     joinedAsOfRound = (Round 3)
 
@@ -352,7 +351,7 @@ testSvOnboardingConfirmedExpiry = do
           newSvParty = svX
           newSvName = "svX"
           newParticipantId = "svX-participant-id"
-          newSvRewardWeight = 19
+          newSvRewardWeight = 0 -- sv weight is not used for on-ledger sv right owners
           reason = "because"
 
   executeAllConfirmedActions app
@@ -375,6 +374,10 @@ testSvOnboardingConfirmedExpiry = do
   [(_, dsoRules)] <- query @DsoRules dso
   Map.keys (dsoRules.offboardedSvs) === [svX]
 
+lookupSvOperatorParty : Text -> DsoRules -> Optional Party
+lookupSvOperatorParty name dsoRules =
+  fmap fst (find (\(_, info) -> info.name == name) $ Map.toList dsoRules.svs)
+
 
 testSvNameUniquenessNotInDevNet : Script ()
 testSvNameUniquenessNotInDevNet = do
@@ -383,7 +386,7 @@ testSvNameUniquenessNotInDevNet = do
 
   [(dsoRulesCid, dsoRules)] <- query @DsoRules dso
 
-  (fromSome (Map.lookup sv2 dsoRules.svs)).name === "sv2"
+  lookupSvOperatorParty "sv2" dsoRules === Some sv2
 
   svX <- allocateParty "svX"
   submitMustFail dso $ exerciseCmd dsoRulesCid $ DsoRules_AddSv svX "sv2" 19 "sv2-participant-id" (Round 3)
@@ -395,12 +398,12 @@ testSvOverwriteInDevNet = do
 
   [(dsoRulesCid, dsoRules)] <- query @DsoRules dso
 
-  (fromSome (Map.lookup sv1 dsoRules.svs)).name === "sv1"
-  (fromSome (Map.lookup sv2 dsoRules.svs)).name === "sv2"
+  lookupSvOperatorParty "sv1" dsoRules === Some sv1
+  lookupSvOperatorParty "sv2" dsoRules === Some sv2
   svX <- allocateParty "svX"
 
   -- overwriting existing sv successfully (using a new participant node)
-  submit dso $ exerciseCmd dsoRulesCid $ DsoRules_AddSv svX "sv2" 19 "svX-participant-id" (Round 3)
+  submit dso $ exerciseCmd dsoRulesCid $ DsoRules_AddSv svX "sv2" 0 "svX-participant-id" (Round 3)
 
   [(_, dsoRules)] <- query @DsoRules app.dso
   dsoRules.offboardedSvs === Map.fromList [(sv2, OffboardedSvInfo "sv2" "sv2-participant-id")]
@@ -410,7 +413,7 @@ testSvOverwriteInDevNet = do
   [(dsoRulesCid, _)] <- query @DsoRules app.dso
 
   -- overwriting existing sv again
-  submit dso $ exerciseCmd dsoRulesCid $ DsoRules_AddSv svX2 "sv2" 19 "svX2-participant-id" (Round 3)
+  submit dso $ exerciseCmd dsoRulesCid $ DsoRules_AddSv svX2 "sv2" 0 "svX2-participant-id" (Round 3)
 
   [(_, dsoRules)] <- query @DsoRules app.dso
   dsoRules.offboardedSvs === Map.fromList
@@ -429,7 +432,7 @@ testSvReonboarding = do
 
   [(_, dsoRules)] <- query @DsoRules dso
 
-  Map.keys dsoRules.svs === [sv1, sv2, sv3, sv4]
+  operatorParties dsoRules === [sv1, sv2, sv3, sv4]
 
   initiateAndAcceptVote app [sv1, sv2, sv3] $
     ARC_DsoRules with
@@ -439,14 +442,13 @@ testSvReonboarding = do
   executeAllDefinitiveVotes app
 
   [(_, dsoRules)] <- query @DsoRules dso
-  Map.keys dsoRules.svs === [sv1, sv2, sv3]
+  operatorParties dsoRules === [sv1, sv2, sv3]
   Map.keys dsoRules.offboardedSvs === [sv4]
   checkSvContractInvariants app
 
   -- reonboard sv4
   let token = "mock-token"
   sv4New <- allocateParty "sv4New"
-
   svSubmits app sv1 $ \cid -> exerciseCmd cid $
     DsoRules_StartSvOnboarding "sv4" sv4New "sv4New-participant-id" token sv1
   forA_ [sv1,sv2,sv3] $ \sv ->
@@ -458,7 +460,8 @@ testSvReonboarding = do
           newSvName = "sv4"
           newParticipantId = "sv4New-participant-id"
           reason = "because the sv4 owner still has the right to operate a node"
-          newSvRewardWeight = 19 -- we use 19 as a dummy reward weight for testing
+          newSvRewardWeight = 0 -- sv weight is not used for on-ledger sv right owners
+
   executeAllConfirmedActions app
   [(svOnboardingConfirmedCid, _)] <- query @SvOnboardingConfirmed dso
 
@@ -466,7 +469,6 @@ testSvReonboarding = do
 
   [(amuletRulesCid, _)] <- query @AmuletRules dso
   [(dsoRulesCid, _)] <- query @DsoRules dso
-
   submit (actAs sv4New <> readAs dso) $ exerciseCmd dsoRulesCid DsoRules_AddConfirmedSv with
     sv = sv4New
     svOnboardingConfirmedCid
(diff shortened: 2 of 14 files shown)


moritzkiefer-da:
@meiersi-da I'd appreciate a first review when you have time. It's not in as good a state as I'd like to (see the many caveats in the description) but given your upcoming vacation, I'd like some alignment beforehand and I'm not sure when I'll find the time to address all the caveats.

daml/splice-dso-governance/daml/Splice/DSO/SvRightOwner.daml
```diff
@@ -0,0 +1,45 @@
+module Splice.DSO.SvRightOwner where
+
+import DA.List
+import Splice.Types
+
+-- | Information about a super validator right owner.
+data SvRightOwnerInfo = SvRightOwnerInfo
+  with
+    governanceParty : Party -- ^ The governance party of an SV that can vote
```
meiersi-da:
```suggestion
    governanceParty : Party -- ^ The governance party representing the SV rights owner
```

daml/splice-dso-governance/daml/Splice/DSO/SvRightOwner.daml
```diff
@@ -0,0 +1,45 @@
...
+  rewardWeight >= 0 &&
+  sum (map snd beneficiaries) <= 1.0 &&
+  all (\(_, beneficiaryWeight) -> beneficiaryWeight > 0.0) beneficiaries &&
+  all (\(beneficiary,_) -> beneficiary /= governanceParty) beneficiaries &&
+  unique (map fst beneficiaries)
+
+
+data SvNodeOperatorInfo = SvNodeOperatorInfo 
+  with
+    party : Party
```
meiersi-da:
```suggestion
    operatorParty : Party  -- ^ Party representing the operator's SV node.
```

daml/splice-dso-governance/daml/Splice/DSO/SvRightOwner.daml
```diff
@@ -0,0 +1,45 @@
...
+    participantId : Text
```
meiersi-da:
```suggestion
    participantId : Text -- ^ Identity of the participant component of the SV's node.
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -150,6 +157,13 @@ data DsoSummary = DsoSummary with
       -- ^ The number of votes required for considering a confirmation, or a request for a vote
   deriving (Eq, Show)
 
+data DsoSummaryV2 = DsoSummaryV2
+  with
+    totalVoteWeight : Int
+    requiredVoteWeight : Int
+    votersWithWeight : [(Text, Int)]
```
meiersi-da: What's the text domain here? Why not a `TextMap`?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -150,6 +157,13 @@ data DsoSummary = DsoSummary with
...
+  deriving (Eq, Show)
```
meiersi-da, line 166: Side note: now that we can, should we switch to explicit serializability annotations to allow changing this kind of type going forward?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -547,39 +579,36 @@ template DsoRules with
...
     choice DsoRules_OffboardSv : DsoRules_OffboardSvResult
       with
         sv : Party
       controller dso
       do
-        require "There is more than one sv" (Map.size svs > 1)
-        case Map.lookup sv svs of
-          None -> fail "Not a sv"
-          Some info -> do
+        require "There is more than one sv" (TextMap.size svRightOwners > 1)
```
meiersi-da: Don't we need more than one SV node operator?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -547,39 +579,36 @@ template DsoRules with
...
-            -- and instead just choose the sv with the lexicographically smallest name.
-            let (newDsoDelegate, newEpoch)
-                  | sv == dsoDelegate = (head (Map.keys newSvs), epoch + 1)
-                  | otherwise        = (dsoDelegate, epoch)
+            let newSvs = TextMap.fromList (filter (\(_, v) -> (party <$> v.nodeOperatorInfo) /= Some sv) (TextMap.toList svRightOwners))
             let offboardedSvInfo = OffboardedSvInfo with
-                    name = info.name
-                    participantId = info.participantId
+                    name = name
+                    participantId = (fromSome (info.nodeOperatorInfo)).participantId
```
meiersi-da: The code change here doesn't look exactly trivial. What do you think about instead having an enum/flag that determines whether to use old or new style node management, and guard old and new choices on it. I believe that would make it easier to reason about. In particular, if the direction is one way: once we switch to the new choices the old ones will never the used again.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -588,21 +617,20 @@ template DsoRules with
...
-          None -> fail "SV is not an SV"
-          Some info -> do
-            previousReport <- fetchAndArchive (ForSvNode with dso; sv; svName = info.name) previousReportCid
-            create SvStatusReport with
-              dso
-              sv
-              svName = info.name
-              number = previousReport.number + 1
-              status = Some status
+        name <- getAndValidateSvNodeOperatorParty this (Some sv)
```
meiersi-da: When reading "sv" there is always the question: is this meant to be a governance party or is it meant to be a node operator party?

wdyt, about always using "svOperator" and "svRightsOwner" to clarify this in code?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -727,13 +754,14 @@ template DsoRules with
         sv : Optional Party
       controller sv
       do
-        _ <- getAndValidateSvParty this sv
-        let s = summarizeDso this
+        _ <- getAndValidateSvNodeOperatorParty this sv
+        let s = summarizeDso OperatorVote this
         let forDso = ForDso with dso
-        require "Enough confirmations" (length confirmationCids >= s.requiredNumVotes)
+        -- FIXME: Needs potential changing if operator weights are not always 1.0
```
meiersi-da: Do we have a CIP that pushes in that direction? At many of the lower levels (e.g., Canton thresholds we have fixed weight 1.0 for voting-like decisions). I suggest we rather keep the invariant that operators are all equally weighted at the operational level.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -755,9 +783,7 @@ template DsoRules with
         targetEffectiveAt : Optional Time
       controller requester
       do
-        requesterName <- case requester `Map.lookup` svs of
-          None -> fail "Requester is not an SV"
-          Some info -> pure (info.name)
+        requesterName <- getAndValidateVotingParty this requester action
```
meiersi-da: VotingParty vs. GovernanceParty vs. RightOwner ... let's decide on one term and use it

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -840,34 +864,33 @@ template DsoRules with
...
               ]
-        let abstainingSvs = [ name | name <- activeSvs, not (name `Map.member` request.votes) ]
-        let (yays, nays) = partition (.accept) validVotes
-        let numYays = length yays
-        let numNays = length nays
-        let numVotes = numYays + numNays
+        let abstainingSvs = [ name | (name, _) <- s.votersWithWeight, not (name `Map.member` request.votes) ]
+        let (yays, nays) = partition (\(_, v) -> v.accept) validVotes
+        let yayWeight = sum $ map (\(name, _) -> fromSome (TextMap.lookup name voteWeightMap)) yays
+        let nayWeight = sum $ map (\(name, _) -> fromSome (TextMap.lookup name voteWeightMap)) nays
```
meiersi-da: `fromSome` hints at a potential streamlining: perform the lookup as part of partitioning the votes, and thus also clarify that only active SV rights owners are counted.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -942,11 +965,13 @@ template DsoRules with
...
       controller sv
       do
-        _ <- getAndValidateSvParty this sv
+        _ <- getAndValidateSvNodeOperatorParty this sv
         let forDso = ForDso with dso
         -- validate and archive non-sv votes
         forA_ nonSvVoteCids $ \voteCid -> do
           vote <- fetchAndArchive forDso voteCid
+          -- FIXME: probably we want non node operators to be able to
+          -- do amulet price votes.
```
meiersi-da: Yeah, I'd even expect that some might want it to be an explicit flag as to whether they allow their operator to vote for the amulet price.

Also brings up the question: should the median be determined in a weighted fashion as well? Probably...

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1038,7 +1064,9 @@ template DsoRules with
         contactPoint : Optional Text
       controller sponsor
       do
-        require "Sponsor is an SV" (sponsor `Map.member` svs)
+        -- FIXME: should this work for governance parties? maybe not that relevant given that long-term
+        -- this should be replaced by staking based onboarding.
```
meiersi-da: Yeah, short-term I'd though expect that they all should have the right to create onboarding secrets.

daml/splice-dso-governance/daml/Splice/DsoBootstrap.daml
```diff
@@ -60,28 +62,31 @@ template DsoBootstrap with
...
+                    governanceParty = sv1Party
+                    nodeOperatorInfo = Some SvNodeOperatorInfo with
+                      party = sv1Party
+                      participantId = sv1ParticipantId
+                    voteWeight = 1
+                    rewardWeight = sv1RewardWeight
+                    rewardNodeOperator = sv1Name
+                    joinedAsOfRound = fromOptional (Round 0) result.initialRound
+                    beneficiaries = []
+                )
```
meiersi-da: There seems to be a mapping from an `svs` value to an `svRightOwners` value. I'd suggest to encode that in `SvRightOwner.daml` and use it both here, and in the `DsoRules` to encode the switch to the new fields.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1954,12 +2152,25 @@ createPerSvPartyContracts dso newSvParty newSvName synchronizerNodes amuletPrice
...
-
+getAndValidateSvNodeOperatorParty : DsoRules -> Optional Party -> Update Text
+getAndValidateSvNodeOperatorParty _ None = fail "no SV party provided"
+getAndValidateSvNodeOperatorParty rules (Some sv) = do
+  case find (\(_, info) -> (party <$> info.nodeOperatorInfo) == Some sv) (TextMap.toList rules.svRightOwners) of
+    None -> fail ("SV party " <> show sv <> " is not an SV node operator")
+    Some (name, _) -> pure name
+
+getAndValidateSvGovernanceParty : DsoRules -> Optional Party -> Update Text
+getAndValidateSvGovernanceParty _ None = fail "no SV party provided"
```
meiersi-da:
```suggestion
getAndValidateSvGovernanceParty _ None = fail "no SV governance party provided"
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1954,12 +2152,25 @@ createPerSvPartyContracts dso newSvParty newSvName synchronizerNodes amuletPrice
...
       synchronizerNodes
 
-getAndValidateSvParty : DsoRules -> Optional Party -> Update Party
-getAndValidateSvParty _ None = fail "no SV party provided"
-getAndValidateSvParty rules (Some sv) = do
-  require "SV party is an actual SV" (sv `Map.member` rules.svs)
-  pure sv
-
+getAndValidateSvNodeOperatorParty : DsoRules -> Optional Party -> Update Text
+getAndValidateSvNodeOperatorParty _ None = fail "no SV party provided"
```
meiersi-da:
```suggestion
getAndValidateSvNodeOperatorParty _ None = fail "no SV operator party provided"
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1954,12 +2152,25 @@ createPerSvPartyContracts dso newSvParty newSvName synchronizerNodes amuletPrice
...
+getAndValidateSvNodeOperatorParty rules (Some sv) = do
+  case find (\(_, info) -> (party <$> info.nodeOperatorInfo) == Some sv) (TextMap.toList rules.svRightOwners) of
```
meiersi-da:
```suggestion
  case find (\(_, info) -> (operatorParty <$> info.nodeOperatorInfo) == Some sv) (TextMap.toList rules.svRightOwners) of
```

took me a while to understand that `party` refers to the `party` field in `nodeOperatorInfo`.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1878,48 +2069,55 @@ svHasBeenOnboardedBefore svName this =
...
+        beneficiaries = []
   create this with
-    svs = Map.insert newSvParty svInfo svs
+    svRightOwners = TextMap.insert newSvName svInfo svRightOwners
 
-createPerSvContracts : DsoRules -> DsoRules_AddSv -> Update ()
-createPerSvContracts DsoRules{..} DsoRules_AddSv{..} = do
+createSvRightOwnerContracts : DsoRules -> Text -> Round -> Update ()
+createSvRightOwnerContracts DsoRules{..} newSvName joinedAsOfRound = do
   void $ create SvRewardState with
```
meiersi-da, line 2273: missing work: adjusting the comments on the existing contracts to call out whether they concern rights owners or operators.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1747,25 +1936,22 @@ pruneAtLeastOne now schedule =
...
+    -- FIXME: Consider if the math really translates well to weights
+    requiredVoteWeight =
+      ceiling ((intToDecimal (totalVoteWeight + f + 1)) / 2.0)
+    votersWithWeight
   where
-    numSvs = Map.size dsoRules.svs
-    f = floor ((intToDecimal (numSvs - 1)) / 3.0)
+    votersWithWeight = case voteType of
+      GovernanceVote -> [(name, info.voteWeight) | (name, info) <- TextMap.toList dsoRules.svRightOwners, info.voteWeight > 0]
+      OperatorVote -> [(name, 1) | (name, info) <- TextMap.toList dsoRules.svRightOwners, isSome info.nodeOperatorInfo] -- FIXME: revisit what weight we want here
```
meiersi-da: I originally thought that `nodeOperatorInfo` is the information about who is the node operator for an SV right owner. This line here makes me doubt that. Why is it OK to use SV right owner names to count node operator votes? Wouldn't we want exactly one vote from every operator, and wouldn't operators be identified by `operatorParty`-id?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1731,13 +1821,112 @@ template DsoRules with
...
+            -- We archive the reward state as part of offboarding to avoid having to track offboardedSvs for non-operators.
+            -- In theory reonboarding an SV immediately could let it mint rewards twice for the same round.
+            -- In practice, reonboarding that quickly seems unrealistic in practice.
+            -- FIXME: Alternatively we could maybe track all offboarded SVs on a separate singleton contract, don't really see that it provides enough value though.
+            fetchAndArchive (ForSv with dso; svName = name) rewardStateCid
+            create this with svRightOwners = TextMap.delete name svRightOwners
+          None -> abort ("No SV with name " <> show name)
+
+    -- FIXME: Consider if we want this as a vote or go through an SvOnboardingConfirmed style flow.
+    -- Vote is probably easier to explain but harder to transition to from the current state.
```
meiersi-da: Couldn't we also distinguish between allowing a node operator to be onboarded, and then the actual onboarding flow?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -517,15 +543,20 @@ template DsoRules with
...
     svs : Map.Map Party SvInfo
+      -- ^ Deprecated in favor of svRightOwners
     offboardedSvs : Map.Map Party OffboardedSvInfo
+      -- ^ Only set for sv node owners to handle DSO party offboarding
     dsoDelegate : Party
       -- ^ __Deprecated__ in favor of delegateless automation.
     config : DsoRulesConfig
     initialTrafficState: Map.Map Text TrafficState -- ^ Map from participant/mediator ID to its traffic state at the time of synchronizer bootstrapping. Used for testing, empty in prod.
     isDevNet : Bool
+    svRightOwners : TextMap.TextMap SvRightOwnerInfo
```
meiersi-da: I'm not sure we want to deprecate the `svs` field. It seems that an interesting alternative is to keep that as the field that stores the state of the node operators. With that both `svs` and `offboardedSvs` make sense.

We can then focus `svRightOwner` purely on tracking `svRightOwners` state. Part of that state is which operator represents the right owner. I suspect we also need to distinguish between the long-term textual ids used to identify node operators and the shorter-term party-ids used to identify their node.

We probably need the same distinction for SV rights owners. 

We can probably just extend our existing convention of talking about SV operator names and parties to SV rights owner names and parties.

meiersi-da: Side note: I'd enforce invariants across the two fields in the ensure clause.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1341,45 +1370,106 @@ template DsoRules with
...
+            weight
+            round = openRound.round
+
+        return DsoRules_ReceiveSvRewardCouponResult with
+            svRewardState = newRewardStateCid
+            svRewardCoupons = couponCids
+
+    nonconsuming choice DsoRules_ReceiveSvRewardCouponV2 : ()
+      with
+        sv : Party
```
meiersi-da: I assume this is an operator party is it?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1341,45 +1370,106 @@ template DsoRules with
...
+        openRoundCid : ContractId OpenMiningRound
+        svRewardStates : TextMap.TextMap (ContractId SvRewardState)
```
meiersi-da: Could we track only one reward state per operator, and have the operator trigger the reward minting and distribute it all in one go? That seems a lot simpler.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1731,13 +1821,112 @@ template DsoRules with
...
+      controller oldGovernanceParty
+      do
+        case TextMap.lookup name svRightOwners of
+          None -> abort ("No sv with name " <> show name)
+          Some info -> do
+            require "governance parties match" (info.governanceParty == oldGovernanceParty)
+            create this with
+              svRightOwners = TextMap.insert name (info with governanceParty = newGovernanceParty) svRightOwners
+
+    -- FIXME: Consider if we want separate choices for updating weights, parties, …. Not sure we need to from an authorization pov if it has dso controller. Maybe
```
meiersi-da: Reminds me a lot of changing individual fields in our configs. Should we use `Patchable` to achieve this?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1731,13 +1821,112 @@ template DsoRules with
...
+        name : Text
+        info : SvRightOwnerInfo
+         -- fixme: don't vote on this directly but vote on an instruction that does not yet fix the round
+      controller dso
+      do
+        case TextMap.lookup name svRightOwners of
+          Some _ -> abort ("SV with name " <> name <> " already exists")
+          None -> do
+            ensureNoOperator info
+            createSvRightOwnerContracts this name info.joinedAsOfRound
```
meiersi-da: Consider whether we need the `joinedAsOfRound`.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1731,13 +1821,112 @@ template DsoRules with
...
+            create this with
+              svRightOwners = TextMap.insert name info svRightOwners
+
+    choice DsoRules_RemoveSvRightOwner : ContractId DsoRules
+      with
+        name : Text
+        rewardStateCid : ContractId SvRewardState
+        -- fixme: don't vote on this directly but vote on an instruction that does not yet fix the cid
```
meiersi-da: If we keep `SvRewardState` per SV node operator, then we get rid of this complication.

meiersi-da (commented):
Thanks a lot for this first cut. Very valuable to see the code.

I did review all code at a high-level, but didn't check the precise logic, as I believe that logic will change a lot if we manage to reuse more of our existing infra to track "SV node operator state" and manage to constrain the additional fields to only track the additional "SV rights owner state" required to make all workflows tick.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: Yeah I agree, the voting CIP is very ambigous on that though. I would hope we can sell it though without much trouble.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: good point, will fix

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: makes sense :+1:

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: good catch, will change

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: Yeah I think that goes in the direction I suggested for the migration path. Will change. I was first trying to get all old choices to work against setting svRightOwners but I think that approach doesn't make a whole lot of sense.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: I agree but I'm unsure how to apply that here. `sv` is the name used in the existing choice argument. I can't change that.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: but this choice is used for both operator and governance votes so I can't use a specific term?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: agreed will refactor

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: maybe, in a world where IP whitelists still exist having the people that control IP whitelists control onboarding secrets doesn't seem that off. but either way easy enough to change.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: yeah I would expect weighted medians.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: Right now we have:

1. A list of SV rights owners
2. Each SV right owner can also be an operator and in that case the information on their node is in nodeOperatorInfo.

So this assumes that DA-1/DA-2 are separate right owners for example. And node operators can still be uniquely identified by the SV name.

The alternative would be that we allow multiple operators per rights owner. That seems like an overall more complex model to me though and I'm not sure it's really better than having something like DA-1/DA-2 as separate rights owners.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: It is, I'm a bit wary of having half of the choices call this `sv` and half of the choices use more specific names (and we can't change the existing choices), that seems more confusing than helpful

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: I see two reasons why I prefer keeping SvRewardState per owner and not operator:

1. It seems to make it easier to change the operator that can mint for you. I guess you can just accept that you can you don't care about double minting on that and then changing the operator isn't all that hard either way.
2. With CIP 105 we'll need to start tracking more things per SV right owner, e.g., the amount minted so far. I'm not saying this necessarily literally needs to be tracked on SvRewardState but I think it definitely needs to be tracked per SV right owner and be preserved when changing the operator through which you mint (so unlike temporary double minting you can't ignore this). That seems to point at tracking things directly per SV right owner to be the better direction to me.

Complexity is a fair point, but it does seem like it is fully isolated to this one choice which seems pretty easy to test so I'm not all that concerned about it.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: Interesting idea, I do find it personally a lot easier to follow the code when it's all tracked in one place. It also was very valuable in forcing adjustments of the existing code to be explcit in whether we want operators or right owners. admittedly that may have been just a one time benefit.

> I suspect we also need to distinguish between the long-term textual ids used to identify node operators and the shorter-term party-ids used to identify their node.

I don't really understand what you mean. We already have the long term SV name and then more temporary governance and operator parties. Governance parties you can either forward as the governing party or change through vote. Node operator parties you can only change through reonboarding which matches what we currently support in practice.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: yeah was thinking the same, seems to make sense.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: note that svrewardstate was already tied to sv name not operator before and got preserved on operator changes for that name

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: yeah that may be a decent option

daml/splice-dso-governance/daml/Splice/DsoRules.daml
meiersi-da: Ah good point. Consider then just adding comments on the param.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
meiersi-da: yes. I understood later what you meant by that distinction. resolving.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
meiersi-da: If we go for the convention of using the proper name where possible, and a comment otherwise, I think we're sort of OK. I'd then probably also just create a `let` binding right at the start of the choice to have the proper name in the remainder.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
meiersi-da: > I see two reasons why I prefer keeping SvRewardState per owner and not operator:
> 
> 1. It seems to make it easier to change the operator that can mint for you. I guess you can just accept that you can you don't care about double minting on that and then changing the operator isn't all that hard either way.

Oh I overlooked the double-minting aspect. Yeah that might be problematic, but likely not too difficult to solve.

> 3. With CIP 105 we'll need to start tracking more things per SV right owner, e.g., the amount minted so far. I'm not saying this necessarily literally needs to be tracked on SvRewardState but I think it definitely needs to be tracked per SV right owner and be preserved when changing the operator through which you mint (so unlike temporary double minting you can't ignore this). That seems to point at tracking things directly per SV right owner to be the better direction to me.> 
> Complexity is a fair point, but it does seem like it is fully isolated to this one choice which seems pretty easy to test so I'm not all that concerned about it.

Hmm... yeah. Difficult to say which split and code organization is better. I was looking for an easy explanation of what's going on and couldn't find it, which is why I was pointing in that direction.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: I like the let option, will switch to that

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: given that we're fully moving away from sv sponsored onboarding, I've gone for not changing this and only allowing operators to onboard until then.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: resolving in favor of https://github.com/canton-network/splice/issues/6489

moritzkiefer-da:
@meiersi-da once you caught up this should be ready for another review

daml/splice-dso-governance/daml/Splice/DSO/SvRightOwner.daml
```diff
@@ -0,0 +1,31 @@
...
+data SvRightOwnerInfo = SvRightOwnerInfo
+  with
+    rightOwnerParty : Party -- ^ The party representing the SV rights owner
+    voteWeight : Int
+      -- ^ the weight of an SV for voting, can be zero
+    rewardWeight : Int
+      -- ^ the weight of an SV for reward minting, can be zero
+    rewardNodeOperatorName : Text
+      -- ^ Name of the SV node operator through which rewards are minted.
+      -- For rewards to be minted both the node operator.
```
meiersi-da, line 18: runaway text

daml/splice-dso-governance/daml/Splice/DSO/SvRightOwner.daml
```diff
@@ -0,0 +1,31 @@
...
+    beneficiaries : [(Party, Decimal)]
+      -- ^ List of beneficiaries and their percentages for reward distribution. Must not include rightOwnerParty. Remainder goes to rightOwnerParty.
+      -- FIXME: probably want to move this to a separate contract
+      -- FIXME: Comment for reviewer: This is deliberately a percentage not a weight. With automatic weight adjustments absolutel weights don't work anymore.
+      -- Beneficiaries are also more of an exception than the rule now that all actual SVs are directly represented as SVs and beneficiaries
+      -- are only if an SV does not want to mint directly under their SV party.
```
meiersi-da: Consider adding the usual constraints wrt max 20 and non-negative percentages that add up to `1.0`.

daml/splice-dso-governance/daml/Splice/DSO/SvState.daml
```diff
@@ -43,7 +43,7 @@ data RewardState = RewardState with
 -- | State of reward collection for a sv identified by their sv name.
 template SvRewardState with
     dso : Party
-    svName : Text
+    svName : Text -- ^ sv right owner name
```
meiersi-da, line 46: consider adjusting the comments on all types in this module wrt whether the `sv` reference refers to a rights owner or a node operator.

meiersi-da: nit: capitalize SV for consistency

daml/splice-dso-governance/daml/Splice/DsoBootstrap.daml
```diff
@@ -35,6 +37,8 @@ template DsoBootstrap with
     initialTrafficState: Map.Map Text TrafficState
     isDevNet : Bool
     initialRound : Optional Int
+    sv1RightOwnerName : Optional Text
+    sv1VoteWeight : Optional Int
```
meiersi-da, line 41: Is there a good location in the code to explain the distinction made between rights owner and node operator and to which of the two concepts we associate key parameters like vote and reward weight?

daml/splice-dso-governance/daml/Splice/DsoBootstrap.daml
```diff
@@ -60,28 +64,32 @@ template DsoBootstrap with
...
+                name = sv1Name
+                svRewardWeight = if isSome sv1RightOwnerName then 0 else sv1RewardWeight
+                participantId = sv1ParticipantId
+                joinedAsOfRound
               offboardedSvs = Map.empty
               dsoDelegate = sv1Party
               config
               initialTrafficState
               isDevNet
+              svRightOwners = flip fmap sv1RightOwnerName $ \rightOwnerName -> TextMap.fromList [
```
meiersi-da:
```suggestion
              svRightOwners = forA sv1RightOwnerName $ \rightOwnerName -> TextMap.fromList [
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -518,19 +542,53 @@ data TrafficState = TrafficState with
...
+
+requireNoOnLedgerSvRightOwners : DsoRules -> Update ()
+requireNoOnLedgerSvRightOwners this =
+  require "svRightOwners are not tracked on-ledger" (not $ onLedgerSvRightOwners this)
+
 template DsoRules with
     dso : Party
     epoch : Int
     svs : Map.Map Party SvInfo
+      -- ^ Deprecated in favor of svRightOwners
```
meiersi-da: Is this really deprecated? I thought this field now represents the SV operators.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -518,19 +542,53 @@ data TrafficState = TrafficState with
...
     offboardedSvs : Map.Map Party OffboardedSvInfo
+      -- ^ Only set for sv node owners to handle DSO party offboarding
```
meiersi-da, line 566: lets consistently use one of "SV operators" or "SV node owners"

meiersi-da: my vote goes to "SV operators"

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -518,19 +542,53 @@ data TrafficState = TrafficState with
...
     config : DsoRulesConfig
     initialTrafficState: Map.Map Text TrafficState -- ^ Map from participant/mediator ID to its traffic state at the time of synchronizer bootstrapping. Used for testing, empty in prod.
     isDevNet : Bool
+    svRightOwners : Optional (TextMap.TextMap SvRightOwnerInfo) -- ^ Map from sv right owner name to info about the right owner
+
   where
     ensure
-      config.numUnclaimedRewardsThreshold > 0
+      config.numUnclaimedRewardsThreshold > 0 &&
+      all (\(_, v) -> validSvRightOwnerInfo this v) (TextMap.toList (getSvRightOwners this)) &&
```
meiersi-da: you can use https://github.com/canton-network/splice/blob/7886fe89c454014827ea7c69b132f2c1ecc5fb88/token-standard/splice-token-standard-utils/daml/Splice/TokenStandard/Utils.daml#L84 if you like. That was the best place that I found for exporting these missing functions as part of the TS work.

meiersi-da: or even better use `Foldable.all`, as https://docs.daml.com/daml/stdlib/Prelude.html#type-da-internal-lf-textmap-11691

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1745,16 +1902,199 @@ template DsoRules with
...
           expectedDso = dso
         pure DsoRules_BootstrapExternalPartyConfigStateResult
 
+    choice DsoRules_MigrateToOnLedgerSvRightOwners : ContractId DsoRules
+      with
+        svOperator : Party
+      controller svOperator
+      do
+        _ <- getAndValidateSvNodeOperatorParty this (Some svOperator)
+        require "svRightOwners is not already on-ledger" (isNone this.svRightOwners)
```
meiersi-da, line 1909:
```suggestion
        require "svRightOwners are not already on-ledger" (isNone this.svRightOwners)
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1745,16 +1902,199 @@ template DsoRules with
...
+        let svRightOwnerFromOperator svOperator info = SvRightOwnerInfo with
+              rightOwnerParty = svOperator
+              rewardNodeOperatorName = info.name
+              rewardWeight = info.svRewardWeight
+              beneficiaries = []
+                -- ^ FIXME: Consider how we want the migration to work
+                -- option 1: Set it to empty here and then have automation in the sv app that sets it to the ones locally configured.
+                --           Then through two votes with same effectivity reduce weight & remove beneficiary + a separate vote to add new sv right owner
+                -- option 2: Switchover through a vote that sets beneficiaries for everyone
```
meiersi-da: I'm inclined to go for this option as it moves from one consistent state to the next one. The form to configure the transition should not be too much work, or not?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1745,16 +1902,199 @@ template DsoRules with
...
+                -- option 3: two-step process, first setup sv right owners through votes (probably a batched vote in some form) and then have a second vote to actually start using right owners (needs a new config flag to indicate whether right owners take effect)
+              voteWeight = 1
```
meiersi-da: Not sure that this is the right default. If we go for Option 2 then I'd suggest that the vote also inputs the the vote weights; i.e. it should just input the whole `svRightOwners` value.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1745,16 +1902,199 @@ template DsoRules with
...
+        newRightOwnerParty : Party
+      controller oldRightOwnerParty
+      do
+        requireOnLedgerSvRightOwners this
+        case TextMap.lookup name (getSvRightOwners this) of
+          None -> abort ("No sv with name " <> show name)
+          Some info -> do
+            require "right owner parties match" (info.rightOwnerParty == oldRightOwnerParty)
+            create this with
+              svRightOwners = Some (TextMap.insert name (info with rightOwnerParty = newRightOwnerParty) (getSvRightOwners this))
```
meiersi-da: What about the option of grouping the properties that the right owner themselves can change into a record and the use `Patchable` to change that analogous to the other kinds of config changes that we have?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1745,16 +1902,199 @@ template DsoRules with
...
+      controller svOperator
+      do
+        _ <- getAndValidateSvNodeOperatorParty this (Some svOperator)
+        requireOnLedgerSvRightOwners this
+        instruction <- fetchAndArchive (ForDso dso) instructionCid
+        case TextMap.lookup instruction.rightOwnerName (getSvRightOwners this) of
+          Some _ -> do
+            -- We archive the reward state as part of offboarding to avoid having to track offboardedSvs for non-operators.
+            -- In theory reonboarding an SV immediately could let it mint rewards twice for the same round.
+            -- In practice, reonboarding that quickly seems unrealistic in practice.
```
meiersi-da, line 2027:
```suggestion
            -- We archive the reward state for SV right owners as part of offboarding to avoid having to track their offboarding in `DsoRules` as we do for SV node operators.
            -- In theory reonboarding an SV rights owner immediately could let it mint rewards twice for the same round.
            -- In practice, reonboarding that quickly seems unrealistic, as it requires a vote by all SV right owners to complete within 20 minutes.
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1745,16 +1902,199 @@ template DsoRules with
...
+      controller svOperator
+      do
+        _ <- getAndValidateSvNodeOperatorParty this (Some svOperator)
+        requireOnLedgerSvRightOwners this
+        instruction <- fetchAndArchive (ForDso dso) instructionCid
+        assertDeadlineExceeded "AddSvRightOwnerInstruction.expiresAt" instruction.expiresAt
+        pure ()
+
+
+template AddSvRightOwnerInstruction
```
meiersi-da, line 2058: explain purpose of this template so readers don't have to infer it from the code

meiersi-da: also below

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1745,16 +1902,199 @@ template DsoRules with
...
+        svOperator : Party
+      controller svOperator
+      do
+        _ <- getAndValidateSvNodeOperatorParty this (Some svOperator)
+        requireOnLedgerSvRightOwners this
+        instruction <- fetchAndArchive (ForDso dso) instructionCid
+        assertDeadlineExceeded "RemoveSvRightOwnerInstruction.expiresAt" instruction.expiresAt
+        pure ()
+
+    nonconsuming choice DsoRules_ExpireAddSvRightOwnerInstruction : ()
```
meiersi-da, line 2045: I wish we had native expiry...

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -2067,3 +2430,10 @@ instance Patchable LogicalSynchronizerUpgradeSchedule where
         upgradeTime = patch new.upgradeTime base.upgradeTime current.upgradeTime
         newPhysicalSynchronizerSerial = patch new.newPhysicalSynchronizerSerial base.newPhysicalSynchronizerSerial current.newPhysicalSynchronizerSerial
         newPhysicalSynchronizerProtocolVersion = patch new.newPhysicalSynchronizerProtocolVersion base.newPhysicalSynchronizerProtocolVersion current.newPhysicalSynchronizerProtocolVersion
+
+operatorParties : DsoRules -> [Party]
```
meiersi-da, line 2436: nit: location seems a bit off ==> push to the other functions taking `DsoRules` as input.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1972,12 +2322,25 @@ createPerSvPartyContracts dso newSvParty newSvName synchronizerNodes amuletPrice
...
-getAndValidateSvParty _ None = fail "no SV party provided"
-getAndValidateSvParty rules (Some sv) = do
-  require "SV party is an actual SV" (sv `Map.member` rules.svs)
-  pure sv
-
+getAndValidateSvNodeOperatorParty : DsoRules -> Optional Party -> Update Text
+getAndValidateSvNodeOperatorParty _ None = fail "no SV operator party provided"
+getAndValidateSvNodeOperatorParty rules (Some sv) = do
+  case Map.lookup sv rules.svs of
+    None -> fail ("SV party " <> show sv <> " is not a registered SV node operator")
```
meiersi-da, line 2319:
```suggestion
    None -> fail ("party " <> show sv <> " is not a registered SV node operator")
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1972,12 +2322,25 @@ createPerSvPartyContracts dso newSvParty newSvName synchronizerNodes amuletPrice
...
+    Some svInfo -> pure svInfo.name
+
+getAndValidateSvRightOwnerParty : DsoRules -> Optional Party -> Update Text
+getAndValidateSvRightOwnerParty _ None = fail "no SV right owner party provided"
+getAndValidateSvRightOwnerParty rules (Some sv) = do
+  case find (\(_, info) -> info.rightOwnerParty == sv) (TextMap.toList (getSvRightOwners rules)) of
+    None -> fail ("SV party " <> show sv <> " is not a right owner party")
```
meiersi-da:
```suggestion
    None -> fail ("party " <> show sv <> " is not a right owner party")
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1936,19 +2285,20 @@ dsoRules_addSv this0 arg@DsoRules_AddSv{..} = do
   create this with
     svs = Map.insert newSvParty svInfo svs
 
-createPerSvContracts : DsoRules -> DsoRules_AddSv -> Update ()
-createPerSvContracts DsoRules{..} DsoRules_AddSv{..} = do
+createSvRightOwnerContracts : DsoRules -> Text -> Round -> Update ()
+createSvRightOwnerContracts DsoRules{..} newSvName joinedAsOfRound = do
```
meiersi-da:
```suggestion
createSvRightOwnerContracts DsoRules{..} svRightOwnerName joinedAsOfRound = do
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1899,34 +2245,37 @@ ensureNeverOperatedNode newSvParty this = do
...
+-- factored out from the choice to avoid mistakes from having the fields of DsoRules{..} in scope
+dsoRules_addSv : DsoRules -> DsoRules_AddSv -> Update (ContractId DsoRules)
+dsoRules_addSv this0 DsoRules_AddSv{..} = do
+  ensureNeverOperatedNode newSvParty this0
+  this@DsoRules{..} <- checkSvNotAlreadyOnboarded this0 newSvName
+  if onLedgerSvRightOwners this
+    then require "rewardWeight must be 0 when onboarding a new SV operator with on-ledger right owners" (newSvRewardWeight == 0)
+    else
+      unless (svHasBeenOnboardedBefore newSvName this) $
+      createSvRightOwnerContracts this newSvName joinedAsOfRound
```
meiersi-da:
```suggestion
        createSvRightOwnerContracts this newSvName joinedAsOfRound
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1899,34 +2245,37 @@ ensureNeverOperatedNode newSvParty this = do
...
-        require "SV is not currently onboarded" (isNone $ lookupSvInfoByName newSvName this0)
-        pure this0
+        require "SV is not currently onboarded" (isNone $ lookupSvOperatorInfoByName svOperatorName this)
+        pure this
 
-  -- create per-sv contracts if they have never been onboarded
-  unless (svHasBeenOnboardedBefore newSvName this) $
-    createPerSvContracts this arg
+-- factored out from the choice to avoid mistakes from having the fields of DsoRules{..} in scope
+dsoRules_addSv : DsoRules -> DsoRules_AddSv -> Update (ContractId DsoRules)
```
meiersi-da, line 2250:
```suggestion
dsoRules_addSvOperator : DsoRules -> DsoRules_AddSv -> Update (ContractId DsoRules)
```

If this suggestion matches the use of this function, then I'd suggest to also push the 'operator' qualification into the functions it calls.

daml/splice-dso-governance-test/daml/Splice/Scripts/TestOnboarding.daml
```diff
@@ -144,7 +143,7 @@ testSvOnboarding = do
           newSvName = "svX"
           newParticipantId = "svX-participant-id"
           reason = "because"
-          newSvRewardWeight = 19 -- we use 19 as a dummy reward weight for testing
+          newSvRewardWeight = 0 -- weight is not used for on-ledger sv right owners
```
meiersi-da, line 146:
```suggestion
          newSvRewardWeight = 0 -- reward weights on the operator state are zero when using on-ledger sv right owners
```

daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
```diff
@@ -129,39 +134,52 @@ initDecentralizedSynchronizerWithAmuletPrice isDevNet initialRound amuletPrice o
...
+        openMiningRoundCid = earliestOpenRoundCid
+        svOperator = sv1
   -- check that the SvStatusReports are present
   checkSvContractInvariants app
 
   pure (app, dso, (sv1, sv2, sv3, sv4))
 
 
-getSvInfoByParty : AmuletApp -> Party -> Script SvInfo
+getSvInfoByParty : AmuletApp -> Party -> Script (Text, SvInfo)
```
meiersi-da, line 179: Why do we now need to return that text? Also why not get rid of `getSvInfoByParty` in favor of `getSvInfoByOperatorParty`?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -152,6 +157,13 @@ data DsoSummary = DsoSummary with
       -- ^ The number of votes required for considering a confirmation, or a request for a vote
   deriving (Eq, Show)
 
+data DsoSummaryV2 = DsoSummaryV2
```
meiersi-da, line 161: Add deprecation note to the old summary type.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -456,7 +480,7 @@ data VoteRequestOutcome
 --
 -- | A vote cast by an SV.
 data Vote = Vote with
-    sv : Party -- ^ The SV party used to submit the vote.
+    sv : Party -- ^ The SV party used to submit the vote. Can be a right owner party or an operator party depending on the vote type.
```
meiersi-da, line 484: If that's the case should we require the two sets of parties to be distinct? Or do we rely on the vote type to know how to resolve the party?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -648,22 +707,23 @@ template DsoRules with
...
       with
-        sv : Party
+        sv : Party -- ^ sv operator party
         synchronizerId : Text
         newNodeConfig : SynchronizerNodeConfig
         nodeStateCid : ContractId SvNodeState
       controller sv
       do
+        let svOperator = sv
+        getAndValidateSvNodeOperatorParty this (Some svOperator)
```
meiersi-da, line 708: Side note: I find the large number of changes in `DsoRules` rather difficult to review. Should we split this ultimately into a change that only does the initial renames and then a follow up that adds the right owners?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
meiersi-da: Probably the latter... and it might not be that relevant, as the `voterName` is part of the map that stores the votes in the request.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -915,19 +971,20 @@ template DsoRules with
...
     nonconsuming choice DsoRules_UpdateAmuletPriceVote : DsoRules_UpdateAmuletPriceVoteResult
       with
-        sv : Party
+        sv : Party -- ^ sv operator party
         voteCid : ContractId AmuletPriceVote
         amuletPrice : Decimal
       controller sv
       do
-        pastVote <- fetchAndArchive (ForOwner with dso; owner = sv) voteCid
+        let svOperator = sv -- ^ for now we still tie amulet price votes to operators, it may make sense to change it to right owners eventually
```
meiersi-da, line 973: OK. Did we already validate with SteerCo that postponing this is OK?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1398,17 +1474,82 @@ template DsoRules with
                 svRewardState = newRewardStateCid
                 svRewardCoupons = couponCids
 
+    nonconsuming choice DsoRules_ReceiveSvRewardCouponV2 : ()
+      with
+        sv : Party -- ^ sv operator party
+        openRoundCid : ContractId OpenMiningRound
+        svRewardStates : TextMap.TextMap (ContractId SvRewardState)
+      controller sv
```
meiersi-da, line 1476:
```suggestion
      controller svOperator
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1398,17 +1474,82 @@ template DsoRules with
...
+      do
+        requireOnLedgerSvRightOwners this
+        let svOperator = sv
+        operatorName <- getAndValidateSvNodeOperatorParty this (Some svOperator)
```
meiersi-da, line 1480:
```suggestion
        operatorName <- getAndValidateSvOperatorParty this (Some svOperator)
```

potential streamlining opportunity

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1398,17 +1474,82 @@ template DsoRules with
...
+        let info = fromSome (TextMap.lookup operatorName (getSvRightOwners this))
+        now <- getTime
+        openRound <- fetchReferenceData (ForDso with dso) openRoundCid
+        require "OpenRound is open" (openRound.opensAt <= now)
+
+        -- FIXME: Do we want to enforce that the operator has to pass in all states that have configured them as a node operator?
+        -- given that you do rely on the operator anyway to mint rewards maybe not required.
+        -- if we do enforce it, we need to handle cases where the lastRoundCollected is out of sync which can happen on operator switches.
```
meiersi-da: I'm leaning to not require that given our discussion on SV staking.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1398,17 +1474,82 @@ template DsoRules with
...
+      with
+        sv : Party -- ^ sv operator party
+        openRoundCid : ContractId OpenMiningRound
+        svRewardStates : TextMap.TextMap (ContractId SvRewardState)
+      controller sv
+      do
+        requireOnLedgerSvRightOwners this
+        let svOperator = sv
+        operatorName <- getAndValidateSvNodeOperatorParty this (Some svOperator)
+        let info = fromSome (TextMap.lookup operatorName (getSvRightOwners this))
```
meiersi-da:
```suggestion
        let operatorInfo = fromSome (TextMap.lookup operatorName (getSvRightOwners this))
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1398,17 +1474,82 @@ template DsoRules with
...
+        now <- getTime
+        openRound <- fetchReferenceData (ForDso with dso) openRoundCid
+        require "OpenRound is open" (openRound.opensAt <= now)
+
+        -- FIXME: Do we want to enforce that the operator has to pass in all states that have configured them as a node operator?
+        -- given that you do rely on the operator anyway to mint rewards maybe not required.
+        -- if we do enforce it, we need to handle cases where the lastRoundCollected is out of sync which can happen on operator switches.
+        forA_ (TextMap.toList svRewardStates) $ \(svName, rewardStateCid) -> do
+          svInfo <- case TextMap.lookup svName (getSvRightOwners this) of
```
meiersi-da:
```suggestion
          rightOwnerInfo <- case TextMap.lookup svName (getSvRightOwners this) of
```

daml/splice-dso-governance/daml/Splice/DsoRules.daml
meiersi-da: do we even use that value?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
meiersi-da: same for `svName`

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1398,17 +1474,82 @@ template DsoRules with
...
+            None -> fail ("SV " <> svName <> " is not an SV")
+            Some svInfo -> do
+              require ("SV " <> svName <> " is minting rewards through " <> show operatorName) (svInfo.rewardNodeOperatorName == operatorName)
+              pure svInfo
+          rewardState <- fetchAndArchive (ForSv with dso; svName = svName) rewardStateCid
```
meiersi-da: consider defining `ForSvOperator` and `ForSvRightOwner`

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1398,17 +1474,82 @@ template DsoRules with
...
+          let state = rewardState.state
+              beneficiaries = svInfo.beneficiaries
+          let (configuredBeneficiaryCouponWeights, remainingWeight) =
+                foldl
+                  (\(acc, remainingWeight) (beneficiary, weight) ->
+                     let beneficiaryWeight = floor (weight * intToDecimal info.rewardWeight)
+                     in ((beneficiary, beneficiaryWeight) :: acc, remainingWeight - beneficiaryWeight))
+                  ([], info.rewardWeight)
+                  beneficiaries
```
meiersi-da, line 1498: that's rather dense code. Consider factoring it out into `SvRightOwner.daml` and adding a unit test for it.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1398,17 +1474,82 @@ template DsoRules with
...
+            ("Round " <> show openRound.round <> " is greater than the last round a reward has been received for " <> show state.lastRoundCollected)
+            (state.lastRoundCollected < openRound.round)
+
+          newRewardStateCid <- create rewardState with
+            state = RewardState with
+              lastRoundCollected = openRound.round
+              numRoundsCollected = state.numRoundsCollected + 1
+              numRoundsMissed =
+                state.numRoundsMissed + (openRound.round.number - state.lastRoundCollected.number - 1)
+              numCouponsIssued = state.numCouponsIssued + length beneficiaryCouponWeights
```
meiersi-da, line 1512: side note: I wonder how staking and beneficiaries should interact. It almost seems that the stakers would be the beneficiaries.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1398,17 +1474,82 @@ template DsoRules with
...
+
+          -- check weights and issue rewards
+          require "Sum of beneficiary weights matches" (info.rewardWeight == sum (map snd beneficiaryCouponWeights))
```
meiersi-da: why do we check the sum against the rewardWeight of the operator?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1398,17 +1474,82 @@ template DsoRules with
...
+
+          couponCids <- forA beneficiaryCouponWeights $ \(beneficiary, weight) ->
+            create SvRewardCoupon with
+              dso
+              sv = svInfo.rightOwnerParty
```
meiersi-da, line 1520: should we also record the operator that created the coupon?

meiersi-da (commented):
Great work @moritzkiefer-da! The overall structure looks good to me and we're now down to completing all design decisions and polishing the code and its test coverage.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1745,16 +1902,199 @@ template DsoRules with
...
+      do
+        requireOnLedgerSvRightOwners this
+        case TextMap.lookup name (getSvRightOwners this) of
+          None -> abort ("No sv with name " <> show name)
+          Some info -> do
+            require "right owner parties match" (info.rightOwnerParty == rightOwnerParty)
+            create this with
+              svRightOwners = Some (TextMap.insert name (info with beneficiaries = beneficiaries) (getSvRightOwners this))
+
+    choice DsoRules_UpdateSvRightOwnerInfo : ContractId DsoRules
```
meiersi-da: I suspect that we'd actually want the option to update multiple of these in one go, so that reweighting and the like can be done in a single vote, or not?

daml/splice-dso-governance/daml/Splice/DSO/SvRightOwner.daml
moritzkiefer-da: will do :+1:

daml/splice-dso-governance/daml/Splice/DSO/SvState.daml
moritzkiefer-da: will make another pass, I did do it on all choices but probably forgot some data types

daml/splice-dso-governance/daml/Splice/DsoBootstrap.daml
moritzkiefer-da: probably on DsoRules itself where we define the list is a good place, will add something

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: yeah brain fart, I think I've used sv node operator in most places which I think I like even a bit more than just sv operators as it makes it very obvious

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: good catch, leftover from a prior iteration

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: Yeah probably the better option.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: Realistically the way CIPs and dev fund grants are staged atm we'll first do the switch to on-ledger SV weights and do voted weights as a separate step (so this PR gets split into). Given that I actually would just keep it as is for now since the migration from on-ledger weights to weighted voting will look a bit different.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: good idea, will do

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: > Or do we rely on the vote type to know how to resolve the party?

yep that, I would deliberately not require them to be distinct. I think that's gonna make the migration much easier.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: not yet but the CIP also doesn't talk about it at all, will add a note somewhere to check.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: sounds like a good idea :+1:

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: So far we just handle it through separate votes with the same effectivity. I think batched votes make sense but I'm a bit hesitant to blow up this PR even further as it seems like a somewhat orthogonal change.

meiersi-da: I wasn't thinking of batching the vote. I was just thinking of passing in a `TextMap SvRightOwnerInfo` and apply the normal patching semantics.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
meiersi-da: Sounds good.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: Hm maybe that's easy enough, still a bit worried about the SV UI for that but I guess we can also keep it in the Daml code and initially only set it for a single right owner.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: sounds buggy will add some tests and fix

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: seems reasonable

daml/splice-dso-governance/daml/Splice/DSO/SvRightOwner.daml
```diff
@@ -0,0 +1,62 @@
...
+  with
+    rightOwnerParty : Party -- ^ The party representing the SV rights owner
+    voteWeight : Int
+      -- ^ the weight of an SV for voting, can be zero
+    rewardWeight : Int
+      -- ^ the weight of an SV for reward minting, can be zero
+    rewardNodeOperatorName : Text
+      -- ^ Name of the SV node operator through which rewards are minted.
+      -- For rewards to be minted both the node operator.
+    beneficiaries : [(Party, Decimal)]
```
meiersi-da, line 19: Can we make this a record, so later extra data about beneficiaries can be added too?

moritzkiefer-da: good idea :+1:

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1745,16 +1890,193 @@ template DsoRules with
...
-        _ <- getAndValidateSvParty this (Some sv)
+        let svOperator = sv
+        _ <- getAndValidateSvNodeOperatorParty this (Some svOperator)
         _ <- fetchAndArchive (ForDso dso) instructionCid
         _ <- exercise amuletRulesCid AmuletRules_BootstrapExternalPartyConfigState with
           openMiningRoundTriple
           expectedDso = dso
         pure DsoRules_BootstrapExternalPartyConfigStateResult
 
+    choice DsoRules_MigrateToOnLedgerSvRightOwners : ContractId DsoRules
```
drdo, line 1904: Should `SvRewardState` be created here for each new SV Right Owner?

Possibly also cleaning up the old `SVRewardState`s for the SV Nodes.

moritzkiefer-da: good catch that is missing indeed

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1745,16 +1890,193 @@ template DsoRules with
...
+        requireOnLedgerSvRightOwners this
+        svRightOwner <- fetchAndArchive (ForSv with dso; svName = name) svRightOwnerCid
+        require "right owner parties match" (svRightOwner.info.rightOwnerParty == rightOwnerParty)
+        create svRightOwner with
+          info.beneficiaries = beneficiaries
+
+    nonconsuming choice DsoRules_UpdateSvRightOwnerInfo : ContractId SvRightOwner
+      with
+        name : Text
+        rightOwnercid : ContractId SvRightOwner
```
drdo, line 1952: Capitalization

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -121,6 +123,10 @@ data DsoRules_ActionRequiringConfirmation
     -- ^ Create BootstrapExternalPartyConfigStateInstruction
   | SRARC_UpdateFeaturedAppRight DsoRules_UpdateFeaturedAppRight
     -- ^ Update a specific featured app right.
+  | SRARC_DsoRules_UpdateSvRightOwnerInfo DsoRules_UpdateSvRightOwnerInfo
```
drdo, line 126: Is there a reason why this doesn't go through an Instruction?
Couldn't we get into a dangling reference problem here as well?

moritzkiefer-da: good point makes sense to go through an instruction

daml/splice-dso-governance/daml/Splice/DsoBootstrap.daml
```diff
@@ -60,28 +64,29 @@ template DsoBootstrap with
...
               dsoDelegate = sv1Party
               config
               initialTrafficState
               isDevNet
+              svRightOwners = sv1RightOwnerName <&> \name -> [name]
+        let info = sv1RightOwnerName <&> \rightOwnerName -> SvRightOwnerInfo with
+                rightOwnerParty = sv1Party
+                voteWeight = fromOptional 0 sv1VoteWeight
+                rewardWeight = sv1RewardWeight
+                rewardNodeOperatorName = rightOwnerName
```
drdo, line 86: Should this be `sv1Name` rather than `rightOwnerName`?

moritzkiefer-da: I don't think so, sv1Name is the node operator name, sv1RightOwnerName/rightOwnerName is the right owner name that we set here.

drdo: Right, but this _is_ the node operator name.

moritzkiefer-da: oh I see right this is a typo

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1745,16 +1890,193 @@ template DsoRules with
...
+        if rightOwnerName `elem` getSvRightOwners this
+          then abort ("SV with name " <> rightOwnerName <> " already exists")
+          else
+            create AddSvRightOwnerInstruction with
+              dso
+              rightOwnerName
+              info
+              expiresAt = now `addRelTime` hours 1
+
+    choice DsoRules_ExecuteAddSvRightOwnerInstruction : ContractId DsoRules
```
drdo, line 1979: I believe this can create orphaned SV Right Owners (i.e. we don't check that the operator exists).

PR #6277, opened 2026-07-06
helm, info, status: Check recent events for Scan health

giner:
Checking recent events instead of snapshots makes monitoring more precise (5 minutes vs 4 hours).


isegall-da (approved):
Thanks

PR #6292, opened 2026-07-07
unbreak backports

krzysztofczyz-da:


.github/workflows/do_backports.yml
```diff
@@ -13,6 +13,10 @@ jobs:
     outputs:
       branches: ${{ steps.parse_backport_comments.outputs.backport_branches }}
     steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```
moritzkiefer-da, line 16: do we also need a checkout below before the uses: backport_workflow.yml?

moritzkiefer-da (approved):
thx

.github/workflows/do_backports.yml
krzysztofczyz-da: Nope workflows and actions are different in that regard

PR #6293, opened 2026-07-07
enable configuring multi-zone node pools for apps to help with google resource issues

mblaze-da:


canton-network-da:
# [backport] Reminder #
- [x] release-line-0.6.11
- [ ] release-line-0.6.10
- [ ] release-line-0.6.9
- [ ] release-line-0.6.8

martinflorian-da (approved):
Looks reasonable to me! Did you do a preview somehow to confirm it really sets the zones as expected?

cluster/configs/shared/base.yaml
```diff
@@ -10,13 +10,15 @@ cluster:
       # Should be set to a lower number (currently 8) on CI clusters that do neither of those.
       maxNodes: 20
       nodeType: n4d-standard-16
+      zones: '*'
```
nicu-da: what's the expected thing that we do for prod clusters? for those we must not deploy to any new zones as that would brick performance because of sql latency

mblaze-da: Good point, I'll make sure that the behavior for prod clusters remains unchanged.

martinflorian-da: Fair point that setting the default like this could be a bit dangerous... but I don't see a great alternative tbh.

For the configuration format itself it sounds reasonable to me to be explicit about which zones we want. We could implicitly assume that "none" means "zone a" or something like that, but that sounds a bit sketchy to me / trying to be too smart.

nicu-da: previously we were defauling to the env variables that define the zone, and if that wasn't defined then I think we left it out and default to the cluster default config (some clusters are zonal). any reason not to have the same behavior?

note that I am fine with the change as long as we already prep a PR for the internal change and check the diffs for expected to make sure we are compatible with the prod clusters

mblaze-da: I've changed it so that we don't alter the current behavior on prod clusters

nicu-da (approved):

mblaze-da:
> Looks reasonable to me! Did you do a preview somehow to confirm it really sets the zones as expected?

@martinflorian-da I run a preview locally and it looked fine

cluster/configs/shared/base.yaml
martinflorian-da: nice!

PR #6296, opened 2026-07-07
[ci] feat: change storage URL

krzysztofczyz-da:
Source cometbft drivers from public google bucket rather than artifactory

## Rationale

We want to remove all dependency on artifactory for the splcie repo, so all artifacts are pulled from ghcr.io or public locations


isegall-da:
@krzysztofczyz-da any reason this was not merged?

PR #6297, opened 2026-07-07
Source cometbft drivers from public google bucket rather than artifactory

dasormeter:
Source cometbft drivers from public google bucket rather than artifactory

## Rationale

We want to remove all dependency on artifactory for the splcie repo, so all artifacts are pulled from ghcr.io or public locations


dasormeter:
dupe - closed in favor of https://github.com/canton-network/splice/pull/6296

martinflorian-da:
@dasormeter You meant to close this PR, right?

PR #6299, opened 2026-07-07
Make postgres version configurable in Pulumi

OriolMunoz-da:
See https://daholdings.slack.com/archives/C03J6NCPSE4/p1783431509338959 for results on scratch. tl;dr: downtime of ~15m observed, failing to check with enterprise plus because of gcp zone being out of resources


nicu-da (approved):
Nice, thanks

PR #6357, opened 2026-07-09
Log providers with wrong vetting state

rautenrieth-da:
Ideally the list is empty most of the time, and there's some overlap with the metric for hidden coupons, but logging the list before exercising the choice helps with debugging.


meiersi-da (approved):
Thanks.

PR #6358, opened 2026-07-09
add a log ignore for transaction rejection in an unvetting race scenario

mblaze-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/9108


mblaze-da:
# [backport]
- [x] release-line-0.6.11

project/ignore-patterns/canton_before_shutdown.ignore.txt
nicu-da, line 1: it should be part of the canton_log.ignore.txt right?

mblaze-da: If the error in the workflow says
```
Found problems in log/canton_before_shutdown.clog
```
then doesn't that mean I need to match the `.clog` file name?

nicu-da: no, I think we strip the suffix, and also strip the before_shutdown from the log file names, and then find  a log ignore file based on that

nicu-da (approved):

PR #6390, opened 2026-07-13
helm, info: Make status work after tightening pod security

giner:
Pods run with user id 1001 by default now. Container images may not have a user with such id and thus no home directory. Replace HOME by /tmp to make prom2json work again.


isegall-da (approved):
tx

PR #6000, opened 2026-06-17
Traffic-based App Rewards: Switch on for default test network

adetokunbo:
Fixes #5871

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If an upgrade test is required, comment `/upgrade_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a logical synchronizer upgrade test is required (from canton-3.5), comment `/lsu_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/canton-network/splice/blob/main/docs/src/release_notes.rst).
- [x] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/canton-network/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [x] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #6392, opened 2026-07-13
Enable non zero round BFT bootstrap

### LINKED ISSUE #6110: Add an integration test that bootstraps the test network at a non-zero round

adetokunbo:
Fixes #6110 

This was stacked on #6000 which contained the fix for the title portion of #6110
The is PR addresses the extended portion mentioned in [this comment](https://github.com/canton-network/splice/issues/6110#issuecomment-4861523152)

#### Cluster Testing

#### PR Guidelines
- [x] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/canton-network/splice/blob/main/docs/src/release_notes.rst).

#### Merge Guidelines


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/NonZeroRoundBootstrapBftTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,240 @@
...
+  "SV triggers complete the reward pipeline for the initial round" in { implicit env =>
+    import definitions.GetRewardAccountingActivityTotalsResponse.members.RewardAccountingActivityTotalsOk
+
+    // Advance rounds so the reward pipeline runs for the initial round:
+    // SV1's scan computes reward totals from seeded activity data,
+    // and SVs confirm CalculateRewardsV2 using the root hash.
+    advanceTimeForRewardAutomationToRunForCurrentRound
+    actAndCheck(
+      "Advance to next round opening",
+      advanceRoundsToNextRoundOpening,
```
dfordivam, line 81: As we discussed [here](https://github.com/canton-network/splice/issues/6110#issuecomment-4861537808), the thing we want to test is whether other SVs can obtain the totals of the init round. So in this case `addDummySvWithFakeScanUrl` must happen before `advanceRoundsToNextRoundOpening`?

dfordivam: The comment above refers to the init round 
> We also need to handle scenario where only one SV has the data of the init round;

adetokunbo: Fixed, PTAL

adetokunbo:
@dfordivam PTAL

Note that the new test runs quickly, faster than `TrafficBasedRewardsTimeBasedIntegrationTest`

Although it specifically tests the behavior in the presence of an extra SV, isn't it also covering the basic BFT multi-SV functionality that that test currently checks?

If so, to reduce overall integration testing latency, (and thereby reduce the chance of future flakes), should this PR also switch `TrafficBasedRewardsTimeBasedIntegrationTest` back to being a single SV test?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/NonZeroRoundBootstrapBftTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,278 @@
...
+    // advancement. This raises BFT f from 0 to 1, so the initial
+    // round's reward pipeline runs under the higher quorum.
+    addDummySvWithFakeScanUrl()
+
+    // The dummy SV's participant doesn't exist on the sequencer, so
+    // this trigger would poll indefinitely logging "does not yet have
+    // a traffic state". Pause it to avoid log noise on CI.
+    env.svs.local.foreach(
+      _.dsoAutomation.trigger[SvOnboardingUnlimitedTrafficTrigger].pause().futureValue
+    )
```
rautenrieth-da: Shouldn't this be called before `addDummySvWithFakeScanUrl`? 

I assume the trigger is guaranteed to have finished setting up unlimited traffic for all existing SVs at the beginning of the test body (there's a `waitForSvToObtainUnlimitedTraffic` in the SV app initialization).

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/NonZeroRoundBootstrapBftTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,278 @@
...
+    )(
+      "dummy SV5 is in DsoRules and raises BFT f from 0 to 1",
+      _ => {
+        val info = sv1Backend.getDsoInfo()
+        info.dsoRules.payload.svs.size() shouldBe 5
+      },
+    )
+
+    // Pause ALL delegate-based triggers to prevent DsoRules contract
+    // churn during the SetSynchronizerNodeConfig submission.
```
rautenrieth-da: Why do we need to prevent churn?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1016,7 +1017,11 @@ class BftScanConnection(
       GetRewardAccountingActivityTotalsResponse(
         RewardAccountingActivityTotalsUndetermined(status = "Undetermined")
       )
-    val callConfig = BftCallConfig.default(scanList.scanConnections)
+    val callConfig =
+      if (initialRound.contains(roundNumber))
+        BftCallConfig.randomSingleCall(scanList.scanConnections)
```
rautenrieth-da: Shouldn't this select the scan of the founding sv instead of a random one? This will take a couple of retries until the random choice picks the one scan that can respond, no?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1064,7 +1069,11 @@ class BftScanConnection(
       GetRewardAccountingRootHashResponse(
         RewardAccountingRootHashUndetermined(status = "Undetermined")
       )
-    val callConfig = BftCallConfig.default(scanList.scanConnections)
+    val callConfig =
+      if (initialRound.contains(roundNumber))
+        BftCallConfig.randomSingleCall(scanList.scanConnections)
+      else
+        BftCallConfig.default(scanList.scanConnections)
```
rautenrieth-da: At one point, we discussed an alternative implementation:

1. Ask ALL scans for activity totals
2. Filter out responses that say "CannotProvide" (but keep all other responses, including errors)
3. Compute a BFT response from the remaining responses (with adjusted quorum)

We have implemented this for a different endpoint at https://github.com/canton-network/splice/blob/main/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala#L420.

This is does more network calls than your implementation. On the other hand, it transparently handles all situations where some scans report "CannotProvide" (e.g., if at some later point many new SVs join in the same round) and does not require the `initialRound` parameter.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/NonZeroRoundBootstrapBftTimeBasedIntegrationTest.scala
adetokunbo: I've removed this.  Prior to recent changes, there was a problem with the `DsoRules` contract being stale leading to lots of `CONTRACT_NOT_FOUND` errors and slower processing and test failures; upon retesting now, this no longer occurs.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
adetokunbo: This has been removed; the implementation has been re-designed as suggested in the next comment.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/NonZeroRoundBootstrapBftTimeBasedIntegrationTest.scala
adetokunbo: Done

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
adetokunbo: I think I've implemented this: PTAL

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1347,6 +1383,24 @@ object BftScanConnection {
     def randomSingleCall(connections: ScanConnections): BftCallConfig =
       default(connections).copy(requestsToDo = 1, targetSuccess = 1)
 
+    /** Config for the second phase of a probe-filter-consensus call.
+      * `n = withData.size` is the size of the filtered set; quorum
+      * scales with data availability. Callers must invoke with a
+      * non-empty `withData`; the `n<=0` guard exists so
+      * `enoughAvailableScans` rejects cleanly if that contract slips.
+      */
+    def forWithDataOnly(
```
rautenrieth-da: We already have `BftCallConfig.forAvailableData` for this purpose, please see if we can reuse it. See my suggested documentation of the method below.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1152,6 +1120,74 @@ class BftScanConnection(
...
+    * quorum for n=1; larger sets defer to `bftCallWithScanUris`
+    * with `BftCallConfig.forWithDataOnly`. Returns `undetermined`
+    * with no URIs when the set is empty or consensus fails.
+    *
+    * Note: the delegated `bftCallWithScanUris` retries on
+    * `ConsensusNotReached`. Because the `call` closure here is a
+    * cached lookup, every retry produces byte-identical responses —
+    * so a genuine disagreement re-runs `logDisagreements` a bounded
+    * number of times before ultimately failing. The extra WARNs are
+    * benign; consumers see the final Undetermined either way.
```
rautenrieth-da: Oof, good point! The extra WARNs are a bit annoying but I don't see how to easily avoid them, so let's leave it like that.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1347,6 +1383,24 @@ object BftScanConnection {
...
+    /** Config for the second phase of a probe-filter-consensus call.
+      * `n = withData.size` is the size of the filtered set; quorum
+      * scales with data availability. Callers must invoke with a
+      * non-empty `withData`; the `n<=0` guard exists so
+      * `enoughAvailableScans` rejects cleanly if that contract slips.
+      */
+    def forWithDataOnly(
+        withData: Seq[SingleScanConnection]
+    ): BftCallConfig = {
+      val n = withData.size
```
rautenrieth-da: At least for the purpose of determining `f`, `n` should include scans with indeterminate responses (where the call returned a failed future). Or in other words, assume that every scan that didn't respond at all could have responded with an OK response with disagreeing data.

Otherwise a temporary network outage (or a targeted attack) reduces the BFT guarantees.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1152,6 +1120,74 @@ class BftScanConnection(
...
+            case Some(data) => Future.successful(data)
+            case None => Future.failed(BftScanConnection.IgnoreResponse(scan.url))
+          },
+        endpoint = endpoint,
+        callConfig = BftCallConfig.forWithDataOnly(withData.keys.toSeq),
+        disagreementLogLevel = Level.WARN,
+      ).transformWith {
+        case Success((value, consensusUris)) =>
+          Future.successful((responseFor(value), consensusUris))
+        case Failure(_) => Future.successful((undetermined, Nil))
```
rautenrieth-da: This is silently converting a consensus mismatch into a successful "Undetermined" response. Usually we return the failed future on mismatches, which would result in a HTTP 502 Bad Gateway response.

I don't see anything wrong with returning "Undetermined", it's just more code and inconsistent with the other endpoints - was there a reason why we need this behavior?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1015,15 +1015,15 @@ class BftScanConnection(
...
-    * So simple equality comparison on responses is not possible, and we treat
-    * the two non-Ok responses as a "no response" by throwing IgnoreResponse so
-    * that this does not cause grouping in executeCall.
-    *
-    * And if no response could be obtained via bft we respond with 'Undetermined'
+  /** Bootstrap-safe reward-accounting read using two-phase
+    * probe-filter-consensus. A single-phase BFT read stalls at
+    * bootstrap because only sv1 has data (n=4 → targetSuccess=2, but
+    * only 1 Ok is possible). Instead we probe every open scan first,
+    * discard Undetermined / CannotProvide / network errors (probe
```
rautenrieth-da: This is suspicious. Conceptually, we should only remove peers that definitely said they cannot provide a response.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1347,6 +1383,24 @@ object BftScanConnection {
...
+      val n = withData.size
+      val f = if (n <= 0) 0 else (n - 1) / 3
+      BftCallConfig(
+        connections = withData,
+        requestsToDo = n,
+        targetSuccess = if (n <= 0) 0 else f + 1,
+      )
+    }
+
     def forAvailableData(
```
rautenrieth-da, line 1434:
```suggestion
    /** A configuration for a BFT call where some peers cannot provide a response and wish
      * to be excluded from the consensus. Only peers that successfully respond that they
      * cannot provide data are excluded, unavailable peers are treated as if responding with disagreeing data.
      *
      * A standard BFT call (see [[BftCallConfig.default]]) tolerates up to `f` Byzantine/faulty
      * peers out of `3f+1` total and thus requires `f+1` matching responses out of `2f+1` requests.
      *
      * This method keeps the same fault-tolerance intent as a standard call.
      * I.e., it still requires `f+1` matching responses, only decreasing that number if it
      * is lower than the number of peers that did not opt out of consensus.
      *
      * E.g., in a network with 4 peers (f=1), it will still require `f+1 = 2` matching responses,
      * unless only 1 peer has the data and all other peers have opted out.
      *
      * Note that keeping the same fault-tolerance intent means that as the number of peers with data decreases,
      * the BFT call becomes increasingly sensitive to failing or unavailable peers.
      * In the extreme case, if the number of peers with data is smaller than `f+1`, it will require
      * ALL peers to successfully respond with the same data.
      */
    def forAvailableData(
```

Unrelated to your PR, but this method could use a documentation.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1347,6 +1383,24 @@ object BftScanConnection {
...
+      * `n = withData.size` is the size of the filtered set; quorum
+      * scales with data availability. Callers must invoke with a
+      * non-empty `withData`; the `n<=0` guard exists so
+      * `enoughAvailableScans` rejects cleanly if that contract slips.
+      */
+    def forWithDataOnly(
+        withData: Seq[SingleScanConnection]
+    ): BftCallConfig = {
+      val n = withData.size
+      val f = if (n <= 0) 0 else (n - 1) / 3
```
rautenrieth-da:
```suggestion
      val f = (n - 1) / 3
```

`(0 - 1) / 3 = 0`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1347,6 +1383,24 @@ object BftScanConnection {
...
+      BftCallConfig(
+        connections = withData,
+        requestsToDo = n,
+        targetSuccess = if (n <= 0) 0 else f + 1,
```
rautenrieth-da:
```suggestion
        targetSuccess = f + 1,
```
This will still have `enoughAvailableScans == false`

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/NonZeroRoundBootstrapBftTimeBasedIntegrationTest.scala
```diff
@@ -0,0 +1,265 @@
...
+
+/** Tests that SV automation triggers can complete the reward
+  * pipeline for the initial round when bootstrapping TBAR at a
+  * non-zero round with BFT f >= 1.
+  *
+  * Adds a dummy 5th SV to increase the BFT quorum from 1 to 2.
+  */
+class NonZeroRoundBootstrapBftTimeBasedIntegrationTest
+    extends IntegrationTestWithIsolatedEnvironment
+    with HasExecutionContext
```
rautenrieth-da, line 39: This suite looks good, but also expensive to run and to maintain. Do you think we would have enough test coverage if we replaced this whole suite by the following changes?
- In `TrafficBasedRewardsSvAppTimeBasedIntegrationTest`, assert that scans return the correct Ok/CannotProvide responses for the initial round
- In `BftScanConnectionTest`, test all kinds of situations and edge cases with the modified BFT behavior

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
dfordivam: Agree, I believe this is what we discussed, to only remove the `CannotProvide`

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/NonZeroRoundBootstrapBftTimeBasedIntegrationTest.scala
dfordivam: note that `TrafficBasedRewardsSvAppTimeBasedIntegrationTest` does not bootstrap with TBAR enabled. So having this is probably the easiest way forward. The other two integ tests which bootstrap with TBAR are 1 SV right now

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
adetokunbo: no - I've reverted to using the Failed Future

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
adetokunbo: As suggested, this now reuses `forAvailableData` — `forWithDataOnly` is a wrapper that supplies `unavailableCount` and overrides `requestsToDo` to sample all withData peers.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
adetokunbo: Agreed — the cached retries produce byte-identical responses, so the extra WARNs are benign.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
adetokunbo: Fixed: Includes unavailable peers in `n` — `forWithDataOnly(withData, unavailable)` computes `n = withData.size + unavailable`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
adetokunbo: Done: now drops only `CannotProvide` — the probe returns a three-state `ProbeVerdict` (WithData / WithoutData / Unavailable); WithoutData is dropped from `n`, Unavailable is kept.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
adetokunbo: Done: now propagates `HttpErrorWithHttpCode(BadGateway)` on consensus failure, matching `getMigrationInfo`; `Undetermined` is returned only when the probed set is empty.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/NonZeroRoundBootstrapBftTimeBasedIntegrationTest.scala
adetokunbo: As Divam points out, no other test exercises the f≥1 quorum path at bootstrap so is likely needed; the PR updates `BftScanConnectionTest` to cover more edge-cases, am happy to add any others that are necessary

adetokunbo:
@rautenrieth-da PTAL, I've addressed the comments, and also removed the ignore patterns that were needed because scan bft warnings affected other integration tests

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1350,20 +1402,65 @@ object BftScanConnection {
...
+      * ALL peers to successfully respond with the same data.
+      *
+      * `unavailableCount` names peers known to be unavailable outside of `connections.failed`
+      * (e.g. classified as such by an earlier probe round). They count toward `n` — and thus
+      * toward `f` — as possibly-disagreeing peers, but are not sampled or expected to respond.
+      */
     def forAvailableData(
         connections: ScanConnections,
         dataAvailable: SingleScanConnection => Boolean,
+        unavailableCount: Int = 0,
```
rautenrieth-da: Isn't this the same as
```scala
connections.open.count(c=>!dataAvailable(c))
```
?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1350,20 +1402,65 @@ object BftScanConnection {
...
     )(implicit loggingContext: ErrorLoggingContext): BftCallConfig = {
-      val f = connections.f
+      val n = connections.open.size + connections.failed + unavailableCount
```
rautenrieth-da: This method is really hard to review. Depending on the call site, terms like "connections.open", "connections.failed", "with data", and "unavailable" have a different meaning, and we don't have precise definitions for them.

The other call sites also do not have the notion of "data will never be available for this scan", and they use different endpoints for determining what scans wish to opt out and for actually querying data. 

Even though I could not find any actual bug, I don't feel comfortable with the state of the code.

I think I reusing this method was not a good idea after all, sorry about this mess. What do you think of keeping `forAvailableData` as it was before (maybe just include the comment block) and adding a separate function like you had before?

To think through the edge cases, I did end up writing a sketch of the implementation myself. I have pushed it to https://github.com/canton-network/splice/pull/7115/changes, if you're interested.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1160,6 +1135,83 @@ class BftScanConnection(
...
+    if (withData.isEmpty) Future.successful((undetermined, Nil))
+    else if (withData.sizeIs == 1 && unavailableCount == 0) {
+      val (scan, value) = withData.iterator.next()
+      Future.successful((responseFor(value), List(scan.url)))
+    } else
+      bftCallWithScanUris[Value](
+        call = scan =>
+          withData.get(scan) match {
+            case Some(data) => Future.successful(data)
+            case None => Future.failed(BftScanConnection.IgnoreResponse(scan.url))
```
rautenrieth-da: We will never hit this, right? This callback will only be called for connections in `withData`.

I would have expected that we delete `BftScanConnection.IgnoreResponse` now. It was used for abstaining from voting, which we have replaced by removing the abstaining connections from the pool.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1160,6 +1135,83 @@ class BftScanConnection(
...
+      undetermined: Response,
+      responseFor: Value => Response,
+  )(implicit
+      ec: ExecutionContext,
+      tc: TraceContext,
+  ): Future[(Response, List[Uri])] = {
+    val withData = probe.withData
+    val unavailableCount = probe.unavailable.size + scanList.scanConnections.failed
+    if (withData.isEmpty) Future.successful((undetermined, Nil))
+    else if (withData.sizeIs == 1 && unavailableCount == 0) {
```
rautenrieth-da: Do we need these special cases? This version is obviously much more performant, but we also lose some logging/metrics. More special cases also make code maintenance more difficult. I would have expected that `bftCallWithScanUris` correctly handles cases where we have 0 or 1 connections.

For the empty `withData`, what do you think of not returning `undetermined` and instead letting the bft call fail with some "not enough scans available" error? This would have no effect on `CalculateRewardsTrigger`, a HTTP 200 response with Undetermined and a HTTP 502 response will both cause the trigger task to fail in the same way.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
rautenrieth-da: Thinking more about this: warnings in production are bad, so we should try our best to not warn in situations cased by random network issues.

Since we do not retry at all, this method will be more vulnerable to such issues. For example, if the network is down during the first stage, the whole call will fail with something like `Only 0 scan instances can be used (out of 4 configured ones)`.

Can we keep track of what peers returned Unavailable in the first phase, and include them in the list of scan connections to use for the actual bft call?

I'm not fully sure how this will play out in practice. `bftCall()` retries up to 12 times, so it will "only" cover an unavailability period of 10-30sec, but that is still better than no retries at all.

rautenrieth-da (commented):
Thanks! I didn't realize how much work the BFT changes are. The whole class was initially made for dumb BFT calls. Special treatment for eventually consistent endpoints or endpoints where some peers will never be able to provide data have been added later and it has become difficult to reason about.

As noted in https://github.com/canton-network/splice/pull/6392/changes#r3915250415, I think we should not reuse the existing implementation after all.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
adetokunbo: Thanks! - #7115 makes a lot of sense, so I've followed its shape: `forAvailableData` is untouched, the two-phase logic goes through `ConnectionsWithProbeVerdicts` + `BftCallConfig.forOptionalResponses` via a shared `consensusOverProbe` helper

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
adetokunbo: Done: dropped both short-circuits — empty returns BadGateway (matching `getMigrationInfo`) and metrics fire uniformly.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
adetokunbo: Done: this now re-queries unavailable peers in phase 2 so `bftCallWithScanUris`'s retry loop can catch peers that recover during a transient outage.

adetokunbo: Actually, this did not work as intended, and I had to revert because various integration tests failed because an Exception escaped.  Looking into an alternative way of implementing this now.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1091,42 +1099,78 @@ class BftScanConnection(
...
+          case BftScanConnection.ProbeVerdict.WithData(cached) => Future.successful(cached)
+          case _ => liveCall(scan)
+        },
+      endpoint = endpoint,
+      callConfig = callConfig,
+      disagreementLogLevel = Level.WARN,
+      // "Not enough scans" here means the probe didn't gather enough cached
+      // responses to satisfy BFT quorum — benign during bootstrap or transient
+      // peer unavailability. The caller sees HTTP 502 and retries.
+      notEnoughScansLogLevel = Level.INFO,
```
rautenrieth-da: I don't understand this comment. "not enough scans" can only happen if every single scan opted out (i.e. returned CannotProvide), no? Why do we need use a different log level here?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
adetokunbo: Unfortunately, so far I have not got this to work without additional complexity — the retry-through-Unavailable approach kept failing CI in new ways after each fix. So sticking with cache-only phase 2, and addressing the alerting concern at the source: `notEnoughScansLogLevel = Level.INFO` on the reward-accounting reads keeps the "Only X scan instances" message from tripping prod.

adetokunbo:
@rautenrieth-da I've addressed all the earlier comments except one (see comment above).  PTAL

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
rautenrieth-da: Answering my own question: In `BftCallConfig.forOptionalResponses`, connections that returned "don't know yet" are removed from the list of connections to use, but still contribute to the computation of `n` and `f`.

rautenrieth-da (approved):
Thanks! The retry behavior is not ideal - we only ask each scan once, so the inner retry loop won't get us out of situations where enough scans initially return "don't know yet".

That's fine for the TBAR endpoints, as we have infinite retries (without warnings) in the trigger. We can improve the retry behavior in the future.

PR #6419, opened 2026-07-15
Fix performance readme

jagathweerasinghe-da:
[static]

In ScanStore, we ingest both ACS + TxLog: 
https://github.com/canton-network/splice/blob/b38cd6edcc544cd01e245f00880160de4c94eeba/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/DbMultiDomainAcsStore.scala#L1696


OriolMunoz-da (approved):
ty

PR #6427, opened 2026-07-15
WIP: cip 0105 externally implemented governance locks

jonored:
WIP code for possible direction for CIP-0105 locking implementation using the existing AmuletAllocationV2 unmodified, for reference only. Do not merge.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,135 @@
...
+
+
+-- | The intention is to treat the overall aggregate lock for a SV as if it is
+-- "the settlement" for the allocations locked to it, and use iterated
+-- settlement to execute updates on the committed allocations as needed.
+-- This approach would permit most of the specific governance locking and
+-- vesting to be in a separate dar from amulet only needed for SVs and
+-- interested observers.
+
+template AggregateLock
```
moritzkiefer-da: unclear to me why this is a separate template, ideally this would be done through the withdraw choice. If that is not an option, it seems like this could just be a choice on ExternalPartyAmuletRules

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,135 @@
...
+        controller authorizers
+        do
+          locked <- view <$> fetch lockedCid
+          withdrawTo <- view <$> fetch withdrawToCid
+          ownerParty <- whenNone locked.allocation.authorizer.owner $
+            assertFail "The requirement 'governance locked allocations must be owned by basic accounts' was not met"
+          let GovernanceLockedControllers{..} = governanceLockedControllersFromMeta locked.meta ownerParty
+          require "Unlock controller must be one of the options" $ any (\conj -> all (`elem` authorizers) conj) unlockControllerSets
+          require "Not allowed to unlock to a different party than locked the funds" $ locked.allocation.authorizer == withdrawTo.allocation.authorizer
+          if not allowImmediateUnlock then require "Unlock without vesting is only allowed when specifically enabled" False else pure ()
```
moritzkiefer-da: I don't understand what this field is for. When do you want unlock without vesting?

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,135 @@
...
+      with
+        factoryCid : ContractId SettlementFactory
+        lockedCid : ContractId Allocation
+        withdrawToCid : ContractId Allocation
+        authorizers : [ Party ]
+        amount : Decimal
+      where
+        controller authorizers
+        do
+          locked <- view <$> fetch lockedCid
```
moritzkiefer-da: use the various checked fetch variants, CI is also gonna fail on this

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,135 @@
...
+isValidAggregateLockedAllocation dso alloc =
+  and [ alloc.allocation.admin == dso
+      , alloc.allocation.committed
+      , null alloc.allocation.transferLegSides
+      , TM.lookup "cip-105/type" alloc.allocation.meta.values == Some "aggregateLock"
+      , alloc.expiresAt == Some maxBound
+      ]
+
+isValidVestingLockedDestination : Party -> V2.AllocationView -> Bool
+isValidVestingLockedDestination dso alloc =
```
moritzkiefer-da, line 34: this seems to be the same validation as isValidAggregateLockedAllocation?

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,135 @@
...
+                  extraTransferLegSides =
+                    [ TransferLegSide with
+                      transferLegId
+                      side = SenderSide
+                      otherside = withdrawTo.allocation.authorizer
+                      amount
+                      instrumentId
+                      meta = emptyMetadata
+                    ]
+                  nextIterationFunding = Some $ TM.singleton instrumentId newLockedAmount
```
moritzkiefer-da: The math doesn't add up for me. `locked` locks up `currentLockedAmount`. You are now trying to unlock `amount`. So `newLockedAmount` is currentLockedAmount - amount. But then you are trying to create two allocations with nextIterationFunding for that amount?

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,135 @@
...
+    dso: Party
+    instrumentId : Text
+    allowImmediateUnlock : Bool
+  where
+    signatory dso
+    nonconsuming choice AggregateLock_Unlock : SettlementFactory_SettleBatchResult
+      with
+        factoryCid : ContractId SettlementFactory
+        lockedCid : ContractId Allocation
+        withdrawToCid : ContractId Allocation
```
moritzkiefer-da: Can you add some comments on how you expect the two allocations to be used? it's not clear to me. Or even better add the test script.

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,135 @@
...
+  and [ alloc.allocation.admin == dso
+      , alloc.allocation.committed
+      , null alloc.allocation.transferLegSides
+      , TM.lookup "cip-105/type" alloc.allocation.meta.values == Some "aggregateLock"
+      , alloc.expiresAt == Some maxBound
+      ]
+
+type ControllerSets = [[Party]]
+
+partiesFromText : Text -> Optional [Party]
```
moritzkiefer-da, line 48: there are a bunch of utilities already e.g. `parseCommaSeparated`, let's reuse those instead of reimplementing half of them.

daml/splice-amulet/daml/Splice/AggregateLock.daml
cgibbard: In (https://docs.google.com/document/d/1VGtgNSNHrHnSLggqxlvhCTFrMPgaeCs2OLEJ8Fq9Ar4/edit?tab=t.0) under the Aggregated Locking Requirements, FR-25 says "It must be possible to require that locked coins only be unlocked through the vesting process." which is phrased in such a way as to suggest that this requirement is an optional possibility.

FR-26 says "It must be possible to permit immediate unlocks of all coins locked to an Aggregate Lock, to support Featured App non-approval; when in this state the vesting process is not required."

I might be misinterpreting how an implementation of that would look (just getting spun up on the details of this project still), but this might be related.

jonored: When you've tried to set up a Featured App, locked the required amulet to satisfy the threshold requirement, and then failed to actually become a Featured App at this time. CIP-0116 says that "if an application is not approved, CC may be moved immediately", and I am defaulting to "implement things in the CIP on-chain if reasoable". No objection to specifying a different process for handling that scenario.

daml/splice-amulet/daml/Splice/AggregateLock.daml
jonored: Should have a different type field and should also be checking that a schedule is defined, but it should definitely share the common pieces.

daml/splice-amulet/daml/Splice/AggregateLock.daml
moritzkiefer-da: ok that makes sense to handle. But your implementation only seems to handle the case where `allowImmediateUnlock = True` and otherwise just fail in the require instead of doing vesting?

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,232 @@
...
+    nonconsuming choice VestingLock_Withdraw : () -- TODO: Return type possibly AllocationResult
+      with
+        factoryCid : ContractId SettlementFactory
+        authorizers : [Party]
+        vestingLockCid : ContractId Allocation
+      where
+        controller authorizers
+        do
+          -- remainingLocked stays in nextIterationFunding
+          vestingLock <- view <$> fetch vestingLockCid
```
moritzkiefer-da: no naked fetches

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,232 @@
...
+          now <- getTime
+
+          -- TODO: Probably add valid unlockController set check?
+          -- TODO: Potentially do case statement with nextIterationFunding optionals
+          require "There are no remaining funds to withdraw"
```
moritzkiefer-da: assertion text is the wrong way around, you should state the property that should hold. so "there are remaining funds to withdraw"

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,232 @@
...
+              ]
+    -- Choice temporarily lives here and it will be moved, testing/draft purposes atm
+    -- Note: this whole choice is WIP
+    nonconsuming choice VestingLock_Withdraw : () -- TODO: Return type possibly AllocationResult
+      with
+        factoryCid : ContractId SettlementFactory
+        authorizers : [Party]
+        vestingLockCid : ContractId Allocation
+      where
+        controller authorizers
```
moritzkiefer-da: there are zero checks on authorizers

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,232 @@
...
+        do
+          locked <- view <$> fetch lockedCid
+          withdrawTo <- view <$> fetch withdrawToCid
+          ownerParty <- whenNone locked.allocation.authorizer.owner $
+            assertFail "The requirement 'governance locked allocations must be owned by basic accounts' was not met"
+          let GovernanceLockedControllers{..} = governanceLockedControllersFromMeta locked.meta ownerParty
+          require "Unlock controller must be one of the options" $ any (\conj -> all (`elem` authorizers) conj) unlockControllerSets
+          require "Not allowed to unlock to a different party than locked the funds" $ locked.allocation.authorizer == withdrawTo.allocation.authorizer
+          if not allowImmediateUnlock then require "Unlock without vesting is only allowed when specifically enabled" False else pure ()
```
moritzkiefer-da:
```suggestion
          require "Unlock without vesting is only allowed when specifically enabled" allowImmediateUnlock
```

token-standard/examples/splice-test-token-v2-test/daml/Splice/Testing/Tokens/TestTokenV2/TestAllocationHappyV2.daml
```diff
@@ -45,6 +47,31 @@ test_allocation_WithdrawV2 = test_allocation_withdrawCancel_generic $ \TestEnv {
         meta = emptyMetadata
   pure ()
 
+-- This is a temporary test script
+test_vesting_lock_WithdrawV2 : Script ()
+test_vesting_lock_WithdrawV2 = do
```
moritzkiefer-da: not testing anything, let's please focus on adding at least some very basic testing before doing any more code changes.

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
```diff
@@ -0,0 +1,109 @@
...
+
+testAggregateLockHappyPath : Script ()
+testAggregateLockHappyPath = do
+  env@TestEnv{..} <- setupTest
+  now <- getTime
+  let dso = env.instrId.admin
+
+  AmuletRegistryV2.tapFaucet registriesEnv.amuletV2 bob 1800.0
+  
+  aggLock <- submit (actAs [alice, dso]) $ createCmd AggregateLock with
```
moritzkiefer-da: you can't assume that you can submit with both alice and dso as actAs. Getting authorization right is a crucial part in Daml models so let's please not "cheat" here

moritzkiefer-da: in this case I also don't understand why you even need actAs alice, the only signatory seems to be DSO.

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
```diff
@@ -0,0 +1,109 @@
...
+lockForAggregate te amount = lockForGovernance te (TM.fromList [(te.instrId.id, amount)]) $ Metadata $ TM.fromList
+  [ ("cip-105/type", "aggregateLock") ]
+
+lockForVesting : TestEnv -> Party -> ContractId AggregateLock -> Script AllocationInstructionResult
+lockForVesting te = lockForGovernance te TM.empty $ Metadata $ TM.fromList
+  [ ("cip-105/type", "vestingLock")
+  , ("cip-105/vestingPeriod", "365") ]
+
+testAggregateLockHappyPath : Script ()
+testAggregateLockHappyPath = do
```
moritzkiefer-da: did you run the test? it fails with
```
  Failed with status: UNHANDLED_EXCEPTION/DA.Exception.AssertionFailed:AssertionFailed: The requirement 'Must be a valid empty vesting unlock allocation' was not met.
```

token-standard/splice-token-standard-utils/daml/Splice/TokenStandard/Utils/Internal/Conversions.daml
```diff
@@ -27,6 +27,9 @@ module Splice.TokenStandard.Utils.Internal.Conversions (
   partiesToMeta,
   dropMeta,
   validateNoMeta,
+  -- TODO: Is exporting the below cool?
```
moritzkiefer-da: yes that's ok

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
```diff
@@ -0,0 +1,166 @@
...
+  aggLock <- submit (actAs dso) $ createCmd AggregateLock with
+    dso
+    instrumentId = env.instrId.id
+    vestingSchedule = VestingSchedule_SV
+  Some aggLockDisclosure <- queryDisclosure dso aggLock
+
+  -- Lock some funds for the aggregate
+  AllocationInstructionResult { output = AllocationInstructionResult_Completed locked } <- lockForAggregate env 1000.0 bob aggLock
+
+  -- Check total for the aggregate; needs to be implemented.
```
moritzkiefer-da: let's actually add that

daml/splice-amulet/daml/Splice/AggregateLock.daml
moritzkiefer-da: comment still applies

daml/splice-amulet/daml/Splice/AggregateLock.daml
moritzkiefer-da: comment still applies

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+    totalUnlockPeriod : Decimal = intToDecimal . convertRelTimeToMicroseconds $ subTime endDate startDate
+    unlockPeriodElapsed : Decimal = intToDecimal . convertRelTimeToMicroseconds $ subTime currentDateTime startDate
+    totalVestedRatioElapsed : Decimal = unlockPeriodElapsed / totalUnlockPeriod
+    withdrawnAmount = initialAmount - nextIterationFundAmount
+
+-- | The intention is to treat the overall aggregate lock for a SV as if it is
+-- "the settlement" for the allocations locked to it, and use iterated
+-- settlement to execute updates on the committed allocations as needed.
+-- This approach would permit most of the specific governance locking and
+-- vesting to be in a separate dar from amulet only needed for SVs and
```
moritzkiefer-da: separate dar comment doesn't make much sense imho

daml/splice-amulet/daml/Splice/AggregateLock.daml
moritzkiefer-da: comment still applies, add some comments on what the allocations are supposed to represent

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+-- interested observers.
+
+template AggregateLock
+  with
+    dso: Party
+    instrumentId : Text
+    vestingSchedule : VestingSchedule
+  where
+    signatory dso
+    nonconsuming choice AggregateLock_Unlock : SettlementFactory_SettleBatchResult
```
moritzkiefer-da: we are still missing the step to turn this into something the SV has to explicitly accept

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+  | VestingSchedule_Immediate
+  deriving (Show, Eq, Ord)
+
+validateVestingSchedule : VestingSchedule -> Time -> Metadata -> Bool
+validateVestingSchedule VestingSchedule_SV now meta =
+  let
+    start = decodeTime $ fromSomeNote "start must exist" (TM.lookup "cip-105/vestingLock.startDate" meta.values)
+    end = decodeTime $ fromSomeNote "end must exist" (TM.lookup "cip-105/vestingLock.endDate" meta.values)
+  in
+    now < start && end `subTime` start == days 365 + hours 6
```
moritzkiefer-da: why 6 hours?

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+governanceLockedControllersFromMeta : Metadata -> Party -> GovernanceLockedControllers
+governanceLockedControllersFromMeta meta acctParty = GovernanceLockedControllers with
+    unlockControllerSets = controllerSetFromMetaMap "cip-105/unlockControllerSets"
+    substituteControllerSets = controllerSetFromMetaMap "cip-105/substituteControllerSets"
+  where
+    controllerSetFromMetaMap key = fromOptional [[acctParty]] $ controllerSetFromMeta =<< key `TM.lookup` meta.values
+
+data VestingSchedule
+  = VestingSchedule_SV
+  | VestingSchedule_Immediate
```
moritzkiefer-da, line 90: not actually used anywhere

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+import Splice.Util
+
+import DA.Optional
+import DA.Text
+import qualified DA.List as L
+import qualified DA.TextMap as TM
+import DA.Time
+
+isValidAggregateLockedAllocation : Party -> V2.AllocationView -> Bool
+isValidAggregateLockedAllocation dso alloc =
```
moritzkiefer-da, line 24: probably want to validate executors as well. And that something is actually locked up.

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+          withdrawTo <- view <$> fetchCheckedInterface (ForDso with dso) withdrawToCid
+          ownerParty <- whenNone locked.allocation.authorizer.owner $
+            assertFail "The requirement 'governance locked allocations must be owned by basic accounts' was not met"
+          let GovernanceLockedControllers{..} = governanceLockedControllersFromMeta locked.meta ownerParty
+          require "Unlock controller must be one of the options" $ any (\conj -> all (`elem` authorizers) conj) unlockControllerSets
+          require "Not allowed to unlock to a different party than locked the funds" $ locked.allocation.authorizer == withdrawTo.allocation.authorizer
+          require "Must be a valid governance-locked allocation" $ isValidAggregateLockedAllocation dso locked
+          require "Must be a valid empty vesting unlock allocation" $ isValidVestingLockedDestination dso withdrawTo
+          require "Must have amounts reserved within the next settlement iteration" $ isSome locked.allocation.nextIterationFunding
+          now <- getTime
```
moritzkiefer-da: use the assertDeadlineExceeded and friend assertions here instead of getTime

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+          require "Vesting allocation must have a correct schedule configuration" $ validateVestingSchedule vestingSchedule now withdrawTo.allocation.meta
```
moritzkiefer-da: it's not really clear why we need the user to create the vesting allocation first instead of creating it in this choice on the fly. In the end the user shouldn't be able to determine any of the parameters in there.

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+                      side = ReceiverSide
+                      otherside = locked.allocation.authorizer
+                      amount
+                      instrumentId
+                      meta = emptyMetadata
+                    ]
+                  nextIterationFunding = Some $ TM.singleton instrumentId amount
+              ]
+    -- Choice temporarily lives here and it will be moved, testing/draft purposes atm
+    -- Note: this whole choice is WIP
```
moritzkiefer-da: please add comments for things you expect to change, it's hard to review anything properly otherwise

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+        authorizers : [Party]
+        vestingLockCid : ContractId Allocation
+        extraArgs : ExtraArgs
+      where
+        controller authorizers
+        do
+          vestingLock <- view <$> fetchCheckedInterface (ForDso with dso) vestingLockCid
+          now <- getTime
+          ownerParty <- whenNone vestingLock.allocation.authorizer.owner $
+            assertFail "The requirement 'governance locked allocations must be owned by basic accounts' was not met"
```
moritzkiefer-da: is that not always enforced for AmuletAllocation in the `ensure` clause?

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+
+          require "Must be a valid vesting unlock allocation" $ isValidVestingLockedDestination dso vestingLock
+          require "There must be an initialAmount set for the vestingLock"
+            $ TM.member "cip-105/vestingLock.initialAmount" vestingLock.allocation.meta.values
+          let GovernanceLockedControllers{..} = governanceLockedControllersFromMeta vestingLock.meta ownerParty
+          require "Unlock controller must be one of the options" $ any (\conj -> all (`elem` authorizers) conj) unlockControllerSets
```
moritzkiefer-da: that still seems to allow for additional authorizers that are not in your unlock controller set, why?

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+
+          let mNextIterationFunding = case vestingLock.allocation.nextIterationFunding of
+                None -> None
+                Some nextIterationFundingTM ->
+                  TM.lookup instrumentId nextIterationFundingTM
+          require ("VestingLock's nextIterationFunding should have remaining funds to withdraw for instrumentId " <> show instrumentId)
+            $ isSome mNextIterationFunding
+          let nextIterationFundingAmount = fromSome mNextIterationFunding
```
moritzkiefer-da:
```suggestion
          nextIterationFundingAmount = case TM.lookup instrumentId =<<(vestingLock.allocation.nextIterationFunding) of
                None -> abort "VestingLock's nextIterationFunding should have remaining funds to withdraw for instrumentId " <> show instrumentId
                Some amount -> pure amount
```
first doing an isNone check and then using fromSome is always a code smell

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+                      amount = availableWithdrawAmount
+                      instrumentId
+                      meta = emptyMetadata
+                    ]
+                  nextIterationFunding = nextIterationFunding
+              ]
+
+          -- We should have one result from settleBatch
+          require "VestingUnlock_Withdraw should result in one allocation result. " $
+            L.length batchResults.allocationSettleResults == 1
```
moritzkiefer-da: similar as above with the isNone + fromSome, don't first do an assertion and then a partial match but instead just pattern match directly

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
```diff
@@ -0,0 +1,166 @@
...
+    amount = initialAmountToUnlock
+    extraArgs = enriched.arg.extraArgs
+
+  allocations <- query @AmuletAllocationV2 bob
+
+  debug allocations
+
+  assert $ length allocations == 2
+
+  -- Need to check other than via inspection that the two allocations actually exist with correct values.
```
moritzkiefer-da: let's add it

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
```diff
@@ -0,0 +1,166 @@
...
+  Some aggLockDisclosure <- queryDisclosure dso aggLock
+
+  -- Lock some funds for the aggregate
+  AllocationInstructionResult { output = AllocationInstructionResult_Completed locked } <- lockForAggregate env 1000.0 bob aggLock
+
+  -- Check total for the aggregate; needs to be implemented.
+ 
+  now <- getTime
+  let initialAmountToUnlock = 365.25
+  AllocationInstructionResult { output = AllocationInstructionResult_Completed vestingInput } <- lockForVesting initialAmountToUnlock (addRelTime now $ minutes 5) env bob aggLock
```
moritzkiefer-da: why + 5 min

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
```diff
@@ -0,0 +1,166 @@
...
+
+  -- Time needs to pass to be able to withdraw from the VestingLock
+  passTime (days 10)
+
+  (AllocationResult vestingAllocResult tmHoldingCids _) <- submit (actAs bob <> disclose aggLockDisclosure <> discloseMany' enriched.disclosures ) $ exerciseCmd aggLock $ VestingLock_Withdraw with
+    factoryCid = enriched.factoryCid
+    authorizers = [ bob ]
+    vestingLockCid = toInterfaceContractId vestingLockAllocationCid
+    extraArgs = enriched.arg.extraArgs
+  debug vestingAllocResult
```
moritzkiefer-da: let's remove all these random debugs

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
```diff
@@ -0,0 +1,166 @@
...
+
+  let holdingCids = fromSome $ TM.lookup env.instrId.id tmHoldingCids
+  holdings <- queryInterface @Holding bob
+  -- The holdings would include both locked/unlocked, so we filter
+  let [(_, (Some unlockedHolding))] = filter (\(hcid, (Some h)) -> isNone h.lock && hcid `elem` holdingCids) holdings
+  debug unlockedHolding
+
+  let tolerance = 0.01
```
moritzkiefer-da: what's the tolerance for? I don't see why you can't do an exact computation in your test

daml/splice-amulet/daml/Splice/AggregateLock.daml
cgibbard: A year is 365.25 days

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
cgibbard: There was a condition in validateVestingSchedule which said now < start which would not be satisfied unless there was a strict time difference there. Unless there's a good reason, that maybe ought to be now <= start, and I don't see any reason in our design doc, but perhaps @jonored knows better?

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
cgibbard: This tolerance is because the actual number is extremely close to, but not exactly 10. However, it might be best just to copy the computed number into the test, given that Decimal is fixed point, an exact equality comparison is not as bad as it might otherwise be for floating point. To get a nice round number of tokens here, we'd need to very precisely compute the time interval we waited for (there is presumably some number of microseconds at which the value is exactly 10).

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
moritzkiefer-da: <= seems fine

daml/splice-amulet/daml/Splice/AggregateLock.daml
moritzkiefer-da: ok then we should add a comment. But other places in splice us 365 days e.g. the roundsPerYear computation in Issuance.daml so I'd lean towards being consistent with that. I don't think in practice the difference matters.

daml/splice-amulet/daml/Splice/AggregateLock.daml
dbzguy56: I think we are just keeping the choice here temporarily until we move it under the Withdraw choice or ExternalPartyAmuletRules

daml/splice-amulet/daml/Splice/AggregateLock.daml
dbzguy56: 365.25 was specified within the CIP [here](https://github.com/canton-foundation/cips/blob/main/cip-0105/cip-0105.md#2-specification), should we look to change either that or the code itself?

moritzkiefer-da: thanks for checking, if the CIP is that precise then let's just stick with the CIP definition and just add a comment.

scripts/setup-mono-package.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
moritzkiefer-da, line 1: change is merged, please rebase and remove from this pr

daml/splice-amulet/daml/Splice/AmuletAllocationV2.daml
```diff
@@ -230,7 +230,9 @@ computeAllocationExpiryInternal transferConfig oldExpiresAt settlementDeadline =
       | maxTime `subTime` oldExpiresAt <= maxTTL = maxTime
       | otherwise = oldExpiresAt `addRelTime` maxTTL
 
+-- FIXME: the exception to allow Some maxBound is for infinite-duration governance locks, and should probably be limited more explicitly to only those cases.
```
meiersi-da: Why can't you use `settlementDeadline = None`?

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,267 @@
...
+    availAmount = if currentDateTime >= endDate
+      then (nextIterationFundAmount, 0.0)
+      else (availableWithdrawAmount, remainingVestingAmount)
+    availableWithdrawAmount = (initialAmount * totalVestedRatioElapsed) - withdrawnAmount
+    remainingVestingAmount = nextIterationFundAmount - availableWithdrawAmount
+    meta = alloc.allocation.meta.values
+    -- The initial amount represents the total that was locked initially when thrown into vesting state
+    initialAmount : Decimal = fromSome . parseDecimal . fromSome $ TM.lookup "cip-105/vestingLock.initialAmount" meta
+    startDate : Time  =  decodeTime . fromSome $ TM.lookup "cip-105/vestingLock.startDate" meta
+    endDate : Time = decodeTime . fromSome $ TM.lookup "cip-105/vestingLock.endDate" meta
```
meiersi-da, line 116: these `fromSome` don't give good error messages. Let's define a helper that parses these metadata fields and reports the field that was missing or could not be parsed in the error message.

daml/splice-amulet/daml/Splice/AggregateLock.daml
dbzguy56: My PR is merged in, and hopefully the correct variant should be correct in the new: https://github.com/canton-network/splice/pull/6427/changes#diff-42bf48da99a7b4ebce12dcb026f648b6e7214a957e5380695103d9c2a839015cR116

daml/splice-amulet/daml/Splice/AggregateLock.daml
dbzguy56: Should be fixed below. Do we want to do the same with `controllerSetFromMeta` and make a helper utility function for semi-colons?

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
dbzguy56: Done

daml/splice-amulet/daml/Splice/AggregateLock.daml
dbzguy56: Done: https://github.com/canton-network/splice/pull/6427/changes#diff-42bf48da99a7b4ebce12dcb026f648b6e7214a957e5380695103d9c2a839015cR68

daml/splice-amulet/daml/Splice/AggregateLock.daml
dbzguy56: Should be done in a much cleaner format now: https://github.com/canton-network/splice/pull/6427/changes#diff-42bf48da99a7b4ebce12dcb026f648b6e7214a957e5380695103d9c2a839015cR200

daml/splice-amulet/daml/Splice/AggregateLock.daml
dbzguy56: Should be much cleaner: https://github.com/canton-network/splice/pull/6427/changes#diff-42bf48da99a7b4ebce12dcb026f648b6e7214a957e5380695103d9c2a839015cR264

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
dbzguy56: 5 mins is removed, and we now check `<=`: https://github.com/canton-network/splice/pull/6427/changes#diff-42bf48da99a7b4ebce12dcb026f648b6e7214a957e5380695103d9c2a839015cR69

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
dbzguy56: Done

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
dbzguy56: Tolerance has been removed and we now have an exact Decimal instead: https://github.com/canton-network/splice/pull/6427/changes#diff-c33c401bd06c2f019c9f1901af1c6d7225d6f1a3193e4b7d6d06da3262069646R141

daml/splice-amulet/daml/Splice/AggregateLock.daml
jonored: Current in-progress plan is to remove at least the choices (moving to allocation withdraw), and probably the whole template. The remaining pieces that need to land somewhere else are early unlock support (which may be it's own tiny "allow early unlock" template provided as context to withdraw) and producing an appropriate key to group lock allocations with.

daml/splice-amulet/daml/Splice/AggregateLock.daml
jonored: lockedCid was the active lock and withdrawTo was a fresh allocation to hold the withdrawn-to-vesting funds. Removed with the move to integrate into allocation withdraw, where the contract operates explicitly with the holder's authority as well as whatever is in the flexible controller lists.

daml/splice-amulet/daml/Splice/AggregateLock.daml
jonored: Discussed and agreed given the additional context, removal of most or all is in progress.

daml/splice-amulet/daml/Splice/AggregateLock.daml
jonored: This choice doesn't necessarily have rights to create allocations in the holder's name; that stops being an issue when it's moved to the withdraw choice.

scripts/setup-mono-package.sh
cgibbard: Yeah, that was an accident because I'd previously cherry-picked it onto the branch. Sorry!

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
jonored: Total check in the test is on the way.

daml/splice-amulet/daml/Splice/AggregateLock.daml
jonored: Referenced now, but might be removed in favor of a different mechanism to handle FA quick unlock given removal of any on-chain "whole aggregate" contract.

daml/splice-amulet/daml/Splice/AggregateLock.daml
dbzguy56: Changes with this will be pushed soon

daml/splice-amulet/daml/Splice/AggregateLock.daml
dbzguy56: @jonored mentioned to me that this should be fixed

daml/splice-amulet/daml/Splice/AggregateLock.daml
dbzguy56: Should be removed because you are correct there

daml/splice-amulet/daml/Splice/AggregateLock.daml
jonored: No checks to do now that we're running in the context of the allocation with the holder's authorization and can create a new one as well as update balances on the old one; we still use `getTime` and `now` to create the times in the new allocation.

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
jonored: Should be there now.

daml/splice-amulet/daml/Splice/AggregateLock.daml
dbzguy56: This comment should be outdated since the template is now gone

token-standard/splice-token-standard-utils/daml/Splice/TokenStandard/Utils/Internal/Allocations.daml
```diff
@@ -374,7 +375,18 @@ settlementFactoryV2_settleBatchDefaultImpl
...
+
+settlementFactoryV2_settleBatchDefaultImplNoSelf
+  : (AllocationV2.AllocationView -> AllocationV2.Allocation_Settle -> Update ExtraArgs)
+     -- ^ Function to compute the actual extra argument to use for settling an allocation.
+     -- Use this for example to redact unrelated choice-context data from the extra arguments.
+  -> Party
+     -- ^ Admin of the allocations and the settlement factory
+  -> AllocationV2.SettlementFactory_SettleBatch
+  -> Update AllocationV2.SettlementFactory_SettleBatchResult
+settlementFactoryV2_settleBatchDefaultImplNoSelf getFilteredExtraArgs admin arg = do
```
meiersi-da, line 389: Why do we need this?

daml/splice-amulet/daml/Splice/AmuletAllocationV2.daml
```diff
@@ -230,12 +233,23 @@ computeAllocationExpiryInternal transferConfig oldExpiresAt settlementDeadline =
...
       | otherwise = oldExpiresAt `addRelTime` maxTTL
 
+-- FIXME: the exception to allow Some maxTime is for infinite-duration governance locks, and should probably be limited more explicitly to only those cases.
 computeAllocationExpiry : TransferConfigV2 Amulet -> Time -> Optional Time -> Update Time
-computeAllocationExpiry transferConfig oldExpiresAt settlementDeadline = do
-  let expiresAt = computeAllocationExpiryInternal transferConfig oldExpiresAt settlementDeadline
-  assertWithinDeadline "allocation.expiresAt" expiresAt
-  pure expiresAt
+computeAllocationExpiry transferConfig oldExpiresAt settlementDeadline
+  | settlementDeadline == Some maxComparableTime = pure maxComparableTime
```
meiersi-da, line 239: This looks problematic: it opens up the option for normal allocations to not expire at all.

I'd suggest we check whether the allocation is an SvLock, and if it is set expiry to maxComparableTime

daml/splice-amulet/daml/Splice/AggregateLock.daml
meiersi-da: still open

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,290 @@
...
+    let vestingLock = view $ toInterface @V2.Allocation a
+        vestingLockCid = toInterfaceContractId @V2.Allocation aCid
+    now <- getTime
+
+    require "Must be a valid vesting unlock allocation" $ isValidVestingLockedDestination vestingLock.allocation.admin vestingLock
+    require "There must be an initialAmount set for the vestingLock"
+      $ TM.member "cip-105/vestingLock.initialAmount" vestingLock.allocation.meta.values
+
+    ownerParty <- whenNone vestingLock.allocation.authorizer.owner $ -- Mostly to safely get ownerParty
+      assertFail "The requirement 'governance locked allocations must be owned by basic accounts' was not met"
```
meiersi-da, line 217: use `Splice.TokenStandard.Utils.regularAccountOwner`

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,290 @@
...
+    let GovernanceLockedControllers{..} = governanceLockedControllersFromMeta vestingLock.meta ownerParty
```
meiersi-da, line 218: I know that the requirements document requires unlock authorizations specifiable per lock (see https://docs.google.com/document/d/13zl8ILEWq6CvSALk2LA79I2rE1PHfni1EFf5ywr-Aik/edit?tab=t.0#bookmark=id.gd3iob401nuc). However, I question the validity of the requirement in its current form: implementing support for custody in every app that should support custodial wallets does not scale, as every app then needs special workflows and the wallets need to integrate with them. The scalable option is for custodial wallets to take custody of the party's private key, and restrict the actions that a party can take using that key; e.g., by requiring the four-eyes principle for signing transactions for that party.

This way the custody implementation is separated from the apps that the party uses.

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,290 @@
...
+
+    checkControllerSet arg.actors withdrawControllerSets
+
+    (instrumentId, nextIterationFunding) <- case TM.toList <$> vestingLock.allocation.nextIterationFunding of
+        Some [a] -> pure a
+        _ -> assertFail "The requirement 'Must reserve a single amount of a single token' was not met"
```
meiersi-da, line 224: I'd expect that we lookup that amulet instrument id here. See like checks in `ExternalPartyAmuletRules` where the allocation factory and settlement is implemented.

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,290 @@
...
+    require "There must be an initialAmount set for the vestingLock"
+      $ TM.member "cip-105/vestingLock.initialAmount" vestingLock.allocation.meta.values
```
meiersi-da, line 226: I'm wary of repeating the parsing of metadata within the core logic. I think we'd be better off to push the parsing of metadata into the allocation factory, and use proper types internally to represent these special allocations. We can store that value on the `AmuletAllocationV2` in a dedicated `allocationKind : Option AllocationKind` field and then define `AllocationKind` as a sum-type over the records that represent the different SV locking kinds of allocations; and later the FA locking kinds of allocations.

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,290 @@
...
+                  instrumentId
+                  meta = emptyMetadata
+              , TransferLegSide with
+                  transferLegId
+                  side = ReceiverSide
+                  otherside = vestingLock.allocation.authorizer
+                  amount = availableWithdrawAmount
+                  instrumentId
+                  meta = emptyMetadata
+              ]
```
meiersi-da, line 275:
```suggestion
```
not required to return funds to the owner

daml/splice-amulet/daml/Splice/AmuletAllocationV2.daml
```diff
@@ -230,12 +233,23 @@ computeAllocationExpiryInternal transferConfig oldExpiresAt settlementDeadline =
...
+    pure expiresAt
 
+isGovernanceLocked alloc = TextMap.member "cip-105/type" alloc.allocation.meta.values
+
+emptyCloneWithMeta : AmuletAllocationV2 -> Time -> Metadata -> Update (ContractId V2.Allocation)
+emptyCloneWithMeta src now meta = toInterfaceContractId <$> create src with
+  lockedAmulet = None
+  createdAt = now
+  allocation = src.allocation with
+    meta
```
meiersi-da, line 252: I'd probably turn it around and copy all fields explicitly to be sure that we don't copy too much. For example, right now the transfer legs and next iteration funding would be copied. We probably actually only need to specify the owner from the `AggregateLock` code and can then specify the rest.

daml/splice-amulet/daml/Splice/AmuletAllocationV2.daml
meiersi-da: OK, saw the FIXME. Yes I agree.

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
```diff
@@ -0,0 +1,289 @@
...
+-- bypassing the aggregate-lock unlock path.  Used to exercise the vesting
+-- withdrawal flow in isolation.
+lockForVesting : Decimal -> Time -> TestEnv -> Party -> Script AllocationInstructionResult
+lockForVesting initialAmount now te =
+  lockForGovernance te (TM.fromList [(te.instrId.id, initialAmount)]) $ Metadata $ TM.fromList
+    [ ("cip-105/type", "vestingLock")
+    , ("cip-105/vestingLock.startDate", encodeTime now)
+    , ("cip-105/vestingLock.endDate", encodeTime $ addRelTime now $ days 365 + hours 6)
+    , ("cip-105/vestingLock.initialAmount", show initialAmount)
+    ]
```
meiersi-da, line 64: This would also become simpler if we had an explicit type to represent a vesting lock together with functions to encode and decode to/from metadata.

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
```diff
@@ -0,0 +1,289 @@
...
+withAggregateWithdrawContext env registries =
+    flip fmap registries $ \reg ->
+      reg { v2Api = fmap updateApi reg.v2Api }
+  where
+    addContext a = do
+      baseCtxt <- a
+      extraContext <- getExternalPartyConfigStateContext env.registriesEnv.amuletV2
+      pure $ baseCtxt <> extraContext
+    updateApi api =
+      api { ggetAllocation_WithdrawContext = \a -> addContext . api.ggetAllocation_WithdrawContext a }
```
meiersi-da, line 79: This change should go into the `AmuletRegistryV2.getAllocation_WithdrawContext`. That function should check the shape of the allocation and add the required extra context if it is an SV lock.

meiersi-da (commented):
Thanks a lot. This goes in the expected direction. I believe we need some more polishing rounds, but overall I expect that we can use this approach to represent the locked amounts.

The polishing that I believe would be most valuable to see is to factor out the codecs for "SV locks" and "vesting SV unlocks", so that the logic can be written against proper datatypes.

Once that's done it should also be easier to push the test coverage, as the core logic becomes easier to understand.

Independently, we need to invest in a PoC of the Daml code for the SV rewards issuance. That's pending though on aligning how that flow should work overall.

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
dbzguy56, line 64: TODO @dbzguy56: we also want to assert the presence of `cip-105/for-benefit-of` under the metadata, without `for-benefit-of`, we won't be able to associate the lock with any SVs

daml/splice-amulet/daml/Splice/AggregateLock.daml
```diff
@@ -0,0 +1,290 @@
...
+    -- We should have one result from settleBatch
+    case settleResult.allocationSettleResults of
+      [allocationResult] -> pure allocationResult
+      _ -> abort "VestingUnlock_Withdraw should result in one allocation result. "
+
+-- Takes the choice metadata as well to enable override to allow quick unlock
+decodeVestingSchedule : Metadata -> Metadata -> Update VestingSchedule
+decodeVestingSchedule _withdrawMeta lockMeta = do
+  case TM.lookup "cip-105/vestingSchedule" lockMeta.values of
+    Some "SV" -> pure VestingSchedule_SV
```
dbzguy56, line 288: TODO @dbzguy56 : Would we want to add `VestingSchedule_Immediate` here (and in `isValidAggregateLockedAllocation`) as a pattern match so that we are able to hit the workflow? Otherwise the "Immediate" variant is never hit and fails to unlock.

daml/splice-amulet/daml/Splice/AggregateLock.daml
jonored: We're using this to calculate new field values at this point; is there still an objection to using getTime?

daml/splice-amulet/daml/Splice/AggregateLock.daml
jonored: I certainly have no objection to that; this is still mostly written as if it were a downstream consumer of AmuletAllocationV2, but it isn't anymore and we should take advantage of that. We still reuse the settlement machinery for transitions and the allocation factory is still a bit simpler, so we do still have a reason for them to be the same template instead of just implement the same interfaces.

daml/splice-amulet/daml/Splice/AggregateLock.daml
jonored: They're unreachable, but if we don't mind putting governance-lock-specific fields into AmuletAllocationV2 then it gets considerably cleaner.

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
jonored: Certainly before we consider merge; this being here was largely just expediency to "tests run and pass".

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
jonored: Agree, and also agree that should be handled in the allocation factory given the rest of the context.

daml/splice-amulet/daml/Splice/AggregateLock.daml
jonored: I don't mind revisiting that requirement, but at least both are straightforward. I suspect most of the uses of non-default authorizations like this go through some custom code already, implementing whatever agreement they are building on top of this allocation as a primitive; at least agreement via a non-Amulet-but-standard token to collect authorization but multiparty on this level to represent constrained consent from some market seems plausible within the token standard without requiring key custody or anything plausibly transfer-like. Avoiding anything that could be construed as a transfer was a fairly big deal in discussions.

daml/splice-amulet/daml/Splice/AggregateLock.daml
jonored: Ok. I want to see the test run before I put that change in a commit but I have no doubt you are correct.

daml/splice-amulet/daml/Splice/AmuletAllocationV2.daml
jonored: Reasonable. This particular trick to decouple the AmuletAllocationV2 and AggregateLock modules was a bit hasty and that's certainly copying more than we want.

token-standard/splice-token-standard-utils/daml/Splice/TokenStandard/Utils/Internal/Allocations.daml
jonored: Avoids needing a contract ID for ExternalPartyAmuletRules added to the context for withdraw with from governance locks. We could also just add it and call the normal one.

daml/splice-amulet/daml/Splice/Amulet.daml
```diff
@@ -618,6 +618,15 @@ template UnclaimedActivityRecord
         pure UnclaimedActivityRecord_DsoExpireResult with unclaimedRewardCid
 
 
+template SvRewardMintReceipt with
+    dso : Party
+    amount : Decimal
+    beneficiary : Party
```
meiersi-da, line 624:
```suggestion
    svRightOwnerName : Text -- ^ SV right owner whose lifetime rewards increased.
```

I don't think we can use parties to identify the SV rights owner whose reward state needs to be updated with the life-time reward. The stable identifier for SV rights owners is their name.

meiersi-da: This also implies that we need to extend the coupon representing an allocation out of the unclaimed rewards with the SV right owner to whom the reward minting should be booked when the coupon is claimed and minted. Once we do so, we can also properly attribute the coupon claiming to minting an SV reward in `AmuletRules_Transfer`.

token-standard/splice-token-standard-utils/daml/Splice/TokenStandard/Utils/Internal/Allocations.daml
meiersi-da: Where possible I'd suggest that we call the actual interface method instead of its implementation to avoid divergences. So yes, let's remove this one, and actually call the interface method for settling a batch.

daml/splice-amulet-test/daml/Splice/Scripts/TestAggregateLocks.daml
meiersi-da: In my experience, mirroring the production information flows matters during the design stage, as many of the obstacles for particular implementation ideas come from the fact the the information flow doesn't work out as expected. So let's align the information flows to prod in the design phase already.

daml/splice-amulet/daml/Splice/Amulet.daml
meiersi-da: Furthermore, we'll also need to store the SV right owner name on the `SvRewardCoupon`.

dbzguy56: Starting to implement the changes through the branch/PR here: https://github.com/obsidiansystems/splice/pull/7/changes

PR #6428, opened 2026-07-15
Fix missing TransferInstruction in TestTokenV2SettlementIntegrationTest

OriolMunoz-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/9174
Bob is just an observer so it can be slow to see the TransferInstruction


pawelperek-da (approved):
bravo

PR #6431, opened 2026-07-15
remove the usage of the splitPostgresInstances flag from the SV apps deployment as it no longer changes the number of instances

mblaze-da:
cluster test: https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/74395


isegall-da (commented):
There seem to be quite a few other places that we have the same flag, can we remove it consistently across the board?

mblaze-da:
> There seem to be quite a few other places that we have the same flag, can we remove it consistently across the board?

@isegall-da in splitwell and validator1 it actually does what it is supposed to (reduces the number of DB instances). So I'd be hesitant about removing it there, unless we don't care anymore.

nicu-da:
> > There seem to be quite a few other places that we have the same flag, can we remove it consistently across the board?
> 
> @isegall-da in splitwell and validator1 it actually does what it is supposed to (reduces the number of DB instances). So I'd be hesitant about removing it there, unless we don't care anymore.

Can we check if it's actually enabled on any cluster?

PR #6438, opened 2026-07-15
add BFT enforcement for bulk storage objects

### LINKED ISSUE #6495: Add BFT checks as a condition for moving objects from staging to committed in bulk storage

isegall-da:
Fixes https://github.com/canton-network/splice/issues/6495

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/BulkStorageCommitFromStaging.scala
```diff
@@ -42,7 +105,7 @@ class BulkStorageCommitFromStaging[T](
     (T, Seq[ObjectKeyAndChecksum]),
     NotUsed,
   ] = {
-    Flow[(T, Seq[ObjectKeyAndChecksum])].mapAsync(parallelism = 1) { case (t, obj) =>
+    Flow[(T, Seq[ObjectKeyAndChecksum])].flatMapConcat { case (t, obj) =>
```
isegall-da, line 110: This was leaking pipelines, it created a new one that did not get destroyed when the main one was.

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -1810,6 +1810,30 @@ paths:
         "501":
           $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/501"
 
+  /v0/history/bulk/checksums:
+    post:
+      tags: [ internal, pre-alpha, scan ]
+      x-jvm-package: scan
+      operationId: "getBulkObjectChecksums"
+      description: |
+        **Under Development, do not use in production yet** Get checksums for bulk history objects. Uses both staging and committed objects.
```
ray-roestenburg-da, line 1819:
```suggestion
        **Under Development, do not use in production yet** Get checksums for bulk history objects. Searches for object_keys in both staging and committed objects.
```
Minor suggestion

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/util/HasPeerBftScanConnection.scala
```diff
@@ -0,0 +1,82 @@
...
+              )
+            } yield {
+              connectionVar = Some(connection)
+              connection
+            }
+        }
+      }
+    }
+
+  protected def closeScanConnection(): Option[SyncCloseable] =
```
ray-roestenburg-da, line 73: This doesn't really close, it creates a SyncCloseable (could be confusing)

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/util/HasPeerBftScanConnection.scala
```diff
@@ -0,0 +1,82 @@
...
+      retryProvider: RetryProvider,
+      loggerFactory: NamedLoggerFactory,
+  )(implicit
+      tc: TraceContext,
+      ec: ExecutionContextExecutor,
+      mat: Materializer,
+      httpClient: HttpClient,
+      templateJsonDecoder: TemplateJsonDecoder,
+  ): Future[BftScanConnection] =
+    blocking {
```
ray-roestenburg-da, line 46: the mutex.exclusive looks like a special mutex to especially not having to use `blocking`:

```
 Lock to be used instead of blocking synchronized
```

 so I think the blocking should be removed.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/util/HasPeerBftScanConnection.scala
```diff
@@ -0,0 +1,82 @@
...
+import org.lfdecentralizedtrust.splice.environment.{RetryProvider, SpliceLedgerClient}
+import org.lfdecentralizedtrust.splice.http.HttpClient
+import org.lfdecentralizedtrust.splice.scan.admin.api.client.BftScanConnection
+import org.lfdecentralizedtrust.splice.scan.config.ScanAppClientConfig
+import org.lfdecentralizedtrust.splice.scan.store.ScanStore
+import org.lfdecentralizedtrust.splice.util.TemplateJsonDecoder
+
+import scala.concurrent.{ExecutionContextExecutor, Future, blocking}
+
+trait HasPeerBftScanConnection {
```
ray-roestenburg-da, line 22: I'd make this a class that owns one connection instead of a trait.

As a trait, the connection state is mixed invisibly into the host class, and the dependencies are method parameters, but only the first caller's arguments are ever used, so a second call site passing a different store or config silently gets the wrong connection.

The mixer also has to remember to wire closeScanConnection() into its closing sequence.

Suggestion:

```
  class PeerBftScanConnection(store: ScanStore, svName: String, ...) extends AutoCloseable {
    @volatile private var started = false
    lazy val connection: Future[BftScanConnection] = {
      started = true
      BftScanConnection.peerScanConnection(...)
    }
    override def close(): Unit = if (started) connection.foreach(_.close())
  }
```

Constructor params pin the dependencies, lazy val gives thread-safe create-once semantics (no mutex, no var, no blocking), caching the future avoids creating two connections on concurrent first calls, and whoever constructs it closes it. Both return a closed connection after close(). But here that's a local owner bug (use-after-close), not a silent failure when calling getOrCreateScanConnection.

If we ever want this fully airtight (post-close access fails loudly with AbortedDueToShutdown, and close() waits for an in-flight creation instead of racing it), extending `FlagCloseable` and wrapping the accessor in `synchronizeWithClosingF` fixes that. For a connection closed once at node shutdown I don't think we need that here though.

ray-roestenburg-da (commented):
Looks good overall, please have a look at my comments re: `HasPeerBftScanConnection`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/util/HasPeerBftScanConnection.scala
isegall-da: good idea. I replaced the `lazy val` with a `def` after all though, as we need a traceContext there, which is not available at construction time. Otherwise, I applied the suggestion now.

isegall-da:
@rautenrieth-da @stephencompall-DA 
can I ask one of you for a re-review in light of @ray-roestenburg-da 's vacation please?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/BulkStorageCommitFromStaging.scala
```diff
@@ -21,19 +23,57 @@ class BulkStorageCommitFromStaging[T](
...
     logger.debug(
       s"Checking BFT agreement for objects: ${objects.map(_.key).mkString(", ")}"
     )
-    Future.successful(true)
+    if (appConfig.bftCheckEnabled) {
+      for {
+        connection <- scanConnection.connection
+        bft <- connection.getBulkObjectChecksums(objects.map(_.key)).map(Some(_)).recoverWith {
+          case ex @ HttpErrorWithHttpCode(code, _) =>
+            if (code == StatusCodes.BadGateway) {
```
rautenrieth-da, line 44: Why do we need a special case for `BadGateway` errors?

isegall-da:
> @rautenrieth-da @stephencompall-DA can I ask one of you for a re-review in light of @ray-roestenburg-da 's vacation please?

ping.
Or maybe @OriolMunoz-da ?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/BulkStorageReader.scala
```diff
@@ -160,6 +160,19 @@ class BulkStorageReader(
...
 
+  def getObjectChecksums(
+      objectKeys: Seq[String]
+  ): Future[Seq[Option[String]]] = {
+    for {
+      committed <- committedS3Connection.getChecksums(objectKeys)
+      staging <- stagingS3Connection.getChecksums(objectKeys)
+    } yield {
+      objectKeys.map { key =>
+        committed.find(_.key == key).orElse(staging.find(_.key == key)).map(_.checksum)
```
rautenrieth-da, line 171: What if the committed and staging checksums both exist but differ? Do we want to log a warning?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/BulkStorageCommitFromStagingTest.scala
```diff
@@ -54,37 +78,261 @@ class BulkStorageCommitFromStagingTest
...
+          logEntries =>
+            forAtLeast(1, logEntries)(
+              _.message should include(
+                "All objects are known to the BFT peers, but the checksums do not match"
+              )
+            ),
+        )
+
+      } catch {
+        case ex: Throwable =>
```
rautenrieth-da: When would the above throw an exception?

rautenrieth-da (approved):
Thanks!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/BulkStorageCommitFromStaging.scala
isegall-da: Because we use `BadGateway` for "consensus not reached", which in this case is not something we want to throw an exception on, just treat as "false", and retry later. Will add a debug log message to act as documentation for that.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/BulkStorageReader.scala
isegall-da: hmm.. Since the only way to get something to committed is by copying it from staging, that would be a very unexpected case. We'd definitely want that to even be an error, but I wonder if it's even worth adding the complexity of checking that here...

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/BulkStorageCommitFromStagingTest.scala
isegall-da: I think I added this before realizing the actual leakage of pipelines was in https://github.com/canton-network/splice/pull/6438#discussion_r3625395669, will remove.

PR #6469, opened 2026-07-20
Include last failure in the circuit breaker exception

### LINKED ISSUE #6285: circuit breakers should keep the last failure and be logged as part of the rejection stacktrace
currently we just log a stacktrace that the circuit breaker is open, and it's a bit of a pain to find the first root issue. 
so keeping the last actual exception in the circuit breaker and adding that as the parent of the circuit breaker open exception would make our life a lot easier

nicu-da:
[ci]

fixes https://github.com/canton-network/splice/issues/6285

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
@@ -21,10 +21,11 @@ import com.digitalasset.canton.topology.PartyId
 import com.digitalasset.canton.tracing.TraceContext
 import io.grpc.StatusRuntimeException
 import org.apache.pekko.actor.Scheduler
-import org.apache.pekko.pattern.{CircuitBreaker, CircuitBreakerOpenException}
+import org.apache.pekko.pattern.CircuitBreaker
 import org.lfdecentralizedtrust.splice.config.CircuitBreakerConfig
 
 import java.util.concurrent.atomic.AtomicReference
+import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success}
 
@@ -40,6 +41,7 @@ class SpliceCircuitBreaker(
 ) extends NamedLogging {
 
   private val lastFailure: AtomicReference[Option[CantonTimestamp]] = new AtomicReference(None)
+  private val lastException: AtomicReference[Option[Throwable]] = new AtomicReference(None)
 
   private val errorCategoriesToIgnore: Set[ErrorCategory] = Set(
     InvalidIndependentOfSystemState,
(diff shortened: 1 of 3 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
```diff
@@ -106,8 +110,11 @@ class SpliceCircuitBreaker(
         if (!isFailureIgnored(exception)) {
           underlying.fail()
           lastFailure.set(Some(clock.now))
```
martinflorian-da, line 112: For my education: why is it better that we set these after actually failing or succeeding? Slightly lower response latency?

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreakerTest.scala
```diff
@@ -70,11 +69,37 @@ class SpliceCircuitBreakerTest
...
+          cb.isOpen shouldBe true
+        }
+      }
+
+      val future3 = cb.withCircuitBreaker(Future.successful("should not reach here"))
+      whenReady(future3.failed) { ex =>
+        ex shouldBe a[SpliceCircuitBreakerOpenException]
+        ex.getCause shouldBe lastFailure
+        ex.getCause.getMessage shouldBe "root cause failure"
+      }
```
martinflorian-da, line 100: Do we want to test also that the cause gets cleared on success? Otherwise we could get some confusing false cause on some race maybe, the next time if fails?

martinflorian-da (approved):
Looks reasonable to me, thank you!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
nicu-da: That's the place when we actually know if it failed or not right? How else would you handle that for the future

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreakerTest.scala
nicu-da: adding

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/SpliceCircuitBreaker.scala
martinflorian-da: possibly I'm even more confused that I thought; I was wondering why we call `underlying.fail()` before we persist what the failure was about.

So if we have two requests in close succession, the first one closes the circuit breaker, the second one then immediately gets blocked, but since we closed the circuit breaker before setting the cause it gets no cause.

Ah yes but the alternative would then risk us circuit breaking slightly later than we should; which actually sounds worse.

I think my good on this question, thanks :)

PR #6488, opened 2026-07-21
add details about CI and automatic upgrade tests to the instructions on bumpting Canton

mblaze-da:


martinflorian-da (approved):
Thanks! I think I'd just skip mentioning step 5 but don't feel strongly about it.

PR #6491, opened 2026-07-21
Add proposal search functionality

### LINKED ISSUE #3025: Add a search window to SV UI
Add a search box that allows the user to:

* Paste a vote proposal contract ID and return the proposal

timpel-fcs:
### Summary

Fixes #3025

- Add `ProposalSearch` following UIs textfield styling below Governance page header with local input and `?q=` URL sync for valid full contract IDs
- Enter on a valid contract ID navigates to the proposal details page
- Filter Action Required, Inflight Votes, and Vote History client-side by exact contract ID; paginate through closed vote history (500 accepted + 500 rejected per page) when not already in loaded data, using the same lookup as the proposal details page (accepted workaround until server-side search is available)
- Validate search input as a 138-character hexadecimal contract ID before filtering or updating the URL

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/sv/frontend/src/__tests__/governance/governance-page.test.tsx
+++ b/apps/sv/frontend/src/__tests__/governance/governance-page.test.tsx
@@ -1,6 +1,6 @@
 // Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { render, screen, waitFor, within } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor, within } from '@testing-library/react';
 import { describe, expect, test } from 'vitest';
 import { SvConfigProvider } from '../../utils';
 import userEvent from '@testing-library/user-event';
@@ -8,7 +8,13 @@ import dayjs from 'dayjs';
 import { dateTimeFormatISO } from '@canton-network/splice-common-frontend-utils';
 import App from '../../App';
 import { navigateToGovernancePage } from '../helpers';
-import { voteResultsAmuletRules, voteResultsDsoRules } from '../mocks/constants';
+import {
+  activeProposalCid,
+  closedVoteCid,
+  voteResultsAmuletRules,
+  voteResultsDsoRules,
+} from '../mocks/constants';
+import { CONTRACT_ID_VALIDATION_MESSAGE } from '../../utils/proposalSearch';
 
 type UserEvent = ReturnType<typeof userEvent.setup>;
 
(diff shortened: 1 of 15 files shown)


ray-roestenburg-da:
I don't think this PR is acceptable. I would have expected some more questions around this issue in it's original form, cause it's rather lacking in terms of defining **how** the search should work and how it should be visualized, it just states a couple of requirements for many possible fitting solutions. I changed it today to just "Paste a vote proposal contract ID and return the proposal", to avoid ambiguity, and simplify this task for now. This PR now comes across as if you just pushed the requirements into AI and did not think about asking for more feedback to get to a fitting solution, or point out problems with the issue. For instance, how will this work across infinite scroll, it doesn't really make sense to search only on the client side and be ok with incomplete results, for complete results you would have to scroll through everything? How does it make sense to take a calendar date and apply them to (createdAt, effectiveAt and threshold)?

timpel-fcs:
@ray-roestenburg-da I picked this up a few days ago so I did not see the updated issue scope. All of these functions were required in the original issue, though you're right of course I should have reconfirmed here regarding more details as the scope was indeed lacking rather than working based on estimations.

For styling I tried to follow the Figma design of other components in the UI as there is no clear design for the search functionality itself on the board, though I am happy to circle back to the design team here if required.

For the date search functionality I decided that there should be no big issue with matching all 3 possible date strings, I assumed that there should not be much overlap between different vote proposal timelines so the number of matches should be small enough and this would allow the users to search more easily based on date estimations. Of course if this assumption is wrong, we will need to improve the search or remove it based on the updated issue scope.

Regarding the vote history infinite scroll limitation with the client side search, this is indeed not handled in the current approach. The action required and in-flight sections already load the full result set so a client-side search is not an issue here. Regarding the vote history search I currently see the following limitations:
- go via `listVoteRequestResults` and paginate through until a match is found is a request heavy approach that does not scale well, if we want to enable partial matching for this it would require a full search of the data set on each query
- go via `useVoteRequestResultByCid` which loads 500 accepted + 500 rejected gives a limited data set and may lead to no results found even though the id is accurate, this is currently accepted for the proposal detail page and directly used here (Enter navigates to the detail page, which uses useVoteRequestResultByCid)
The issue stems from the backend not allowing to filter the vote history by cid, in order to build a sophisticated search this would have to be adjusted but I was under the impression this issue was scoped to frontend-only changes. I would appreciate some further input from you here on how to best proceed.

I'm happy to remove the date and name search options here to align this to the reduced scope of the issue and will change this PR to draft as it needs reworking.

katiepang-da:
Thinking about the searching component on an infinite scroll:
- in general infinite scroll is meant for users to casually scroll through items with little to zero constraints in mind, this helps them scroll through easier without adding friction to click for more 
- however when it comes to adding search, this may not be the best idea for many reasons including loading time. losing where one was prior to a search

Rethinking possibilities with search in mind:
- pagination to provide easier location for users to keep track when searching 
- pagination gives less page load so it will return results faster 
- can also break up the search fields per item (ie: createdDate vs ExpiredDate separately) this can help reduce the results to be more scanable 

Given the underlying problem is a data fetching problem, it might be a better idea to fix this issue. If we need a bandage for now, perhaps an idea could be what we do similar to infinite scroll where we fetch all the records and then organize the results based on the filter criteria 

cc @pawelperek-da i hope this summarizes what we talked about in slack lol

not sure if this answers the issue in this ticket but something to think about - can discuss further in the weekly sync

timpel-fcs:
Screenshot 2026-08-03 at 15 37 27
Screenshot 2026-08-03 at 16 54 28
Screenshot 2026-08-03 at 16 54 46
Screenshot 2026-08-03 at 16 55 00
Screenshot 2026-08-03 at 15 37 13

pawelperek-da (approved):
Thanks!

PR #6503, opened 2026-07-22
Extend LSU cancellation test to mimic actual cancellations

nicu-da:
[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


nicu-da:
test failure reproduced the issue

nicu-da:
I do wonder if we should just move the endpoint to the operator auth endpoint instead of the admin one

moritzkiefer-da (approved):
nice thanks, changing auth seems fairly reasonable. we do allow you to schedule an lsu in the ui so saying you need different auth to cancel it feels weird

nicu-da:
> nice thanks, changing auth seems fairly reasonable. we do allow you to schedule an lsu in the ui so saying you need different auth to cancel it feels weird

will just do it in a separate PR

nicu-da:
@moritzkiefer-da asking for a re-review here because I had to redo the whole cancellation mechanism and found some more issue

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvAdminHandler.scala
```diff
@@ -46,12 +46,6 @@ class HttpSvAdminHandler(
     withSpan(s"$workflowId.cancelLogicalSynchronizerUpgrade") { _ => _ =>
       for {
         decentralizedSynchronizer <- dsoStore.getDsoRules().map(_.domain)
-        sequencerId <- synchronizerNodeService.sequencerAdminConnection().flatMap(_.getSequencerId)
-        _ <- participantAdminConnection
```
moritzkiefer-da, line 50: can you explain why just changing this to use the right key doesn't work?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/lsu/LsuCancellationTrigger.scala
```diff
@@ -0,0 +1,108 @@
...
+import org.lfdecentralizedtrust.splice.environment.SynchronizerNode.LocalSynchronizerNodes
+import org.lfdecentralizedtrust.splice.environment.TopologyAdminConnection.TopologyTransactionType
+import org.lfdecentralizedtrust.splice.sv.LocalSynchronizerNode
+import org.lfdecentralizedtrust.splice.sv.lsu.LsuCancellationTrigger.LsuCancellationTask
+import org.lfdecentralizedtrust.splice.sv.onboarding.SynchronizerNodeReconciler
+import org.lfdecentralizedtrust.splice.sv.onboarding.SynchronizerNodeReconciler.SynchronizerNodeState.OnboardedImmediately
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class LsuCancellationTrigger(
```
moritzkiefer-da: that trigger seems like a fair amount of complexity. At this point, I'm wondering what do we actually gain from removing the successor announcements? They are tied to the physical synchronizer id so nothing should get confused by old successors sticking around. And you're actually increasing the overall size of the topology state instead of reducing it by publishing a Remove.

nicu-da: Fair point, it's mostly a clean-up step but the successor id will be incremented anyway so we can get rid of this

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvAdminHandler.scala
nicu-da: it would work but if we want to remove the successors we have to do it for all the svs  not only those that used the API

moritzkiefer-da: ah I see, one more reason imho to just accept the successors sticking around.

nicu-da: done

moritzkiefer-da (approved):
:shipit:

PR #6515, opened 2026-07-22
Add data_table_name to ACS snapshots

### LINKED ISSUE #6262: E1.1 Schema prep (flyway)
Schema prep (Flyway)

Scope: ALTER TABLE acs_snapshot ADD COLUMN table_name TEXT (NULL = legacy acs_snapshot_data; non-NULL = per-snapshot table).

Acceptance: Migration applies; legacy snapshots remain readable.

OriolMunoz-da:
Fixes https://github.com/canton-network/splice/issues/6262


apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V073__acs_snapshot_table_name.sql
```diff
@@ -0,0 +1,14 @@
+alter table acs_snapshot
+  -- the name 'table_name' is not reserved, but it is a PostgreSQL keyword, so we use a different name to avoid confusion
+  add column data_table_name text default null,
+  -- these values won't be set anymore
+  drop constraint acs_snapshot_first_row_id_fkey,
+  drop constraint acs_snapshot_last_row_id_fkey,
+  alter column first_row_id drop not null,
+  alter column last_row_id drop not null,
```
OriolMunoz-da, line 8: I didn't find mention of this in the design, but it's necessary

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V073__acs_snapshot_table_name.sql
```diff
@@ -0,0 +1,14 @@
+alter table acs_snapshot
+  -- the name 'table_name' is not reserved, but it is a PostgreSQL keyword, so we use a different name to avoid confusion
+  add column data_table_name text default null,
```
OriolMunoz-da, line 3: this seems like a sensible deviation from the design

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
OriolMunoz-da, line 1: Included the minimal changes to validate:
> legacy snapshots remain readable.

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V073__acs_snapshot_table_name.sql
```diff
@@ -0,0 +1,14 @@
+alter table acs_snapshot
```
OriolMunoz-da, line 1: by the way: the table `acs_incremental_snapshot` probably needs the table name for the same reason this one does
OR
if it's derived from `target_record_time`, then so can `acs_snapshot` and therefore we don't need to add the column

wdyt?

OriolMunoz-da: and also: this makes me consider _not_ merging this to main until the rest of E1 is done, because I expect other surprises to popup

ray-roestenburg-da: sounds good to wait for merging

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V073__acs_snapshot_table_name.sql
ray-roestenburg-da: yes that's fine

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V073__acs_snapshot_table_name.sql
ray-roestenburg-da: ok

ray-roestenburg-da: thanks

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -258,9 +267,9 @@ class AcsSnapshotStore(
       templates: Seq[PackageQualifiedName],
   )(implicit tc: TraceContext): Future[QueryAcsSnapshotResult] = {
     for {
-      snapshot <- storage
+      snapshotTrait <- storage
```
ray-roestenburg-da, line 270: was this on purpose? Trait in the name is a little strange for me

ray-roestenburg-da: oh you mean like the kind of snapshot, ok

ray-roestenburg-da (approved):
LGTM, but as you say, maybe better to wait with merging

PR #6534, opened 2026-07-23
Remove double splitwell test setup

OriolMunoz-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/9285

surely if we run setup twice (which looks like a bad merge conflict resolution?), we will encounter weird flakes such as topology transactions being proposed twice as part of `withMultiSyncFeatureFlag`  running twice


moritzkiefer-da (approved):
but more is better?

PR #6563, opened 2026-07-26
Public Sequencer: Refactor JoiningNodeInitializer.scala to execute the /v0/onboard/sv/start HTTP request before connecting to synchronizer

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/6528


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
```diff
@@ -0,0 +1,34 @@
+package org.lfdecentralizedtrust.splice.integration.tests
+
+import com.digitalasset.canton.HasExecutionContext
```
pasindutennage-da, line 3: the purpose of this test, just for this PR, is to make sure that refactoring joining SV code doesn't break anything.

it still doesn't have anything to do with permissioning.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/sv1/SV1Initializer.scala
```diff
@@ -20,7 +20,7 @@ import com.digitalasset.canton.lifecycle.CloseContext
 import com.digitalasset.canton.logging.NamedLoggerFactory
 import com.digitalasset.canton.participant.synchronizer.SynchronizerConnectionConfig
```
pasindutennage-da, line 21: Previously, SV1 init logic was merged to main instead of the feature branch. (because we decided to go for a feature branch after some time.)

Since, Permission logic is not implemented in this new feature branch, yet, I just commented those SV1 RestrictedOpen calls, so that sv1 starts in UnrestrictedOpen mode.

Changing back to RestrictedOpen mode will be done in a next PR, with correct integration test.

(the goal of this PR is just joiningNodeINitializer refactoring and to make sure things still work in the unrestrictedOpen mode)

daml/splice-dso-governance/daml/Splice/SvOnboarding.daml
```diff
@@ -24,7 +28,6 @@ template SvOnboardingRequest with
     expiresAt : Time                -- ^ When this contract can be archived even if the onboarding did not succeed.
   where
     signatory dso
-    observer candidateParty
```
pasindutennage-da, line 27: Deleting a field in an existing template creates problems, however, SvOnboardingRequest is more of a temporary contract rather than a long persisted contract, so I assumed things won't blow up.

Reason for deletion is; in the new flow, when SV2 sends /start, it still has not registered in the sycnrhonizer, and hence causing a DAML error.

Wondering if its needed to make this backward compatible -- to add a new field to a template, "Optional" is the prefered way, but what about deleting?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
```diff
@@ -192,7 +192,7 @@ class HttpSvPublicHandler(
                     s"Party ${token.candidateParty} does not have the same namespace than its participant ${token.candidateParticipantId}."
                   )
                 )
-              } else if (!isCandidatePartyHostedOnParticipant)
+              } else if (!isCandidatePartyHostedOnParticipant && !config.permissionedSynchronizer)
```
moritzkiefer-da, line 195: can you track somewhere to carefully think through if there are security implications to not having this check?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
```diff
@@ -0,0 +1,175 @@
...
+    override val tracer: Tracer,
+) extends PollingParallelTaskExecutionTrigger[SvOnboardingObserverTrigger.Task] {
+
+  import SvOnboardingObserverTrigger.Task
+
+  override protected def retrieveTasks()(implicit
+      tc: TraceContext
+  ): Future[Seq[Task]] = {
+    for {
+      confirmations <- store.listSvOnboardingConfirmed()
```
moritzkiefer-da, line 43: you also want to do the topology check here and filter stuff out. Otherwise you keep refetching the contract and calling completeTask only to then do nothing which produces a log mess.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
```diff
@@ -0,0 +1,175 @@
...
+            .flatMap { txs =>
+              val governancePackageId =
+                SvOnboardingConfirmed.TEMPLATE_ID_WITH_PACKAGE_ID.getPackageId
+
+              val hasVetted = txs.exists { tx =>
+                tx.mapping match {
+                  case vp: VettedPackages if vp.participantId == participantId =>
+                    vp.packages.exists(_.packageId == governancePackageId)
+                  case _ => false
+                }
```
moritzkiefer-da: can we replace all this by a check through PackageVersionSupport?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
```diff
@@ -0,0 +1,175 @@
...
+        case Some(participantId) =>
+          // query for VettedPackages
+          participantAdminConnection
+            .listAllTransactions(
+              store = TopologyStoreId.Synchronizer(dsoRules.domain),
+              timeQuery = TimeQuery.HeadState,
+              includeMappings = Set(TopologyMapping.Code.VettedPackages),
+            )
+            .flatMap { txs =>
+              val governancePackageId =
```
moritzkiefer-da: checking exact package ids doesn't work with upgrading

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
```diff
@@ -0,0 +1,175 @@
...
+              val hasVetted = txs.exists { tx =>
+                tx.mapping match {
+                  case vp: VettedPackages if vp.participantId == participantId =>
+                    vp.packages.exists(_.packageId == governancePackageId)
+                  case _ => false
+                }
+              }
+
+              if (hasVetted) {
+                // Apply a random backoff up to 10 seconds to prevent all SVs from submitting simultaneously
```
moritzkiefer-da: that doesn't look right, you should get the random delay by inheriting from SvTaskBasedTrigger

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
```diff
@@ -0,0 +1,175 @@
...
+                val delayMillis = Random.nextLong(10000L)
+                DelayUtil
+                  .delayIfNotClosing(
+                    "sv-onboarding-observer-delay",
+                    FiniteDuration(delayMillis, TimeUnit.MILLISECONDS),
+                    this,
+                  )
+                  .onShutdown(())
+                  .flatMap { _ =>
+                    isStaleTask(task).flatMap { isStale =>
```
moritzkiefer-da: same for the staleness check

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  override protected def isStaleTask(task: Task)(implicit
+      tc: TraceContext
+  ): Future[Boolean] = {
+    store.multiDomainAcsStore
+      .lookupContractById(SvOnboardingConfirmed.COMPANION)(task.contractId)
+      .map {
+        case None => true
+        case Some(c) => c.payload.svPartyIsObserver.toScala.exists(_.booleanValue())
```
moritzkiefer-da: this can never happen can it? You're checking this in retrieveTasks already. And to change it you would neet to archive and create a new contract. So there is no way this can change for a contract you fetched in retrieveTasks

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/sv1/SV1Initializer.scala
```diff
@@ -489,26 +489,27 @@ class SV1Initializer(
...
-                        ParticipantPermission.Submission,
-                        None,
-                        None,
-                      ),
-                      serial = PositiveInt.one,
-                      isProposal = false,
-                    )
-                    .map(Some(_))
-                } else Future.successful(None)
+//              sv1PermissionTx <-
```
moritzkiefer-da: let's just delete it, too easy for it to get stuck in a commented state and we produce a mess

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1154,8 +1157,20 @@ template DsoRules with
...
         return DsoRules_ConfirmSvOnboardingResult with ..
 
+    nonconsuming choice DsoRules_MakeSvOnboardingConfirmedObserver : DsoRules_MakeSvOnboardingConfirmedObserverResult
+      with
+        svOnboardingConfirmedCid : ContractId SvOnboardingConfirmed
+        sv : Party
+      controller sv
+      do
+        _ <- getAndValidateSvParty this (Some sv)
+        res <- exercise svOnboardingConfirmedCid SvOnboardingConfirmed_MakeObserver
```
moritzkiefer-da, line 1179: you want something like a checkedFetch (ForDso svOnboardingConfirmedCid) to check that it's for the right dso party.

daml/splice-dso-governance/daml/Splice/SvOnboarding.daml
moritzkiefer-da: add a boolean to indicate whether the party is an observer. Set that to false going forward. Changing observers on the fly is just asking for something to blow up in prod.

pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
pasindutennage-da: Thank you.

Added a point to tracking issue. 

I will think about the implications.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/sv1/SV1Initializer.scala
pasindutennage-da: Fixed

daml/splice-dso-governance/daml/Splice/DsoRules.daml
pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
pasindutennage-da: fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
pasindutennage-da: fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
pasindutennage-da: fixed

pasindutennage-da:
@moritzkiefer-da 

Thank you for the feedback. I addressed all of them. PR is ready for another round of feedback.

Thanks

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
```diff
@@ -0,0 +1,110 @@
...
+  override protected def retrieveTasks()(implicit
+      tc: TraceContext
+  ): Future[Seq[Task]] = {
+    for {
+      confirmations <- store.listSvOnboardingConfirmed()
+      unobserved = confirmations.filter(co =>
+        !co.payload.svPartyIsObserver.toScala.exists(_.booleanValue())
+      )
+
+      readyTasks <- Future.sequence(unobserved.map { c =>
```
moritzkiefer-da: I think you're spawning your futures in parallel here with unbounded parallelism which is generally not what you want. Use something like MonadUtil.sequentialTraverse or one of the ones with bounded parallelism (in this case given that there is rarely more than one SV onboardig just sequentialTraverse should be totally fine)

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
```diff
@@ -174,21 +174,70 @@ class JoiningNodeInitializer(
...
+            ),
+            participantId.uid.namespace,
           )
-        ),
-      ).tupled
+      )
+
+      _ <-
+        if (config.permissionedSynchronizer && isBootstrapping) {
+          sendOnboardingRequest(svParty, dsoPartyId)
```
moritzkiefer-da, line 236: what's actually the argument for not using the new onboarding flow in all cases? You need it in a feature branch because it requires a Daml change but it seems like it should work just fine for unpermissioned synchronizers and you get much broader test coverage that way.

daml/splice-dso-governance/daml/Splice/SvOnboarding.daml
```diff
@@ -4,27 +4,32 @@
...
+    token : Text                                -- ^ An encoded and signed `SvOnboardingToken` that confirms the candidate's identity.
+    sponsor : Party                             -- ^ The established SV node that created this contract.
     dso: Party
-    expiresAt : Time                -- ^ When this contract can be archived even if the onboarding did not succeed.
+    expiresAt : Time                            -- ^ When this contract can be archived even if the onboarding did not succeed.
+    candidatePartyIsObserver : Optional Bool    -- ^ Controls whether the candidateParty can see this contract yet.
   where
     signatory dso
-    observer candidateParty
+    observer if (fromOptional False candidatePartyIsObserver) then [candidateParty] else []
```
moritzkiefer-da: this is the wrong way around, you need to default to the old behavior when the flag is not set for backwards compatibility

moritzkiefer-da (approved):
thx

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
pasindutennage-da: @moritzkiefer-da 

`BootstrapPackageConfigDarUploadIntegrationTest` and similar tests that use `private val initialPackageConfig = InitialPackageConfig.minimumInitialPackageConfig` fail because of using older packages (e.g. `UNKNOWN_INFORMEES error`).

I blindly tried to just bump the DsoGovernence `minimumInitialization: DarResource`, however, it resulted in propogating errors (packages that depend on that too needs to be upgraded, carefully).

So, I had 2 options: Just add @Ignore on top of the problamatic tests, or hide implementation behind a feature flag.

WDYT?

daml/splice-dso-governance/daml/Splice/SvOnboarding.daml
pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/SvOnboardingObserverTrigger.scala
pasindutennage-da: FIxed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/onboarding/joining/JoiningNodeInitializer.scala
moritzkiefer-da: ideally we would have a version check instead of a feature flag then so the tests using old versions fall back to the prior daml code. not super urgent though

PR #6584, opened 2026-07-28
Public Sequencer: Add a new trigger to SV app that creates ParticipantSynchronizerPermission topology txs based on MemberTraffic contracts.

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/6573


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/GrantValidatorPermissionTrigger.scala
```diff
@@ -0,0 +1,62 @@
...
+      ]
+  )(implicit tc: TraceContext): Future[TaskOutcome] = {
+    val payload = task.payload
+    val synchronizerId =
+      com.digitalasset.canton.topology.SynchronizerId.tryFromString(payload.synchronizerId)
+    val participantId = ParticipantId.tryFromProtoPrimitive(payload.memberId)
+
+    for {
+      _ <-
+        participantAdminConnection.ensureParticipantSynchronizerPermission(
```
moritzkiefer-da: This permissions based on the existence of the contract but completely ignores the amount you purchased. We need to require a minimum amount purchased (not that you need to sum across multiple contracts check how the traffic reconciliation contract handles this). Eventually this should be a daml parameter probably in amulet config. Maybe start with just a local sv app config parameter for now and add it to the tracking issue to move this to Daml.

pasindutennage-da: @moritzkiefer-da 
Thank you.
Fixed. Ready for another review :)

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -470,6 +470,7 @@ case class SvAppBackendConfig(
     // from running concurrently against the same database.  Only disable for migration scenarios
     // where intentional overlap is required.
     instanceLockEnabled: Boolean = true,
+    minMemberTrafficToOnboardValidator: Long = 100000L,
```
moritzkiefer-da, line 473: can you add a point to the tracking issue to callibrate this value

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SynchronizerFeesTestUtil.scala
```diff
@@ -95,6 +95,96 @@ trait SynchronizerFeesTestUtil extends TestCommon {
...
 
+  /** Have a validator purchase traffic on behalf of a target member. */
+
+  def buyMemberTrafficFor(
+      payerApp: ValidatorAppBackendReference,
+      targetMember: Member,
+      trafficAmount: Long,
+  )(implicit env: SpliceTestConsoleEnvironment): AmuletOperationOutcome = {
+    val ts = env.environment.clock.now
+    val payerParty = payerApp.getValidatorPartyId()
```
moritzkiefer-da: this seems confusingly complex, why do we not just use wallet.createBuyTrafficRequest?

moritzkiefer-da (approved):
thx

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
pasindutennage-da: Thanks, added a bullet to tracking Issue.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/util/SynchronizerFeesTestUtil.scala
pasindutennage-da: Fixed.

PR #6591, opened 2026-07-28
[ci] Retry event history sanity check to avoid scan ingestion race

ray-roestenburg-da:


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/plugins/EventHistorySanityCheckPlugin.scala
```diff
@@ -34,7 +35,13 @@ class EventHistorySanityCheckPlugin(
   ): Unit = {
     val initializedScans = environment.scans.local.filter(_.is_initialized)
     if (initializedScans.nonEmpty) {
-      compareEventHistories(initializedScans)
+      // Scans may still be ingesting the final updates (e.g. an SV onboarding) when the
+      // environment is torn down, so retry instead of failing on a partial history.
+      eventually(compareEventHistories(initializedScans))(
```
OriolMunoz-da: Can't tell from the comment what this accomplishes:

- If you want to wait until the second SV catches up ("failing on a partial history") with the founder, that's unnecessary, because compareEventHistories already does that:
```
val minSize = Math.min(founderHistoryToUse.size, otherScanHistory.size)
      val (otherComparable, otherRestDebug) = otherScanHistory.splitAt(minSize)
      val (founderComparable, founderRestDebug) = founderHistoryToUse.splitAt(minSize)
```

- If you're referring to the onboarding ("final updates (e.g. an SV onboarding"):
```
// we have to exclude all history before the otherScan was onboarded because:
      // - they will be missing verdicts for everything that happened before they joined (we don't do backfilling yet)
      // - they will have extra verdicts that do not involve the DSO party as part of onboarding, so the founder doesn't have them
      val founderHistorySinceOtherOnboarding = completeFounderHistory
        .dropWhile(item =>
```

ray-roestenburg-da: The failure this fixes is a third case, not either of those. In run 30083179744 the plugin threw `IllegalStateException`: SV sv2-… doesn't appear in founder history, the founder's history was missing sv2's `DsoRules_AddSv`. The minSize truncation only runs after both histories are lined up in `compareEventHistories`, and the onboarding `dropWhile` is exactly the code that throws when the common starting point is missing. The founder can be missing a long-ingested update because the events endpoint caps served history at `min(update cursor, verdict cursor)` (in `ScanEventStore.getCurrentMigrationCap`), and verdict ingestion from the mediator lags. eventually re-fetches on each attempt, so it just waits for the verdict cursor to catch up. I updated the comment.

OriolMunoz-da (approved):
thanks for the explanation!

PR #6601, opened 2026-07-29
[ci] Recover accepted DUPLICATE_COMMAND in SpliceLedgerConnection ded…

ray-roestenburg-da:
…up path

A dedup'd submission that times out client-side but succeeds on the ledger returns
`ALREADY_EXISTS`/`DUPLICATE_COMMAND` with `accepted=true` on retry. `RetryProvider` classified that
as retryable, so the call retried until its budget ran out inside the 24h dedup window and surfaced
as HTTP 409.

**Recovery.** `SpliceLedgerConnection.clientSubmit` now reads the completed transaction back by its
`completion_offset` and returns the original result. Opt-in per submission and set only by the HTTP
handlers: automation needs the duplicate to surface so its own retry re-runs the staleness check.

**Retry classification.** `RetryFor` gains `duplicateCommandIsFatal`, set on `ClientCalls` and
`InitializingClientCalls`. This is a latency improvement rather than a correctness fix, an
`accepted=true` duplicate no longer reaches the classifier at all, so it only affects `accepted=false`,
where the command genuinely never ran and 409 is the right answer, now returned at once instead of
after 12 retries.

**Not covered:** concurrent duplicates, where nothing has completed yet. Those fail with
`SUBMISSION_ALREADY_IN_FLIGHT`, which carries no offset to read back, so one call wins as before.

Fixes a real failure on main: post-merge run 30567495326 (`RollForwardLsuIntegrationTest`) shows a tap
going DEADLINE_EXCEEDED → `SUBMISSION_ALREADY_IN_FLIGHT` → `DUPLICATE_COMMAND accepted=true,
completion_offset=245` → 409 after 20.4s. Also see https://github.com/DACH-NY/cn-test-failures/issues/9435

Duplicate wallet operations with the same command id now return 200 with the original result. Release
note included; four integration tests that asserted the old failure were updated.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/RetryProvider.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/RetryProvider.scala
@@ -600,9 +600,18 @@ object RetryProvider {
               case _ => false
             }
 
+            val isDuplicateCommand = errorDetails.exists {
+              case detail: ErrorDetails.ErrorInfoDetail =>
+                (detail.errorCodeId: String) == "DUPLICATE_COMMAND"
+              case _ => false
+            }
+
             errorCategory match {
               // Pruning errors fall under FAILED_PRECONDITION which we usually retry but there is no chance to recover from it so we instead treat it as a fatal error.
               case _ if isPruningError => fatalError
+              // DUPLICATE_COMMAND: accepted duplicates are recovered centrally in SpliceLedgerConnection;
+              // rejected ones can never succeed within the dedup window — don't retry either.
+              case _ if isDuplicateCommand => fatalError
               case Some(cat) if cat.retryable.nonEmpty || extraRetryableCategories.contains(cat) =>
                 //  don't log the stack traces of transient gRPC exceptions to make the logs less noisy.
                 val msg =
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/RetryProvider.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/RetryProvider.scala
@@ -472,6 +472,7 @@ final class RetryProvider(
             "operation" -> operationId
           ),
           this,
+          retryConfig.duplicateCommandIsFatal,
         ),
       )
     }
@@ -516,6 +517,7 @@ object RetryProvider {
       metricsFactory: LabeledMetricsFactory,
       additionalMetricsLabels: Map[String, String],
       flagCloseable: FlagCloseable,
+      duplicateCommandIsFatal: Boolean,
   ) extends ExceptionRetryPolicy {
     // Additional categories that are not marked as retryable but we
     // can safely retry since we know there are other apps or
@@ -609,9 +611,9 @@ object RetryProvider {
             errorCategory match {
               // Pruning errors fall under FAILED_PRECONDITION which we usually retry but there is no chance to recover from it so we instead treat it as a fatal error.
               case _ if isPruningError => fatalError
-              // DUPLICATE_COMMAND: accepted duplicates are recovered centrally in SpliceLedgerConnection;
-              // rejected ones can never succeed within the dedup window — don't retry either.
-              case _ if isDuplicateCommand => fatalError
+              // Accepted duplicates are recovered centrally in SpliceLedgerConnection; a rejected
+              // one can never succeed within the dedup window, so client calls give up here.
+              case _ if isDuplicateCommand && duplicateCommandIsFatal => fatalError
               case Some(cat) if cat.retryable.nonEmpty || extraRetryableCategories.contains(cat) =>
                 //  don't log the stack traces of transient gRPC exceptions to make the logs less noisy.
                 val msg =
@@ -812,6 +814,7 @@ object RetryProvider {
         metricsFactory: LabeledMetricsFactory,
         additionalMetricsLabels: Map[String, String],
         flagCloseable: FlagCloseable,
+        duplicateCommandIsFatal: Boolean,
     ): ExceptionRetryPolicy
   }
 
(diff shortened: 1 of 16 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/RetryProvider.scala
```diff
@@ -600,9 +600,18 @@ object RetryProvider {
               case _ => false
             }
 
+            val isDuplicateCommand = errorDetails.exists {
+              case detail: ErrorDetails.ErrorInfoDetail =>
+                detail.errorCodeId == "DUPLICATE_COMMAND"
```
stephencompall-DA:
```suggestion
                (detail.errorCodeId: String) == "DUPLICATE_COMMAND"
```

stephencompall-DA: After #6283 I think more paranoia about `==` is generally good

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
```diff
@@ -777,6 +778,23 @@ class SpliceLedgerConnection(
...
+    val statusProto = GrpcStatusProto.fromThrowable(ex)
+    if (statusProto == null) None
+    else
+      ErrorDetails
+        .from(statusProto)
+        .collectFirst {
+          case ErrorInfoDetail(errorCodeId, metadata)
+              if errorCodeId == "DUPLICATE_COMMAND" &&
+                metadata.get("accepted").contains("true") =>
+            metadata.get("completion_offset").map(_.toLong)
```
stephencompall-DA:
```suggestion
            metadata.get("completion_offset").flatMap(co => Try(co.toLong).toOption)
```

stephencompall-DA (approved):

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/RetryProvider.scala
```diff
@@ -600,9 +600,18 @@ object RetryProvider {
               case _ => false
             }
 
+            val isDuplicateCommand = errorDetails.exists {
```
moritzkiefer-da, line 605: I'm a bit nervous about having this so deeply engrained. For most of our logic retrying is actually exactly what you want. In particular all the triggers are gonna do exactly the right thing, they see a duplicate, they retry and they discover that the task is stale and it all works fine.

The part that is broken is the 3 commands or so that live outside of triggers but still have builtin retries. Can we enable the custom retry behavior only for these?

ray-roestenburg-da: Sure let me try that tomorrow

ray-roestenburg-da: Added duplicateCommandIsFatal: Boolean so it is not used in triggers.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
```diff
@@ -998,6 +1016,20 @@ class SpliceLedgerConnection(
                     .withCause(ex.getCause)
                     .asRuntimeException
                 }
+                .recoverWith {
+                  case ex: StatusRuntimeException
```
moritzkiefer-da, line 1032: I think you also need to guard this behind the flag so it doesn't get used in triggers. In a trigger you have

retry(
staleness check,
ledger command submission,
)

right now you're still eating it in command submission and don't make it to the retry

moritzkiefer-da (approved):
thx

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SpliceLedgerConnection.scala
ray-roestenburg-da: ok let me have a look

ray-roestenburg-da: https://github.com/canton-network/splice/pull/6601/changes/1f28161a1f2074407070ab26d4a5e421684acb19 Fixes it.

ray-roestenburg-da: Please check the fix, recovery is now opt-in per submission and off by default, so a trigger gets
the duplicate back as an error and its own `retry(staleness check, submission)` re-runs as before.

ray-roestenburg-da: Enabled at HTTP handlers:

- `HttpWalletHandler`
- `HttpWalletHandlerUtil`
- `HttpValidatorAdminHandler`
- `HttpSvPublicHandler`

Everything else keeps the default, including `WalletPreapprovalSweepTrigger` and
`AcceptTransferPreapprovalProposalTrigger`.

The app-init paths have not been changed: `SvApp`, `SV1Initializer`, `SplitwellApp`,
`SpliceUtil`, `ValidatorUtil` — submit dedup'd commands under `InitializingClientCalls`, I left them off to keep this minimal.

I've kept the `RetryProvider` change, as a latency improvement, with the recovery in place an `accepted=true` duplicate never reaches the
classifier, so only `accepted=false` is affected, where the command genuinely never ran
and 409 is the right answer, now returns immediately instead of after 12 retries.

moritzkiefer-da (approved):

PR #6618, opened 2026-07-30
[ci] Close channel and snapshot stream on synchronous failures in Seq…

ray-roestenburg-da:
…uencerAdminConnection

(Noticed a possible leak, did not create an issue for it)

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
@@ -331,7 +331,7 @@ class SequencerAdminConnection(
       // a throw before the onComplete callback is registered would leak the channel
       case NonFatal(e) =>
         channel.close()
-        throw e
+        Future.failed(e)
     }
   }
 
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
@@ -68,6 +68,7 @@ import java.nio.file.{Files, Path}
 import java.util.{Base64, Collections}
 import scala.concurrent.{ExecutionContextExecutor, Future, blocking}
 import scala.jdk.CollectionConverters.*
+import scala.util.control.NonFatal
 import org.lfdecentralizedtrust.splice.store.bulk.ZstdGroupedWeight
 
 /** Connection to the subset of the Canton sequencer admin API that we rely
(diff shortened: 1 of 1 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
```diff
@@ -161,13 +162,20 @@ class SequencerAdminConnection(
...
+    try {
+      runCmd(
+        SequencerAdminCommands.InitializeFromLsuPredecessor(
+          inputStream,
+          staticSynchronizerParameters,
+          ignorePsidCheck,
+        )
+      ).andThen(_ => inputStream.close())
+    } catch {
+      // a synchronous failure never completes the future, so close the stream here
```
nicu-da: Do we have an example of how can this happen? The issue would be actually worse than this as the exception would escape the future context and probably propagate all the way to the executor service and just be silently ignored. So we actually have to make sure that we don't throw any exceptions outside the future context if we return a future

ray-roestenburg-da: The catch only happens if the exception happens in synchronous execution, so anything that fails before actually creating the future, in any of the code called here.

nicu-da: Well yes, but that means that runCmd can have a failure that is thrown outside the future, so we need to actually fix that as no client of this api will catch exceptions thrown outside the future

ray-roestenburg-da: I did not do this as a response to a specific error in runCmd but simply as a way of preventing leak of inputstream.

ray-roestenburg-da: (So I don't see how this extra handling is worse, at the least we don't leak the inputstream)

nicu-da: If this can happen and leaking the input stream is the least of our worries, as we will simply lose failures. 
So either it cannot happen or we need to actually fix runCmd, as this is the equivalent of tending to your broken nail when you have an open wound.

ray-roestenburg-da: Right, so runCmd can't actually throw it looks like, I think you are right, I'll remove this change

ray-roestenburg-da: removed, please have another look @nicu-da

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
```diff
@@ -302,29 +303,36 @@ class SequencerAdminConnection(
...
-        request
-          .createRequestInternal()
-          .getOrElse(throw new IllegalStateException("Unable to create internal request.")),
-        (req: OnboardingStateV2Request, obs: StreamObserver[OnboardingStateV2Response]) =>
-          stub.onboardingStateV2(req, obs),
-      )
-      .map { response =>
-        val proto: ByteString = response.onboardingStateForSequencer
-        PekkoByteString(proto.asReadOnlyByteBuffer())
+    try {
```
ray-roestenburg-da, line 306: In this case there is definitely more that could fail synchronously I think @nicu-da and would prefer to have the try catch in place.

nicu-da: if we're sure that it can fail then on the catch we should return a Future.failed not rethrow

ray-roestenburg-da: yeah ok that makes sense, let me change that.

ray-roestenburg-da: Fixed

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/SequencerAdminConnection.scala
```diff
@@ -302,29 +303,36 @@ class SequencerAdminConnection(
...
       }
-      .via(
-        ZstdGroupedWeight(compressionLevel = 3, minSize = chunkSize.toLong)
-      ) // 3 is the default zstd compression level
-    val storageObject = source.runWith(sink)
-    storageObject.onComplete { _ =>
-      channel.close()
+      storageObject
+    } catch {
+      // a throw before the onComplete callback is registered would leak the channel
```
nicu-da, line 331: note that the throw would still need to be outside the future context, as even if the future fails before the onComplete is called, it will just run immediately if the future is already completed

nicu-da (approved):

PR #6621, opened 2026-07-30
Add the ability to restrict whitelisting only to the required hosts

nicu-da:
Instead of allowing access through the general whitelist we add specific scan/sv app/sequencer/cantonbft allow lists that target only the ip that are required and keep the general whitelisting only for internal ips.

cantonbft is exposed only to the svs

[static]

drafting this and will merge and enable by default when i come back from holiday 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/pulumi/infra/src/istio.ts
+++ b/cluster/pulumi/infra/src/istio.ts
@@ -3,13 +3,11 @@
 import * as gcp from '@pulumi/gcp';
 import * as k8s from '@pulumi/kubernetes';
 import * as pulumi from '@pulumi/pulumi';
-import * as assert from 'assert/strict';
 import {
   allSvsToDeployBasic,
   coreSvsToDeployBasic,
 } from '@canton-network/splice-pulumi-common-sv/src/svConfigsBasic';
 import { cometBFTExternalPort } from '@canton-network/splice-pulumi-common-sv/src/synchronizer/cometbftConfig';
-import { spliceConfig } from '@canton-network/splice-pulumi-common/src/config/config';
 import { mergeWith } from 'lodash';
 
 import {
@@ -25,7 +23,11 @@ import {
   isDevNet,
   isMainNet,
 } from '../../common';
-import { clusterBasename, infraConfig, loadIPRanges } from './config';
+import { clusterBasename, infraConfig } from './config';
+import { configureIstioGatewayPolicies, installAppWhitelisting } from './whitelisting';
+import { loadInternalWhitelistedIps, loadIPRanges } from './whitelisting/ipRanges';
+import { configurePublicInfo } from './whitelisting/publicInfo';
+import { configurePublicTokenRegistry } from './whitelisting/publicTokenRegistry';
 
 interface ConfiguredIstio {
   allResources: pulumi.Resource[];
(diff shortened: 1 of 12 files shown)


canton-network-da:
# [backport] Reminder #
- [ ] release-line-0.6.14
- [ ] release-line-0.6.13
- [ ] release-line-0.6.12
- [ ] release-line-0.6.11

martinflorian-da:
This is for https://github.com/DACH-NY/canton-network-internal/issues/359 , correct?

nicu-da:
Yes, somehow I missed linking the issue

cluster/pulumi/infra/src/whitelisting/sequencer.ts
```diff
@@ -0,0 +1,51 @@
...
+import { loadIPRanges } from './ipRanges';
+import { CLUSTER_INGRESS_NS, createIstioIpAllowPolicies, istioIngressSelector } from './policies';
+
+export function configureSequencerWhitelist(): pulumi.Output<pulumi.Resource[]>[] {
+  const dnsNames = [getDnsNames().cantonDnsName, getDnsNames().daDnsName];
+  const migrations = DecentralizedSynchronizerUpgradeConfig.runningMigrations();
+
+  const publicApiHosts = allSvsToDeployBasic.flatMap(sv =>
+    migrations.flatMap(migration =>
+      dnsNames.map(dns => `sequencer-${migration.id}.${sv.ingressName}.${dns}`)
```
martinflorian-da: TIL that you also need to add `hostname:*`, not just `hostname`, otherwise clients that dial `hostname:443` (at least some grpc implementations, if not all?) will not be caught.

https://istio.io/latest/docs/ops/best-practices/security/#writing-host-match-policies

cluster/pulumi/infra/src/istio.ts
```diff
@@ -975,6 +826,7 @@ export function configureIstio(
     ingressNs.ns
   );
   const sequencerFlowControl = configureSequencerFlowControl(ingressNs.ns);
+  installAppWhitelisting();
```
martinflorian-da: Any reason not to pass `ingressNs.ns` here as well instead of hard coding that name? Seems more consistent and I'm also not sure whether we don't lose some dependency information otherwise so that pulumi could decide to create things out of order.

martinflorian-da (commented):
Thanks! I need a bit more time to think this through...

Any reason not to run a cluster test to confirm it doesn't break the status quo?

nicu-da:
> Thanks! I need a bit more time to think this through...
> 
> Any reason not to run a cluster test to confirm it doesn't break the status quo?

sure let me do that. 
i did run a manual deployment with the flag off and then updated to turn it on

cluster/pulumi/infra/src/whitelisting/sequencer.ts
martinflorian-da, line 1: So "architecture"-wise, you're splitting by service and then by visibility.

I wonder if you split first by visibility and then by service (so that this file here would be split into the "SVs-only" and "all validators" parts) the end result on the cluster might be more clear + you could avoid duplicating the full list of IPs (i.e., you'd call `loadIPRanges()` only once possibly).

martinflorian-da (approved):
Thanks!

martinflorian-da:
> Thanks!

(And yes I'd still suggest ensuring that a CI-run cluster test actually finishes...)

cluster/pulumi/infra/src/whitelisting/sequencer.ts
nicu-da: Is there any real downside to reading the ip ranges multiple times? Would be easy to factor out but I don't see the benefit tbh

nicu-da:
> > Thanks!
> 
> (And yes I'd still suggest ensuring that a CI-run cluster test actually finishes...)

Sure, once gcp plugs in more servers

cluster/pulumi/infra/src/whitelisting/sequencer.ts
martinflorian-da: I'm less concerned about the reading than about the state on k8s. So a benefit would be "it's easier to understand/edit when you're looking at the CRs". Maybe there is also some minor speed-up of deploys, but that would be marginal I expect.

Also the code might become a bit simpler to read as you see all the "public" and "SV-only" things in one place. If we keep that as a convention it will be a bit harder to accidentally whitelist something for the wrong audience, or to forget that some service is whitelisted in a given way.

nicu-da: I tend to disagree, the way I see it we always refer of certain services/apps being whitelisted, not specifically to whom we open the whitelist (and this is also the more granular view thus you get much better visibility when you list all the access filter you have defined in k8s for example). 
That way you couple the whitelisting itself with the app deployment (I see this whitelisting as moving away in the future to reside next to the service definition itself).

> Also the code might become a bit simpler to read as you see all the "public" and "SV-only" things in one place. If we keep that as a convention it will be a bit harder to accidentally whitelist something for the wrong audience, or to forget that some service is whitelisted in a given way.

This imo just increases the chances of adding an endpoint to the wrong whitelist and opening it up by mistake.

martinflorian-da: >  couple the whitelisting itself with the app deployment (I see this whitelisting as moving away in the future to reside next to the service definition itself).

That is a nice idea!

Happy to move ahead like this then; no strong feelings here.

PR #6629, opened 2026-07-30
Public Sequencer: Add scan endpoint to lookup if a given participant id is permissioned

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/6628


moritzkiefer-da (approved):
thx

PR #6631, opened 2026-07-31
add per ip rate limits

### LINKED ISSUE #6599: Allow specifying custom ratelimits for provided IPs, in Istio
## Background

Currently, we have setup scan rate limits using the following config.

```
  /registry/metadata/v1/info:
    name: registry-metadata-info
    type: limited
    maxTokens: 720
    tokensPerFill: 720
    fillInterval: 60s
    perIpLimits:
      maxTokens: 120
      tokensPerFill: 120
      fillInterval: 60s
```

This configuration supports setting two types of rate limit buckets for a given end point.

1. per-endpoint maxTokens: This is the maximum rate allowed across **all** source IPs, for this endpoint.
2. per-endpoint-per-ip maxTokens: This is the maximum rate for a given source IP, for this endpoint.

We opted the above two modes to satisfy two properties.

1. To enforce maximum resource usage per-end point
2. To enforce fairness across IPs (to avoid one IP taking the entire available capacity)

## Problem

In our test clusters and also in production, we end up sending higher amount of traffic from some IPs, compared to rest of IPs used in that network. Examples are as follows.
 
1. CILR: Multi-validators host many validators (and hence many participants), and hence naturally send more traffic than a single validator node. 
2. DevNet:  Multi-validators send more traffic than a single validator node. Also on DevNet we test with higher load compared to the MainNet, hence, naturally per-ip limits configured for MainNet are not enough.

## Non-Solution 

Bumping the rate limits will fix the problem, however, it increases the attack surface, hence not preferred. 

## Solution 

We need to support configuring rate limits for a specific IP, so that for IPs that are known to send higher amount of traffic will get higher rate limits. An example configuration should look like the following.

```
  /registry/metadata/v1/info:
    name: registry-metadata-info
    type: limited
    maxTokens: 720
    tokensPerFill: 720
    fillInterval: 60s
    perIpLimits:
      maxTokens: 120
      tokensPerFill: 120
      fillInterval: 60s
    ipLimit192-68-78-50:
      maxTokens: 220
      tokensPerFill: 220
      fillInterval: 60s
    ipLimit192-68-78-51:
      maxTokens: 250
      tokensPerFill: 220
      fillInterval: 60s 

```

krzysztofczyz-da:
Fixes https://github.com/canton-network/splice/issues/6599

This PR adds:
- per IP rate limits
- validation for the schema

example usage:
```
perIpLimits:
  maxTokens: 120
  tokensPerFill: 120
  fillInterval: 60s
  overrides:
    test:
      ips:
        - 192.68.78.50
      maxTokens: 220
      tokensPerFill: 220
      fillInterval: 60s
```

Note: the change is backwards-compatible.
Relevant doc: [link](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter#using-rate-limit-descriptors-for-local-rate-limiting)
Example config change: [link](https://github.com/canton-network/splice/pull/6631/changes/0dc6baf444fa541f2a2d470fff38c4dad33e55a4)

### DIFF
--- a/cluster/deployment/mock/config.yaml
+++ b/cluster/deployment/mock/config.yaml
@@ -34,18 +34,6 @@ multiValidator:
 sv:
   synchronizer:
     skipInitialization: true
-  scan:
-    externalRateLimits:
-      rateLimits:
-        /registry/metadata/v1/info:
-          perIpLimits:
-            overrides:
-              test:
-                ips:
-                  - 192.68.78.50
-                maxTokens: 250
-                tokensPerFill: 250
-                fillInterval: 60s
 splitwell:
   maxDarVersion: '0.1.8'
   participantPruningSchedule:
--- a/cluster/deployment/mock/config.yaml
+++ b/cluster/deployment/mock/config.yaml
@@ -34,6 +34,18 @@ multiValidator:
 sv:
   synchronizer:
     skipInitialization: true
+  scan:
+    externalRateLimits:
+      rateLimits:
+        /registry/metadata/v1/info:
+          perIpLimits:
+            overrides:
+              test:
+                ips:
+                  - 192.68.78.50
+                maxTokens: 250
+                tokensPerFill: 250
+                fillInterval: 60s
 splitwell:
   maxDarVersion: '0.1.8'
   participantPruningSchedule:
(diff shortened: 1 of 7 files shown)


moritzkiefer-da (approved):
thanks! lgtm. I suggest adding it to the mock config just so we have an example.

And then let's backport to 0.7.0 and set it on devnet to allow traffic from our multi-validators.

PR #6648, opened 2026-08-01

canton-network-da:


mblaze-da:
@martinflorian-da WDYT about making merging these PRs a responsibility of the person currently on CI monitoring? The impact is on CI infra and that is their responsibility either way. Currently it has been a bit awkward that we merge this and notify them regardless.

martinflorian-da:
@mblaze-da I like that idea; we might still need to have default reviewers and ask those to hot potato this to the right person on rotation

pawelperek-da (approved):
yup

PR #6654, opened 2026-08-03
Public Sequencer: Add a decentralized automation to SV app to look for ValidatorLicenseRequest and create ValidatorLicense

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/6653


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
```diff
@@ -0,0 +1,161 @@
...
+      )
+
+      partyToParticipant <- participantAdminConnection.listPartyToParticipant(
+        store = Some(TopologyStoreId.Synchronizer(synchronizerId)),
+        filterParty = validatorParty.filterString,
+      )
+
+      outcome <- partyToParticipant.headOption.flatMap(_.mapping.participantIds.headOption) match {
+        case None =>
+          Future.successful(
```
moritzkiefer-da: don't you want to reject and archive in that case or something like that?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
```diff
@@ -0,0 +1,161 @@
...
+              s"Skipping as participant ID for $validatorParty is not yet known to local topology"
+            )
+          )
+        case Some(participantId) =>
+          participantAdminConnection
+            .listParticipantSynchronizerPermission(synchronizerId, participantId.filterString)
+            .flatMap { permissions =>
+              permissions.headOption.map(_.mapping) match {
+                case None =>
+                  Future.successful(
```
moritzkiefer-da: if the participant can create the license request how can it not have been permissioned?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
```diff
@@ -0,0 +1,161 @@
...
+                    TaskSuccess(
+                      s"Skipping as participant $participantId does not have ParticipantSynchronizerPermission"
+                    )
+                  )
+                case Some(permission) =>
+                  val isLoginAfterPassed =
+                    permission.loginAfter.forall(loginAfter => context.clock.now >= loginAfter)
+
+                  if (isLoginAfterPassed) {
```
moritzkiefer-da: I don't really see why we need this, licenses are not there to control permissioning. the only reason we delay them is because we literally cannot create them before the party is in the topology state. once it is I don't think we need to be careful here.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
```diff
@@ -0,0 +1,161 @@
...
+  )(implicit tc: TraceContext): Future[TaskOutcome] = {
+
+    val action = new ARC_DsoRules(
+      new SRARC_GrantValidatorLicense(
+        new DsoRules_GrantValidatorLicense(reqCid)
+      )
+    )
+
+    for {
+      queryResult <- dsoStore.lookupConfirmationByActionWithOffset(svParty, action)
```
moritzkiefer-da: I think you need to check for both confirmations and existing licenses and then take the minimum offset of both.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
pasindutennage-da: This is a tricky -- in the eyes of *this sv*, if it has no partyToParticipant, then it can mean 2 things: 

1) SV participant is lagging: SV cannot make a decision, so just log and give up 
2) the ValidatorLicenseRequest is a bad request: Archive it

Above 2 states in the eyes of *this SV* are the same/indistinguishable, and hence, the  safe way is to Archive the ValidatorLicenseRequest, however, using `SRARC_` voting, by 2f+1 SVs. Is this what you meant?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
pasindutennage-da: again, this can happen with out of order message delievery -- topology propogation is slow, while ValidatorLicenseRequest was delievered first.

Unless Canton has a strick causal ordering of contract delievery and topology delievery (permissiontopology is received before ValidatorLicneseRequest), it is possible to end up in this branch.

so again, just dropping it is the safe way?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
moritzkiefer-da: > SV participant is lagging: SV cannot make a decision, so just log and give up

Always good to think about lagging nodes. I'm not sure that is a problem here though.

The guarantee you get is that a sequencer will deliver you messages in the order they got sequenced.

Here we have two messages:
1. The topology transaction
2. The Daml transaction

The Daml transaction cannot go through if the party is not known so we know the topology transaction has been sequenced before.

So if you lag you would not have seen both but you cannot have received the Daml transaction without having received the topology transaction.

If you're concerned about the local node reordering things after receiving it from the sequencer that's also not an issue, the participant also needs to process topology first or would blow up on the transaction as it does not know the party.

pasindutennage-da: Thanks, removed PartyToParticipant and Permission checks

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
pasindutennage-da: Removed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
```diff
@@ -0,0 +1,125 @@
...
+            )
+          )
+        case (Some(_), _) =>
+          Future.successful(
+            TaskSuccess(
+              s"Skipping as confirmation from $svParty is already created for granting validator license to $validatorParty"
+            )
+          )
+        case (None, None) =>
+          val minOffset = Seq(licenseResult.offset).foldLeft(confirmationResult.offset)(math.min)
```
moritzkiefer-da:
```suggestion
          val minOffset = math.min(licenseResult.offset, confirmationResult.offset)
```
am I missing something? I don't understand what you're trying to do with the more complex version

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
```diff
@@ -0,0 +1,125 @@
...
+        _.exerciseDsoRules_ConfirmAction(
+          svParty.toProtoPrimitive,
+          action,
+        )
+      )
+      outcome <- (confirmationResult.value, licenseResult.value) match {
+        case (_, Some(_)) =>
+          Future.successful(
+            TaskSuccess(
+              s"Skipping as a ValidatorLicense already exists for $validatorParty"
```
moritzkiefer-da: We're still missing the part where we reject in those cases, otherwise those contracts just accumulate and you pile up more and more.

moritzkiefer-da (approved):
thx

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
pasindutennage-da: Sure, will add.

for `reject`, we would need  `SRARC_` type voting by 2f+1 SVs right, or do you think whether a single SV can unilaterally achieve the request?

moritzkiefer-da: yep srarc vote si right

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/confirmation/ValidatorLicenseRequestTrigger.scala
pasindutennage-da: Fixed

PR #6660, opened 2026-08-04
add rate limits by CIDRs

### LINKED ISSUE #6640: per CIDR rate limits
After completion of https://github.com/canton-network/splice/issues/6599 we've realized that we would actually like to have a per CIDR rate limit so that we can whitelist a range of IPs.
The point of this ticket is to implement this.

Then change the ratelimits on devnet to use this new config (like in https://github.com/DACH-NY/canton-network-internal/pull/6342)

krzysztofczyz-da:
Fixes https://github.com/canton-network/splice/issues/6640


canton-network-da:
# [backport] Reminder #
- [ ] release-line-0.7.0
- [ ] release-line-0.6.14
- [ ] release-line-0.6.13
- [ ] release-line-0.6.12

martinflorian-da:
@krzysztofczyz-da FYI you didn't actually approve a cluster test; was this on purpose?

cluster/deployment/mock/config.yaml
```diff
@@ -46,6 +46,20 @@ sv:
...
+          # Example: give a CIDR range a shared higher limit.
+          # This uses Envoy's remote_address_match action, which checks the
+          # connection's remote address against the configured CIDR ranges.
+          perCidrLimits:
+            maxTokens: 500
+            tokensPerFill: 500
+            fillInterval: 60s
+            overrides:
+              validator-net:
+                cidrs:
```
martinflorian-da, line 61: Did we consider removing `ips:` altogether? IPs are just `/32` CIDRs, so this would reduce complexity. Is there some downside here I'm missing?

martinflorian-da: Also, more nit level but for consistency with e.g. https://github.com/canton-foundation/configs-private which will likely become our upstream source of truth somehow someday: can we call these "IP ranges" in the config format? So `perIpRangeLimit` etc.

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
```diff
@@ -123,6 +170,37 @@ export function validateIpLimits(pathPrefix: string, rateLimit: LocalLimit<Match
...
+export function validateCidrLimits(pathPrefix: string, rateLimit: LocalLimit<MatchedLimits>): void {
+  if (!rateLimit.perCidrLimits) {
+    return;
+  }
+
+  const cidrs: { cidr: string; overrideKey: string }[] = [];
+
+  Object.entries(rateLimit.perCidrLimits.overrides || {}).forEach(([overrideKey, override]) => {
+    override.cidrs.forEach(cidr => {
+      cidrs.push({ cidr, overrideKey });
```
martinflorian-da, line 185: Does this mean that if we have something like this:

```
          perCidrLimits:
            maxTokens: 500
            tokensPerFill: 500
            fillInterval: 60s
            overrides:
              validator-net:
                cidrs:
                  - 1.2.3.4/24
                  - 5.6.7.8/24
```

...then we can have 500 req/minute for 1.2.3.4 **and** 500 req/minute for 5.6.7.8?

If so my feedback is that this likely going to be misunderstood; and then we might also just not make this an array but allow only a single range?

That said, I think I like the idea of bundling multiple IPs into one override. Is there a way for them to share the same rate, so that in above example we'd get 500 req/minute for **either** 1.2.3.4 **or** 5.6.7.8.

In any case, my strongest ask here is to document the semantics so it's clear what multiple CIDRs means.

cluster/deployment/mock/config.yaml
krzysztofczyz-da: > Did we consider removing ips: altogether?

The only thing about this is I wanted to make sure the change is backwards compatible to not nuke something on accident.
> can we call these "IP ranges" in the config format? So perIpRangeLimit etc.

If we want a breaking change that wouldn't be a problem and `perIpRangeLimit` SGTM.

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
krzysztofczyz-da: I've changed the semantics so that all CIDRs under a single override key share one token bucket. In your example, validator-net would get 500 req/min total, shared across `1.2.3.4/24` and `5.6.7.8/24`.
I also added comments documenting how this works.

cluster/deployment/mock/config.yaml
martinflorian-da: Breaking changes are super fine here! We're the only users I know of, and it's going to be breaking in a loud way so likely we'll notice and adapt our configs.

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
```diff
@@ -43,10 +46,47 @@ type LocalLimit<L> = {
   name: string;
 } & L;
 
-// This is arbitrary, but must not match any limit `name` used for an EnvoyFilter
-// above. All existing manual YAML entries use 'client_ip' so this is the nicest
-// migration away from always specifying that.
-const clientIpEntryKey = 'client_ip';
```
pasindutennage-da: @krzysztofczyz-da perhaps for future reference, can we comment here why we got rid of client_ip and xff?

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
```diff
@@ -43,10 +46,47 @@ type LocalLimit<L> = {
...
+  const [network, prefixStr] = cidr.split('/');
+  const prefix = parseInt(prefixStr, 10);
+  const networkLong = ipToLong(network);
+  const hostBits = 32 - prefix;
+  const hostCount = 1 << hostBits;
+  const networkMask = ~(hostCount - 1) >>> 0;
+  const normalizedNetworkLong = networkLong & networkMask;
+  const broadcastLong = normalizedNetworkLong | ((hostCount - 1) >>> 0);
+  return { networkLong: normalizedNetworkLong, prefix, broadcastLong };
+}
```
pasindutennage-da, line 102: so this means, if ever 2 ranges intersect, then one should divide it into 3 CIDRs so that it doesn't?

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
```diff
@@ -43,10 +46,47 @@ type LocalLimit<L> = {
...
 
-// This is arbitrary, but must not match any limit `name` used for an EnvoyFilter
-// above. All existing manual YAML entries use 'client_ip' so this is the nicest
-// migration away from always specifying that.
-const clientIpEntryKey = 'client_ip';
+// Envoy native descriptor key for IP-range-based rate limiting using
+// remote_address_match. The action checks the connection's remote address
+// against configured IP ranges and produces this descriptor entry.
+const remoteAddressMatchKey = 'remote_address_match';
+const remoteAddressMatchDefaultValue = 'per-ip-range-default';
```
pasindutennage-da, line 56: Don't we need to add checks to make sure per-ip-range-default is  reserved? (i.e. what if someone names the limits by this?)

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
```diff
@@ -43,10 +46,47 @@ type LocalLimit<L> = {
...
+    address_prefix: network,
+    prefix_len: { value: parseInt(prefixStr, 10) },
+  };
+}
+
+function ipToLong(ip: string): number {
+  return ip.split('.').reduce((acc, octet) => (acc << 8) + parseInt(octet, 10), 0) >>> 0;
+}
+
+function parseCidr(cidr: string): { networkLong: number; prefix: number; broadcastLong: number } {
```
pasindutennage-da, line 73: did you check whether this doesn't break with 0.0.0.0/0?

pasindutennage-da (changes requested):
Thank you.

PR makes sense to me. 

@martinflorian-da  and or @nicu-da  Can one of you please take one more look, before approval?

We also need to manually test this

Use small rate limits like 5 req/min so that you can trigger the rate limits. 

Lets make sure we cover all cases though 

1. a range with just 1 CIDR,
2. multiple CIDRs to see if they all fall in to same bucket
3. no CIDR so that we make sure the default per-ip still works,
4. check per-end point rate limits works as before
5. check for global rate limit

cluster/configs/shared/rate-limits/token-registry.yaml
```diff
@@ -5,7 +5,7 @@ rateLimits:
     maxTokens: 720
     tokensPerFill: 720
     fillInterval: 60s
-    perIpLimits:
+    perIpRangeLimit:
       maxTokens: 120
       tokensPerFill: 120
       fillInterval: 60s
```
nicu-da, line 11: @pasindutennage-da lets make sure we also document more explicitly that we rate limit on a 60s interval

nicu-da: also we seem to be rate limiting per endpoint, what's our plan on documenting this for the SVs? which endpoints to include and how we notify of new endpoints

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
```diff
@@ -100,41 +140,43 @@ function validateEndpointCoverage(
...
+  Object.entries(rateLimit.perIpRangeLimit.overrides || {}).forEach(([overrideKey, override]) => {
+    override.ipRanges.forEach(ipRange => {
+      ipRanges.push({ ipRange, overrideKey });
     });
   });
 
-  if (duplicates.length > 0) {
-    throw new Error(`${pathPrefix}: duplicate IPs in per-IP rate limits: ${duplicates.join(', ')}`);
+  const overlaps: string[] = [];
+  for (let i = 0; i < ipRanges.length; i++) {
```
nicu-da: let simplify this by using a foreach? (and remove the i,j notations)

nicu-da (commented):
overall lgtm, leaving the more in depth review to @pasindutennage-da and the cluster test

cluster/configs/shared/rate-limits/token-registry.yaml
pasindutennage-da: @nicu-da 

Sure, will make 1 minute explicit in docs.

Right now, we do it coarse grained - just mentioning that all registry/ end points are public and that each end point should have x global limit and y per ip limit

Obviously we should at some point go for more fine grained: higher limits for end points that are constantly hit. 

How about once this PR is merged, I make a google doc, and there we agree on a better structure?

nicu-da: sgtm! 
my  issue with mentioning the limits per endpoint is that that's pretty non descriptive, I think we need more specific docs.

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
krzysztofczyz-da: added 👌

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
krzysztofczyz-da: makes sense - added & tested 👌

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
krzysztofczyz-da: Changed the bit magic a bit to make it work and added a test - thanks ❤️ 
Also added comments to make it clearer what it does 👌

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
krzysztofczyz-da: Yes, that's to ensure that we don't have one IP in many buckets which probably would be confusing.

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
krzysztofczyz-da: Done 👌

pasindutennage-da:
@krzysztofczyz-da Thanks.

Lets first manually test if rate limits are hit for all 5 cases above?

martinflorian-da:
> We also need to manually test this

@pasindutennage-da fwiw we gave up on testing also other istio rate limiting things... Perhaps better to merge this as is to avoid it rotting / becoming stale, and revisit once we have better tooling. I just added a pointer on https://github.com/DACH-NY/canton-network-internal/issues/1936 (FYI also @nicu-da)

image

cluster/pulumi/common/src/ratelimit/rateLimitSchema.ts
```diff
@@ -9,21 +9,35 @@ export const BucketRateLimitSchema = z.object({
...
+  cidr => {
+    const parts = cidr.split('/');
+    if (parts.length !== 2) {
+      return false;
+    }
+    const [network, prefixStr] = parts;
+    if (isIP(network) !== 4) {
+      return false;
+    }
+    const prefix = parseInt(prefixStr, 10);
```
pasindutennage-da: Nits: Check if `parseInt` will handle non-numeric trailings correctly.  If not validator before passing.

cluster/pulumi/common/src/ratelimit/rateLimitSchema.ts
```diff
@@ -9,21 +9,35 @@ export const BucketRateLimitSchema = z.object({
   fillInterval: z.string(),
 });
 
-const Ipv4AddressSchema = z.string().refine(ip => isIP(ip) === 4, {
-  message: 'Expected IPv4 address',
-});
+const IpRangeSchema = z.string().refine(
+  cidr => {
+    const parts = cidr.split('/');
```
pasindutennage-da, line 14: Nits: Updated schema assumes that there is always x.x.x.x/32 format for single IPs. Perhaps add a more descriptive message to explicitly warn that single ips must have /32 ?

pasindutennage-da (approved):

cluster/pulumi/common/src/ratelimit/rateLimitSchema.ts
krzysztofczyz-da: added

cluster/pulumi/common/src/ratelimit/rateLimitSchema.ts
krzysztofczyz-da: added 👌

PR #6705, opened 2026-08-06
Close stream in TemplateJsonDecoder

OriolMunoz-da:
Randomly stumbled upon these in a heap dump. Not a big problem because of the cache preventing them from leaking all over, but worth fixing.


moritzkiefer-da (approved):
thx

Issue #6690, opened 2026-08-06
TBAR: In 4SV simtime test bootstrap, the round advancement get stuck due to missing meta row


dfordivam:
For the SVs other than the first SV; the initialization of the `AppActivityRecordMetaT` table requires observing the round data from `ScanRewardsReferenceStore`.

In the absence of meta row, the reward processing would be stuck endlessly as the SVs own scan http-handler would return `Undetermined` for the root-hash and totals, and it won't do BFT read from SV0. Moreover the meta row won't be added unless more verdicts get ingested.

The root cause of the this is the `advanceRoundsToNextRoundOpening` API, being used in most of the simtime tests, would get be stuck waiting for the summarizing round to finish processing, and it will prevent any further network activity/verdict ingestion, which would have eventually created the meta row, and thus allowed BFT read to happen.

dfordivam:
Note that the issue surfaced after the [TBAR: Guard meta row insertion on traffic summaries](https://github.com/canton-network/splice/pull/6579/changes#top) #6579  PR, because before this PR the meta-row's `earliest_ingested_round` would have been initialized via `lastArchivedRoundO` right when the round 0 was archived. But this working behaviour must have been flaky as the verdict ingestion does not wait for the `ScanRewardsReferenceStore` to have ingested the round 0 archival. So after the 6579 the issue started happening consistently and not as a flake.

adetokunbo:
As discussed, part of resolution of this will be to implement the solution proposed for #6065

adetokunbo:
FYI #6708 when complete should resolve this, now that #6392 is submitted.  
In particular, [see this comment](https://github.com/canton-network/splice/pull/6708#discussion_r3739758001) and some of the others

PR #6579, opened 2026-07-28
TBAR: Guard meta row insertion on traffic summaries

dfordivam:
Fixes #6463 
Fixes #6499 

- Add test for first SV scenario
- Fix a comment
#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If an upgrade test is required, comment `/upgrade_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a logical synchronizer upgrade test is required (from canton-3.5), comment `/lsu_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/canton-network/splice/blob/main/docs/src/release_notes.rst).
- [ ] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/canton-network/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [ ] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #6694, opened 2026-08-06
Traffic-Based App Rewards: Clearly distinguish not-ingestable from not-yet-ingested

adetokunbo:
As discussed earlier this is

- a part of the solution for #6690
- discussed previously as a potential solution for #6065

The new method on `AppActivityStore` moves the logic for determining the result from the handler into the trait; it returns a case class of two values because that looks clearer;  alternatively, the result could be a simple boolean; let me know if that's preferable

#### Cluster Testing
- [ ] If a cluster test is required, comment `/cluster_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If an upgrade test is required, comment `/upgrade_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a hard-migration test is required (from the latest release), comment `/hdm_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.
- [ ] If a logical synchronizer upgrade test is required (from canton-3.5), comment `/lsu_test` on this PR to request it, and ping someone with access to the DA-internal system to approve it.

#### PR Guidelines
- [ ] Include any change that might be observable by our partners or affect their deployment in the [release notes](https://github.com/canton-network/splice/blob/main/docs/src/release_notes.rst).
- [x] Specify fixed issues with `Fixes #n`, and mention issues worked on using `#n`
- [ ] Include a screenshot for frontend-related PRs - see [README](https://github.com/canton-network/splice/blob/main/TESTING.md#running-and-debugging-integration-tests) or use your favorite screenshot tool

#### Merge Guidelines
- [x] Make the git commit message look sensible when squash-merging on GitHub (most likely: just copy your PR description).

PR #6708, opened 2026-08-07
Scan app: Activity meta row earliest_round fixes

dfordivam:
This does some fixes to the changes introduced in #6579, and partly fixes the issue described in #6690

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
@@ -321,7 +321,18 @@ class DbAppActivityRecordStore(
 
     // earliestRound: the oldest round open at the earliest record_time of this batch.
     // or (-1) on firstSV, as it is expected to have complete data for the first round.
-    val earliestRound = if (isFirstSv) Some(-1L) else firstActiveRoundO
+    //
+    // Its possible that we are processing a batch of verdicts where the
+    // earliest record_time is before the opensAt time of the earliest ingested
+    // rounds in the ScanRewardsReferenceStore. In this case firstActiveRoundO
+    // would be None, and it is always safe to consider the round info from
+    // items or the lastArchivedRoundO as a fallback.
+    val earliestRound =
+      if (isFirstSv) Some(-1L)
+      else
+        firstActiveRoundO
+          .orElse(items.map(_.roundNumber).minOption)
+          .orElse(lastArchivedRoundO)
 
     // lastArchived: the highest round archived as of this verdict batch.
     //   - From the caller when available
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/ScanVerdictIngestionService.scala
@@ -256,7 +256,7 @@ class ScanVerdictIngestionService(
                   }
                   firstActiveRoundO <- recordTimes.minOption match {
                     case Some(minRecordTime) =>
-                      appActivityComputation.lookupActiveOpenMiningRound(minRecordTime)
+                      appActivityComputation.lookupEarliestIngestedRoundCandidate(minRecordTime)
                     case None => Future.successful(None)
                   }
                   lastArchivedRoundO <- recordTimes.maxOption match {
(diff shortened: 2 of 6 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbAppActivityRecordStore.scala
```diff
@@ -321,7 +321,18 @@ class DbAppActivityRecordStore(
...
+    // earliest record_time is before the opensAt time of the earliest ingested
+    // rounds in the ScanRewardsReferenceStore. In this case firstActiveRoundO
+    // would be None, and it is always safe to consider the round info from
+    // items or the lastArchivedRoundO as a fallback.
+    val earliestRound =
+      if (isFirstSv) Some(-1L)
+      else
+        firstActiveRoundO
+          .orElse(items.map(_.roundNumber).minOption)
+          .orElse(lastArchivedRoundO)
```
rautenrieth-da, line 335: I don't like the fact that we changed this part of the code so many times already, and that I can't immediately tell if it's correct now. Can you please check my understanding below?

1. We are assuming that scan never has any holes in activity records. During a short period after SV bootstrap, scan _might_ be missing activity records. However, once the first activity record is ingested in the store, all subsequent activity records _must_ (eventually) be ingested as well.
2. Column `earliest_ingested_round` in `table app_activity_record_meta` stores the round at which the activity record ingestion started. It is used to answer the question whether the database contains all app activity records for a round, in the three flavors `earliestRoundWithCompleteAppActivity`, `earliestIngestedRound`, and `assertCompleteActivity`. We don't use the column for anything else. 
3. The value for the `earliest_ingested_round` column is:
    1. For the founding SV, it's always the first round. The founder starts ingesting all data from genesis and can never miss an activity record.
    2. For other SVs, it's the round that was open as of the first verdict in a batch (even if that verdict didn't produce any activity record), or the round associated with the first computed activity record, or the last round that was archived as of the last verdict in a batch.
4. The reason for the complex logic for the non-founding SVs is that we want to initialize the meta row as soon as possible, even if the rounds around SV bootstrap have no app activity records at all.

If the above is correct, can we merge https://github.com/canton-network/splice/pull/6694 first (to make sure the SV falls back to BFT reads if it hasn't initialized the meta row yet) and then replace the complex logic with the following?
```scala
val earliestRound: Long = if (isFirstSv) -1L else items.map(_.roundNumber).min
```

dfordivam: 1. True, there should never be holes.

2. Yes, sort of, It is ok if no activity records were computed for a round. we used to have the earliest and latest round tied to activity records, though now it is based on the ability to have computed the activity records.
    Which simply means the meta row now answers whether we observed the traffic-summaries and had the FeaturedAppRight data in our store.

    If we observed a round open and archive, and if we had both traffic-summaries and FeaturedAppRight data, then the meta table must track it.

3. i. Its `initialRound - 1`, so that we consider `initialRound` complete
   ii. It is a round for which we can say for sure that we had the ability to compute the activity records.

4. Yes, although we have a couple of different requirements, both related to BftRead.

   we consider `initialRound` as a special case where all SVs will fallback to single peer BftRead. (Tim has a draft PR for this)

   For rounds `initialRound + 1` onwards we have the following situation
   i. For tests the #6694 will be sufficient as we only need the fallback to BftRead, and n = 1 for 4 SV topology.

   ii. For testnet bootstrap, the n > 1 for BftRead. we can fix this in two ways, either relax BftRead's n > 1 for not just the `initialRound` but first few rounds, or initialize the `earliest_ingested_round` to be `initialRound`.

     Now initialize the `earliest_ingested_round` to be `initialRound` is easily doable on the testnet as the `initialRound` is 24h and all SVs are supposed to onboard by the time it is archived.
   In fact this would have happened pre-6579, as the `lastArchivedRoundO` would most likely have been `initialRound`

     But with the changes done in 6579 this initialization broke; as the `firstActiveRoundO` was incorrectly computed using `lookupActiveOpenMiningRounds`, which is wrong in this case as this API is too conservative for this use-case. And this PR is fixing this mistake.

     OTOH if we assume that `lastArchivedRoundO` will most likely be `initialRound`, (which is a fair assumption given the 24h long onboarding period). we can in fact simplify the `earliestRound` initialization to this, and not bother with computing the `firstActiveRoundO`. 

```
    val earliestRound =
      if (isFirstSv) Some(-1L)
      else
          items.map(_.roundNumber).minOption
          .orElse(lastArchivedRoundO)
```

   Side note: For rounds `initialRound` and `initialRound + 1`, by definition we can never have any activity records, as both opened at network start. So `items` will always be empty.

dfordivam:
This PR can be simplified further once we start adding round data to `VerdictT` being done in [Add round number to verdicts](https://github.com/canton-network/splice/pull/6794#top)#6794

adetokunbo (commented):
Should this change include an integration test update?  

E.g, had #6579 switched the `TrafficBasedAppRewardsTimeBasedIntegrationTest` from a 1SV to a 4SV environment, the issue this PR addresses could have been detected earlier

PR #6711, opened 2026-08-07
Make Pulumi pass the correct migrationId for Validator-stable-old 

pasindutennage-da:
Fix https://github.com/DACH-NY/canton-network-internal/issues/6411

### DIFF
--- a/cluster/pulumi/common-validator/src/config.ts
+++ b/cluster/pulumi/common-validator/src/config.ts
@@ -1,6 +1,12 @@
 // Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { DeployValidatorRunbook, EnvVarConfigSchema, K8sResourceSchema, KmsConfigSchema, LogLevelSchema } from '@canton-network/splice-pulumi-common/src/config';
+import {
+  DeployValidatorRunbook,
+  EnvVarConfigSchema,
+  K8sResourceSchema,
+  KmsConfigSchema,
+  LogLevelSchema
+} from '@canton-network/splice-pulumi-common/src/config';
 import { clusterSubConfig } from '@canton-network/splice-pulumi-common/src/config/config';
 import { z } from 'zod';
 
--- a/cluster/pulumi/common-validator/src/config.ts
+++ b/cluster/pulumi/common-validator/src/config.ts
@@ -1,14 +1,18 @@
 // Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
-import {
-  DeployValidatorRunbook,
-  EnvVarConfigSchema,
-  K8sResourceSchema,
-  KmsConfigSchema,
-  LogLevelSchema,
-} from '@canton-network/splice-pulumi-common/src/config';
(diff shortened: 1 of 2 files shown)


token-standard/package-lock.json
```diff
@@ -42,23 +42,23 @@
     },
     "cli/daml.js/daml-stdlib-DA-Time-Types-1.0.0": {
```
pasindutennage-da, line 43: did a make clean-all && make build

moritzkiefer-da (changes requested):
please add an integration test that configures the sv and scan app with a non-zero migration id and check that a traffic purchase works (and confirm that it fails without this)

pasindutennage-da:
@moritzkiefer-da 

> please add an integration test that configures the sv and scan app with a non-zero migration id and check that a traffic purchase works

`SV1Initializer`:  `domainMigrationId <- resolveDomainMigrationId(Future.successful(0L))`

Hence asserting against scan.getMigrationId() just evaluates to 0 == 0. This won't catch any future regressions.

I don't see an exposed config parameter to force a non-zero migration ID on boot. Any tips? 

Should I do some expensive `upgrade` first (which sounds a bit out of scope for MemberTraffic test)?

moritzkiefer-da:
> Hence asserting against scan.getMigrationId() just evaluates to 0 == 0. This won't catch any future regressions.

maybe just add a config flag to overwrite in testing

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/automation/TopupMemberTrafficTrigger.scala
```diff
@@ -59,7 +59,6 @@ class TopupMemberTrafficTrigger(
     clock: Clock,
     walletManager: UserWalletManager,
     scanConnection: BftScanConnection,
-    domainMigrationId: Long,
```
nicu-da, line 62: why is this needed though? shouldn't the trigger be initialized with the same migration id that's resolved from scan?

nicu-da (changes requested):
I don't think we're solving anything here, the id is read from the db or from scan, there's no configurable migration id anymore.

The issue is that we updated the migration id for a version that still had the migration id in the config, so we should just fix that deployment

apps/validator/src/main/scala/org/lfdecentralizedtrust/splice/validator/automation/TopupMemberTrafficTrigger.scala
pasindutennage-da: @nicu-da 

Trigger should be initialized with the resolved value, however, trigger should dynamically fetch it from scan in future (otherwise, it will keep using a stale id.) Thats the central idea of this PR, and the attached issue details the reasoning.

nicu-da: @pasindutennage-da the migration id is static and will never change anymore, what's the situation where we see it changing?

moritzkiefer-da:
> The issue is that we updated the migration id for a version that still had the migration id in the config, so we should just fix that deployment

The problem we saw was that new validators on cilr submitted traffic purchases with migration id 0. Definitely possible I misdiagnosed but it is a very bad issue so we definitely need to do something™.

nicu-da (commented):
as discussed in dm, lets just add an additional_config to the validator app in cilr for the old validator

PR #6740, opened 2026-08-10
Bump aws_version

rautenrieth-da:
The AWS SDK pulls in netty as a dependency, which triggers some vulnerability warnings. Even though they do not apply in practice, they cause noise.

This PR bumps `aws_version` to the same version as Canton.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2087,6 +2087,7 @@ def mergeStrategy(oldStrategy: String => MergeStrategy): String => MergeStrategy
     case path if path.endsWith("scala-collection-compat.properties") => MergeStrategy.first
     // Don't really care about the notice file so just take any.
     case "META-INF/FastDoubleParser-NOTICE" => MergeStrategy.first
+    case "META-INF/FastDoubleParser-LICENSE" => MergeStrategy.first
     case "META-INF/license/LICENSE.boringssl.txt" => MergeStrategy.first
     case path if path.endsWith("/OSGI-INF/MANIFEST.MF") => MergeStrategy.first
     case x =>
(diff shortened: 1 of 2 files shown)


martinflorian-da (approved):
Thanks!

Answering my own question: I don't see how this can impact customers, as no Splice code is talking to AWS directly even if you use AWS KMS.

rautenrieth-da:
> as no Splice code is talking to AWS directly even if you use AWS KMS.

We do use the AWS library in [S3BucketConnection](https://github.com/canton-network/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala#L15-L19), which is used for writing cold storage data from the scan app. Are you saying we should include the change in release notes?

martinflorian-da:
> > as no Splice code is talking to AWS directly even if you use AWS KMS.
> 
> We do use the AWS library in [S3BucketConnection](https://github.com/canton-network/splice/blob/main/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala#L15-L19), which is used for writing cold storage data from the scan app. Are you saying we should include the change in release notes?

Ah I see... but the circle of external users for those code paths seem very slim at the moment. So I don't think we need release notes. (Part of my rationale was also that we're not testing AWS KMS things, so have a slightly higher chance of accidentally introducing a regression; that argument also doesn't seem to apply here?)

rautenrieth-da:
> Part of my rationale was also that we're not testing AWS KMS things, so have a slightly higher chance of accidentally introducing a regression; that argument also doesn't seem to apply here?

We don't use the KMS part in splice at all, only the S3 (cloud storage) part. And even that is not enabled in production yet.

PR #6757, opened 2026-08-12
Reapply "[ci] Bump the development-dependencies group across 1 directory with 8 updates"

OriolMunoz-da:
Reverts canton-network/splice#6726

js-yaml 4.2 electric boogaloo

also [fixes bug in the fix](https://github.com/canton-network/splice/pull/6757/changes#diff-7a98bde75a2d2e75bc89fada90e0ec8c36e19ccadba61f6f8ab565a48b01a066)


cluster/pulumi/common/tsconfig.eslint.json
```diff
@@ -0,0 +1 @@
+{"extends": "./tsconfig.json"}
```
moritzkiefer-da, line 1: dumb question, why do we need this now?

moritzkiefer-da (approved):
thx what could go wrong, maybe still worth running a scratch deployment?

cluster/pulumi/common/tsconfig.eslint.json
OriolMunoz-da: absolutely no idea, pinging @krzysztofczyz-da to answer this when he's back from vacay

OriolMunoz-da:
waiting until cimain is fixed before merging because I don't need more noise

PR #6764, opened 2026-08-12
Extend splice rate limits

nicu-da:
Valid for scan and sv app

Add global rate limiter (previosly we rate limited only each individual operation), which is enabled by default. Add the ability to rate limit also per ip for each oepration (disabled by default). The global rate limiter has this option enabled by default.

Extend the rate limiter to check a longer interval (60s). The previous behavior was checking only the last 1s, this is still in place and works as a burst limiter, allowing for shorter burts but the longer interval enforces a lower limit for the configured interval.

[ci]

fixes https://github.com/DACH-NY/canton-network-internal/issues/6295

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/main/resources/application.conf
+++ b/apps/app/src/main/resources/application.conf
@@ -14,3 +14,6 @@ pekko.http.server.parsing.ignore-illegal-header-for =
     "x-real-ip"
     "etag" ]
 pekko.http.client.parsing.ignore-illegal-header-for = ["etag"]
+# expose the remote address so that it can be used for per client IP rate limiting
+# when no X-Forwarded-For/X-Real-Ip header is set by an upstream proxy
+pekko.http.server.remote-address-attribute = on
--- a/apps/app/src/main/resources/application.conf
+++ b/apps/app/src/main/resources/application.conf
@@ -14,7 +14,3 @@ pekko.http.server.parsing.ignore-illegal-header-for =
     "x-real-ip"
     "etag" ]
 pekko.http.client.parsing.ignore-illegal-header-for = ["etag"]
-# expose the remote address so that it can be used for per client IP rate limiting
-# when no trusted X-Envoy-External-Address (nor X-Forwarded-For/X-Real-Ip) header is
-# set by an upstream proxy
-pekko.http.server.remote-address-attribute = on
--- a/apps/app/src/main/resources/application.conf
+++ b/apps/app/src/main/resources/application.conf
@@ -15,5 +15,6 @@ pekko.http.server.parsing.ignore-illegal-header-for =
     "etag" ]
 pekko.http.client.parsing.ignore-illegal-header-for = ["etag"]
 # expose the remote address so that it can be used for per client IP rate limiting
-# when no X-Forwarded-For/X-Real-Ip header is set by an upstream proxy
+# when no trusted X-Envoy-External-Address (nor X-Forwarded-For/X-Real-Ip) header is
+# set by an upstream proxy
 pekko.http.server.remote-address-attribute = on
--- a/apps/app/src/test/resources/include/scans/_scan.conf
+++ b/apps/app/src/test/resources/include/scans/_scan.conf
@@ -34,6 +34,15 @@
       default {
         rate-per-second = 200
       }
+      # the global limiter caps the total request rate across all operations; disabled in
+      # tests so that it doesn't interfere with the per-operation limits exercised here.
+      global {
+        enabled = false
+        rate-per-second = 200
+        per-client-ip {
+          enabled = false
+        }
+      }
       rate-limiters {
         getAcsSnapshot.rate-per-second = 2
         # High defaults as we call this from UpdateHistorySanityCheckPlugin
--- a/apps/app/src/test/resources/include/scans/_scan.conf
+++ b/apps/app/src/test/resources/include/scans/_scan.conf
@@ -31,14 +31,10 @@
       getAcsSnapshot = 1 minute
     }
     rate-limiting {
-      default {
-        rate-per-second = 200
-      }
       # the global limiter caps the total request rate across all operations; disabled in
       # tests so that it doesn't interfere with the per-operation limits exercised here.
       global {
         enabled = false
-        rate-per-second = 200
         per-client-ip {
           enabled = false
         }
--- a/apps/common/src/main/java/com/google/common/util/concurrent/BurstyRateLimiterFactory.java
+++ b/apps/common/src/main/java/com/google/common/util/concurrent/BurstyRateLimiterFactory.java
@@ -1,0 +1,24 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.google.common.util.concurrent;
+
+/**
+ * Shim that constructs a Guava {@link RateLimiter} backed by {@code SmoothBursty} with a custom
+ * maximum burst duration.The relevant constructor SmoothRateLimiter.SmoothBursty are package-private
+ */
+public final class BurstyRateLimiterFactory {
+
+    private BurstyRateLimiterFactory() {
+    }
+
+    /**
+     * Creates a bursty {@link RateLimiter} that sustains {@code permitsPerSecond} on average while
+     * allowing bursts of up to {@code permitsPerSecond * maxBurstSeconds} permits after idle periods.
+     */
+    public static RateLimiter create(double permitsPerSecond, double maxBurstSeconds) {
+        RateLimiter rateLimiter = new SmoothRateLimiter.SmoothBursty(RateLimiter.SleepingStopwatch.createFromSystemTimer(), maxBurstSeconds);
+        rateLimiter.setRate(permitsPerSecond);
+        return rateLimiter;
+    }
+}
--- a/apps/common/src/main/java/com/google/common/util/concurrent/BurstyRateLimiterFactory.java
+++ b/apps/common/src/main/java/com/google/common/util/concurrent/BurstyRateLimiterFactory.java
@@ -5,7 +5,8 @@ package com.google.common.util.concurrent;
 
 /**
  * Shim that constructs a Guava {@link RateLimiter} backed by {@code SmoothBursty} with a custom
- * maximum burst duration.The relevant constructor SmoothRateLimiter.SmoothBursty are package-private
+ * maximum burst duration. It lives in this package because the relevant {@code
+ * SmoothRateLimiter.SmoothBursty} constructor is package-private.
  */
 public final class BurstyRateLimiterFactory {
 
@@ -17,7 +18,9 @@ public final class BurstyRateLimiterFactory {
      * allowing bursts of up to {@code permitsPerSecond * maxBurstSeconds} permits after idle periods.
      */
     public static RateLimiter create(double permitsPerSecond, double maxBurstSeconds) {
-        RateLimiter rateLimiter = new SmoothRateLimiter.SmoothBursty(RateLimiter.SleepingStopwatch.createFromSystemTimer(), maxBurstSeconds);
+        RateLimiter rateLimiter =
+                new SmoothRateLimiter.SmoothBursty(
+                        RateLimiter.SleepingStopwatch.createFromSystemTimer(), maxBurstSeconds);
         rateLimiter.setRate(permitsPerSecond);
         return rateLimiter;
     }
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
@@ -14,12 +14,24 @@ case class RateLimitersConfig(
     /** Per-operation overrides of the overall `default` rate limiter. */
     rateLimiters: Map[String, SpliceRateLimitConfig.WithPerClientIp],
     global: SpliceRateLimitConfig.WithPerClientIp = RateLimitersConfig.DefaultGlobal,
+    /** Name of the HTTP header set by a trusted reverse proxy that carries the real client IP. This header must be set - and any
+      * client-provided value overwritten - by infrastructure the client cannot bypass, otherwise it
+      * can be spoofed. When present and parseable as an IP literal it takes precedence over the
+      * client-controlled `X-Forwarded-For`/`X-Real-Ip` headers. Set to an empty string to disable
+      * trusting a proxy header and only rely on `X-Forwarded-For`/`X-Real-Ip`/the remote address.
+      */
+    trustedClientIpHeader: String = RateLimitersConfig.DefaultTrustedClientIpHeader,
 ) {
   def forRateLimiter(name: String): SpliceRateLimitConfig.WithPerClientIp =
     rateLimiters.getOrElse(name, default)
 }
 
 object RateLimitersConfig {
+
+  /** Header set by the Envoy sidecar/ingress (Istio) to the trusted external client address that
+    * Envoy computes from its trusted-hops configuration.
+    */
+  val DefaultTrustedClientIpHeader: String = "x-envoy-external-address"
 
   private val DefaultGlobal: SpliceRateLimitConfig.WithPerClientIp =
     SpliceRateLimitConfig.WithPerClientIp(
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
@@ -3,11 +3,27 @@
 
 package org.lfdecentralizedtrust.splice.config
 
-import org.lfdecentralizedtrust.splice.util.SpliceRateLimitConfig
+import org.lfdecentralizedtrust.splice.util.{PerAttributeRateLimitConfig, SpliceRateLimitConfig}
 
 case class RateLimitersConfig(
-    default: SpliceRateLimitConfig,
-    rateLimiters: Map[String, SpliceRateLimitConfig],
+    /** Overall rate limiter applied per operation. Used when there is no operation-specific override
+      * in `rateLimiters`. The embedded `perClientIp` limiter is disabled by default; enable it to
+      * additionally limit per client IP.
+      */
+    default: SpliceRateLimitConfig.WithPerClientIp,
+    /** Per-operation overrides of the overall `default` rate limiter. */
+    rateLimiters: Map[String, SpliceRateLimitConfig.WithPerClientIp],
+    global: SpliceRateLimitConfig.WithPerClientIp = RateLimitersConfig.DefaultGlobal,
 ) {
-  def forRateLimiter(name: String): SpliceRateLimitConfig = rateLimiters.getOrElse(name, default)
+  def forRateLimiter(name: String): SpliceRateLimitConfig.WithPerClientIp =
+    rateLimiters.getOrElse(name, default)
+}
+
+object RateLimitersConfig {
+
+  private val DefaultGlobal: SpliceRateLimitConfig.WithPerClientIp =
+    SpliceRateLimitConfig.WithPerClientIp(
+      ratePerSecond = 200,
+      perClientIp = PerAttributeRateLimitConfig(),
+    )
 }
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/ClientIpDirectives.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/ClientIpDirectives.scala
@@ -1,0 +1,69 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package org.lfdecentralizedtrust.splice.http
+
+import org.apache.pekko.http.scaladsl.model.headers.{`X-Forwarded-For`, `X-Real-Ip`}
+import org.apache.pekko.http.scaladsl.model.{AttributeKeys, RemoteAddress}
+import org.apache.pekko.http.scaladsl.server.Directive1
+import org.apache.pekko.http.scaladsl.server.Directives.*
+
+object ClientIpDirectives {
+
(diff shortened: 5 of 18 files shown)


docs/src/release_notes_upcoming.rst
```diff
@@ -30,3 +30,35 @@
 
         - During the creation of ``AmuletRules_SetConfig`` proposal, for the ``rewardConfig``
           config field form improve the field descriptions and use drop-downs.
+
+    - Scan & SV App
+
+        - HTTP rate limiting has been extended with a global rate limiter applied across all
+          operations, optional per-client-IP rate limiting (enabled by default at the global level),
+          and an additional sustained rate limit enforced over a longer window on top of the existing
+          per-second burst limit. The client IP is derived from the ``X-Forwarded-For``/``X-Real-Ip``
```
martinflorian-da: Pointer from standup while it's fresh:

That header maybe not great; at the least we should confirm that an attacker can't fool our own istio; see this experiment by @krzysztofczyz-da : https://github.com/canton-network/splice/issues/6640#issuecomment-5217188469

nicu-da: Added better support for this, by default for istion and improved the release notes

docs/src/release_notes_upcoming.rst
```diff
@@ -30,3 +30,55 @@
...
+          address only for requests that did not pass through the ingress. These can be tuned via
+          the following ``rate-limiting`` config keys:
+
+          .. warning::
+
+             When per-client-IP rate limiting is enabled, operators must ensure that the client IP
+             used for rate limiting cannot be spoofed. Either configure
+             ``rate-limiting.trusted-client-ip-header`` to a trusted, non-spoofable header set by
+             your ingress/proxy (e.g. ``x-envoy-external-address`` for Istio deployments), or ensure
+             that the ``X-Forwarded-For`` header contains the actual client IP as its first value
```
martinflorian-da, line 41: > first value

Is that a standard behavior? I.e., push to front instead of append at end?

apps/app/src/main/resources/application.conf
```diff
@@ -14,3 +14,7 @@ pekko.http.server.parsing.ignore-illegal-header-for =
     "x-real-ip"
     "etag" ]
 pekko.http.client.parsing.ignore-illegal-header-for = ["etag"]
+# expose the remote address so that it can be used for per client IP rate limiting
+# when no trusted X-Envoy-External-Address (nor X-Forwarded-For/X-Real-Ip) header is
+# set by an upstream proxy
+pekko.http.server.remote-address-attribute = on
```
martinflorian-da: So that is clearly spoofable. I'm have mixed feelings about even covering that route as it seems to offer some false sense of security...

apps/app/src/test/resources/include/scans/_scan.conf
```diff
@@ -34,6 +34,15 @@
       default {
         rate-per-second = 200
       }
+      # the global limiter caps the total request rate across all operations; disabled in
+      # tests so that it doesn't interfere with the per-operation limits exercised here.
+      global {
+        enabled = false
+        rate-per-second = 200
```
martinflorian-da: nit nit nit: consider defaulting to 0 or something so we can skip these lines; being forces to set a rate when `enabled = false` triggers OCD

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
```diff
@@ -3,11 +3,39 @@
...
+      */
+    default: SpliceRateLimitConfig.WithPerClientIp,
+    /** Per-operation overrides of the overall `default` rate limiter. */
+    rateLimiters: Map[String, SpliceRateLimitConfig.WithPerClientIp],
+    global: SpliceRateLimitConfig.WithPerClientIp = RateLimitersConfig.DefaultGlobal,
+    /** Name of the HTTP header set by a trusted reverse proxy that carries the real client IP. This header must be set - and any
+      * client-provided value overwritten - by infrastructure the client cannot bypass, otherwise it
+      * can be spoofed. When present and parseable as an IP literal it takes precedence over the
+      * client-controlled `X-Forwarded-For`/`X-Real-Ip` headers. Set to an empty string to disable
+      * trusting a proxy header and only rely on `X-Forwarded-For`/`X-Real-Ip`/the remote address.
```
martinflorian-da, line 21:
```suggestion
      * trusting a proxy header and only rely on `X-Forwarded-For`/`X-Real-Ip`/the remote address (not recommended).
```

nit

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
```diff
@@ -3,11 +3,39 @@
...
 package org.lfdecentralizedtrust.splice.config
 
-import org.lfdecentralizedtrust.splice.util.SpliceRateLimitConfig
+import org.lfdecentralizedtrust.splice.util.{PerAttributeRateLimitConfig, SpliceRateLimitConfig}
 
 case class RateLimitersConfig(
-    default: SpliceRateLimitConfig,
-    rateLimiters: Map[String, SpliceRateLimitConfig],
+    /** Overall rate limiter applied per operation. Used when there is no operation-specific override
+      * in `rateLimiters`. The embedded `perClientIp` limiter is disabled by default; enable it to
```
martinflorian-da, line 10: Is this comment lying to me? Below I see `perClientIp = PerAttributeRateLimitConfig()`, so we do set per-IP limiting by default?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/ClientIpDirectives.scala
```diff
@@ -0,0 +1,69 @@
...
+object ClientIpDirectives {
+
+  /** Extracts the address of the client the request originated from, if it can be determined.
+    *
+    * The address is taken from the first of the following sources that yields an address:
+    *   1. the `trustedClientIpHeader` (if configured and parseable as an IP literal), which is set
+    *      by a trusted reverse proxy and hence cannot be spoofed by the client,
+    *   1. the client-controlled `X-Forwarded-For` header,
+    *   1. the client-controlled `X-Real-Ip` header,
+    *   1. the remote address of the transport connection, if the server is configured to expose it.
```
martinflorian-da: Possibly a stupid question, but if we can use this shouldn't it have higher precedence than client-controlled headers that are more easily spoofed?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/HttpRateLimiter.scala
```diff
@@ -19,35 +24,96 @@ class HttpRateLimiter(
...
 
     import org.apache.pekko.http.scaladsl.server.Directives.*
 
-    extractRequestContext.flatMap { _ =>
-      if (rateLimiter.markRun()) {
+    HttpRateLimiter.extractClientIpKey(trustedClientIpHeader).flatMap { clientIp =>
+      // The global limiters are checked first so that a request rejected globally does not consume
+      // budget from the per-operation limiters.
+      val allowed =
+        globalLimiter.markRun() &&
```
martinflorian-da: So.... an attacker could single-handedly deplete the global limit here, before even starting to get limited by IP? Is this really what we want? It looks like an easy way to do a DoS to me...

apps/common/src/main/java/com/google/common/util/concurrent/BurstyRateLimiterFactory.java
martinflorian-da, line 1: There is also https://github.com/canton-network/splice/blob/c57f1686ef468810613346f05f239524c7e338da/canton/community/common/src/main/scala/com/digitalasset/canton/util/RateLimiter.scala#L30 but not actually sure if we can reuse that here.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/HttpRateLimiter.scala
martinflorian-da: If I'm right - can we also cover that via a test?

martinflorian-da (commented):
Thanks! I'm still not fully satisfied with my review (devil hidden in the detail here...), but some of the comments I already have seem worth an earlier discussion.

Maybe to increase our confidence in all of this rate limiting - have we thought about some cluster-based tests? I guess something like "different source IPs" will be quite hard to pull off unfortunately.

docs/src/release_notes_upcoming.rst
nicu-da: The client ip should be the first value followed by however many other further nodes are downstream.
This also implies that you should drop any value received from the client

apps/app/src/main/resources/application.conf
nicu-da: fair point, this is low value. will just drop it.

apps/app/src/test/resources/include/scans/_scan.conf
nicu-da: it's set for completeness if you want to modify it in the future in the tests, it has a code default value

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
nicu-da: wouldn't say that it's not recommended. if you do use it then you should sanitize the request when first received

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
nicu-da: not sure where you see that 

```
      perClientIp: PerAttributeRateLimitConfig = PerAttributeRateLimitConfig.Disabled,
```

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/ClientIpDirectives.scala
nicu-da: No, as that value will most likely be just the sidecar/ingress ip from within the  cluster

nicu-da: that's why i am just fully removing it

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/HttpRateLimiter.scala
nicu-da: Good point, moving the ip rate limiters to run first makes more sense

apps/common/src/main/java/com/google/common/util/concurrent/BurstyRateLimiterFactory.java
nicu-da: No, that one has no window interval and also use an atomic update of the state which worries me a bit in term of performance

docs/src/release_notes_upcoming.rst
moritzkiefer-da: > The client ip should be the first value followed by however many other further nodes are downstream.
> This also implies that you should drop any value received from the client

This is the opposite of how envoy handles this. Envoy always takes the right most address and you tell it how many it can ignore on the right (it's not fully clear to me if it even modifies XFF if you set numTrustedProxies > 1 before passing it downstream).

You're saying it's up to you to strip on the left or it's trivially forgeable as a user-set xff will appear before any correct IP.

Any reason why we don't follow the envoy style approach of taking the right most one and letting users pass in an int in the config for how many you skip? that seems like a safer default.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/HttpRateLimiter.scala
martinflorian-da: Actually don't we want to always mark both?

martinflorian-da: What to the sequencer caps do?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
martinflorian-da: Ah I think I more generally confused between `default` and `global`!

So for `global` we have per-IP limiting... which then also extends to this `default`.

Bear with me... (naming suggestions incoming)

`default` => `perOperationDefault`

`rateLimiters` => `perOperationOverrides` (or just `perOperation`)

`global` => `global`

... and then I still struggle to remember the specific interplay between the 3, and I just read the code a few moments ago 😅 Not sure if the comment can be made to explain which one exactly hits when (once we agree on that).

docs/src/release_notes_upcoming.rst
nicu-da: I debated doing that, but most likely that will blow up fast as you would rate limit on the right most ip if you don't configure it right and that will be the same ip for everyone as it's some internal ip.
Reason why I chose the left most is because even if you do that, you just bypass the per client ip and eventually hit the global rate limiter. So the impact is pretty limited and you will realize the issue pretty quickly (compared to whitelisting).
Adding the number of hops as a config would also mean tying the network topology in the app config and needing to keep that config in sync.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
nicu-da: so this just adds the global key, I agree that the default, rateLimiters naming is bad but I wouldn't break compatibility for now just to fix it

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/HttpRateLimiter.scala
nicu-da: Caps don't really have anything in common with this, they get updated on the read path and reject on the write path.

nicu-da: Ideally we want to have  the global configured limit of successful requests, so marking both would break that.

docs/src/release_notes_upcoming.rst
moritzkiefer-da: hm yeah maybe fair, I'm not sure how realistic it is for people to strip things before the real IP but maybe doesn't matter. For SVs we can just coordinate enough that they all have a working config.

apps/app/src/test/resources/include/scans/_scan.conf
martinflorian-da: mmm explicitly configuring a value that is not actually used seems a bit smelly; I'd remove if we easily can; easy enough to add back later if indeed we want to extend for some test

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
martinflorian-da: fair

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
martinflorian-da: Happy to do it in a separate PR but IMO it would be a reasonable investment to make this already complicated subject a bit easier to understand.

> break compatibility

I don't think many people set these explicitly via `ADDITIONAL_CONFIG_` and similar... and if we time to the right version they might not even be allowed to complain (as per [new versioning logic](https://docs.google.com/document/d/1WH3nZp-fCzAQcvY1N3nRhgjD6QmX5xQjg3a4KmN-3lA/)).

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/HttpRateLimiter.scala
martinflorian-da: Nice I think I understand it now; boolean logic is hard :)

Thank and thanks for switching this around; it makes sense to me now!

docs/src/release_notes_upcoming.rst
martinflorian-da: @nicu-da Did we start making a list of all the things we're asking SVs to configure in a specific way ... so we remember when we get to updating our docs? Please don't say "in my head" because these days I'm afraid we might not always have access to that :)

docs/src/release_notes_upcoming.rst
```diff
@@ -5,4 +5,80 @@
...
+          and an additional sustained rate limit enforced over a longer window on top of the existing
+          per-second burst limit. The client IP is taken from the trusted, non-spoofable
+          ``X-Envoy-External-Address`` header set by the Envoy/Istio ingress, falling back to the
+          client-controlled ``X-Forwarded-For``/``X-Real-Ip`` headers and finally the remote
+          address only for requests that did not pass through the ingress. These can be tuned via
+          the following ``rate-limiting`` config keys:
+
+          .. warning::
+
+             When per-client-IP rate limiting is enabled, operators must ensure that the client IP
```
```suggestion
             When per-client-IP rate limiting is enabled, SV operators must ensure that the client IP
```

docs/src/release_notes_upcoming.rst
```diff
@@ -5,4 +5,80 @@
...
+             your ingress/proxy (e.g. ``x-envoy-external-address`` for Istio deployments), or ensure
+             that the ``X-Forwarded-For`` header contains the actual client IP as its first value
+             and cannot be spoofed by clients. Otherwise, clients may bypass per-client-IP limits or
+             cause other clients to be throttled by forging these headers.
+
+
+          - ``rate-limiting.trusted-client-ip-header``: name of the trusted proxy header carrying
+            the real client IP (default ``x-envoy-external-address``). Override it for non-Istio
+            deployments, or set it to an empty string to disable trusting a proxy header and only
+            rely on ``X-Forwarded-For``/``X-Real-Ip``/the remote address.
```
```suggestion
            rely on ``X-Forwarded-For`` / ``X-Real-Ip`` / the remote address.
```

docs/src/release_notes_upcoming.rst
```diff
@@ -5,4 +5,80 @@
...
+    - Scan & SV App
+
+        - HTTP rate limiting has been extended with a global rate limiter applied across all
+          operations, optional per-client-IP rate limiting (enabled by default at the global level),
+          and an additional sustained rate limit enforced over a longer window on top of the existing
+          per-second burst limit. The client IP is taken from the trusted, non-spoofable
+          ``X-Envoy-External-Address`` header set by the Envoy/Istio ingress, falling back to the
+          client-controlled ``X-Forwarded-For``/``X-Real-Ip`` headers and finally the remote
+          address only for requests that did not pass through the ingress. These can be tuned via
+          the following ``rate-limiting`` config keys:
```
martinflorian-da: It actually seems a bit too verbose to describe all of this in the release notes...

Not just too verbose for release notes (as opposed to linked docs), but also you're expecting people to make the jump to "how to configure on my Helm-based deployment" themselves; I don't think we typically do this?

I wonder if we can just:

- keep release notes rather short; I'd keep the warning, maybe extend it though with a full `ADITIONAL_CONFIG_*` snippet
- make sure the in-code comments are understandable also for non-engineer operators; I'm not opposed to us pointing to the right scala file in the release notes.
- at some point we'll need actual docs that also describe how to set these on Helm-based deployments, if needed... so I'd save the more elaborate explanation below to then

martinflorian-da (approved):
Thanks!

My confidence is not super duper so I'd appreciate someone else also having a careful look.

That said I see value in early testing on CILR.

As for "cluster-based testing" - moving to some future 1:1.

docs/src/release_notes_upcoming.rst
nicu-da: for now we have https://github.com/DACH-NY/canton-network-internal/issues/6475 where we also cover the ip check
after we get the stuff from Pasindu we'll expand

apps/app/src/test/resources/include/scans/_scan.conf
nicu-da: agree, removed it

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/config/RateLimitersConfig.scala
nicu-da: yup, let me do another PR after this one to clean-up the config api

docs/src/release_notes_upcoming.rst
nicu-da: fair point, my thoughts were the following: we don't have docs for it yet, and those will be added only later so for now I am hijacking this to document it. Let me just remove the whole config keys (except the header config).

docs/src/release_notes_upcoming.rst
nicu-da: It is sort of already under sv / scan app (still aded though because I have no strong feelings about it)

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/ClientIpDirectives.scala
```diff
@@ -0,0 +1,64 @@
...
+    val trimmedHeaderName = headerName.trim
+    if (trimmedHeaderName.isEmpty) provide(None)
+    else
+      // matched case-insensitively (and locale independently) as the configured header name is not
+      // required to be lowercase
+      optionalHeaderValueByName(trimmedHeaderName).map(_.flatMap(parseIpLiteral))
+  }
+
+  private val forwardedForClientIp: Directive1[Option[RemoteAddress]] = {
+    optionalHeaderValuePF { case `X-Forwarded-For`(Seq(address, _*)) => address }
```
pasindutennage-da, line 42: Ultra Pro Max Nit: would it make sense to log a warning, if X-Forwarded-For is used in the chain? If ever there is spoofing, then it will be useful to debug?

nicu-da: Hmm, I think it would be useful, adding some log every x minutes if the configured header is not found. Let me add a tech debt for it

pasindutennage-da (approved):

PR #6769, opened 2026-08-12
Public Sequencer: Add a DevNet only HTTP endpoint to SV App, that creates free MemberTraffic contracts for the provided participantId, using its own test CC

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/6763

### DIFF
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
@@ -334,7 +334,7 @@ class HttpSvPublicHandler(
       } else {
         Future.failed(
           HttpErrorHandler.notImplemented(
-            "DevNet traffic purchasing self-service is only available in DevNet."
+            "Traffic purchasing self-service is only available in DevNet."
           )
         )
       }
--- a/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
+++ b/apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
@@ -315,7 +315,91 @@ class HttpSvPublicHandler(
       respond: r0.DevNetBuyMemberTrafficResponse.type
   )(
       body: definitions.DevNetBuyMemberTrafficRequest
-  )(extracted: TraceContext): Future[r0.DevNetBuyMemberTrafficResponse] = ???
+  )(extracted: TraceContext): Future[r0.DevNetBuyMemberTrafficResponse] = {
+    implicit val traceContext: TraceContext = extracted
+    withSpan(s"$workflowId.devNetBuyMemberTraffic") { _ => _ =>
+      if (isDevNet) {
+        for {
+          participantId <- ParticipantId.fromProtoPrimitive(
+            body.participantId,
+            "participant_id",
+          ) match {
+            case Right(id) => Future.successful(id)
+            case Left(err) =>
+              Future.failed(HttpErrorHandler.badRequest(s"Invalid participant ID: $err"))
+          }
+          amuletRules <- dsoStore.getAmuletRules()
+          dsoRules <- dsoStore.getDsoRules()
+          openRound <- dsoStore.getLatestActiveOpenMiningRound()
+
+          _ = logger.info(s"AmuletRules DevNet Tap by $svUserName")
+
+          tapCmd = amuletRules.exercise(
+            _.exerciseAmuletRules_DevNet_Tap(
+              svParty.toProtoPrimitive,
+              config.devNetPublicSetupTapAmount.bigDecimal,
+              openRound.contractId,
+            )
+          )
+
+          tapResult <- dsoStoreWithIngestion
+            .connection(SpliceLedgerConnectionPriority.Low)
+            .submit(
+              actAs = Seq(svParty),
+              readAs = Seq.empty,
+              update = tapCmd,
+            )
+            .withSynchronizerId(dsoRules.domain)
+            .noDedup
+            .yieldResult()
+
+          amuletCid = tapResult.exerciseResult.amuletSum.amulet
+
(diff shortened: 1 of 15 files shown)


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/config/SvAppConfig.scala
```diff
@@ -485,6 +485,8 @@ case class SvAppBackendConfig(
     // where intentional overlap is required.
     instanceLockEnabled: Boolean = true,
     minMemberTrafficToOnboardValidator: Long = 100000L,
```
pasindutennage-da, line 487: updated tracking issue to remind myself to move these to daml

nicu-da (approved):
Can we also make sure the default rate limits in the sv app for this endpoint are adjusted down (I think we do the same for the tap endpoint)

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
```diff
@@ -307,6 +307,40 @@ class HttpSvPublicHandler(
...
+    * Protection: Rate limiting, endpoint only used for DevNet
+    */
+  override def devNetBuyMemberTraffic(
+      respond: r0.DevNetBuyMemberTrafficResponse.type
+  )(
+      body: definitions.DevNetBuyMemberTrafficRequest
+  )(extracted: TraceContext): Future[r0.DevNetBuyMemberTrafficResponse] = {
+    implicit val traceContext: TraceContext = extracted
+    withSpan(s"$workflowId.devNetBuyMemberTraffic") { _ => _ =>
+      if (isDevNet) {
```
moritzkiefer-da: this should also be behind the permissioned synchronizer flag

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
```diff
@@ -850,6 +884,75 @@ class HttpSvPublicHandler(
...
+        )
+        .withSynchronizerId(dsoRules.domain)
+        .noDedup
+        .yieldResult()
+
+      amuletCid = tapResult.exerciseResult.amuletSum.amulet
+
+      _ = logger.info(s"AmuletRules BuyMemberTraffic by $svUserName for $participantId")
+
+      buyCmd = amuletRules.exercise(
```
moritzkiefer-da: this is not gonna work, you're racing with the wallet automation that will try to merge amulets. You need to do this in a single transaction.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
```diff
@@ -850,6 +884,75 @@ class HttpSvPublicHandler(
...
+      dsoRules <- dsoStore.getDsoRules()
+      roundTriple <- dsoStore.getOpenMiningRoundTriple()
+      openRound = roundTriple.oldest
+
+      _ = logger.info(s"AmuletRules DevNet Tap by $svUserName")
+
+      tapCmd = amuletRules.exercise(
+        _.exerciseAmuletRules_DevNet_Tap(
+          svParty.toProtoPrimitive,
+          config.devNetPublicSetupTapAmount.bigDecimal,
```
moritzkiefer-da: shouldn't this calculate the amount based on the required traffic and then convert it?

pasindutennage-da:
@nicu-da 

> Can we also make sure the default rate limits in the sv app for this endpoint are adjusted down (I think we do the same for the tap endpoint)

Sure, in Public Sequencer backend tracking issue, I already have a bullet point for this, and will be done, but in a different PR.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/SvSvStore.scala
```diff
@@ -38,7 +41,7 @@ trait SvSvStore extends AppStore {
       : org.lfdecentralizedtrust.splice.store.MultiDomainAcsStore.ContractFilter[
         org.lfdecentralizedtrust.splice.sv.store.db.SvTables.SvAcsStoreRowData,
         AcsInterfaceViewRowData.NoInterfacesIngested,
-      ] = SvSvStore.contractFilter(key)
+      ] = SvSvStore.contractFilter(key, config.map(_.permissionedSynchronizer).getOrElse(false))
```
moritzkiefer-da: I don't think guarding filters by the flag is gonna work well, I think you just always need to ingest it and also bump the store version to force reingestion

moritzkiefer-da (approved):
thanks

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/store/SvSvStore.scala
pasindutennage-da: Fixed

PR #6777, opened 2026-08-12
Fix WebSocket JWT extraction dropping tokens due to unstrimmed whitespace

### LINKED ISSUE #6776: WebSocket JWT auth on `/v2/state/active-contracts` silently fails to authenticate
We ran into this while paging through `/v2/state/active-contracts` over WebSocket, as recommended in the docs (https://docs.canton.network/global-synchronizer/reference/api-configuration#websockets) for result sets too large for the non-streaming endpoint.

## What happens

Connecting a WebSocket client to `/v2/state/active-contracts` with subprotocols `daml.ws.auth` and `jwt.token.<TOKEN>` (the documented auth mechanism for this endpoint), e.g.:

```js
const ws = new WebSocket(
  `wss://<host>/v2/state/active-contracts`,
  ['daml.ws.auth', `jwt.token.${token}`]
);
```

or via `wscat`:

```
wscat -c "wss://<host>/v2/state/active-contracts" -s "daml.ws.auth" -s "jwt.token.<TOKEN>"
```

The handshake succeeds fine (`HTTP/1.1 101 Switching Protocols`, `Sec-WebSocket-Protocol: daml.ws.auth` correctly echoed back). But as soon as we send the request body, the first response frame is:

```json
{"code":"NA","cause":"A security-sensitive error has been received","correlationId":"...","traceId":"...","context":{},"resources":[],"errorCategory":-1,"grpcCodeValue":16,"retryInfo":null,"definiteAnswer":null}
```

`grpcCodeValue: 16` is `UNAUTHENTICATED` — even though the exact same JWT works fine via the standard `Authorization: Bearer` header on non-WS endpoints.

## Why

The JWT extraction for WebSocket connections lives in `Endpoints.scala` (`canton/community/ledger/ledger-json-api/.../http/json/v2/Endpoints.scala`), currently around line 447-456:

```scala
auth
  .apiKey(header[Option[String]]("Sec-WebSocket-Protocol"))
  .map { bearer =>
    val tokenPrefix = "jwt.token." // TODO (i21030) test this
    bearer
      .map(_.split(",").toSeq)
      .getOrElse(Seq.empty)
      .filter(_.startsWith(tokenPrefix))
      .map(_.substring(tokenPrefix.length))
      .headOption
      .map(Jwt.apply)
  }(_.map(_.token))
```

It splits `Sec-WebSocket-Protocol` on `,` but never trims the resulting strings. RFC 6455 clients (browsers, `ws`, `wscat`) send this header as a comma-**space** separated list, so `"daml.ws.auth, jwt.token.<TOKEN>".split(",")` gives `["daml.ws.auth", " jwt.token.<TOKEN>"]` — note the leading space on the second element. `.startsWith("jwt.token.")` then fails, the filter drops it, and `.headOption` returns `None`. The JWT is silently dropped, the request goes out unauthenticated, and the ledger API rejects it.

The pre-existing `// TODO (i21030) test this` comment suggests this path never actually had a passing end-to-end test with a real client, which is presumably how this slipped through.

(We also considered whether pekko-http itself was stripping the header before tapir could read it, since the handshake completing but auth still failing could look that way — but checking `tapir-pekko-http-server`'s `PekkoServerRequest` shows the header is preserved in the request and decoded before the WebSocket upgrade directive runs. So it's just the missing `.trim()`.)

## PR

Opened #6777 with the fix and a few regression tests.

omaiko-intellecteu:
## Summary

Fixes #6776.

`Sec-WebSocket-Protocol` is a comma-**space** separated list per RFC 6455 (`"daml.ws.auth, jwt.token.<TOKEN>"`). The JWT extraction in `Endpoints.scala` split on `,` but never trimmed the resulting elements, so every element after the first kept a leading space and failed the `startsWith("jwt.token.")` check. The JWT was silently dropped and the ledger API call was rejected as `UNAUTHENTICATED` (grpc code 16), even though the same token works fine via the standard `Authorization` header.

- Extracts the parsing logic into `Endpoints.extractWsJwtToken` and trims each element after splitting.
- Resolves the outstanding `// TODO (i21030) test this`.
- Adds `EndpointsTest.scala` covering: RFC 6455 comma-space form, no-space form, token-first ordering, token-only, extra surrounding/internal whitespace, absent header, no matching subprotocol, and empty header value.

## Test plan

- [x] New unit tests in `EndpointsTest.scala` (`extractWsJwtToken`) cover the reported case and related edge cases.
- [ ] CI

### DIFF
--- a/canton/community/ledger/ledger-json-api/src/main/scala/com/digitalasset/canton/http/json/v2/Endpoints.scala
+++ b/canton/community/ledger/ledger-json-api/src/main/scala/com/digitalasset/canton/http/json/v2/Endpoints.scala
@@ -434,6 +434,21 @@ object Endpoints {
   val wsSubprotocol: Header =
     sttp.model.Header("Sec-WebSocket-Protocol", "daml.ws.auth")
 
+  private val wsJwtTokenPrefix = "jwt.token."
+
+  // Extracts the JWT from a `Sec-WebSocket-Protocol` header value such as
+  // "daml.ws.auth, jwt.token.<TOKEN>" (RFC 6455 clients join subprotocols with ", ").
+  // Elements must be trimmed before matching the prefix, since the leading space on
+  // every element but the first would otherwise make `startsWith` fail to find the token.
+  def extractWsJwtToken(headerValue: Option[String]): Option[Jwt] =
+    headerValue
+      .map(_.split(",").toSeq.map(_.trim))
+      .getOrElse(Seq.empty)
+      .filter(_.startsWith(wsJwtTokenPrefix))
+      .map(_.substring(wsJwtTokenPrefix.length))
+      .headOption
+      .map(Jwt.apply)
+
   lazy val baseEndpoint: Endpoint[CallerContext, Unit, Unit, Unit, Any] = endpoint
     .securityIn(
       auth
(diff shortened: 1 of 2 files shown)


moritzkiefer-da:
@omaiko-intellecteu this is the wrong repository. The canton fork in splice is not actually used for the participant, we only reuse some of the infra in splice. You need to make a PR against the Canton repository.

PR #6792, opened 2026-08-16
Public Sequencer: Create ValidatorUnpermission template and DAML tests

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/6791


daml/splice-amulet/daml/Splice/ValidatorLicense.daml
```diff
@@ -255,9 +258,39 @@ template ValidatorLicenseRequest with
         assertDeadlineExceeded "expiresAt" expiresAt
         return ValidatorLicenseRequest_ExpireResult {}
 
+-- | ValidatorUnpermission contract is created alongside the ValidatorLicense to manage
+-- the validator's permissions on the synchronizer.
+-- We opted to create ValidatorUnpermission together with validatorLicense, instead of later (for e.g. when validator is
```
martinflorian-da:
```suggestion
-- We create ValidatorUnpermission together with validatorLicense, instead of later (for e.g. when validator is
```

martinflorian-da: Also I believe there needs to be `, e.g.,` around `e.g.` (also below if you want to change that)

daml/splice-amulet/daml/Splice/ValidatorLicense.daml
```diff
@@ -255,9 +258,39 @@ template ValidatorLicenseRequest with
...
+-- unpermissioned for the first time),
+-- to avoid possible race conditions (for e.g. when more that 2 validator unpermission vote requests are initiated
```
martinflorian-da: Dumb question: If we create both contracts at the same time and there is a 1:1 mapping between them, why did we opt for having a dedicated `Unpermission` contract instead of extending `ValidatorLicense` with fields to cover unpermissioning?

martinflorian-da (commented):
Thanks! I'm mainly curious here; @cocreature needs to review all Daml for this project anyway (unless we change this explicitly).

daml/splice-amulet/daml/Splice/ValidatorLicense.daml
moritzkiefer-da: The problem with ValidatorLicense is that it has the validator as an observer. That gives all kind of anoying possibilities to stop the SVs from updating it.

daml/splice-amulet/daml/Splice/ValidatorLicense.daml
```diff
@@ -255,9 +258,39 @@ template ValidatorLicenseRequest with
...
+-- concurrently) and to avoid needing periodic merge logic.
+template ValidatorUnpermission
+  with
+    dso : Party
+    validator : Party
+    loginAfter : Optional Time
+    revoked : Bool
+  where
+    signatory dso
+    observer validator
```
moritzkiefer-da: I don't think we want the observer on this, otherwise if that node unvets the package you can't block it. I was hoping we don't even need to create this eagerly but can only create it when actually blocking a node.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1071,6 +1076,31 @@ template DsoRules with
...
+        updateResult <- exercise validatorUnpermissionCid ValidatorUnpermission_Update with
+          newLoginAfter
+          newRevoked
+
+        let newValidatorUnpermissionCid = updateResult.newValidatorUnpermissionCid
+
+        if newRevoked then
+          forA_ optValidatorLicenseCid $ \licenseCid -> do
+            _ <- fetchChecked (ForDso with dso) licenseCid
+            void $ exercise licenseCid ValidatorLicense_Withdraw with reason = "Permanently unpermissioned by SV vote"
```
martinflorian-da: Shall we add an assertion that the license matches the unpermission contract? (same validator party?)

martinflorian-da: And btw is there ever a valid case where `newRevoked == true` but `optValidatorLicenseCid == None`?

daml/splice-amulet/daml/Splice/ValidatorLicense.daml
pasindutennage-da: @moritzkiefer-da 

`I was hoping we don't even need to create this eagerly but can only create it when actually blocking a node.`

How strongly are you proposing this?

IMO, early creation significantly simplifies code + avoids voting races and retries + and no need for periodic merge automations. 

I am tempted to justify the additional O(num_validators) storage. 

WDYT?

moritzkiefer-da: > I am tempted to justify the additional O(num_validators) storage.

My concern isn't really around storage. Take a malicious validator. They can get permissioned but never request their license. How do you unpermission them?

pasindutennage-da: Agreed. Thank you.

daml/splice-amulet/daml/Splice/ValidatorLicense.daml
pasindutennage-da: Fixed

daml/splice-dso-governance/daml/Splice/DsoRules.daml
pasindutennage-da: Thank you. I fixed both of them.

daml/splice-amulet/daml/Splice/ValidatorLicense.daml
pasindutennage-da: @moritzkiefer-da All comments addressed. Ready for another review. Thank you

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1071,6 +1079,84 @@ template DsoRules with
...
+          require "optValidatorLicenseCid must be provided when permanently revoking a validator" (isSome optValidatorLicenseCid)
+        else
+          pure ()
+
+        -- Ensure the provided license actually belongs to this validator
+        forA_ optValidatorLicenseCid $ \licenseCid -> do
+          license <- fetchChecked (ForDso with dso) licenseCid
+          require "optValidatorLicenseCid must belong to the same validator" (license.validator == validator)
+
+        -- Handle License Archival
```
moritzkiefer-da: I don't understand why we need this case. Why do we care about revoking the license Validator permissioning doesn't depend on it?

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1071,6 +1079,84 @@ template DsoRules with
...
         return DsoRules_ExpireValidatorLicenseRequestResult
 
+    nonconsuming choice DsoRules_UnpermissionValidator : DsoRules_UnpermissionValidatorResult
+      with
+        validator : Party
+        participantId : Text
+        optValidatorUnpermissionCid : Optional (ContractId ValidatorUnpermission)
+        newLoginAfter : Optional Time
+        newRevoked : Bool
+        optValidatorLicenseCid : Optional (ContractId ValidatorLicense)
```
moritzkiefer-da: pinning the contract id in the vote seems problematic to me if that changes concurrently. Take a look at the instruction pattern that we use for example for BootstrapExternalPartyConfigStateInstruction to avoid pinning contract ids in the vote.

daml/splice-amulet/daml/Splice/ValidatorLicense.daml
```diff
@@ -255,9 +258,36 @@ template ValidatorLicenseRequest with
         assertDeadlineExceeded "expiresAt" expiresAt
         return ValidatorLicenseRequest_ExpireResult {}
 
+-- | ValidatorUnpermission contract is created to manage
+-- the validator's permissions on the synchronizer.
+-- It is created lazily when SVs successfully vote to restrict a validator's permissions.
+template ValidatorUnpermission
```
moritzkiefer-da: given that this is not required by validators can we move it to splice-dso-governance

daml/splice-amulet/daml/Splice/ValidatorLicense.daml
```diff
@@ -255,9 +258,36 @@ template ValidatorLicenseRequest with
...
+  with
+    dso : Party
+    validator : Party
+    participantId : Text
+    loginAfter : Optional Time
+    revoked : Bool
```
moritzkiefer-da: add comments on what these fields mean

daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -1071,6 +1079,84 @@ template DsoRules with
         exercise validatorLicenseRequestCid ValidatorLicenseRequest_Expire
         return DsoRules_ExpireValidatorLicenseRequestResult
 
+    nonconsuming choice DsoRules_UnpermissionValidator : DsoRules_UnpermissionValidatorResult
+      with
+        validator : Party
+        participantId : Text
+        optValidatorUnpermissionCid : Optional (ContractId ValidatorUnpermission)
+        newLoginAfter : Optional Time
+        newRevoked : Bool
```
moritzkiefer-da: I'd remove the `new` prefix from these.

daml/splice-dso-governance/daml/Splice/DsoRules.daml
moritzkiefer-da: alternatively, can we just always create a new unpermission contract and rely on merging it with an existing one instead of needing to update it here (I realize I commented on the wrong contract id, imho we should just not archive the license, I'm taking about the ValidatorUnpermission contract)

moritzkiefer-da (commented):
thanks, mostly there but needs a bit more fine tweaking

daml/splice-amulet/daml/Splice/ValidatorLicense.daml
```diff
@@ -255,9 +258,36 @@ template ValidatorLicenseRequest with
         assertDeadlineExceeded "expiresAt" expiresAt
         return ValidatorLicenseRequest_ExpireResult {}
 
+-- | ValidatorUnpermission contract is created to manage
+-- the validator's permissions on the synchronizer.
+-- It is created lazily when SVs successfully vote to restrict a validator's permissions.
+template ValidatorUnpermission
+  with
+    dso : Party
+    validator : Party
```
moritzkiefer-da: I don't think we need the validator party id, just the participant as that is the one we're unpermissioning.

daml/splice-dso-governance/daml/Splice/DsoRules.daml

daml/splice-dso-governance/daml/Splice/DsoRules.daml
pasindutennage-da: Fixed. Opted to always create a new Unpermission

daml/splice-amulet/daml/Splice/ValidatorLicense.daml
pasindutennage-da: Fixed

daml/splice-amulet/daml/Splice/ValidatorLicense.daml
pasindutennage-da: Fixed

daml/splice-dso-governance/daml/Splice/DsoRules.daml

daml/splice-amulet/daml/Splice/ValidatorLicense.daml
pasindutennage-da: Fixed

pasindutennage-da:
@moritzkiefer-da Thanks for the feedback. All addressed and ready for review.

moritzkiefer-da (approved):
nice thanks!

PR #6801, opened 2026-08-17
helm: allow setting priorityClassName on node pods

### LINKED ISSUE #6800: helm: no way to set priorityClassName on node pods
Running more than one Splice node on shared Kubernetes infrastructure is the normal case, not an edge case: an operator typically runs DevNet and TestNet nodes alongside MainNet so it can validate a release before taking it to production, and node providers run nodes for several parties on shared capacity. In both situations the pods end up co-resident.

There is currently no way to express that the MainNet pods matter more. The charts expose `nodeSelector`, `affinity` and `tolerations`, but not `spec.priorityClassName`, so every Splice pod sits at priority 0. Under node pressure the kubelet has nothing to distinguish a production validator from a DevNet one, and the scheduler has nothing to preempt with — a DevNet pod can win a scheduling race, or survive an eviction, at the expense of MainNet. The usual Kubernetes answer to exactly this is a PriorityClass per tier, which the charts do not currently allow.

Setting `priorityClassName` in values today is a silent no-op rather than an error: the charts leave `additionalProperties` unset in `values.schema.json`, so the key validates fine but is never read by the templates. That makes it easy to believe it has been configured when it has not.

I have a patch that adds it alongside the existing placement values, including in the shared `splice-util-lib.postgres-metrics` helper.

SLoeuillet:
Fixes #6800.

**Motivation:** running several Splice nodes on shared Kubernetes infrastructure is the normal operator topology — DevNet and TestNet alongside MainNet so a release can be validated before production, and node providers running nodes for several parties on shared capacity. Today every Splice pod sits at priority 0, so nothing expresses that MainNet outranks the rest: under node pressure the kubelet cannot tell a production validator from a DevNet one, and the scheduler has nothing to preempt with. A PriorityClass per tier is the standard Kubernetes answer, and the charts currently make it unreachable.

Adds an optional `priorityClassName`, rendered next to the existing `nodeSelector` / `affinity` / `tolerations`, and threaded through the shared `splice-util-lib.postgres-metrics` helper so the metrics deployment picks it up too. Nothing is rendered when the value is unset, so this is a no-op for anyone who does not set it.

It is declared in `values.schema.json` only for the charts that already declare `tolerations` there, so this does not start declaring placement values for charts that currently leave them undeclared.

Excluded: `splice-postgres` and `splice-istio-gateway` (both deprecated), and the dev/demo charts (`cn-docs`, `splice-info`, `splice-load-tester`, `splice-party-allocator`, `splice-splitwell-*`). Happy to extend to those if you would rather have the value everywhere.

Tested with `helm unittest` on all seven charts touched: existing tests still pass, plus new set/unset tests for `splice-participant` and `splice-validator`.

### DIFF
--- a/cluster/helm/splice-global-domain/templates/mediator.yaml
+++ b/cluster/helm/splice-global-domain/templates/mediator.yaml
@@ -155,6 +155,9 @@ spec:
                 echo "trying to create postgres database {{ .Values.mediator.persistence.databaseName }}, last error: $errmsg";
                 sleep 2;
               done
+      {{- with .Values.priorityClassName }}
+      priorityClassName: {{ . }}
+      {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
@@ -227,5 +230,5 @@ data:
 {{ .Files.Get "files/logback.xml" | indent 4 }}
 ---
 {{- if .Values.enablePostgresMetrics }}
-{{- include "splice-util-lib.postgres-metrics" (dict "persistence" .Values.mediator.persistence "namespace" .Release.Namespace "nodeSelector" .Values.nodeSelector "affinity" .Values.affinity "tolerations" .Values.tolerations "serviceAccountName" .Values.serviceAccountName ) }}
+{{- include "splice-util-lib.postgres-metrics" (dict "persistence" .Values.mediator.persistence "namespace" .Release.Namespace "nodeSelector" .Values.nodeSelector "affinity" .Values.affinity "tolerations" .Values.tolerations "priorityClassName" .Values.priorityClassName "serviceAccountName" .Values.serviceAccountName ) }}
 {{- end}}
--- a/cluster/helm/splice-global-domain/templates/mediator.yaml
+++ b/cluster/helm/splice-global-domain/templates/mediator.yaml
@@ -156,7 +156,7 @@ spec:
                 sleep 2;
               done
       {{- with .Values.priorityClassName }}
-      priorityClassName: {{ . }}
+      priorityClassName: {{ . | quote }}
       {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
--- a/cluster/helm/splice-util-lib/templates/_helpers.tpl
+++ b/cluster/helm/splice-util-lib/templates/_helpers.tpl
@@ -67,6 +67,7 @@ valueFrom:
 {{- $nodeSelector := .nodeSelector }}
 {{- $affinity := .affinity }}
 {{- $tolerations := .tolerations }}
+{{- $priorityClassName := .priorityClassName }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -127,6 +128,9 @@ spec:
                 echo "Waiting for database {{ $persistence.databaseName }}, at hostname {{ $persistence.host }}, port {{ $persistence.port | default 5432 }} to be accessible. Last error: $errmsg"
                 sleep 2;
             done
+      {{- with $priorityClassName }}
+      priorityClassName: {{ . }}
+      {{- end }}
       {{- with $nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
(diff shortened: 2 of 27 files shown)


cluster/helm/splice-util-lib/templates/_helpers.tpl
```diff
@@ -127,6 +128,9 @@ spec:
                 echo "Waiting for database {{ $persistence.databaseName }}, at hostname {{ $persistence.host }}, port {{ $persistence.port | default 5432 }} to be accessible. Last error: $errmsg"
                 sleep 2;
             done
+      {{- with $priorityClassName }}
+      priorityClassName: {{ . }}
```
martinflorian-da:
```suggestion
      priorityClassName: {{ . | quote }}
```

I think it might otherwise get confused in the exotic case that the priority class is named "42"; maybe best to check by adding a helm test case whether this makes sense.

cluster/helm/splice-global-domain/templates/mediator.yaml
```diff
@@ -227,5 +230,5 @@ data:
 {{ .Files.Get "files/logback.xml" | indent 4 }}
 ---
 {{- if .Values.enablePostgresMetrics }}
-{{- include "splice-util-lib.postgres-metrics" (dict "persistence" .Values.mediator.persistence "namespace" .Release.Namespace "nodeSelector" .Values.nodeSelector "affinity" .Values.affinity "tolerations" .Values.tolerations "serviceAccountName" .Values.serviceAccountName ) }}
+{{- include "splice-util-lib.postgres-metrics" (dict "persistence" .Values.mediator.persistence "namespace" .Release.Namespace "nodeSelector" .Values.nodeSelector "affinity" .Values.affinity "tolerations" .Values.tolerations "priorityClassName" .Values.priorityClassName "serviceAccountName" .Values.serviceAccountName ) }}
```
martinflorian-da, line 233: That is a bit complex - shall we add a test case that covers this?

martinflorian-da (approved):
Thanks!

Main question: Did you test this on a cluster? Can you confirm that including `priorityClassName` in a deployment definition does what you expect it does?

More off-topic:

>  running several Splice nodes on shared Kubernetes infrastructure is the normal operator topology

Are you (and others) aware that this is mode of operating is 1. problematic for security reasons (compromise of a DevNet node gives an attacker easy access to MainNet nodes as well) and 2. not actively being tested by the Splice team? We might need to update our public docs here about these points...

cluster/helm/splice-util-lib/templates/_helpers.tpl
SLoeuillet: Good catch — applied, and you were right to ask for the test.

I took the quoting wider than the suggestion: the helper is one of **13** sites that emit this scalar (the other twelve are the chart templates, each emitting it independently), so quoting only `_helpers.tpl` would have left a class named `42` broken everywhere else. All of them are quoted now.

The test earns its place: `renders a numeric-looking priorityClassName as a string` on `splice-validator` and `splice-participant` **fails against the unquoted template** and passes once quoted — I checked both directions rather than assuming.

cluster/helm/splice-global-domain/templates/mediator.yaml
SLoeuillet: Agreed, and added — two tests on `mediator_test.yaml`: `priorityClassName` reaching the postgres-metrics sidecar deployment through the `splice-util-lib.postgres-metrics` include, and its absence when unset.

This was worth covering exactly because of the complexity you flagged: the value is passed positionally inside a `dict` alongside six other arguments, so a future edit could drop it with nothing else noticing. I verified the test catches that — removing `"priorityClassName" .Values.priorityClassName` from the include makes it fail.

One wrinkle for whoever reads the test later: the postgres-metrics deployment shares its name with a Service and a ServiceMonitor, so `documentSelector` on `metadata.name` matches three documents and errors with `multiple indexes found`. The tests select on `spec.template.metadata.labels.app` instead.

martinflorian-da (approved):
Thank you and thank you for confirming that you tested end-to-end!

Please resolve the merge conflict and ping me or @isegall-da so we can click merge for you. ("On Slack" works too.)

PR #6803, opened 2026-08-17
helm: make the Postgres role and bootstrap database configurable for validators

### LINKED ISSUE #6802: helm: Postgres role and bootstrap database are hardcoded in the pg-init containers
Since `splice-postgres` is deprecated and unsupported after 2026-11-12, every operator currently using it has to move to a Postgres instance they provision themselves before that date. The charts do not currently let them do that cleanly.

The pg-init and wait containers connect with a hardcoded `--username=cnadmin --dbname=cantonnet`, and the postgres-exporter sidecar in `splice-util-lib` with a hardcoded `DATA_SOURCE_USER: cnadmin`.

That assumes the target instance has a `cnadmin` superuser and a database literally named `cantonnet`. Both hold for `splice-postgres`, because that chart creates them — which is exactly why the assumption has been invisible so far. They do not generally hold for a managed instance (Cloud SQL, RDS, Azure Database) or for any instance where the DBA picks the role and database names. On several managed offerings you cannot create a role called `cnadmin` with real superuser rights at all.

So an operator following the deprecation notice has to either rename their roles and databases to match the charts, or patch the charts. Given the deprecation applies to everyone still on `splice-postgres`, this seems likely to come up repeatedly between now and November.

It is also inconsistent within a single chart today: `splice-global-domain` already accepts `persistence.user` and honours it for the application connection, while its own init containers still connect as `cnadmin` regardless. `splice-scan` and `splice-sv-node` list `user` as required in their schema without using it in their init containers.

Affected: `splice-participant`, `splice-validator`, `splice-scan`, `splice-sv-node`, `splice-domain`, `splice-global-domain`, and the `splice-util-lib` postgres-metrics helper. (`splice-splitwell-app` has the same pattern.)

I have a patch that reads both from values, defaulting to the current behaviour so nothing changes for existing users.

SLoeuillet:
Fixes #6802. **Scope reduced to validator and participant** per the discussion below — the SV charts and splitwell move to a follow-up.

## Why

`splice-postgres` is unsupported after 2026-11-12, so every operator on it has to move to a self-provisioned or managed Postgres. The charts assumed that migration target would offer a `cnadmin` superuser and a `cantonnet` database, which a managed offering may not permit creating at all — so the names blocked the very migration path upstream points at.

`persistence.user` and `persistence.bootstrapDatabaseName` are now read wherever those values were hardcoded, defaulting to `cnadmin` and `cantonnet` so **rendered output is byte-identical unless you set them**.

## The part that was not just the init container

The participant needed more than the pg-init fix. `canton-base`'s `storage.conf` reads `user = ${?CANTON_PARTICIPANT_POSTGRES_USER}`, and the chart never set it — so the node kept connecting as `cnadmin` even with `persistence.user` configured. That is a half-applied configuration: the init container uses the new role, the application does not, and nothing complains. Worse than not offering the value, since it looks configurable.

The image already supports the variable and the compose deployment already passes it (`CANTON_PARTICIPANT_POSTGRES_USER=${SPLICE_DB_USER}`); Helm was the only deployment path missing it. `splice-validator` was never affected — it writes `user` straight into its HOCON storage block.

## Testing

Helm deployment on a `kind` cluster, against a Postgres initialised with `POSTGRES_USER=appuser` / `POSTGRES_DB=bootstrapdb`, so `cnadmin` and `cantonnet` **do not exist**:

- participant `1/1 Running` — `participant_test owner=appuser`, `CANTON_PARTICIPANT_POSTGRES_USER=appuser`, 18 Flyway migrations, `Canton started`, and Postgres reports `appuser -> participant_test (32 conn)`
- validator app completes its DB work — `validator_test owner=appuser`, schema `validator owner=appuser` — then stops at the ledger-API wiring, which was deliberately not configured
- **negative control**: the unpatched charts with identical values loop on `FATAL: password authentication failed for user "cnadmin"` at `Init:0/1`, with no user env var rendered at all

Chart unit tests cover the defaults, an override, and that neither default survives in the rendered command; each was checked to fail without the change rather than assumed.

## Not in this PR

`splice-sv-node`, `splice-scan`, `splice-global-domain` (sequencer + mediator), `splice-domain` and `splice-splitwell-app` carry the same hardcoding and the same fix, but SV deployments have more moving parts and the cluster test coverage for them is not there yet. That work is preserved and will follow once it is.

### DIFF
--- a/cluster/helm/splice-participant/templates/participant.yaml
+++ b/cluster/helm/splice-participant/templates/participant.yaml
@@ -55,6 +55,10 @@ spec:
           value: {{ .Values.persistence.port | quote }}
         - name: CANTON_PARTICIPANT_POSTGRES_SCHEMA
           value: {{ .Values.persistence.schema }}
+        {{- with .Values.persistence.user }}
+        - name: CANTON_PARTICIPANT_POSTGRES_USER
+          value: {{ . | quote }}
+        {{- end }}
         - name: CANTON_PARTICIPANT_POSTGRES_PASSWORD
           {{- if ((.Values.persistence).secretOverrides).postgresPassword }}
           value: {{ .Values.persistence.secretOverrides.postgresPassword | quote }}
@@ -186,7 +190,7 @@ spec:
               {{- end }}
               DB_TO_CREATE='{{ .Values.persistence.databaseName }}'
 
-              until db_exists=$(psql -h {{ .Values.persistence.host }} -p {{ .Values.persistence.port }} --username=cnadmin --dbname=cantonnet -tAc "SELECT 1 FROM pg_database WHERE datname='${DB_TO_CREATE}'" 2>&1); do
+              until db_exists=$(psql -h {{ .Values.persistence.host }} -p {{ .Values.persistence.port }} --username={{ .Values.persistence.user | default "cnadmin" }} --dbname={{ .Values.persistence.bootstrapDatabaseName | default "cantonnet" }} -tAc "SELECT 1 FROM pg_database WHERE datname='${DB_TO_CREATE}'" 2>&1); do
                 echo "trying to connect to postgres, last error: ${db_exists}";
                 sleep 2;
               done
@@ -195,7 +199,7 @@ spec:
                 echo "Database ${DB_TO_CREATE} already exists. Done."
               else
                 {{- if $conflictCheck }}
-                existing=$(psql -h {{ .Values.persistence.host }} -p {{ .Values.persistence.port }} --username=cnadmin --dbname=cantonnet -tAc "SELECT datname FROM pg_database WHERE datname LIKE 'participant%'" | paste -sd ', ' -)
+                existing=$(psql -h {{ .Values.persistence.host }} -p {{ .Values.persistence.port }} --username={{ .Values.persistence.user | default "cnadmin" }} --dbname={{ .Values.persistence.bootstrapDatabaseName | default "cantonnet" }} -tAc "SELECT datname FROM pg_database WHERE datname LIKE 'participant%'" | paste -sd ', ' -)
                 if [ -n "${existing}" ]; then
                   echo "ERROR: Refusing to create participant database '${DB_TO_CREATE}' because the following participant database(s) already exist: ${existing}." >&2
                   echo "       A participant database should only be created during a fresh deployment. Please double-check that this is intended and not a misconfiguration." >&2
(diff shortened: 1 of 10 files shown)


martinflorian-da (approved):
Thank you @SLoeuillet !

I think I would feel slightly better if we could add 1-2 helm tests (perhaps for validator + participant?) that cover the new value, to guard against regressions.

More on the nit side: we could also cover `splitwell.yaml`, for symmetry... (Really we should think about some larger refactoring that avoids all this code duplication, but doesn't have to be in this PR.)

Also please rebase to latest main again to fix your merge conflict and consume the fix for the reason your deployment test failed... sorry for the trouble.

martinflorian-da (changes requested):
Argh apologies for missing that earlier, we should also fix the sequencer:

```
> rg "\-\-dbname=cantonnet" cluster/helm/
cluster/helm/splice-splitwell-app/templates/splitwell.yaml
128:              until errmsg=$(psql -h {{ .Values.persistence.host }} -p {{ .Values.persistence.port }} --username=cnadmin --dbname=cantonnet -c 'create database {{ .Values.persistence.databaseName }}' 2>&1); do

cluster/helm/splice-global-domain/templates/sequencer.yaml
266:              until psql -h "$DB_HOST" -p "$DB_PORT" --username="$DB_USER" --dbname=cantonnet -tAc 'SELECT 1' >/dev/null 2>&1; do
270:              if [ "$(psql -h "$DB_HOST" -p "$DB_PORT" --username="$DB_USER" --dbname=cantonnet -tAc "SELECT 1 FROM pg_database WHERE datname = '${DB_NAME}'")" = "1" ]; then
274:              psql -v ON_ERROR_STOP=1 -h "$DB_HOST" -p "$DB_PORT" --username="$DB_USER" --dbname=cantonnet -c "CREATE DATABASE \"${DB_NAME}\"";
```

And even bigger miss: for participant/domain/mediator/sequencer  `persistence.user` is not actually used for anything but the init container atm? You want to be overriding `POSTGRES_USER` somewhere: https://github.com/digital-asset/canton/blob/72c82f382942e3bda7d6418f72f95b1524f57ca8/docker/canton/images/canton-base/storage.conf#L9

Given that both @isegall-da and myself missed these things on the first pass makes me want to ask you for some e2e confirmation on a cluster that this is working as expected, with dbname overridden... Did you try these changes out?

martinflorian-da (approved):
Perfect, thank you!

PR #6805, opened 2026-08-17
helm: quote validator wallet user ids

### LINKED ISSUE #6804: helm: validator wallet user ids are rendered unquoted, breaking ids that contain '@'
`splice-validator` renders `validatorWalletUsers` into a HOCON fragment unquoted:

```
canton.validator-apps.validator_backend.validator-wallet-users.0 = {{ $user }}
```

and `validatorWalletUser` into an env var value unquoted. Both assume the user id is a bare alphanumeric string.

Two ways that breaks:

- **`@` is not a legal character in an unquoted HOCON string.** Identity providers that use the email address as the user id therefore produce a config the validator app cannot parse. Rendering `validatorWalletUsers: [alice@example.com]` today gives `validator-wallet-users.0 = alice@example.com`.
- **An all-digit id loses its type.** In the HOCON fragment it becomes a number rather than a string; in the env var it renders as a YAML integer, which is not a valid `env[].value` and is rejected on apply.

The workaround is to embed literal double quotes in the value, i.e. `validatorWalletUsers: ["alice@example.com"]` in YAML so that the quotes survive into the rendered HOCON. That works but is easy to get wrong and hard to guess.

This looks like an oversight rather than intent, since the surrounding code already quotes comparable values: `scanAddress` uses `| quote`, and the scan-client `sv-names` / `seed-urls` lists use `| toJson`. Every other bare `= {{ ... }}` in the file is numeric.

Patch to follow.

SLoeuillet:
Fixes #6804.

Adds `| quote` to the two places a validator wallet user id is rendered: the `validatorWalletUsers` HOCON fragment and the `validatorWalletUser` env var value.

Before, with `validatorWalletUsers: [alice@example.com, "12345678"]`:

```
validator-wallet-users.0 = alice@example.com
validator-wallet-users.1 = 12345678
```

`@` is not legal in an unquoted HOCON string, and the second is a number rather than a string. After:

```
validator-wallet-users.0 = "alice@example.com"
validator-wallet-users.1 = "12345678"
```

Same for the env var, where an all-digit id previously rendered as a YAML integer and so was not a valid `env[].value`.

**Behaviour change worth flagging:** anyone who worked around this by embedding literal double quotes in the value (`"alice@example.com"`) must remove them, otherwise the id now comes out double-quoted. Ids that are bare alphanumeric strings are unaffected.

Added `tests/wallet-user-ids_test.yaml` covering the `@` and all-digit cases. Both tests fail on `main` and pass with this change; the 42 existing `splice-validator` tests are untouched and still pass.

Independent of #6801 and #6803 — separate branch off `main`, no overlapping files.


moritzkiefer-da (approved):
thx

PR #6812, opened 2026-08-17
extend splice rate limits 

nicu-da:
re reverts https://github.com/canton-network/splice/pull/6807 and fixes ci 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/ClientIpDirectives.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/ClientIpDirectives.scala
@@ -1,0 +1,69 @@
+// Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package org.lfdecentralizedtrust.splice.http
+
+import org.apache.pekko.http.scaladsl.model.headers.{`X-Forwarded-For`, `X-Real-Ip`}
+import org.apache.pekko.http.scaladsl.model.{AttributeKeys, RemoteAddress}
+import org.apache.pekko.http.scaladsl.server.Directive1
+import org.apache.pekko.http.scaladsl.server.Directives.*
+
+object ClientIpDirectives {
+
+  /** Extracts the address of the client the request originated from, if it can be determined.
+    *
+    * The address is taken from the first of the following sources that yields an address:
+    *   1. the `trustedClientIpHeader` (if configured and parseable as an IP literal), which is set
+    *      by a trusted reverse proxy and hence cannot be spoofed by the client,
+    *   1. the client-controlled `X-Forwarded-For` header,
+    *   1. the client-controlled `X-Real-Ip` header,
+    *   1. the remote address of the transport connection, if the server is configured to expose it.
+    *
+    * @param trustedClientIpHeader
+    *   name of the header set by a trusted reverse proxy, matched case-insensitively. An empty name
+    *   disables trusting a proxy header.
+    */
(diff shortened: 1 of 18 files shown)


martinflorian-da (approved):
d'oh thanks

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/http/ClientIpDirectives.scala
```diff
@@ -0,0 +1,64 @@
...
+    *
+    * @param trustedClientIpHeader
+    *   name of the header set by a trusted reverse proxy, matched case-insensitively. An empty name
+    *   disables trusting a proxy header.
+    */
+  def extractClientIp(trustedClientIpHeader: String): Directive1[Option[RemoteAddress]] =
+    firstDefined(
+      trustedClientIp(trustedClientIpHeader),
+      forwardedForClientIp,
+      realIpClientIp,
```
stas-sbi: Having this hardcoded means that an operator have to know to strip the relevant headers from clients.

What would you say about using client IP from the TCP connection by default and add an option to get it from a header instead? Something like:
```
set-source-ip-from:
  headers: X-Envoy-External-Address, X-Forwarded-For, X-Real-Ip
  trust-from-addresses: cidr1, cidr2, ...
```

giner, line 29: Having this hardcoded means that an operator have to know to strip the relevant headers from clients.

What would you say about using client IP from the TCP connection by default and add an option to get it from a header instead? Something like:
```
set-source-ip-from:
  headers: X-Envoy-External-Address, X-Forwarded-For, X-Real-Ip
  trust-from-addresses: cidr1, cidr2, ...
```

giner: @nicu-da @martinflorian-da

nicu-da: @giner we will add documentation for this. 
No real way to get any real ip from the TCP connection as most likely that would be just some sidecar/loadbalancer/gateway. 
Also the header used is configurable, so you can configure your own trusted header and then make sure the networking infra handles it.

giner: >  No real way to get any real ip from the TCP connection as most likely that would be just some sidecar/loadbalancer/gateway

This depends on the infra setup (it is true that getting client's IP for a TLS connection is not possible because the app doesn't handle SSL termination so there should be something in front of it). In any case it should be up to the operator to decide where they get the IP from.

>  Also the header used is configurable, so you can configure your own trusted header and then make sure the networking infra handles it

It is configurable but it falls back to `X-Forwarded-For` and `X-Real-Ip` without the operator knowing this, neither having a way to disable it.

giner: > No real way to get any real ip from the TCP connection as most likely that would be just some sidecar/loadbalancer/gateway

Thinking about this once again, I guess getting real IP from the TCP connection is not very useful for Splice apps (no SSL termination so no direct connection from clients). May be `set-source-ip-from` should be made mandatory when per-client rate limits are configured.

nicu-da: @giner good point for no ability to disable that, let me add that as an option. 
We will add the docs but for now the release notes are the main documentation.

PR #6819, opened 2026-08-18
recreate node pools on machine type changes to work around a provider bug

mblaze-da:
part of https://github.com/canton-network/splice/issues/6818


OriolMunoz-da (approved):
thanks!

PR #6823, opened 2026-08-18
Restart CometBFT when it starts replaying messages

moritzkiefer-da:
We can't reproduce the issue, so I can't really tell whether it works exactly. What I did test is:

1. Deploy on scratchnet and check that the computed numbers are plausible.
2. Change the script to always fail and check that cometbft gets restarted.
3. The logs show up properly in gcp.

After merging, I would suggest we also at least temporarily deploy a snapshot on one of our mainnet nodes to validate the numbers there as well.

There is also the question of whether we want to fasttrack getting this cometbft change to mainnet for all SVS but that's somewhat orthogonal to the actual change so let's pick that up in slack once we are happy with the change.

I did consider alternatively having this be a process in the cometbft container but in the end a separate sidecar that you can easily filter logs for and don't have to worry about logs interleaving seemed cleaner to me without that much more complexity.

[static]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/cluster/helm/splice-cometbft/tests/cometbft_watchdog_test.yaml
+++ b/cluster/helm/splice-cometbft/tests/cometbft_watchdog_test.yaml
@@ -1,0 +1,98 @@
+# Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+suite: "cometbft watchdog"
+templates:
+  - deployment.yaml
+release:
+  name: global-domain-3
+set:
+  sv1:
+    externalAddress: sv.mock.net:26316
+    nodeId: "4c7c99516fb3309b89b7f8ed94690994c8ec0ab0"
+    publicKey: "H2bcJU2zbzbLmP78YWiwMgtB0QG1MNTSozGl1tP11hI="
+    keyAddress: "9473617BBC80C12F68CC25B5A754D1ED9035886C"
+  node:
+    id: "1234dedbeef1234dedbeef1234dedbeef1234ded"
+    identifier: "global-domain-3-cometbft"
+    externalAddress: "global-domain-3-cometbft.sv.mock.net:26356"
+    keysSecret: "cometbft-keys"
+  genesis:
+    chainId: "mock-3"
+    chainIdSuffix: "1"
+tests:
+  - it: "rejects the chart defaults, so that running without a watchdog is a choice"
+    template: deployment.yaml
+    asserts:
+      - failedTemplate:
+          errorPattern: "missing properties 'sequencerMetricsUrl', 'mediatorMetricsUrl'"
+
+  - it: "leaves the pod untouched when disabled"
+    template: deployment.yaml
+    set:
+      watchdog:
+        enabled: false
+    documentSelector:
+      path: kind
+      value: Deployment
+    asserts:
+      - lengthEqual:
+          path: spec.template.spec.containers
+          count: 1
+      - equal:
+          path: spec.template.spec.containers[0].livenessProbe.httpGet
+          value:
+            path: /health
+            port: rpc
+      - notMatchRegex:
+          path: spec.template.spec.containers[0].command[4]
+          pattern: "restart-requested"
+
+  - it: "deploys the watchdog sidecar when enabled"
+    template: deployment.yaml
+    set:
+      watchdog:
+        enabled: true
+        sequencerMetricsUrl: http://global-domain-3-sequencer:10013/metrics
+        mediatorMetricsUrl: http://global-domain-3-mediator:10013/metrics
+        threshold: 2.5
+        evaluationIntervalSeconds: 900
+    documentSelector:
+      path: kind
+      value: Deployment
+    asserts:
+      - lengthEqual:
+          path: spec.template.spec.containers
+          count: 2
+      - equal:
+          path: spec.template.spec.containers[1].name
+          value: "watchdog"
+      - matchRegex:
+          path: spec.template.spec.containers[1].image
+          pattern: "/cometbft-watchdog:"
+      - contains:
+          path: spec.template.spec.containers[1].env
+          content:
+            name: WATCHDOG_MEDIATOR_METRICS_URL
+            value: "http://global-domain-3-mediator:10013/metrics"
+      - contains:
+          path: spec.template.spec.containers[1].env
+          content:
+            name: WATCHDOG_THRESHOLD
+            value: "2.5"
+      - contains:
+          path: spec.template.spec.containers[1].env
+          content:
+            name: WATCHDOG_EVALUATION_INTERVAL_SECONDS
+            value: "900"
+      - contains:
+          path: spec.template.spec.containers[1].volumeMounts
+          content:
+            mountPath: /watchdog
+            name: watchdog
+      - contains:
+          path: spec.template.spec.volumes
+          content:
+            name: watchdog
+            emptyDir:
+              sizeLimit: 1Mi
--- a/cluster/helm/splice-cometbft/tests/cometbft_watchdog_test.yaml
+++ b/cluster/helm/splice-cometbft/tests/cometbft_watchdog_test.yaml
@@ -21,7 +21,7 @@ set:
     chainId: "mock-3"
     chainIdSuffix: "1"
 tests:
-  - it: "rejects the chart defaults, so that running without a watchdog is a choice"
+  - it: "rejects the chart defaults, so that operators need to explicitly configure the right URLs"
     template: deployment.yaml
     asserts:
       - failedTemplate:
--- a/cluster/helm/splice-cometbft/values-template.yaml
+++ b/cluster/helm/splice-cometbft/values-template.yaml
@@ -111,6 +111,28 @@ mempool:
 # should stakater reloader annotation be added (note: reloader needs to be installed separately)
 enableReloader: true
 
+# Restarts CometBFT when cometbft starts replaying messages
+watchdog:
+  # Enabled by default to make sure noone accidentally forgets to enable it.
+  enabled: true
+  # Required but we don't know the migration id to set sensible defaults.
+  # sequencerMetricsUrl: http://global-domain-0-sequencer:10013/metrics
+  # mediatorMetricsUrl: http://global-domain-0-mediator:10013/metrics
+  pollIntervalSeconds: 30
+  threshold: 5
+  evaluationIntervalSeconds: 300
+  scrapeTimeoutSeconds: 10
+  startupGraceSeconds: 600
+  cooldownSeconds: 600
+  markerFile: /watchdog/restart-requested
+  imageName: "cometbft-watchdog"
+  resources:
+    limits:
+      memory: 128Mi
+    requests:
+      cpu: 0.05
+      memory: 64Mi
+
 securityContexts:
   pod:
     security_context_profile: pod_security_context
@@ -119,3 +141,6 @@ securityContexts:
   containers:
     cometbft:
       security_context_profile: app_security_context
+    watchdog:
+      security_context_profile: app_security_context
+      readOnlyRootFilesystem: true
--- a/cluster/images/cometbft-watchdog/Dockerfile
+++ b/cluster/images/cometbft-watchdog/Dockerfile
@@ -1,0 +1,16 @@
+# Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+FROM python:3.12-slim@sha256:2c941e860699f878900b0edc2403613c234d4b32eda3cc9fa7036991a2a63c4a
+
+LABEL org.opencontainers.image.base.name="python:3.12-slim"
+
+COPY restart-watchdog.py /restart-watchdog.py
+
+COPY target/LICENSE .
+
+# Ensure logs get output unbuffered even when the output is a pipe and don't write byte code in the container.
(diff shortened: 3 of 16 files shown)


cluster/helm/splice-cometbft/tests/cometbft_watchdog_test.yaml
```diff
@@ -0,0 +1,98 @@
...
+  node:
+    id: "1234dedbeef1234dedbeef1234dedbeef1234ded"
+    identifier: "global-domain-3-cometbft"
+    externalAddress: "global-domain-3-cometbft.sv.mock.net:26356"
+    keysSecret: "cometbft-keys"
+  genesis:
+    chainId: "mock-3"
+    chainIdSuffix: "1"
+tests:
+  - it: "rejects the chart defaults, so that running without a watchdog is a choice"
```
martinflorian-da:
```suggestion
  - it: "rejects the chart defaults, so that operators need to explicitly configure the right URLs"
```

cluster/images/cometbft-watchdog/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+FROM python:3.12-slim@sha256:2c941e860699f878900b0edc2403613c234d4b32eda3cc9fa7036991a2a63c4a
```
martinflorian-da, line 4: Hmm that adds a new external dependency that might start triggering vulnerability scanners at some point... (or worse, obviously).

Did you explicitly decide against da-base-image (and installing python there)?

cluster/images/cometbft-watchdog/restart-watchdog.py
```diff
@@ -0,0 +1,296 @@
...
+                    try:
+                        watchdog.request_restart(reason)
+                    except OSError as error:
+                        log("ERROR", f"Failed to write {config.marker_file}: {error}")
+                        return 1
+                    watchdog.reset()
+                    # don't kill cometbft after we just killed it.
+                    quiet_until = time.monotonic() + config.cooldown
+
+        time.sleep(max(0.0, config.poll_interval - (time.monotonic() - cycle_start)))
```
martinflorian-da: Shall we add some randomization, at least at the start (on top of quiet time)?

Synchronization effects where everyone restarts at the exactly the same time also sounds a bit scary to me.

docs/src/release_notes_upcoming.rst
```diff
@@ -21,36 +21,10 @@ release-notes:: Upcoming
...
-            also been added, allowing 100 requests per second burst / 50 sustained across all
-            operations combined, with an embedded per-client-IP limiter allowing 20 requests per
-            second burst / 10 sustained.
+    - CometBFT
+
+        - Added a watchdog to restart cometbft when we detect that it
+          is replaying messages. You must set
+          ``watchdog.sequencerMetricsUrl: global-domain-$SERIAL-sequencer:10013/metrics`` and
+          ``watchdog.mediatorMetricsUrl: http://global-domain-$SERIAL-mediator:10013/metrics`` in the
+          cometbft helm values. If needed, the watchdog can be disabled through ``watchdog.enabled: false``.
```
martinflorian-da, line 64: We should probably also extend `apps/app/src/pack/examples/sv-helm/cometbft-values.yaml` and update the public docs to reflect that?

docs/src/release_notes_upcoming.rst
```diff
@@ -21,36 +21,10 @@ release-notes:: Upcoming
...
-          - SV app: the per-operation burst limit has been lowered from 200 to 20 requests per
-            second, with a new sustained limit of 10 requests per second. A new global limiter has
-            also been added, allowing 100 requests per second burst / 50 sustained across all
-            operations combined, with an embedded per-client-IP limiter allowing 20 requests per
-            second burst / 10 sustained.
+    - CometBFT
+
+        - Added a watchdog to restart cometbft when we detect that it
+          is replaying messages. You must set
+          ``watchdog.sequencerMetricsUrl: global-domain-$SERIAL-sequencer:10013/metrics`` and
```
```suggestion
          ``watchdog.sequencerMetricsUrl: http://global-domain-$SERIAL-sequencer:10013/metrics`` and
```

cluster/helm/splice-cometbft/values-template.yaml
```diff
@@ -111,6 +111,28 @@ mempool:
...
+  enabled: true
+  # Required but we don't know the migration id to set sensible defaults.
+  # sequencerMetricsUrl: http://global-domain-0-sequencer:10013/metrics
+  # mediatorMetricsUrl: http://global-domain-0-mediator:10013/metrics
+  pollIntervalSeconds: 30
+  threshold: 5
+  evaluationIntervalSeconds: 300
+  scrapeTimeoutSeconds: 10
+  startupGraceSeconds: 600
+  cooldownSeconds: 600
```
martinflorian-da, line 126: I wonder if we want a higher default value here... not sure what happens during catchup for example.

Just thinking aloud; easy enough to tweak in the wild.

cluster/images/cometbft-watchdog/restart-watchdog.py
```diff
@@ -0,0 +1,296 @@
...
+        self.poll_interval = env_float("WATCHDOG_POLL_INTERVAL_SECONDS")
+        self.threshold = env_float("WATCHDOG_THRESHOLD")
+        self.evaluation_interval = env_float("WATCHDOG_EVALUATION_INTERVAL_SECONDS")
+        self.scrape_timeout = env_float("WATCHDOG_SCRAPE_TIMEOUT_SECONDS")
+        self.startup_grace = env_float("WATCHDOG_STARTUP_GRACE_SECONDS")
+        self.cooldown = env_float("WATCHDOG_COOLDOWN_SECONDS")
+        self.marker_file = os.environ.get("WATCHDOG_MARKER_FILE")
+        if self.evaluation_interval < self.poll_interval:
+            raise ConfigError(
+                "WATCHDOG_EVALUATION_INTERVAL_SECONDS must be at least "
```
martinflorian-da, line 152:
```suggestion
                "WATCHDOG_EVALUATION_INTERVAL_SECONDS must be less than "
```

That's what the condition says at least?

cluster/images/cometbft-watchdog/restart-watchdog.py
```diff
@@ -0,0 +1,296 @@
...
+    # don't immediately kill cometbft after startup
+    quiet_until = time.monotonic() + config.startup_grace
+    while True:
+        cycle_start = time.monotonic()
+        try:
+            observation = watchdog.poll(cycle_start)
+        except ScrapeError as error:
+            # keep cometbft running if we fail to scrape.
+            log("WARNING", f"Skipping evaluation: {error}")
+            watchdog.reset()
```
martinflorian-da, line 285: So we assume here that "metric fail" means "all is well again for this cycle"?

Possibly this is a good call, just calling out.

cluster/images/cometbft-watchdog/restart-watchdog.py
```diff
@@ -0,0 +1,296 @@
...
+
+def scrape(url, metric, required_labels, timeout):
+    try:
+        request = urllib.request.Request(
+            url, headers={"Accept": "text/plain"}
+        )
+        with urllib.request.urlopen(request, timeout=timeout) as response:
+            charset = response.headers.get_content_charset() or "utf-8"
+            body = response.read().decode(charset, errors="replace")
+    except (urllib.error.URLError, OSError, ValueError) as error:
```
martinflorian-da: Any reason to not just catch all errors here? I don't think we want this sidecar to ever die if we can avoid it?

...and according to an LLM this can also throw `http.client.HTTPException` (not worth my time to verify this claim, but also seems low-risk enough to just add it, or all of `Exception`).

cluster/images/cometbft-watchdog/restart-watchdog.py
```diff
@@ -0,0 +1,296 @@
...
+
+        previous = self.previous
+        self.previous = (now, sequencer, mediator)
+        if previous is None:
+            return None
+        previous_time, previous_sequencer, previous_mediator = previous
+        elapsed = now - previous_time
+        if elapsed <= 0:
+            return None
+        sequencer_rate = counter_increase(previous_sequencer, sequencer) / elapsed
```
martinflorian-da, line 213: Remind me - what happens on a restart? Do these counters re-start from 0? If so we might have a logic bug here that may or may not matter.

martinflorian-da (approved):
Thanks!

WDYT about doing a `/cluster_test` to? Just to avoid breaking CI (not sure if you also deployed the sv runbook for example?)

cluster/images/cometbft-watchdog/Dockerfile
moritzkiefer-da: > Hmm that adds a new external dependency that might start triggering vulnerability scanners at some point... (or worse, obviously).

I briefly htought about it. My reasoning why it's ok is that the vulnerability scanner issues we had are for validators not SVs. Also this is a temporary thing™. So this just seemed easier :shrug:

cluster/images/cometbft-watchdog/restart-watchdog.py
moritzkiefer-da: you already get randomization through when you start polling don't you? but fair enough, I can see if it's easy enough to add something.

cluster/images/cometbft-watchdog/restart-watchdog.py
moritzkiefer-da: > Possibly this is a good call, just calling out.

yeah I'd rather err on the side of not restarting for now.

cluster/images/cometbft-watchdog/restart-watchdog.py
moritzkiefer-da: > Any reason to not just catch all errors here? I don't think we want this sidecar to ever die if we can avoid it?

fair point will change

cluster/images/cometbft-watchdog/restart-watchdog.py
moritzkiefer-da: > Remind me - what happens on a restart? Do these counters re-start from 0? If so we might have a logic bug here that may or may not matter.

counters restart from zero but I'm not sure I see the logic bug? We also reset the data when we trigger a restart.

cluster/helm/splice-cometbft/values-template.yaml
moritzkiefer-da: 600 seems quite long to me so I'd lean towards starting with that and reviisting if needed. Cometbft is also quick to restart.

docs/src/release_notes_upcoming.rst
moritzkiefer-da: > We should probably also extend apps/app/src/pack/examples/sv-helm/cometbft-values.yaml 

will do

> update the public docs to reflect that?

not sure I care enough, we're not gonna get new SVs on cometbft

cluster/images/cometbft-watchdog/restart-watchdog.py
martinflorian-da: I mean if we restart only the mediator, for example, then this side car lives and so does the sequencer counter, but the mediator counter_increase one might be off?

cluster/images/cometbft-watchdog/restart-watchdog.py
moritzkiefer-da: I think I fucked up the condition, fixing that instead.

moritzkiefer-da: wait no it's correct, if it's less we error out. So the error tells you it must not be less.

cluster/images/cometbft-watchdog/restart-watchdog.py
moritzkiefer-da: ah right, let me add a safe guard and just reset it when we see counters decreasing.

cluster/images/cometbft-watchdog/restart-watchdog.py
martinflorian-da: > when you start polling don't you

Hm fair point, but that might not be 100% guaranteed, for example if for some reason we have a synchronized restart some time.

If we agree that randomizing is a good idea it seems better to make that explicit and use a controlled source of random?

cluster/images/cometbft-watchdog/Dockerfile
martinflorian-da: fair, sold

cluster/images/cometbft-watchdog/restart-watchdog.py
martinflorian-da: Ah of course :facepalm: boolean logic hard, sorry

docs/src/release_notes_upcoming.rst
martinflorian-da: to be clear, my point was that you need to manually update the snippet on https://github.com/canton-network/cf-docs/ in order for the example values change to even show up there

cluster/images/cometbft-watchdog/restart-watchdog.py
moritzkiefer-da: done

docs/src/release_notes_upcoming.rst
moritzkiefer-da: added the file, will update in the external repo if I can convince myself to deal with that

cluster/images/cometbft-watchdog/restart-watchdog.py
moritzkiefer-da: added randomization

PR #6830, opened 2026-08-18
Double scan rate limits for `/registry/transfer-instruction/v1/transfer-factory`

martinflorian-da:
Based on experience on MainNet, see DACH-NY/canton-network-internal#6543

This endpoint is cheap, some IP(s) are getting rate limited at 3-4 requests / second.

[static]

### DIFF
--- a/cluster/configs/shared/rate-limits/token-registry.yaml
+++ b/cluster/configs/shared/rate-limits/token-registry.yaml
@@ -52,12 +52,12 @@ rateLimits:
   /registry/transfer-instruction/v1/transfer-factory:
     name: registry-transfer-factory
     type: limited
-    maxTokens: 730
-    tokensPerFill: 730
+    maxTokens: 1440
+    tokensPerFill: 1440
     fillInterval: 60s
     perIpLimits:
-      maxTokens: 120
-      tokensPerFill: 120
+      maxTokens: 240
+      tokensPerFill: 240
       fillInterval: 60s
   /registry/allocation/v2/settlement-factory:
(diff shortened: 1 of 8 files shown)


cluster/configs/shared/rate-limits/token-registry.yaml
```diff
@@ -52,12 +52,12 @@ rateLimits:
   /registry/transfer-instruction/v1/transfer-factory:
     name: registry-transfer-factory
     type: limited
-    maxTokens: 730
```
martinflorian-da, line 55: @pasindutennage-da I'm going to assume that you actually meant to put 720 here, not 730.

cluster/configs/shared/rate-limits/token-registry.yaml
```diff
@@ -52,12 +52,12 @@ rateLimits:
...
-    tokensPerFill: 730
+    maxTokens: 1440
```
moritzkiefer-da, line 55: I assume it's intentional that we bump global and per ip? haven't followed it all that closely

moritzkiefer-da (approved):
thx

cluster/configs/shared/rate-limits/token-registry.yaml
martinflorian-da: IMO either we 2x all endpoints, or we declare that this endpoint is less concerning than others and 2x just it... but then we can 2x both per IP and global IMO, don't see why we should decouple that.

Notably this all feels like shooting in the dark to me :v:

PR #6834, opened 2026-08-18
Add ACS snapshot table indexes

OriolMunoz-da:
Part of https://github.com/canton-network/splice/pull/6742

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotIndexTrigger.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotIndexTrigger.scala
@@ -1,0 +1,68 @@
+package org.lfdecentralizedtrust.splice.scan.automation
+
+import com.digitalasset.canton.data.CantonTimestamp
+import com.digitalasset.canton.lifecycle.FutureUnlessShutdown
+import com.digitalasset.canton.resource.DbStorage
+import com.digitalasset.canton.tracing.TraceContext
+import io.opentelemetry.api.trace.Tracer
+import org.apache.pekko.stream.Materializer
+import org.lfdecentralizedtrust.splice.automation.{SqlIndexInitializationTrigger, TriggerContext}
+import org.lfdecentralizedtrust.splice.scan.store.AcsSnapshotStore
+import org.lfdecentralizedtrust.splice.scan.store.AcsSnapshotStore.AcsTableDDL
+import com.digitalasset.canton.lifecycle.FutureUnlessShutdownImpl.*
+
+import java.time.Instant
+import scala.concurrent.{ExecutionContextExecutor, Future}
+import scala.util.Try
+
+class AcsSnapshotIndexTrigger(storage: DbStorage, store: AcsSnapshotStore, context: TriggerContext)(
+    implicit
+    ec: ExecutionContextExecutor,
+    override val tracer: Tracer,
+    mat: Materializer,
+) extends SqlIndexInitializationTrigger(storage, context) {
+
+  override protected def retrieveNextIndexTasks()(implicit
+      tc: TraceContext
+  ): FutureUnlessShutdown[Seq[SqlIndexInitializationTrigger.IndexAction]] = {
+    store
+      .lookupOldestUnindexedSnapshot()
+      .map { snapshot =>
+        // Statements are safe to retry because of `if not exists`
+        Seq(
+          SqlIndexInitializationTrigger.IndexAction.Create(
+            AcsTableDDL
+              .stakeholderIndexName(snapshot.historyId, snapshot.snapshotRecordTime),
+            AcsTableDDL.stakeholderIndexAction(snapshot.historyId, snapshot.snapshotRecordTime),
+          ),
+          SqlIndexInitializationTrigger.IndexAction.Create(
+            AcsTableDDL
+              .stakeholderTemplateIdIndexName(snapshot.historyId, snapshot.snapshotRecordTime),
+            AcsTableDDL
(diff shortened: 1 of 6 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotIndexTrigger.scala
```diff
@@ -0,0 +1,55 @@
...
+    ec: ExecutionContextExecutor,
+    override val tracer: Tracer,
+    mat: Materializer,
+) extends SourceBasedTrigger[PerTableAcsSnapshot] {
+
+  /** The source from which to consume tasks. */
+  override protected def source(implicit
+      traceContext: TraceContext
+  ): Source[PerTableAcsSnapshot, NotUsed] = Source
+    .repeat(())
```
rautenrieth-da, line 35: Why is this not a PollingTrigger? Isn't `Source.repeat(())` an infinite busy loop that is repeating the `lookupOldestUnindexedSnapshot` query as fast as the database can handle?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/QueryAcsSnapshotPaginationTokenTest.scala
```diff
@@ -28,6 +32,26 @@ class QueryAcsSnapshotPaginationTokenTest extends AnyWordSpec with BaseTest {
     }
   }
 
+  "CreatedAtContractIdAcsSnapshotPaginationToken" should {
```
rautenrieth-da, line 35: This change makes sense, but is unrelated to table indexes

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/AcsSnapshotStoreTest.scala
```diff
@@ -1548,4 +1548,27 @@ class TablePerAcsSnapshotStoreTest extends AcsSnapshotStoreTest {
...
+      _ <- store.insertNewSnapshot(nextTable, DefaultMigrationId, timestamp1)
+      snapshotOpt <- store.lookupSnapshotAtOrBefore(
+        migrationId = DefaultMigrationId,
+        CantonTimestamp.MaxValue,
+      )
+      snapshot = snapshotOpt.valueOrFail("snapshot should've just been created")
+      oldestUnindexedOpt <- store.lookupOldestUnindexedSnapshot()
+      oldestUnindexed = oldestUnindexedOpt.valueOrFail("snapshot should be unindexed")
+      _ = oldestUnindexed should be(snapshot)
+      _ <- store.indexSnapshotStakeholdersTable(oldestUnindexed)
```
rautenrieth-da, line 1570: Call this twice to make sure it's idempotent? `mkStore` uses `storage.underlying`, so the default idempotency enforcement is disabled

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V075__acs_snapshot_table_name.sql
```diff
@@ -16,6 +18,8 @@ alter table acs_snapshot
             -- legacy table
          (first_row_id is not null and last_row_id is not null and creates_table_name is null and stakeholders_table_name is null));
 
+create index acs_snapshot_unindexed on acs_snapshot (history_id, snapshot_record_time) where not indexes_created;
```
rautenrieth-da, line 21: I like this trick, we don't need a separate table for tracking work to do

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/AcsSnapshotIndexTrigger.scala
OriolMunoz-da: > Why is this not a PollingTrigger?

just because of refactors, will change it to one

either way:

> ? Isn't Source.repeat(()) an infinite busy loop that is repeating the lookupOldestUnindexedSnapshot query as fast as the database can handle?

no: downstream should apply backpressure when processing each task in the mapAsync(1) step

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/QueryAcsSnapshotPaginationTokenTest.scala
OriolMunoz-da: ugh, too many branches, mb :facepalm: 
keeping it here since it will be merged anyway, but sorry for the noise...

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -851,6 +880,42 @@ class AcsSnapshotStore(
...
 
+  def indexSnapshotStakeholdersTable(
+      snapshot: PerTableAcsSnapshot
+  )(implicit tc: TraceContext): Future[Unit] = {
+    // This doesn't have to be transactional:
+    // - if the `set indexes_created = true` executes, we know everything succeeded.
+    // - otherwise, on retry the `create index IF NOT EXISTS` will just move on.
+    for {
+      _ <- storage.update(
+        AcsSnapshotDDL.stakeholderIndexAction(
```
rautenrieth-da: Do we need a DDL advisory lock, to address cases where (in tests) multiple apps try to create indexes at the same time, and thus update the same catalog table?

rautenrieth-da (approved):
Thanks!

PR #6868, opened 2026-08-19
Introduce optional use of ComputeClasses

rautenrieth-da:
This PR:

- Add a new config value `kubernetesScheduling.computeClasses.enabled`.
- If the value is true, create two ComputeClasses, and change pod annotations from using the `cn_apps` key to using `loud.google.com/compute-class`. This _should_ trigger the autoscaler to use the ComputeClass definition for choosing which node pool to scale up.

### DIFF
--- a/cluster/pulumi/cluster/src/nodePools.ts
+++ b/cluster/pulumi/cluster/src/nodePools.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import * as gcp from '@pulumi/gcp';
 import * as k8s from '@pulumi/kubernetes';
+import * as pulumi from '@pulumi/pulumi';
 import {
   config,
   GCP_PROJECT,
@@ -69,11 +70,16 @@ type NodeConfigLabelsAndTaints = Pick<
   'labels' | 'taints'
 >;
 
+interface NodePoolWithConfig {
+  pool: gcp.container.NodePool;
+  config: GkeNodePoolConfig;
+}
+
 function installAppsNodePools(
   cluster: string,
   allZones: string[],
   configs: Array<GkeNodePoolConfig>
-): Array<gcp.container.NodePool> {
+): Array<NodePoolWithConfig> {
   const defaultZone = config.optionalEnv('CLOUDSDK_HYPERDISK_NODEPOOL_COMPUTE_ZONE');
   return configs.map((config, index) => {
     const name =
@@ -110,7 +116,7 @@ function installAppsNodePools(
             ...config.labels,
           },
         };
-    return new gcp.container.NodePool(
+    const pool = new gcp.container.NodePool(
       name,
       {
         cluster,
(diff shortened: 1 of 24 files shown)


nicu-da (approved):
looks reasonable, thanks!

cluster/pulumi/cluster/src/nodePools.ts
```diff
@@ -140,9 +158,33 @@ function installInfraNodePools(
         replaceOnChanges: ['nodeConfig.machineType'],
       }
     );
+    return { name, pool };
   });
 }
 
+function installComputeClass(
+  name: string,
+  pools: NamedNodePool[]
```
mblaze-da: Why do we need `NamedNodePool` here instead of referencing `gcp.container.NodePool::name`? The name currently used here lacks the random postfix, doesn't it?

cluster/pulumi/cluster/src/nodePools.ts
```diff
@@ -140,9 +158,33 @@ function installInfraNodePools(
...
+  return new k8s.apiextensions.CustomResource(
+    `compute-class-${name}`,
+    {
+      apiVersion: 'cloud.google.com/v1',
+      kind: 'ComputeClass',
+      metadata: { name },
+      spec: {
+        // Reference the existing, manually-managed node pools rather than
+        // letting GKE auto-provision. Priorities are evaluated in order.
+        priorities: pools.map(p => ({ nodepools: [p.name] })),
```
mblaze-da: IMO, for the sake of legibility of config we should configure priorities explicitely and pull those values from the config. Currently, looking at the config it is not at all apparent that the order matters.

mblaze-da (approved):
thanks!

cluster/pulumi/cluster/src/nodePools.ts
rautenrieth-da: I was trying to preserve backwards compatibility of the config schema, do you have a suggestion how to go from
```
cluster:
  nodePools:
    apps:
      nodeType: "c4d-standard-32"
      minNodes: 0
      maxNodes: 40
      zones: '*'
    additionalApps:
    - nodeType: c4-standard-32
      minNodes: 0
      maxNodes: 40
      zones: '*'
    - nodeType: "c3d-standard-30-lssd"
      minNodes: 0
      maxNodes: 40
      zones: '*'
```
to something that makes the priorities more explicit? Just add a `priority` field to each node pool config?

mblaze-da: Yes, I'd do so and give it some default to make it backwards compatible

rautenrieth-da: The PR was already setting up one priority per node pool, i.e., `apps` first, then `additionalApps[0]`, then `additionalApps[1]`. Changing the priority would be a matter of reshuffling the entries in the cluster config.

I have added one more commit that adds an optional `priority` field to the node pools. Node pools are grouped by priority, then sorted descending by priority. I.e., it allows as to express that N4 and N4D pools have the same priority. The field defaults to `-<index of the node pool in the config>`, so leaving it out behaves sensibly.

PR #6874, opened 2026-08-20
Add PruneRewardAccountingTrigger to delete old reward-accounting data

### LINKED ISSUE #6136: Prune TBAR stores that hold intermediate date
That is the tables backing these two stores:
- https://github.com/canton-network/splice/blob/d5d1108aa222955b4fac192feb6881164911e9cb/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanAppRewardsStore.scala#L15
- https://github.com/canton-network/splice/blob/d5d1108aa222955b4fac192feb6881164911e9cb/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala#L31

The goal is that the performance of the reward computation is independent of the time it has been running. 

Note that the pruning of the `AppActivityStore` will be done separately, as part of the work be the Splice Maintainers to store the verdicts in a bulk format outside of the Scan DB.

dfordivam:
Fixes #6136

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
@@ -1,7 +1,7 @@
 package org.lfdecentralizedtrust.splice.integration.tests
 
 import com.digitalasset.canton.HasExecutionContext
-import com.digitalasset.canton.config.NonNegativeDuration
+import com.digitalasset.canton.config.{NonNegativeDuration, NonNegativeFiniteDuration}
 import com.digitalasset.canton.console.LocalInstanceReference
 import com.digitalasset.canton.data.CantonTimestamp
 import com.digitalasset.canton.lifecycle.CloseContext
@@ -52,6 +52,7 @@ import org.lfdecentralizedtrust.splice.util.{
   AmuletConfigSchedule,
   AmuletConfigUtil,
   ScanTestUtil,
+  SpliceUtil,
   TimeTestUtil,
   TriggerTestUtil,
   WalletTestUtil,
@@ -85,6 +86,10 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
   // event-history consistency check cannot hold here.
   override protected def runEventHistorySanityCheck: Boolean = false
 
+  // Long enough that advancing a few rounds does not reach it
+  private val rewardAccountingRetentionPeriod: NonNegativeFiniteDuration =
+    NonNegativeFiniteDuration(SpliceUtil.defaultInitialTickDuration.asJava.multipliedBy(10))
+
   override def environmentDefinition: SpliceEnvironmentDefinition =
     EnvironmentDefinition
       .simpleTopology4SvsWithSimTime(this.getClass.getSimpleName)
@@ -95,6 +100,11 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
             dryRunVersion = None,
             appRewardCouponThreshold = BigDecimal("0"),
           )
+        )(config)
+      )
+      .addConfigTransform((_, config) =>
+        ConfigTransforms.updateAllScanAppConfigs_(
+          _.copy(rewardAccountingRetentionPeriod = rewardAccountingRetentionPeriod)
         )(config)
       )
       // Prevent wallets from minting RewardCouponV2 before the test
@@ -839,6 +849,28 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
       }
     }
 
+    def confirmPruningMetrics(
+        scan: LocalInstanceReference,
+        atLeastRound: Long,
+        timeUntilSuccess: FiniteDuration = 20.seconds,
+    ): Unit = {
+      def pruningMetricValue(name: String, labels: Map[String, String] = Map.empty): Long =
+        metricValue(scan, s"scan.reward_accounting_pruning.$name", labels)
+
+      eventually(timeUntilSuccess) {
+        pruningMetricValue("pruned_round") should be >= atLeastRound
+        forAll(
+          Seq(
+            "scan_rewards_reference_store_archived",
+            "app_activity_round_totals",
+            "app_reward_round_totals",
+          )
+        ) { table =>
+          pruningMetricValue("deleted_rows", Map("table" -> table)) should be > 0L
+        }
+      }
+    }
+
     // Simulate SV2 scan ingestion lag and confirm that pruning does not happen
     // until the ingestion has caught up.
     val newLowestOpen = pauseScanVerdictIngestionWithin(sv2ScanBackend) {
@@ -850,8 +882,16 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
       // So advancing by 3 rounds is a safe way to achieve this.
       (1 to 3).foreach(_ => advanceRoundsToNextRoundOpening)
 
-      clue(s"sv1 prunes rounds below $newLowestOpen") {
+      clue(s"sv1 retains rounds below $newLowestOpen while within the retention period") {
+        hasUnprunedRewardAccountingDataBelow(sv1Db, sv1HistoryId, newLowestOpen) shouldBe true
+        hasUnprunedArchiveDataForRound(sv1RewardsRefStore, newLowestOpen - 1) shouldBe true
+      }
+
+      advanceTime(rewardAccountingRetentionPeriod.asJava)
+
+      clue(s"sv1 prunes rounds below $newLowestOpen once past the retention period") {
         confirmFullyPruned(sv1Db, sv1HistoryId, sv1RewardsRefStore, newLowestOpen)
+        confirmPruningMetrics(sv1ScanBackend, newLowestOpen - 1)
       }
 
       clue(
@@ -872,6 +912,7 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
         newLowestOpen,
         timeUntilSuccess = 90.seconds,
       )
+      confirmPruningMetrics(sv2ScanBackend, newLowestOpen - 1)
     }
   }
 
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
@@ -58,7 +58,7 @@ import org.lfdecentralizedtrust.splice.util.{
 }
 import org.slf4j.event.Level
 
-import scala.concurrent.duration.DurationInt
+import scala.concurrent.duration.{DurationInt, FiniteDuration}
 import slick.jdbc.canton.ActionBasedSQLInterpolation.Implicits.actionBasedSQLInterpolationCanton
 
 // This test focuses on the SV app side triggers testing
@@ -829,11 +829,12 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
         historyId: Long,
         store: ScanRewardsReferenceStore,
         upperExclusive: Long,
+        timeUntilSuccess: FiniteDuration = 20.seconds,
     ): Unit = {
-      eventually() {
+      eventually(timeUntilSuccess) {
         hasUnprunedRewardAccountingDataBelow(db, historyId, upperExclusive) shouldBe false
       }
-      eventually() {
+      eventually(timeUntilSuccess) {
         hasUnprunedArchiveDataForRound(store, upperExclusive - 1) shouldBe false
       }
     }
@@ -843,6 +844,11 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
     val newLowestOpen = pauseScanVerdictIngestionWithin(sv2ScanBackend) {
       advanceRoundsToNextRoundOpening
       val newLowestOpen = oldestOpenRound
+
+      // We need the archived_at of newLowestOpen + 1 to be lower than the
+      // the active open round's openAt.
+      // So advancing by 3 rounds is a safe way to achieve this.
+      (1 to 3).foreach(_ => advanceRoundsToNextRoundOpening)
 
       clue(s"sv1 prunes rounds below $newLowestOpen") {
         confirmFullyPruned(sv1Db, sv1HistoryId, sv1RewardsRefStore, newLowestOpen)
@@ -858,7 +864,14 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
     }
 
     clue(s"sv2 eventually prunes data once verdict ingestion resumes") {
-      confirmFullyPruned(sv2Db, sv2HistoryId, sv2RewardsRefStore, newLowestOpen)
+      // This can occasionally take longer than the default 20s eventually window
+      confirmFullyPruned(
+        sv2Db,
+        sv2HistoryId,
+        sv2RewardsRefStore,
+        newLowestOpen,
+        timeUntilSuccess = 90.seconds,
+      )
     }
   }
 
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
@@ -46,6 +46,7 @@ import org.lfdecentralizedtrust.splice.sv.automation.delegatebased.{
 }
 import org.lfdecentralizedtrust.splice.scan.admin.api.client.BftScanConnection
 import org.lfdecentralizedtrust.splice.scan.automation.RewardComputationTrigger
+import org.lfdecentralizedtrust.splice.scan.store.ScanRewardsReferenceStore
 import org.lfdecentralizedtrust.splice.sv.config.InitialRewardConfig
 import org.lfdecentralizedtrust.splice.util.{
   AmuletConfigSchedule,
@@ -67,6 +68,8 @@ import slick.jdbc.canton.ActionBasedSQLInterpolation.Implicits.actionBasedSQLInt
 // - BFT read in all three SV app's reward processing triggers
 //
 // - Reporting of mismatches in 'Confirmation' of root-hash
+//
+// - Pruning of reward processing data
 @org.lfdecentralizedtrust.splice.util.scalatesttags.SpliceAmulet_0_1_19
 class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
     extends IntegrationTestWithIsolatedEnvironment
@@ -354,6 +357,8 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
       confirmBftRead(bobParty)
 
       confirmMismatchingRootHashIsFlagged(bobParty)
+
+      advanceRoundsAndConfirmAllPriorRoundsPruned()
   }
 
   // sv2's CalculateRewardsTrigger and SummarizingMiningRoundTrigger report the
@@ -766,6 +771,95 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
         }
       },
     )
+  }
+
+  private def advanceRoundsAndConfirmAllPriorRoundsPruned()(implicit
+      env: SpliceTestConsoleEnvironment
+  ): Unit = {
+    val sv1Db = sv1ScanBackend.appState.storage match {
+      case db: DbStorage => db
+      case _ => fail("Expected DbStorage")
+    }
+    val sv2Db = sv2ScanBackend.appState.storage match {
+      case db: DbStorage => db
+      case _ => fail("Expected DbStorage")
+    }
+    val sv1HistoryId = sv1ScanBackend.appState.eventStore.updateHistory.historyId
+    val sv2HistoryId = sv2ScanBackend.appState.eventStore.updateHistory.historyId
+    val sv1RewardsRefStore = sv1ScanBackend.appState.rewardsReferenceStore
+    val sv2RewardsRefStore = sv2ScanBackend.appState.rewardsReferenceStore
+
+    def hasUnprunedRewardAccountingDataBelow(
+        db: DbStorage,
+        historyId: Long,
+        upperExclusive: Long,
+    ): Boolean = {
+      implicit val closeContext: CloseContext = CloseContext(db)
+      db
+        .querySingle(
+          sql"""select
+                  exists(select 1 from app_activity_party_totals
+                         where history_id = $historyId and round_number < $upperExclusive)
+                  or exists(select 1 from app_activity_round_totals
+                         where history_id = $historyId and round_number < $upperExclusive)
+                  or exists(select 1 from app_reward_party_totals
+                         where history_id = $historyId and round_number < $upperExclusive)
+                  or exists(select 1 from app_reward_round_totals
+                         where history_id = $historyId and round_number < $upperExclusive)
+                  or exists(select 1 from app_reward_batch_hashes
+                         where history_id = $historyId and round_number < $upperExclusive)
+                  or exists(select 1 from app_reward_root_hashes
+                         where history_id = $historyId and round_number < $upperExclusive)
+             """.as[Boolean].headOption,
+          "test.hasUnprunedRewardAccountingDataBelow",
+        )
+        .value
+        .futureValueUS
+        .value
+    }
+
+    def hasUnprunedArchiveDataForRound(
+        store: ScanRewardsReferenceStore,
+        roundNumber: Long,
+    ): Boolean =
+      store.lookupArchivedAtForOpenMiningRound(roundNumber).futureValue.isDefined
+
+    def confirmFullyPruned(
+        db: DbStorage,
+        historyId: Long,
+        store: ScanRewardsReferenceStore,
+        upperExclusive: Long,
+    ): Unit = {
+      eventually() {
+        hasUnprunedRewardAccountingDataBelow(db, historyId, upperExclusive) shouldBe false
+      }
+      eventually() {
+        hasUnprunedArchiveDataForRound(store, upperExclusive - 1) shouldBe false
+      }
+    }
+
+    // Simulate SV2 scan ingestion lag and confirm that pruning does not happen
+    // until the ingestion has caught up.
+    val newLowestOpen = pauseScanVerdictIngestionWithin(sv2ScanBackend) {
+      advanceRoundsToNextRoundOpening
+      val newLowestOpen = oldestOpenRound
+
+      clue(s"sv1 prunes rounds below $newLowestOpen") {
+        confirmFullyPruned(sv1Db, sv1HistoryId, sv1RewardsRefStore, newLowestOpen)
+      }
+
+      clue(
+        s"sv2 does not prune rounds below $newLowestOpen while its verdict ingestion is paused"
+      ) {
+        hasUnprunedArchiveDataForRound(sv2RewardsRefStore, newLowestOpen - 1) shouldBe true
+      }
+
+      newLowestOpen
+    }
+
+    clue(s"sv2 eventually prunes data once verdict ingestion resumes") {
+      confirmFullyPruned(sv2Db, sv2HistoryId, sv2RewardsRefStore, newLowestOpen)
+    }
   }
 
   private def doTransfer(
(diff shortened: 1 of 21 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
```diff
@@ -685,6 +690,95 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
     )
   }
 
+  private def advanceRoundsAndConfirmAllPriorRoundsPruned()(implicit
```
dfordivam, line 790: This test is flaky at the moment, so I am investigating to make it more robust

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanAppRewardsStore.scala
```diff
@@ -1020,6 +1030,49 @@ class DbScanAppRewardsStore(
             )
     """.asUpdate
 
+  /** Deletes all reward-accounting data for a single round across all six
+    * tables, in a single transaction.
+    */
+  def deleteRewardAccountingDataForRound(
```
dfordivam, line 1038: Indexes being used here for reference

```
    "app_reward_batch_hashes_pkey" PRIMARY KEY, btree (history_id, round_number, batch_level, party_seq_num_begin_incl)
    "app_reward_root_hashes_pkey" PRIMARY KEY, btree (history_id, round_number)
    "app_reward_party_totals_pkey" PRIMARY KEY, btree (history_id, round_number, app_provider_party_seq_num)
    "app_reward_round_totals_pkey" PRIMARY KEY, btree (history_id, round_number)
    "app_activity_party_totals_pkey" PRIMARY KEY, btree (history_id, round_number, app_provider_party)
    "app_activity_round_totals_pkey" PRIMARY KEY, btree (history_id, round_number)
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
```diff
@@ -270,6 +270,85 @@ class DbScanRewardsReferenceStore(
...
+                      and active.round <= archived.round
+                  )
+           """.as[Option[Long]].head,
+          "lookupLowestPrunableArchivedRewardRound",
+        )
+      )
+    }
+
+  // Returns number of rows deleted
+  def pruneArchivedDataForRound(
```
dfordivam: This will always be a seq scan

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
```diff
@@ -0,0 +1,118 @@
...
+       *
+       * Also here it is assumed that ScanAppRewardsStore only has the data for
+       * rounds for which we have data in ScanRewardsReferenceStore. Which is a
+       * fair assumption as the reward calculations require the data to be
+       * present in ScanRewardsReferenceStore for that round.
+       */
+      summary <- appRewardsStore.deleteRewardAccountingDataForRound(task.roundNumber)
+      deletedArchiveRows <- rewardsReferenceStore.pruneArchivedDataForRound(task.roundNumber)
+    } yield TaskSuccess(
+      s"Pruned reward accounting data for round ${task.roundNumber}: " +
```
dfordivam: Example log

    "Completed processing with outcome: Pruned reward accounting data for round 11: removed 5 archived rows; and RewardAccountingPruneSummary(2,1,2,1,1,1)."

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
```diff
@@ -0,0 +1,118 @@
...
+              logger.warn(
+                s"This should never happen, could not resolve archived_at for round $roundNumber."
+              )
+              Seq.empty
+            case Some(uptoInclusive) =>
+              if (verdictStore.lastIngestedRecordTime.exists(_ >= uptoInclusive)) {
+                Seq(Task(roundNumber))
+              } else {
+                logger.debug(
+                  s"Skipping pruning as the verdict ingestion (last ingested record time: ${verdictStore.lastIngestedRecordTime}) " +
```
dfordivam: Example log

    "Skipping pruning as the verdict ingestion (last ingested record time: Some(1970-01-01T02:02:00.000131Z)) has not yet caught up to round 11's archivedAt (1970-01-01T02:02:00.000222Z)"

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
```diff
@@ -0,0 +1,158 @@
...
+                    } yield (activeRoundO, archivedAtNextO) match {
+                      case (Some(activeRound), Some(archivedAtNext)) =>
+                        if (
+                          CantonTimestamp
+                            .assertFromInstant(activeRound.payload.opensAt) > archivedAtNext
+                        ) {
+                          Seq(Task(roundNumber))
+                        } else {
+                          logger.debug(
+                            s"Skipping pruning of round $roundNumber as the ingestion's currently active round " +
```
dfordivam: This is the log, though this may be noisy and perhaps can be removed

   "Skipping pruning of round 12 as the ingestion's currently active round 15 opened at 1970-01-01T17:47:20Z, which is not yet past round 13's archivedAt (1970-01-01T17:47:30.000228Z)."

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
dfordivam: Fixed the flakyness, needed more time after `pauseScanVerdictIngestionWithin` block

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
```diff
@@ -270,6 +270,85 @@ class DbScanRewardsReferenceStore(
       case _ => Future.successful(None)
     }
 
+  def lookupArchivedAtForOpenMiningRound(
```
adetokunbo: nit: use override here, to be consistent with `lookupLowestPrunableArchiveRewardRound` below
Also for `pruneArchiveDataForRound`

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
```diff
@@ -685,6 +690,108 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
...
+    }
+
+    clue(s"sv2 eventually prunes data once verdict ingestion resumes") {
+      // This can occasionally take longer than the default 20s eventually window
+      confirmFullyPruned(
+        sv2Db,
+        sv2HistoryId,
+        sv2RewardsRefStore,
+        newLowestOpen,
+        timeUntilSuccess = 90.seconds,
```
adetokunbo: The comment above suggests that it can complete in 20s, but is sometimes much longer; is it possible to also document why there is such a large variance as a means of justifying the 90s ?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
```diff
@@ -0,0 +1,158 @@
...
+import org.lfdecentralizedtrust.splice.scan.store.db.DbScanVerdictStore
+import org.lfdecentralizedtrust.splice.store.UpdateHistory
+import com.digitalasset.canton.data.CantonTimestamp
+import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
+import com.digitalasset.canton.tracing.TraceContext
+import io.opentelemetry.api.trace.Tracer
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/** Periodically deletes reward-accounting data that is no longer needed, one
```
adetokunbo, line 27: Design question:  it's not mentioned in #6136, so asking here: is it confirmed that no retention window is necessary?  The data in these tables is useful for debugging.  IUC, this prunes them aggressively - should there be a configurable capability to retain them for a bit longer?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
```diff
@@ -685,6 +690,108 @@ class TrafficBasedRewardsSvAppTimeBasedIntegrationTest
...
+      (1 to 3).foreach(_ => advanceRoundsToNextRoundOpening)
+
+      clue(s"sv1 prunes rounds below $newLowestOpen") {
+        confirmFullyPruned(sv1Db, sv1HistoryId, sv1RewardsRefStore, newLowestOpen)
+      }
+
+      clue(
+        s"sv2 does not prune rounds below $newLowestOpen while its verdict ingestion is paused"
+      ) {
+        hasUnprunedArchiveDataForRound(sv2RewardsRefStore, newLowestOpen - 1) shouldBe true
```
rautenrieth-da: Should we explicitly trigger one iteration of `PruneRewardAccountingTrigger` here? This might mean that we have to pause the trigger for the entire duration of the test and control them manually. Although unlikely, it could be that sv2 has unpruned data simply because the trigger didn't finish yet.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
rautenrieth-da: Also interested in knowing why it takes so long. Is it verdict ingestion taking so long to catch up, or the pruning trigger?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
```diff
@@ -0,0 +1,158 @@
...
+          verdictStore.lastIngestedRecordTime match {
+            case None =>
+              logger.debug("Skipping pruning as no verdict has been ingested yet.")
+              Future.successful(Seq.empty)
+            case Some(lastIngestedRecordTime) =>
+              rewardsReferenceStore
+                .lookupActiveOpenMiningRounds(Seq(lastIngestedRecordTime))
+                .flatMap(_.get(lastIngestedRecordTime) match {
+                  case None =>
+                    logger.warn(
```
rautenrieth-da: Tests are currently hitting this line. Could it be that the verdict ingestion is ahead of the reference store ingestion so that `lookupActiveOpenMiningRounds(lastIngestedRecordTime)` doesn't have an answer yet?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
```diff
@@ -94,6 +94,30 @@ trait ScanRewardsReferenceStore extends AppStore {
       tc: TraceContext
   ): Future[Seq[Contract[CalculateRewardsV2.ContractId, CalculateRewardsV2]]]
 
+  def lookupArchivedAtForOpenMiningRound(
+      roundNumber: Long
+  )(implicit tc: TraceContext): Future[Option[CantonTimestamp]]
+
+  /** The lowest round number that may be safe to prune
+    * The criterion is based on the prescence of `OpenMiningRound`,
```
rautenrieth-da:
```suggestion
    * The criterion is based on the presence of `OpenMiningRound`,
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
```diff
@@ -0,0 +1,158 @@
...
+  *
+  *   - Archived contracts from `ScanRewardsReferenceStore` archive table
+  *     which archived <= the archival time of the OpenMiningRound
+  *
+  * A round only becomes eligible for pruning once none of its `OpenMiningRound`,
+  * `CalculateRewardsV2`, or `ProcessRewardsV2` contracts -- nor those of any
+  * lower round -- remain active. AND the verdict ingestion has moved past the
+  * `record_time` of OpenMiningRound's archival.
+  */
+class PruneRewardAccountingTrigger(
```
rautenrieth-da, line 47: Can you please add some basic metrics to this trigger? Maybe a `Gauge` with the latest round eligible for pruning, or some meters for the values in `RewardAccountingPruneSummary`?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
```diff
@@ -270,6 +270,85 @@ class DbScanRewardsReferenceStore(
...
+      // even after the pruning.
+      val pqn = PackageQualifiedName.fromJavaCodegenCompanion(OpenMiningRound.COMPANION)
+      futureUnlessShutdownToFuture(
+        storage.query(
+          sql"""select min(archived.round)
+                from #${ScanRewardsReferenceTables.archiveTableName} archived
+                where archived.store_id = $storeId and archived.migration_id = $migrationId
+                  and archived.package_name = ${pqn.packageName}
+                  and archived.template_id_qualified_name = ${pqn.qualifiedName}
+                  and archived.round is not null
```
rautenrieth-da: Can the round ever be null for OpenMingRound contracts?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
```diff
@@ -270,6 +270,85 @@ class DbScanRewardsReferenceStore(
...
+      val storeId = multiDomainAcsStore.acsStoreId
+      val migrationId = multiDomainAcsStore.domainMigrationId
+      // It is important to limit the query for the round to OpenMiningRound
+      // as we prune only the contracts <= OpenMiningRound's archived_at
+      // and the db will likely still have other contracts for this round
+      // even after the pruning.
+      val pqn = PackageQualifiedName.fromJavaCodegenCompanion(OpenMiningRound.COMPANION)
+      futureUnlessShutdownToFuture(
+        storage.query(
+          sql"""select min(archived.round)
```
rautenrieth-da: This relies on the fact that archiving an OpenMiningRound also creates a CalculateRewardsV2 contract in the same transaction, right?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
```diff
@@ -0,0 +1,158 @@
...
+          }
+      }
+
+  override def completeTask(task: Task)(implicit tc: TraceContext): Future[TaskOutcome] =
+    for {
+      /* The two deletes happen on different stores, and the deletes are being
+       * done in independent transactions.
+       * Because both are idempotent operations and re-running either for the
+       * same or an older round is a no-op.
+       * If pruning of either store fails, the task will be retried.
```
rautenrieth-da, line 100: We also need to consider the case where the task fails in between the two pruning operations, and then all retries fail because of intermittent database connection issues.

In that case the task will be abandoned, and a new task eventually created. The logic for creating the new task needs to work even if exactly one of the two independent pruning transactions completed. It looks like this works because we first delete the reward accounting data, and the result of `retrieveTasks` only depends on the TCS store which is pruned second. I would mention this in the comment though.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
```diff
@@ -94,6 +94,30 @@ trait ScanRewardsReferenceStore extends AppStore {
...
+
+  /** Deletes all rows from the archive table with `archived_at <=` the record_time at
+    * which round's `OpenMiningRound` contract was archived.
+    *
+    * Here we don't have a lower bound on the `archived_at`, so this API should
+    * be used with care, preferably ensuring that the deletion happens one round at a time.
+    *
+    * Returns number of rows deleted.
+    */
+  def pruneArchivedDataForRound(
```
rautenrieth-da:
```suggestion
  def pruneArchivedDataBeforeRound(
```
or similar, "for" sounds like it would only delete data for that particular round

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
rautenrieth-da: Good point! I think a configurable retention window with a default of 1 week (or at least a day) would be very useful for debugging. In case of a high severity issue like a root hash mismatch we might want to inspect the state of the database, and it might take a while to get a data dump from all SVs.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
```diff
@@ -0,0 +1,158 @@
...
+    override val tracer: Tracer,
+) extends PollingParallelTaskExecutionTrigger[Task] {
+
+  override protected def context: TriggerContext = triggerContext.copy(
+    config = triggerContext.config.copy(
+      parallelism = 1
+    )
+  )
+
+  override def retrieveTasks()(implicit tc: TraceContext): Future[Seq[Task]] =
```
rautenrieth-da, line 72: The logic in this method is quite complex. Can we unit test it? If you want to avoid setting up the stores with real contracts, you could factor out the implementation into an independent function and then call it with test values, like we did with `AcsSnapshotTrigger.retrieveTaskForCurrentMigration`.

rautenrieth-da (commented):
Thanks! Wasn't easy to review, but looking good in general. Left some minor comments.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
dfordivam: It is verdict ingestion, the cause is some grpc stream issue where it errors out if not consumed, and it need retry by the service.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/db/DbScanRewardsReferenceStore.scala
dfordivam: not really, I was probably trying to make use of the index, but this can be removed as we have another one to cover this.

```
    "scan_rewards_reference_store_archived_round" btree (store_id, migration_id, round) WHERE round IS NOT NULL
    "scan_rewards_reference_store_archived_temporal" btree (store_id, migration_id, package_name, template_id_qualified_name, archived_at) INCLUDE (created_at)
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
dfordivam: yeah now that we even have retention window this did get complex, and even testing it in a unit test was tricky; mostly cos its hard to judge whether the testing is good enough and if the test makes sense.

And I did notice a bug with the in-memory var cache of TcsStore ingestion-start, which is also fixed now.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
dfordivam: renamed to `pruneArchivedUpToRound` as we actually delete the data which was for in store when the round was active.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
dfordivam: Done

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
dfordivam: yeah this was a mistake, changed to a debug. Its a bootstrap related issue, as `lookupActiveOpenMiningRounds` doesn't work for rounds 0, 1 on non initial SVs.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TrafficBasedRewardsSvAppTimeBasedIntegrationTest.scala
dfordivam: Done, now being run manually

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
dfordivam: Done, added `rewardAccountingRetentionPeriod` config with default of 1 week

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
dfordivam: ok, added metrics as described

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/metrics/RewardAccountingPruningMetrics.scala
```diff
@@ -0,0 +1,61 @@
...
+import org.lfdecentralizedtrust.splice.scan.store.db.DbScanAppRewardsStore.RewardAccountingPruneSummary
+
+class RewardAccountingPruningMetrics(metricsFactory: LabeledMetricsFactory)(implicit
+    metricsContext: MetricsContext
+) extends AutoCloseable {
+  private val prefix: MetricName =
+    SpliceMetrics.MetricsPrefix :+ "scan" :+ "reward_accounting_pruning"
+
+  // In normal operation this advances by one round per tick, so it stalling is
+  // the signal that pruning stopped making progress.
```
rautenrieth-da, line 19: Opened https://github.com/canton-network/splice/issues/7186 for implementing the alert.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/ScanRewardsReferenceStore.scala
```diff
@@ -99,6 +95,91 @@ trait ScanRewardsReferenceStore extends AppStore {
...
+        // archived_at of round roundNumber + 1.
+        for {
+          activeRoundO <- lookupActiveOpenMiningRounds(Seq(lastIngestedRecordTime))
+            .map(_.get(lastIngestedRecordTime))
+          archivedAtNextO <- lookupArchivedAtForOpenMiningRound(roundNumber + 1)
+        } yield (activeRoundO, archivedAtNextO) match {
+          case (Some((activeRoundNumber, activeRoundOpensAt)), Some(archivedAtNext)) =>
+            val retainUntil = now.minus(retentionPeriod.asJava)
+            if (activeRoundOpensAt <= archivedAtNext) {
+              logger.debug(
```
rautenrieth-da: In practice (with a retention period of 7 days), we should never run into situations where a round outside of the retention period has to be skipped because the pruned round is too close to the currently active round, no? Can we maybe swap the two branches so we get the "is still within the retention period" log instead of this complicated one?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbScanRewardsReferenceStoreTest.scala
```diff
@@ -426,6 +421,408 @@ class DbScanRewardsReferenceStoreTest
...
+
+          _ <- sync1.archive(pr3, recordTime = ts(230).toInstant)(store.multiDomainAcsStore)
+          afterPr3Archived <- store.lookupLowestPrunableArchivedRewardRound()
+          _ = afterPr3Archived shouldBe Some(3L)
+        } yield succeed
+      }
+
+    }
+
+    "lookupPrunableRewardRound" should {
```
rautenrieth-da, line 643: Is this missing testing the `Skipping pruning of round $roundNumber as round ${roundNumber + 1} has not archived yet.` branch?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
```diff
@@ -0,0 +1,136 @@
...
+    triggerContext: TriggerContext,
+)(implicit
+    override val ec: ExecutionContext,
+    mat: Materializer,
+    override val tracer: Tracer,
+) extends PollingParallelTaskExecutionTrigger[Task] {
+
+  override protected def context: TriggerContext = triggerContext.copy(
+    config = triggerContext.config.copy(
+      parallelism = 1
```
rautenrieth-da, line 62: We could probably also tweak the polling interval. Running the trigger every 30sec is wasteful if it can only prune data once every 10min. I don't think we need to do this now though.

rautenrieth-da (approved):
> TODO: add release notes

This is still missing, right? Otherwise the PR looks good, thanks!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/automation/PruneRewardAccountingTrigger.scala
dfordivam: yeah, a 2 min interval may be good. To clear the backlog at 5x pace should be good enough.

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/store/db/DbScanRewardsReferenceStoreTest.scala
dfordivam: Actually adding a test for this branch turned out to be a little tricky, and not that useful.
The reason being the "ingestion start" of the store need to be before the openAt of round, otherwise we will hit the `case (None, _)` branch.
And I will have to fabricate a store state with an archived contract like `FeaturedAppRight` to exist in the archived table before the `openAt` of the currently open round, while also ensuring that the archived contract of the previous round is not in archive table. This scenario can be created in test, but can't happen if we follow the logic of round advancement and pruning.
Do you see a value in adding this test? or I should instead change the message to "this can never happen"

rautenrieth-da: I see. No need to force it in a test then. Consider bumping the log message to a warning and adding a "This should never happen" to the text.

dfordivam: Ok converted this to a warn, and also added a store bootstrap / ingestion bootstrap test for the other `case (None, _)` branch. This test also acts as a demonstration to assert that the warn wont hit.

PR #6875, opened 2026-08-20
Remove in-app rate limit override for `getDateOfMostRecentSnapshotBefore`

martinflorian-da:
[static]

A client gets rate limited on this on MainNet DA-2. Upon closer inspection, this endpoint triggers a very cheap query answered by an index scan: https://github.com/canton-network/splice/blob/1973921ef727fded6d6d3734b00e7eeca4010b51/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala#L72

This doesn't sound as if special rate-limiting is needed at all.


moritzkiefer-da (approved):
🔥

PR #6894, opened 2026-08-20
Cip 0105 with custom templates POC

jonored:
Implement governance locks with separate templates and _not_ on top of V2 allocations (but _do_ provide the V2 allocation interface for the separate lock templates).

Next steps are to do a proper happy path test and then add a sketch of the token standard V1 interface according to the implementation CIP work-in-progress.

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
+++ b/daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
@@ -1,17 +1,17 @@
 module Splice.Scripts.TestGovernanceLocks where
 
-import DA.Assert
 import DA.List.Total
 import DA.Time
 
-import Splice.Testing.Registries.AmuletRegistryV2 qualified as AmuletRegistryV2
 import Daml.Script
 import Splice.Amulet
 import Splice.AmuletRules
-import Splice.Api.Token.HoldingV1
+import Splice.Api.Token.HoldingV2 qualified as V2
 import Splice.ExternalPartyConfigState
 import Splice.GovernanceLocks
 import Splice.Scripts.TokenStandard.TestAmuletTokenStandardTestEnv
+import Splice.Testing.Registries.AmuletRegistryV2 qualified as AmuletRegistryV2
+import Splice.Testing.TokenStandard.WalletClientV2
 import Splice.Testing.Utils
 
 
@@ -40,7 +40,7 @@ testGovernanceLockHappyPath = do
   (aliceVesting, Some aliceLock) <- governanceUnlockRawChoice env aliceLock (Some 80000.0) unlockTime [alice]
 
   -- Assert that we have a lock remaining of 100k Amulet.
-  assertLockedHoldingAmount alice aliceLock 100000.0
+  checkLockedHolding alice aliceLock 100000.0
 
   (externalPartyContext, disc) <- getExternalPartyTransferContext dso
   submitMustFail (actAs alice <> discloseMany' disc) $ exerciseCmd aliceVesting VestingLock_Withdraw with
@@ -61,10 +61,10 @@ testGovernanceLockHappyPath = do
   (unlocked, Some newVestingLock) <- vestingWithdrawRawChoice env aliceVesting None withdrawTime [alice]
 
   -- Check that we did indeed unlock 10k amulet as expected - 1/8th of the total vesting amount of 80k.
-  assertHoldingAmount alice unlocked 10000.0
+  checkHolding' alice unlocked 10000.0
 
   -- And check that we remain with 7/8ths of the vesting amount - 70k.
-  assertLockedHoldingAmount alice newVestingLock 70000.0
+  checkLockedHolding alice newVestingLock 70000.0
 
   passTime (days 365)
 
@@ -73,19 +73,17 @@ testGovernanceLockHappyPath = do
   (unlockedHolding2, None) <- vestingWithdrawRawChoice env newVestingLock None fullWithdrawTime [alice]
 
   -- Check that we withdrew all and exactly the 70k _remaining_ after the above withdrawal.
-  assertHoldingAmount alice unlockedHolding2 70000.0
+  checkHolding' alice unlockedHolding2 70000.0
 
+-- | Polymorphic checkHolding, to avoid repeated toInterfaceContractId when we have a LockedAmulet or Amulet ContractId.
+checkHolding' : Implements a V2.Holding => Party -> ContractId a -> Decimal -> Script ()
+checkHolding' party holding amount = checkHolding party (toInterfaceContractId holding) amount
 
-
-assertHoldingAmount : Implements a Holding => Party -> ContractId a -> Decimal -> Script ()
-assertHoldingAmount p holdingCid amount = do
-  Some holding <- (queryInterfaceContractId p $ toInterfaceContractId @Holding holdingCid)
-  holding.amount === amount
-
-assertLockedHoldingAmount : (Template t, HasEnsure t, HasField "holding" t (ContractId LockedAmulet)) => Party -> ContractId t -> Decimal -> Script ()
-assertLockedHoldingAmount p lockCid amount = do
+-- | Find the holding for a GovernanceLock or VestingLock, and check that it has the correct amount.
+checkLockedHolding : (Template t, HasEnsure t, HasField "holding" t (ContractId LockedAmulet)) => Party -> ContractId t -> Decimal -> Script ()
+checkLockedHolding p lockCid amount = do
   Some lock <- queryContractId p lockCid
-  assertHoldingAmount p lock.holding amount
+  checkHolding' p lock.holding amount
 
 getExternalPartyTransferContext : Party -> Script (ExternalPartyTransferContext, Disclosures')
 getExternalPartyTransferContext dso = do
@@ -108,11 +106,11 @@ governanceLockRawChoice env owner amount specification = do
   let dso = env.instrId.admin
   [(glfCid, _glf)] <- query @GovernanceLockFactory dso
   disc <- queryDisclosure' dso glfCid
-  inputs <- fmap fst <$> query @Amulet owner -- Excessive, check how the normal registry does it.
+  inputs <- listHoldingCids owner env.instrId
   GovernanceLockFactory_LockResult cid _ <- submitWithTransferContext env (actAs [owner] <> discloseMany' disc) $
     \context -> exerciseCmd glfCid GovernanceLockFactory_Lock with
       amount
-      inputs = toInterfaceContractId <$> inputs
+      inputs = coerceContractId <$> inputs -- These are Amulet and LockedAmulet, but we don't have convenient constraints to show that.
       specification
       context
       owner
--- a/daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
+++ b/daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
@@ -79,10 +79,10 @@ checkHolding' : Implements a V2.Holding => Party -> ContractId a -> Decimal -> S
 checkHolding' party holding amount = checkHolding party (toInterfaceContractId holding) amount
 
 -- | Find the holding for a GovernanceLock or VestingLock, and check that it has the correct amount.
-checkLockedHolding : (Template t, HasEnsure t, HasField "holding" t (ContractId LockedAmulet)) => Party -> ContractId t -> Decimal -> Script ()
+checkLockedHolding : (Template t, HasEnsure t, HasField "lockedAmulet" t (ContractId LockedAmulet)) => Party -> ContractId t -> Decimal -> Script ()
 checkLockedHolding p lockCid amount = do
   Some lock <- queryContractId p lockCid
-  checkHolding' p lock.holding amount
+  checkHolding' p lock.lockedAmulet amount
 
 getExternalPartyTransferContext : Party -> Script (ExternalPartyTransferContext, Disclosures')
 getExternalPartyTransferContext dso = do
--- a/daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
+++ b/daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
@@ -36,8 +36,10 @@ testGovernanceLockHappyPath = do
 
   let unlockTime = now `addRelTime` minutes 5
 
+  -- Unlock 80k Amulet such that 1/8th should result in 10k Amulet withdrawable after 1/8th year.
   (aliceVesting, Some aliceLock) <- governanceUnlockRawChoice env aliceLock (Some 80000.0) unlockTime [alice]
 
+  -- Assert that we have a lock remaining of 100k Amulet.
   assertLockedHoldingAmount alice aliceLock 100000.0
 
   (externalPartyContext, disc) <- getExternalPartyTransferContext dso
@@ -49,13 +51,19 @@ testGovernanceLockHappyPath = do
 
   passTime (days 46)
 
+  -- Chosen to be exactly 1/8th of the specified 365.25 year.
+  -- 1/8th and 80k are reliably exactly representible in fixed-point binary
+  -- with at least three fractional places, so the arithmetic is exact and we can
+  -- comfortably use equality.
 
-  let withdrawTime = unlockTime `addRelTime` (days 45 + hours 15 + minutes 45) -- Chosen to be 1/8th of the specified 365.25 year.
+  let withdrawTime = unlockTime `addRelTime` (days 45 + hours 15 + minutes 45)
 
   (unlocked, Some newVestingLock) <- vestingWithdrawRawChoice env aliceVesting None withdrawTime [alice]
 
+  -- Check that we did indeed unlock 10k amulet as expected - 1/8th of the total vesting amount of 80k.
   assertHoldingAmount alice unlocked 10000.0
 
+  -- And check that we remain with 7/8ths of the vesting amount - 70k.
   assertLockedHoldingAmount alice newVestingLock 70000.0
 
   passTime (days 365)
@@ -64,6 +72,7 @@ testGovernanceLockHappyPath = do
 
   (unlockedHolding2, None) <- vestingWithdrawRawChoice env newVestingLock None fullWithdrawTime [alice]
 
+  -- Check that we withdrew all and exactly the 70k _remaining_ after the above withdrawal.
   assertHoldingAmount alice unlockedHolding2 70000.0
 
 
--- a/daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
+++ b/daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
@@ -1,0 +1,129 @@
+module Splice.Scripts.TestGovernanceLocks where
+
+import DA.Assert
+import DA.List.Total
+import DA.Time
+
+import Splice.Testing.Registries.AmuletRegistryV2 qualified as AmuletRegistryV2
+import Daml.Script
+import Splice.Amulet
+import Splice.AmuletRules
+import Splice.Api.Token.HoldingV1
+import Splice.ExternalPartyConfigState
+import Splice.GovernanceLocks
+import Splice.Scripts.TokenStandard.TestAmuletTokenStandardTestEnv
+import Splice.Testing.Utils
+
+
+testGovernanceLockHappyPath : Script ()
+testGovernanceLockHappyPath = do
+  env@TestEnv{..} <- setupTest
+  let dso = env.instrId.admin
+
+  let someSvSpec = GovernanceLockSpecification with
+        subject = GLS_SuperValidatorRightsOwner "SomeApp"
+
+  submit (actAs dso) $ createCmd GovernanceLockFactory with dso
+
+  AmuletRegistryV2.tapFaucet registriesEnv.amuletV2 alice 180000.0
+  AmuletRegistryV2.tapFaucet registriesEnv.amuletV2 bob 10000.0
+
+  aliceLock <- governanceLockRawChoice env alice 180000.0 someSvSpec
+
+  _bobLock <- governanceLockRawChoice env bob 10000.0 someSvSpec
+
+  now <- getTime
+
+  let unlockTime = now `addRelTime` minutes 5
+
+  (aliceVesting, Some aliceLock) <- governanceUnlockRawChoice env aliceLock (Some 80000.0) unlockTime [alice]
+
+  assertLockedHoldingAmount alice aliceLock 100000.0
+
+  (externalPartyContext, disc) <- getExternalPartyTransferContext dso
+  submitMustFail (actAs alice <> discloseMany' disc) $ exerciseCmd aliceVesting VestingLock_Withdraw with
+    amount = Some 1.0
+    requestedAt = unlockTime
+    context = externalPartyContext
+    actors = [ alice ]
+
+  passTime (days 46)
+
+
+  let withdrawTime = unlockTime `addRelTime` (days 45 + hours 15 + minutes 45) -- Chosen to be 1/8th of the specified 365.25 year.
+
+  (unlocked, Some newVestingLock) <- vestingWithdrawRawChoice env aliceVesting None withdrawTime [alice]
+
+  assertHoldingAmount alice unlocked 10000.0
+
+  assertLockedHoldingAmount alice newVestingLock 70000.0
+
+  passTime (days 365)
+
+  let fullWithdrawTime = unlockTime `addRelTime` days 366
+
+  (unlockedHolding2, None) <- vestingWithdrawRawChoice env newVestingLock None fullWithdrawTime [alice]
+
+  assertHoldingAmount alice unlockedHolding2 70000.0
+
+
+
+assertHoldingAmount : Implements a Holding => Party -> ContractId a -> Decimal -> Script ()
+assertHoldingAmount p holdingCid amount = do
+  Some holding <- (queryInterfaceContractId p $ toInterfaceContractId @Holding holdingCid)
+  holding.amount === amount
+
+assertLockedHoldingAmount : (Template t, HasEnsure t, HasField "holding" t (ContractId LockedAmulet)) => Party -> ContractId t -> Decimal -> Script ()
+assertLockedHoldingAmount p lockCid amount = do
+  Some lock <- queryContractId p lockCid
+  assertHoldingAmount p lock.holding amount
+
+getExternalPartyTransferContext : Party -> Script (ExternalPartyTransferContext, Disclosures')
+getExternalPartyTransferContext dso = do
+  extPartyConfigs <- query @ExternalPartyConfigState dso
+  let (externalPartyConfigState, _) = case maximumOn (\(_, c) -> c.targetArchiveAfter) extPartyConfigs of
+        None -> error "No ExternalPartyConfigState found: have you initialized the Amulet registry?"
+        Some c -> c
+  ecDD <- queryDisclosure' @ExternalPartyConfigState dso externalPartyConfigState
+  let eptc = ExternalPartyTransferContext with externalPartyConfigState; featuredAppRight = None
+  pure (eptc, ecDD)
+
+submitWithTransferContext : TestEnv -> SubmitOptions -> (ExternalPartyTransferContext -> Commands a) -> Script a
+submitWithTransferContext env options commands = do
+  let dso = env.instrId.admin
+  (context, disclosures) <- getExternalPartyTransferContext dso
+  submit (options <> discloseMany' disclosures) $ commands context
+
+governanceLockRawChoice : TestEnv -> Party -> Decimal -> GovernanceLockSpecification -> Script (ContractId GovernanceLock)
+governanceLockRawChoice env owner amount specification = do
+  let dso = env.instrId.admin
+  [(glfCid, _glf)] <- query @GovernanceLockFactory dso
+  disc <- queryDisclosure' dso glfCid
+  inputs <- fmap fst <$> query @Amulet owner -- Excessive, check how the normal registry does it.
+  GovernanceLockFactory_LockResult cid _ <- submitWithTransferContext env (actAs [owner] <> discloseMany' disc) $
+    \context -> exerciseCmd glfCid GovernanceLockFactory_Lock with
+      amount
+      inputs = toInterfaceContractId <$> inputs
+      specification
+      context
+      owner
+  pure cid
+
+governanceUnlockRawChoice : TestEnv -> ContractId GovernanceLock -> Optional Decimal -> Time -> [Party] -> Script (ContractId VestingLock, Optional (ContractId GovernanceLock))
+governanceUnlockRawChoice env cid amount requestedAt actors = do
+  result <- submitWithTransferContext env (actAs actors) $ \context -> exerciseCmd cid GovernanceLock_Unlock with
+    amount
+    context
+    requestedAt
+    actors
+  pure (result.vestingLock, result.governanceLock)
+
+vestingWithdrawRawChoice : TestEnv -> ContractId VestingLock -> Optional Decimal -> Time -> [Party] -> Script (ContractId Amulet, Optional (ContractId VestingLock))
+vestingWithdrawRawChoice env cid amount requestedAt actors = do
+  result <- submitWithTransferContext env (actAs actors) $ \context -> exerciseCmd cid VestingLock_Withdraw with
+    amount
+    context
+    requestedAt
+    actors
+  pure (result.unlocked, result.vestingLock)
+
--- a/daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
+++ b/daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
@@ -44,7 +44,6 @@ testGovernanceLockHappyPath = do
 
   (externalPartyContext, disc) <- getExternalPartyTransferContext dso
   submitMustFail (actAs alice <> discloseMany' disc) $ exerciseCmd aliceVesting VestingLock_Withdraw with
-    amount = Some 1.0
     requestedAt = unlockTime
     context = externalPartyContext
     actors = [ alice ]
@@ -58,7 +57,7 @@ testGovernanceLockHappyPath = do
 
   let withdrawTime = unlockTime `addRelTime` (days 45 + hours 15 + minutes 45)
 
-  (unlocked, Some newVestingLock) <- vestingWithdrawRawChoice env aliceVesting None withdrawTime [alice]
+  (unlocked, Some newVestingLock) <- vestingWithdrawRawChoice env aliceVesting withdrawTime [alice]
 
   -- Check that we did indeed unlock 10k amulet as expected - 1/8th of the total vesting amount of 80k.
   checkHolding' alice unlocked 10000.0
@@ -70,7 +69,7 @@ testGovernanceLockHappyPath = do
 
   let fullWithdrawTime = unlockTime `addRelTime` days 366
 
-  (unlockedHolding2, None) <- vestingWithdrawRawChoice env newVestingLock None fullWithdrawTime [alice]
+  (unlockedHolding2, None) <- vestingWithdrawRawChoice env newVestingLock fullWithdrawTime [alice]
 
   -- Check that we withdrew all and exactly the 70k _remaining_ after the above withdrawal.
   checkHolding' alice unlockedHolding2 70000.0
@@ -125,10 +124,9 @@ governanceUnlockRawChoice env cid unlockAmount unlockAt actors = do
     actors
   pure (result.vestingLock, result.governanceLock)
 
-vestingWithdrawRawChoice : TestEnv -> ContractId VestingLock -> Optional Decimal -> Time -> [Party] -> Script (ContractId Amulet, Optional (ContractId VestingLock))
-vestingWithdrawRawChoice env cid amount requestedAt actors = do
+vestingWithdrawRawChoice : TestEnv -> ContractId VestingLock -> Time -> [Party] -> Script (ContractId Amulet, Optional (ContractId VestingLock))
+vestingWithdrawRawChoice env cid requestedAt actors = do
   result <- submitWithTransferContext env (actAs actors) $ \context -> exerciseCmd cid VestingLock_Withdraw with
-    amount
     context
     requestedAt
     actors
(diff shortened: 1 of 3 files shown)


daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+    dso : Party
+  where
+    signatory dso
+    choice GovernanceLockFactory_Lock : GovernanceLockFactory_LockResult
+      with
+        amount : Decimal
+        inputs : [ContractId Holding]
+        specification : GovernanceLockSpecification
+        context : ExternalPartyTransferContext
+        authorizer : Party
```
meiersi-da:
```suggestion
        owner : Party
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+  = GovernanceLockSpecification with
+      subject : GovernanceLockSubject
+  deriving (Eq, Show, Serializable)
+
+template GovernanceLockFactory
+  with
+    dso : Party
+  where
+    signatory dso
+    choice GovernanceLockFactory_Lock : GovernanceLockFactory_LockResult
```
meiersi-da: Makes a lot of sense to expose this choice from a separate factory for the PoC. I suspect though that ultimately it will be easier to expose it directly from `ExternalPartyAmuletRules`. Creating the singleton contract for the factory, ingesting it into the stores, and exposing it via Scan is an effort that we can save this way.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+          dso
+          owner = authorizer
+          holding
+          specification
+        pure $ GovernanceLockFactory_LockResult with
+          governanceLock
+          changeHolding
+
+  -- TODO: implement V2 AllocationFactory in terms of above.
+  -- TODO: find and use binding point for "magic address" V1 token standard to above.
```
meiersi-da, line 52: It's here: https://github.com/canton-network/splice/blob/3e4bb9a094ad7f91f456b64d8f891827911f5657/daml/splice-amulet/daml/Splice/ExternalPartyAmuletRules.daml#L341-L346

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+    specification : GovernanceLockSpecification
+  where
+    signatory dso, owner
+    choice GovernanceLock_Unlock : GovernanceLock_UnlockResult
+      with
+        amount : Optional Decimal
+        requestedAt : Time
+        context : ExternalPartyTransferContext
+        authorizers : [Party]
+      controller authorizers
```
meiersi-da:
```suggestion
        actors : [Party]
      controller actors
```

nit: authorizers on its own would work well, but in TSv2 authorizer has a specific meaning for allocation, and the actors pattern that we use here is used consistently with the naming proposed in the suggestion.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+      do
+        require "Only the authorizer may unlock" $ authorizers == [owner] -- TODO: custom controllers
```
meiersi-da:
```suggestion
        require "Only the owner may unlock" $ authorizers == [owner] -- TODO: custom controllers
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+        assertWithinDeadline "requestedAt must be in the future for unlock" requestedAt
```
meiersi-da: wdyt about using `unlockAt` or `effectiveAt` as "requested" being in the future is kind of confusing (and "ungrammatical").

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+  = GovernanceLockFactory_LockResult with
+      governanceLock : ContractId GovernanceLock
+      changeHolding : Optional (ContractId Amulet)
+  deriving (Eq, Show, Serializable)
+
+
+template GovernanceLock
+  with
+    dso : Party
+    owner : Party
```
meiersi-da:
```suggestion
    owner : Party  -- ^ Lock owner
    amount : Decimal  -- ^ Locked amount
```

This way the lock becomes self-describing. It may also simplify some checks.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+      controller authorizers
+      do
+        require "Only the authorizer may unlock" $ authorizers == [owner] -- TODO: custom controllers
+        assertWithinDeadline "requestedAt must be in the future for unlock" requestedAt
+
+        (vestingLockedHolding, governanceLock, initialAmount) <- case amount of
+          None -> do -- No split required, reuse the LockedAmulet contract.
+            currentAmount <- fetchAmount (ForOwner with dso; owner) holding
+            pure (holding, None, currentAmount)
+          Some unlockAmount -> do -- Splitting.
```
meiersi-da, line 81: if an amount is given it may also just match the full amount locked. So, there is not necessarily a still locked holding

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+  with
+    dso : Party
+    owner : Party
+    holding : ContractId LockedAmulet
+    specification : GovernanceLockSpecification
+  where
+    signatory dso, owner
+    choice GovernanceLock_Unlock : GovernanceLock_UnlockResult
+      with
+        amount : Optional Decimal
```
meiersi-da: I'm not sure it is worth making this optional. In particular when storing the locked amount on the template it is very easy for the caller to select the full amount themselves if they want to do so.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+        requestedAt : Time
+        context : ExternalPartyTransferContext
+        authorizers : [Party]
+      controller authorizers
+      do
+        require "Only the authorizer may withdraw" $ authorizers == [owner] -- TODO: custom controllers
+        currentAmount <- fetchAmount (ForOwner with dso; owner) holding
+        assertDeadlineExceeded "requestAt must be in the past for withdraw" requestedAt
+        require "withdrawal time must be after the start time" $ requestedAt > startTime
```
meiersi-da: for polish: switch to `require'` from the TokenStandardUtils, which gives better error messages out of the box. Example usage here: https://github.com/canton-network/splice/blob/3e4bb9a094ad7f91f456b64d8f891827911f5657/daml/splice-amulet/daml/Splice/AmuletAllocationV2.daml#L122

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+    owner : Party
+    holding : ContractId LockedAmulet
+    unlockPeriod : RelTime
+    startTime : Time
+    initialAmount : Decimal
+  where
+    signatory dso, owner
+    choice VestingLock_Withdraw : VestingLock_WithdrawResult
+      with
+        amount : Optional Decimal
```
meiersi-da: When would a caller not want to withdraw the maximal amount? Partial unlocks make sense, but partial withdraws not really. The caller pays the traffic cost anyways, so they want to maximize their flexibility.


I'd suggest we remove this argument.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+      unlocked : ContractId Amulet
+  deriving (Eq, Show, Serializable)
+
+calculateAvailableWithdrawAmount : Decimal -> Time -> VestingLock -> Optional Decimal -- None indicates the _full_ amount is available
+calculateAvailableWithdrawAmount currentAmount currentTime VestingLock{ .. }
+    | currentTime > (startTime `addRelTime` unlockPeriod) = None
+    | otherwise = Some $ initialAmount * fractionElapsed - (initialAmount - currentAmount)
+  where
+    fractionElapsed = (currentTime `subTime` startTime) `divRelTime` unlockPeriod
+    divRelTime = \a b -> (intToDecimal $ convertRelTimeToMicroseconds a) `div` (intToDecimal $ convertRelTimeToMicroseconds b)
```
meiersi-da, line 174: wdyt about untangling the concerns in here by separating out the function that computes the vested amount as of a particular time from subtracting the amount that was already withdrawn.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,246 @@
...
+partialUnlock : ForOwner -> ContractId LockedAmulet -> Decimal -> ExternalPartyTransferContext -> Update (ContractId LockedAmulet, ContractId Amulet)
+partialUnlock forOwner lockedAmuletCid amount context = do
+  currentAmount <- fetchAmount forOwner lockedAmuletCid
+  unlockResult <- exercise lockedAmuletCid LockedAmulet_UnlockV2
+  ([newLockedCid], Some unlockedAmulet) <- lockAmulet forOwner [currentAmount - amount] [toInterfaceContractId $ unlockResult.amuletCid] context
+  pure (newLockedCid, unlockedAmulet)
+
+lockAmulet : ForOwner -> [Decimal] -> [ContractId Holding] -> ExternalPartyTransferContext -> Update ([ContractId LockedAmulet], Optional (ContractId Amulet))
+lockAmulet forOwner@ForOwner {..} amounts inputHoldingCids context = do
+  let lockContext = "governance"
```
meiersi-da, line 209: polish: encode type of lock here, and whether it is vesting

meiersi-da (commented):
Nice! Good work and great to see how much simpler the code became with our decision to separate the implementation concerns. Thanks for driving this PoC!

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: I almost added a comment to that effect; agreed. I'm not completely sure whether we ought to expose this particular _choice_ as intended API at all, though, as opposed to factoring it out into a function that we end up calling from the V2 AllocationFactory and V1 TransferFactory choices and this template exists just for tests written against the core.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: I'm not entirely fond of the denormalization from having the amount in both GovernanceLock and in the LockedAmulet, but I don't strenuously object and we can certainly guarantee that they stay in sync.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: Counterargument: both of the planned main public interfaces to this choice use a "withdraw from this agreement entirely" primitive that does not inherently supply an amount for a partial withdrawal, so an optional amount here avoids a bit of duplication. On the other hand, we could disallow V2 withdrawal without an amount in the metadata, which would make V1 and V2 handling differ enough to invalidate that argument.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: I was somewhat deliberately leaving amount = None as the canonical "withdraw everything" argument, rather than branching on amount == this.amount - I'd rather only one representation for a total withdrawal, but if it's the None version then it should reject with a decent message unless amount < this.amount rather than on a pattern match failure.

Although - that check would actually reject unless amount < this.amount - minimumLockAmount, which we need to add anyways, so we'd have to add an exception to that rule for total withdrawal as well.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: Fair enough; this is probably not very useful; I'll drop it.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: Reasonable; good to have the overall total available as a loose function anyways.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: `unlockAt` seems fine to me. It's always a little odd because it's notionally "now" for both governance and vesting, but the bounds are on opposite sides because we're constraining the effective duration that the vesting lock has run to be at maximum the actual ledger time between unlock call and withdrawal time.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: Style aside with regard to normalization, in retrospect this is a hard requirement for the interface views anyways. Will do.

daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
```diff
@@ -0,0 +1,129 @@
...
+  passTime (days 46)
+
+
+  let withdrawTime = unlockTime `addRelTime` (days 45 + hours 15 + minutes 45) -- Chosen to be 1/8th of the specified 365.25 year.
+
+  (unlocked, Some newVestingLock) <- vestingWithdrawRawChoice env aliceVesting None withdrawTime [alice]
+
+  assertHoldingAmount alice unlocked 10000.0
+
+  assertLockedHoldingAmount alice newVestingLock 70000.0
```
cgibbard: It might be good to make these constants into parameters (or just let bind them at the top at least) and calculate them in assertions in terms of the initial locked amounts and other values, just to make the arithmetic of what's being tested explicit. (At least so long as the numbers are behaving themselves like this.)

```
unlockedAmount = 80000
withdrawTimeFactor = 1/8
withdrawTime = scaleRelTime withdrawTimeFactor (days 365 + hours 6)
expectedWithdrawAmount = withdrawTimeFactor * unlockedAmount
expectedLockedHoldingAmount = unlockedAmount - expectedWithdrawAmount
```

Just realised while writing this that there's nothing like scaleRelTime in the library and that's possibly why we've got the arbitrary constants here, but it might be worth writing in terms of the round trip to microseconds...

jonored: I very much want a handful of these tests to go this specific and have a manually calculated set of explicit numbers; I actually spent a not trivial amount of time checking for one that successfully lands on a whole number when run through Decimal. That's not to say that I think they _all_ need to be like this, but I specifically don't want to rely _entirely_ on "that's the right value apart from a few decimal places of odd at the bottom end, so I'll pin that value as what the test checks" or parallel arithmetic running in the test.

In this case, the factor of 1/8 is chosen specifically because it's 1 >> 4 and thus we expect the binary expression of the withdrawal amount to be the total shifted left by four places (which critically _still fits_ in the 128-bit mantissa of Decimal and thus produces an exact answer); I think for Decimal in particular (but not for Numeric in general, because 1 is closer to the low end of the mantissa in half of them) we can shift any integers down by ten bits before there's a possibility that it might not be representable. I'm doing that because we aren't very interested in whether any given fraction of the unlock period is exactly representable, but we _do_ want to make it unlikely that we are making the _same_ error in the test and the implementation and missing a bug. I don't think we need to juggle precision to make the implementation any more careful than it already is, at least.

Pushed a tweak that ought to raise that from ~10 bits of available shift to 30-some; we don't need the "fraction elapsed" to represent values outside of the 0-1 range, and I think it didn't impair readability to just use Numeric 37 for that value.

daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
```diff
@@ -0,0 +1,129 @@
...
+
+  let someSvSpec = GovernanceLockSpecification with
+        subject = GLS_SuperValidatorRightsOwner "SomeApp"
+
+  submit (actAs dso) $ createCmd GovernanceLockFactory with dso
+
+  AmuletRegistryV2.tapFaucet registriesEnv.amuletV2 alice 180000.0
+  AmuletRegistryV2.tapFaucet registriesEnv.amuletV2 bob 10000.0
+
+  aliceLock <- governanceLockRawChoice env alice 180000.0 someSvSpec
```
meiersi-da, line 30: Good call to test directly on the raw choices. They are easier to work with and have a faster feedback loop.

meiersi-da: Consider later just making it an env parameter what API interface is used to execute an action. This way it becomes easy to run the same test code with raw or TSv2 APIs

daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
```diff
@@ -0,0 +1,129 @@
...
+
+assertHoldingAmount : Implements a Holding => Party -> ContractId a -> Decimal -> Script ()
+assertHoldingAmount p holdingCid amount = do
+  Some holding <- (queryInterfaceContractId p $ toInterfaceContractId @Holding holdingCid)
+  holding.amount === amount
+
+assertLockedHoldingAmount : (Template t, HasEnsure t, HasField "holding" t (ContractId LockedAmulet)) => Party -> ContractId t -> Decimal -> Script ()
+assertLockedHoldingAmount p lockCid amount = do
+  Some lock <- queryContractId p lockCid
+  assertHoldingAmount p lock.holding amount
```
meiersi-da: Note there are a bunch of utils: https://github.com/canton-network/splice/blob/8bc2e0756dd6a8da3fa3de75239021934d210801/token-standard/splice-token-standard-v2-test/daml/Splice/Testing/TokenStandard/WalletClientV2.daml#L19-L27

meiersi-da: use these if possible

daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
```diff
@@ -0,0 +1,129 @@
...
+  let dso = env.instrId.admin
+  (context, disclosures) <- getExternalPartyTransferContext dso
+  submit (options <> discloseMany' disclosures) $ commands context
+
+governanceLockRawChoice : TestEnv -> Party -> Decimal -> GovernanceLockSpecification -> Script (ContractId GovernanceLock)
+governanceLockRawChoice env owner amount specification = do
+  let dso = env.instrId.admin
+  [(glfCid, _glf)] <- query @GovernanceLockFactory dso
+  disc <- queryDisclosure' dso glfCid
+  inputs <- fmap fst <$> query @Amulet owner -- Excessive, check how the normal registry does it.
```
meiersi-da: use one of: https://github.com/canton-network/splice/blob/8bc2e0756dd6a8da3fa3de75239021934d210801/token-standard/splice-token-standard-v2-test/daml/Splice/Testing/TokenStandard/WalletClientV2.daml#L13-L17

daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
```diff
@@ -0,0 +1,129 @@
...
+import Splice.Testing.Utils
+
+
+testGovernanceLockHappyPath : Script ()
+testGovernanceLockHappyPath = do
+  env@TestEnv{..} <- setupTest
+  let dso = env.instrId.admin
+
+  let someSvSpec = GovernanceLockSpecification with
+        subject = GLS_SuperValidatorRightsOwner "SomeApp"
```
meiersi-da:
```suggestion
        subject = GLS_SuperValidatorRightsOwner "TestSV"
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,247 @@
...
+import Splice.Expiry
+import qualified Splice.TokenStandard.Utils as TSU
+import Splice.Types
+import Splice.Util
+
+data GovernanceLockSubject
+  = GLS_FeaturedApp with
+      appProviderParty : Party
+  | GLS_SuperValidatorRightsOwner with
+      superValidatorName : Text
```
meiersi-da:
```suggestion
      svName : Text  -- ^ Name of the SV rights owner for whom the funds are locked.
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,247 @@
...
+  deriving (Eq, Show, Serializable)
+
+-- Note: will also contain controllers specifications, when added.
```
meiersi-da:
```suggestion
-- FIXME: add controllers specifications
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,247 @@
...
+data GovernanceLockSpecification
+  = GovernanceLockSpecification with
+      subject : GovernanceLockSubject
+  deriving (Eq, Show, Serializable)
+
+template GovernanceLockFactory
```
meiersi-da:
```suggestion
-- FIXME: move the choices of this factory into `ExternalPartyAmuletRules` once we're happy with them
template GovernanceLockFactory
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,247 @@
...
+  with
+    dso : Party
+    owner : Party
+    holding : ContractId LockedAmulet
+    specification : GovernanceLockSpecification
+  where
+    signatory dso, owner
+    choice GovernanceLock_Unlock : GovernanceLock_UnlockResult
+      with
+        amount : Optional Decimal
```
meiersi-da:
```suggestion
        unlockAmount : Optional Decimal -- ^ Amount to unlock. Defaults to the full amount if not set.
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
meiersi-da: I would aim for V2 to be as easy to use as the compatibility mode. So it should support the default. I'm OK either way wrt the defaulting.

I realize that my main concern is about the implementation: there I'd just compute the concrete amount to unlock up front, and make the code work with that.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,247 @@
...
+        requestedAt : Time
```
meiersi-da:
```suggestion
        unlockAt : Time  -- ^ Time as of which the unlock is to be registered and vesting starts. Must be in the future. 
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,247 @@
...
+        context : ExternalPartyTransferContext
+        actors : [Party]
+      controller actors
+      do
+        require "Only the owner may unlock" $ actors == [owner] -- TODO: custom controllers
```
meiersi-da:
```suggestion
        checkActors [[owner]] -- ^ TODO: custom controllers
```

meiersi-da: From https://github.com/canton-network/splice/blob/8bc2e0756dd6a8da3fa3de75239021934d210801/token-standard/splice-token-standard-utils/daml/Splice/TokenStandard/Utils/Internal/Choices.daml#L157-L160

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
meiersi-da: I've made a suggestion above for the param name and the comment that explains what it means. I usually find that adding these comments helps to relieve the names of carrying all the conceptual load.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
meiersi-da: My optimization priorities are:
1. ease of reasoning about correctness
2. caller UX for happy path
3. quality of error messages

I think this here is a case where we are making reasoning about correctness more difficult to improve the quality of error messages. I wouldn't do that.

Just compute `let actualUnlockAmount = fromOptional amount unlockAmount` and then implement unlocking that using a single codepath that's always used.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,247 @@
...
+  with
+    dso : Party
+    owner : Party
+    holding : ContractId LockedAmulet
+    unlockPeriod : RelTime
+    startTime : Time
+    initialAmount : Decimal
+  where
+    signatory dso, owner
+    choice VestingLock_Withdraw : VestingLock_WithdrawResult
```
meiersi-da, line 127:
```suggestion

    -- | Withdraw all funds that have vested since the last withdrawal
    -- and the time of this request.
    choice VestingLock_Withdraw : VestingLock_WithdrawResult
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,247 @@
...
+      with
+        amount : Optional Decimal
+        requestedAt : Time
```
meiersi-da:
```suggestion
        requestedAt : Time -- ^ Time to use for determining the funds that have vested. Must be in the past.
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,247 @@
...
+        context : ExternalPartyTransferContext
+        actors : [Party]
+      controller actors
+      do
+        require "Only the owner may withdraw" $ actors == [owner] -- TODO: custom controllers
+        currentAmount <- fetchAmount (ForOwner with dso; owner) holding
+        assertDeadlineExceeded "requestAt must be in the past for withdraw" requestedAt
```
meiersi-da:
```suggestion
        assertDeadlineExceeded "requestedAt" requestedAt
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
meiersi-da: There actually seems to be a simpler option available: just compute the funds available for withdrawal at `requestedAt` given the state of the `VestingLock` and forbid withdraw in case there are no funds available.

This takes care in a uniform way of delayed unlocks, repeated withdrawals, and the final withdrawal.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,247 @@
...
+            pure VestingLock_WithdrawResult with
+              vestingLock = None
+              unlocked = unlockResult.amuletCid
+          Some partial -> do -- We have a fraction that is not the whole being withdrawn, do a partial unlock.
+            (locked, unlocked) <- partialUnlock (ForOwner with dso; owner) holding partial context
+            vestingLock <- Some <$> create this with
+              holding = locked
+            pure VestingLock_WithdrawResult with
+              vestingLock
+              unlocked
```
meiersi-da: if you implement `splitLock` such that it returns `Optional (ContractId LockedAmulet)` for the remainder, then you can just use that here as well; and you get a simpler implementation of `_Unlock` as well.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
meiersi-da: You didn't do that yet though, did you?

daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
jonored: This was also required to test before one of the interfaces is implemented, and getting tests ASAP was priority. Factoring out helpers we can switch between the interfaces seems sensible.

daml/splice-amulet-test/daml/Splice/Scripts/TestGovernanceLocks.daml
jonored: I was mostly wanting to check whether we had any that trimmed the input list to the required amount like I expect production to do, but it seems we don't do that in the tests.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: I tried to get that to be simpler but I'm not really happy with it; what we need _here_ is the change from executeExternalPartyTransfer and the amount we aren't relocking, and what we need for Unlock is a new LockedAmulet with the same info that it creates from a TransferOutput and the remainder after the new allocation we create; I didn't see any particularly elegant or straightforward way to merge those paths much more than they already are in the lockAmulet helper.

I think the complete withdrawal branch of this choice actually ends up equivalent to (and probably with a shared implementation with) the Expire choice once the holding has the shorter deadline, anyways; I don't see a reason to disturb it if it's directly usable.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: I'm not trying to optimize for error message quality; I actually do tend to find fewer branches and branches on enums rather than on continuous-ish value comaprisons easier to check for completeness, and this is that single branch on an enum. I was just commenting that it ought to be explicit that unlockAmount == amount is not permitted as it is indeed not in this arrangement.

It's probably not too bad on readability to tuck the subsequent branches away in the functor instance of Option, and it does make it _look_ more unified, so I'll see what I think of that. We can't reuse the holding for the VestingLock anyways, we want to reconfigure it to a shorter duration.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: Not yet; it'll go with changing the VestingLock representation.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: Still WIP; addressing tomorrow, this was a little messier in the helpers than I had time for today.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: Final withdraw probably branches away from this code entirely, as it can just leave the LockedAmulet in place; agreed with regard to just checking that there's anything to withdraw.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: addendum: I tried a few times towards the end of the day to push the decision of whether to make a "still locked" holding in the helpers below, but I did not find a way I was particularly happy with. I will give it one more attempt earlier in the day.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,249 @@
...
+      svName : Text  -- ^ Name of the SV rights owner for whom the funds are locked.
+  deriving (Eq, Show, Serializable)
+
+-- FIXME: add controllers specifications
+data GovernanceLockSpecification
+  = GovernanceLockSpecification with
+      subject : GovernanceLockSubject
+  deriving (Eq, Show, Serializable)
+
+-- FIXME: move the choices of this factory into `ExternalPartyAmuletRules` once we're happy with them
```
meiersi-da: let's do this as part of polishing this PR

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,249 @@
...
+      changeHolding : Optional (ContractId Amulet)
+  deriving (Eq, Show, Serializable)
+
+
+template GovernanceLock
+  with
+    dso : Party
+    owner : Party  -- ^ Lock owner
+    amount : Decimal  -- ^ Locked amount
+    holding : ContractId LockedAmulet
```
meiersi-da:
```suggestion
    lockedAmulet : ContractId LockedAmulet
```

clearly distinguishes it from token standard `Holding`s which comprise both locked and unlocked amulet.

search and replace everywhere

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,249 @@
...
+import Splice.Amulet.TwoStepTransfer (holdingToTransferInputs)
+import Splice.AmuletRules
+import Splice.Api.Token.HoldingV1 (Holding)
+import Splice.Api.Token.MetadataV1
+import Splice.Expiry
+import Splice.TokenStandard.Utils hiding (require)
+import Splice.Types
+import Splice.Util
+
+data GovernanceLockSubject
```
meiersi-da:
```suggestion
data GovernanceLockType
```

this way there's space for a `GLT_ProvisionalFeaturedApp with provider : Party`

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,249 @@
...
+  = GLS_FeaturedApp with
+      appProviderParty : Party
```
meiersi-da:
```suggestion
      provider : Party -- ^ Featured app provider party.
```

meiersi-da: nit: align to rest of codebase

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,249 @@
...
+
+template GovernanceLock
+  with
+    dso : Party
+    owner : Party  -- ^ Lock owner
+    amount : Decimal  -- ^ Locked amount
+    holding : ContractId LockedAmulet
+    specification : GovernanceLockSpecification
+  where
+    signatory dso, owner
```
meiersi-da, line 63:
```suggestion
    signatory dso, owner
    
```
nit: style alignement

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
meiersi-da: OK. I think what still confuses me is that for a lock of `amount` CC there's a difference between calling `Unlock` with `unlockAmount = None` vs `unlockAmount = Some amount`. I'd want a caller to have both options available, and I see no good reason for these two options to work differently.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,249 @@
...
+          unlockPeriod = vestingDuration
+          startTime = unlockAt
+          initialAmount
+
+        pure $ GovernanceLock_UnlockResult with
+          vestingLock
+          governanceLock
+
+  -- TODO: implement V2 Allocation interface on above. Most fields of the view are generated, as they are constants or specific presentations of the above.
+  -- TODO: implement pending V1 transaction for the above, for backwards compatibility.
```
meiersi-da:
```suggestion
  -- TODO: implement pending V1 transfer for the above, for backwards compatibility.
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,249 @@
...
+      vestingLock : ContractId VestingLock
+      governanceLock : Optional (ContractId GovernanceLock)
+  deriving (Eq, Show, Serializable)
+
+template VestingLock
+  with
+    dso : Party
+    owner : Party
+    holding : ContractId LockedAmulet
+    unlockPeriod : RelTime
```
meiersi-da:
```suggestion
    vestingPeriod : RelTime  -- ^ Duration over which the amount vests in a linear fashion. 
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,249 @@
...
+    startTime : Time
+    initialAmount : Decimal
+  where
+    signatory dso, owner
+
```
meiersi-da, line 124: add ensures clause

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,249 @@
...
+  deriving (Eq, Show, Serializable)
+
+template VestingLock
+  with
+    dso : Party
+    owner : Party
+    holding : ContractId LockedAmulet
+    unlockPeriod : RelTime
+    startTime : Time
+    initialAmount : Decimal
```
meiersi-da:
```suggestion
    vestingAmount : Decimal  -- ^ The amount that is vesting.
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,249 @@
...
+      governanceLock : Optional (ContractId GovernanceLock)
+  deriving (Eq, Show, Serializable)
+
+template VestingLock
+  with
+    dso : Party
+    owner : Party
+    holding : ContractId LockedAmulet
+    unlockPeriod : RelTime
+    startTime : Time
```
meiersi-da:
```suggestion
    startTime : Time  -- ^ Time at which vesting starts.
```

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
```diff
@@ -0,0 +1,249 @@
...
+    -- and the time of this request.
+    choice VestingLock_Withdraw : VestingLock_WithdrawResult
+      with
+        requestedAt : Time -- ^ Time to use for determining the funds that have vested. Must be in the past.
+        context : ExternalPartyTransferContext
+        actors : [Party]
+      controller actors
+      do
+        checkActors actors [[owner]] -- ^ TODO: custom controllers
+        currentAmount <- fetchAmount (ForOwner with dso; owner) holding
```
meiersi-da: do we still need this if we structure the code such that a vesting lock stores the amount that is still vesting?

meiersi-da: OK, saw below that you plan to change the structure of the vesting lock repr. I'll skip reviewing the body of this choice.

daml/splice-amulet/daml/Splice/GovernanceLocks.daml
jonored: Mapped it to the None case; my preference for canonical input representations is much weaker than the preferences about code flow, and both are weaker concerns than code reuse if the amount in common between full unlock and partial unlock increases.

PR #6908, opened 2026-08-21
fix stack references and their mocks to allow migrating to split-SV deployment on prod clusters

mblaze-da:


cluster/expected/canton-network/expected.json
```diff
@@ -174,22 +174,12 @@
...
+      "create": "'../../../cluster/pulumi/canton-network/bigquery-cloudsql.sh' create-pub-rep-slot \\\n      --private-network-project=\"test-project\" \\\n      --compute-region=\"europe-west6\" \\\n      --service-account-email=\"undefined\" \\\n      --schema-name=\"scan_sv_1\" \\\n      --tables-to-replicate-joined=\"update_history_creates, update_history_exercises, scan_verdict_store, scan_verdict_transaction_view_store, app_activity_record_store\" \\\n      --postgres-user-name=\"cnadmin\" \\\n      --publication-name=\"update_history_datastream_pub\" \\\n      --replication-slot-name=\"update_history_datastream_r_slot\" \\\n      --replicator-user-name=\"bqdatastream\" \\\n      --postgres-instance-name=\"undefined\" \\\n      --scan-app-database-name=\"scan_sv_1\" \\\n      --flyway-migration-to-wait-for=\"V068__app_activity_record_meta.sql\" \\\n      ",
```
nicu-da: do we understand why this changed?

nicu-da (approved):
thx!

cluster/expected/canton-network/expected.json
mblaze-da: It's because I introduced [this import](https://github.com/canton-network/splice/blob/af2579290e291818e1767af3fb1e76c6cc30b8d6/cluster/pulumi/common/src/dump-config-common.ts#L5) which caused [MOCK_SPLICE_ROOT](https://github.com/canton-network/splice/blob/af2579290e291818e1767af3fb1e76c6cc30b8d6/cluster/pulumi/common/src/utils.ts#L18) to initialise too early... I hate how much all this static initialisation messes up code reuse. I'll try to fix this.

PR #6933, opened 2026-08-24
Simplify istio rate limits

nicu-da:
By default only have a global rate limit

[static]

part of https://github.com/DACH-NY/canton-network-internal/issues/2115

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


cluster/configs/shared/base.yaml
```diff
@@ -259,9 +259,14 @@ sv:
   scan:
     externalRateLimits:
       !include(./rate-limits/v0-acs.yaml;./rate-limits/unlimited.yaml;./rate-limits/public-banned.yaml;./rate-limits/token-registry.yaml)
+      enablePerEndpointRateLimits: false
```
martinflorian-da: So the idea is to keep the functionality around in case we want to use it after all, at least for our deployments?

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.test.ts
```diff
@@ -213,6 +216,61 @@ test('buildRateLimitActions emits per-endpoint and per-IP actions', () => {
   });
 });
 
+test('buildGlobalPerIpRateLimitAction keys only on the non-spoofable client address', () => {
```
martinflorian-da, line 210: what does "keys" even mean in this context 😕

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
```diff
@@ -62,17 +62,18 @@ const minFillIntervalMs = 50;
...
   inboundPort: pulumi.Input<number>;
 
-  /**
-   * Used when no descriptors match the request.
-   * */
   globalLimits: Limits;
+
+  globalPerIpLimits?: Limits;
+
+  enablePerEndpointRateLimits?: boolean;
```
martinflorian-da: Hm can we just skip this flag and simply leave the per-endpoint-rate limits unset if we don't want these?

Having something explicitly set in a config that is actually unused because of a boolean flag somewhere seems a bit meh.

martinflorian-da (approved):
Thanks!

cluster/configs/shared/base.yaml
nicu-da: Yes, pretty much, as a safety precaution

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.test.ts
nicu-da: the key of the attribute used for the ip basically. making sure we don't change something unintentionally to not use the masked address

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
nicu-da: Agree, issue was that it would've made the PR more complex. Let me do it though

cluster/configs/shared/rate-limits/v0-acs.yaml
```diff
@@ -1,11 +0,0 @@
-rateLimits:
-  /api/scan/v0/acs:
-    name: acs
-    type: limited
-    maxTokens: 500         # The per-endpoint global limit (shared between all ips)
-    tokensPerFill: 500
-    fillInterval: 60s
-    perIpLimits: # per-IP limit
-      maxTokens: 10
```
martinflorian-da, line 9: Talking to myself: this should be fine: https://github.com/canton-network/splice/blob/9af44593e63cdb1bc38bdd558ab273c9e70ce180/cluster/images/scan-app/app.conf#L78

cluster/configs/shared/base.yaml
```diff
@@ -258,10 +258,14 @@ multiValidator:
 sv:
   scan:
     externalRateLimits:
-      !include(./rate-limits/v0-acs.yaml;./rate-limits/unlimited.yaml;./rate-limits/public-banned.yaml;./rate-limits/token-registry.yaml)
+      !include(./rate-limits/public-banned.yaml;./rate-limits/token-registry.yaml)
```
martinflorian-da: For token registry - is the medium-term to remove those as well in favor of relying only on the splice rate limits? (Shall we add a TODO if so?)

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
```diff
@@ -210,49 +195,13 @@ function validateEffectiveRateLimits(
...
-      );
-    }
-    throw new Error(errorParts.join('\n'));
-  }
-
   // Filter out banned and unlimited entries
   const effectiveRateLimits = Object.fromEntries(
     Object.entries(args.rateLimits || {}).filter(
       (ent): ent is [string, LocalLimit<MatchedLimits>] => {
         // TODO (#4201): in banned case, implement actual banning with special short-circuit for whitelisted IPs
```
martinflorian-da: Do we need to explicitly track this as nice-to-have for MainNet opening?

martinflorian-da (approved):
Removes more LOC than it adds, must be good. Thank you!

cluster/configs/shared/base.yaml
nicu-da: Let me create an issue to revisit right before opening it up. 
The idea is to remove it unless we explicitly want some higher limits for that part of the API

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
nicu-da: tbh not sure if it's even worth doing. 
we can just set the limit to 0 in the splice rate limiter, as we need to ship it to everyone anyway

cluster/configs/shared/base.yaml
martinflorian-da: > higher limits for that part of the API

Can we do actually do that? Successful requests will not deplete the global and default per-IP buckets?

nicu-da: no, the first match consumes the limit, the global one is basically the "default" and used if nothing else matches 
so we can sort of partition out certain parts of the API

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
martinflorian-da: IMHO that would be a valid way to resolve this, but AFAIU `banned` currently is a NOP, which can mislead us in the future.

cluster/configs/shared/base.yaml
martinflorian-da: Nice!

cluster/pulumi/common/src/ratelimit/envoyRateLimiter.ts
nicu-da: good point, let me just delete the banned stuff

PR #6941, opened 2026-08-24
Remove log ignore that is stale after #6935

martinflorian-da:
I thought about this a bit too late / hit merge on #6935 a bit too quickly...

[force] (is fine because that log message doesn't exist anymore)


OriolMunoz-da (approved):
ty

PR #6942, opened 2026-08-24
[ci] fix race condition in writing s3 objects

isegall-da:
#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
+++ b/apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
@@ -222,6 +222,14 @@ class S3BucketConnection(
     private val parts = TrieMap.empty[Integer, CompletedPart]
     private val md = MessageDigest.getInstance("SHA-256")
 
+    /** The checksum of the whole object. Computing it via a `lazy val` to support idempotent `finish()` calls.
+      */
+    private lazy val objectChecksum: String = Base64.getEncoder.encodeToString(md.digest())
+
+    /** `lazy val` to ensure that multi-part upload is completed at most once.
+      */
+    private lazy val finishResult: Future[Unit] = doFinish()
+
     /** Call this once before uploading a new part.
       *  The content must already be provided for checksums, but will not be uploaded yet.
       */
@@ -273,7 +281,12 @@ class S3BucketConnection(
       }
     }
 
-    def finish(): Future[Unit] = {
+    /** Completes the multi-part upload and stores the object checksum in the object's metadata.
+      * Idempotent, safe to call more than once (will just return the Future from the first call again).
+      */
+    def finish(): Future[Unit] = finishResult
+
+    private def doFinish(): Future[Unit] = {
       require(numParts.get() > 0)
       require(
         parts.size == numParts.get(),
@@ -297,7 +310,7 @@ class S3BucketConnection(
         _ <- s3Client.completeMultipartUpload(completeRequest).asScala
 
         // Copy-in-place of the object to add the final checksum to its metadata
-        metadata = Map("splice-checksum" -> Base64.getEncoder.encodeToString(md.digest()))
+        metadata = Map("splice-checksum" -> objectChecksum)
 
         copyReq = CopyObjectRequest
           .builder()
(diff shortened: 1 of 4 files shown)


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
```diff
@@ -222,6 +222,14 @@ class S3BucketConnection(
     private val parts = TrieMap.empty[Integer, CompletedPart]
     private val md = MessageDigest.getInstance("SHA-256")
 
+    /** The checksum of the whole object. Computing it via a `lazy val` to support idempotent `finish()` calls.
+      */
+    private lazy val objectChecksum: String = Base64.getEncoder.encodeToString(md.digest())
```
rautenrieth-da, line 227: Why does this have to be a lazy val? Isn't `doFinish` the only call site of this, and that whole method should only be called once?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
```diff
@@ -222,6 +222,14 @@ class S3BucketConnection(
...
+
+    /** `lazy val` to ensure that multi-part upload is completed at most once.
+      */
+    private lazy val finishResult: Future[Unit] = doFinish()
```
rautenrieth-da, line 231: Similarly here, `finish()` should only be called once now that we have the `finishingObject` guard, no? Do we need both protections?

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3UploadTest.scala
```diff
@@ -34,6 +40,27 @@ class S3UploadTest extends StoreTestBase with HasS3Mock {
...
+        _ <- o.upload(1, part)
+        _ <- o.finish()
+        checksumAfterFirstFinish <- bucketConnection.getChecksums(Seq("finish-twice"))
+        _ <- o.finish()
+        checksumAfterSecondFinish <- bucketConnection.getChecksums(Seq("finish-twice"))
+        content <- bucketConnection.readFullObject("finish-twice")
+      } yield {
+        checksumAfterFirstFinish.map(_.checksum) should not contain emptyDigest
+        checksumAfterSecondFinish shouldBe checksumAfterFirstFinish
+        new String(content.toArray, "UTF-8") shouldBe content
```
rautenrieth-da: Test is failing, probably because `content` is being shadowed.

```
"idempotency test" was not equal to ByteString(105, 100, 101, 109, 112, 111, 116, 101, 110, 99, 121, 32, 116, 101, 115, 116)
```

rautenrieth-da (approved):

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
isegall-da: If I had to choose only one, I'd keep this one and make this a safe utility to use and drop the protection on the call site. I'm a bit on the fence whether to drop that one though. It adds a tiny bit of complexity there, so I think I tend to agree that double protection is too much, and I'll drop the protection in GroupedWeightS3ObjectFlow

apps/scan/src/test/scala/org/lfdecentralizedtrust/splice/scan/store/bulk/S3UploadTest.scala
isegall-da: Ha, stupid last minute "cleanup" that I pushed untested, thanks.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/S3BucketConnection.scala
isegall-da: Naa, looking at it again, I convinced myself that protecting both makes more sense for future breakages. That way, the utility is a safe library, and the pekko state machine is future-proof against other side-effects of finishing an object that may be non-idempotent. I'll just improve the comments accordingly.

Issue #6904, opened 2026-08-21
Add SV UI support and tests for developmentFundManagerBlacklist


zheli:
Followup issue to #6793 

The SV config voting UI stubs `developmentFundManagerBlacklist` as `null` in `buildAmuletRulesConfigFromChanges`, since `developmentFundManagerBlacklist` is a list of parties, unlike the usual scalar config fields, so it needs dedicated UI and tests. 

For more context, see original comment [#6793](https://github.com/canton-network/splice/pull/6793/changes/BASE..60b4574e911d1b791294d0ed68cee51163539dea#r3828674261)

PR #6793, opened 2026-08-16
Add fund manager blacklist and dev fund coupon mint delay

zheli:
Implement  #6722

Issue #6722, opened 2026-08-07
Delayed collection of DevFund coupons


meiersi-da:
### Motivation  
Change DevFund coupon minting so that the coupons can only be collected after a delay (default 7 days) after they were created. This way the SVs have one week of time to observe and react to unexpected coupons being created by the DevFund manager party. The expected reaction is that the the SVs change the AmuletConfig such that  

* all coupon minting from the coupons created by the misbehaving fund manager is blocked and the coupons thus expire  
* a new fund manager is voted in that can take up operations  
      
### Work Breakdown  
* Daml:  
  * change `AmuletConfig`  
    * add field `developmentFundManagerBlacklist : Optional [Party]`  
    * add field `minDevelopmentFundMintingDelay : Optional RelTime` with a default of 0 days  
      * explicitly setting it is how the roll-out of this change is enforced, as that will block downgrades that lack the delay enforcement logic  
  * add field `mintAfter : Optional Time` to `DevelopmentFundCoupon`  
    * if set, ensure it is smaller than `expiresAt`  
  * change minting in `AmuletRules_Transfer` so that both the blacklist and the time constraint are enforced if set  
  * change `AmuletRules_AllocateDevelopmentFundCoupon` such that  
    * it takes a `mintAfter : Optional Time` parameter  
    * if `minDevelopmentFundMintingDelay` is set, then use `assertWithinDeadline` to validate that `mintAfter` at least `minDevelopmentFundMintingDelay` in the future  
  * test that  
    * blacklisting is enforced  
    * `mintAfter` is enforced on minting  
    * `minDevelopmentFundMintingDelay` is enforced on allocation  
* SV UI:  
  * adjust config voting UI to incorporate the new `AmuletConfig` fields  
* Wallet UI:  
  * adjust coupon display to show `mintAfter`  
  * adjust coupon allocation UI to allow setting `mintAfter`  
    * and set it to a default of `now` \+ `minDevelopmentFundCouponDelay` \+ 1h  
      * the extra 1h is the default prepare submission delay supported  
    * note: might need backend API changes – please investigate as part of doing this work  
* Validator app  
  * adjust `MintingDelegationCollectRewardsTrigger` (for external parties) to only collect `DevelopmentFundCoupons` after their minting delay  
  * adjust `TreasuryService` (for internal parties) to only collect `DevelopmentFundCoupons` after their minting delay  
  
### Roll-out  

[...]

PR #6944, opened 2026-08-24
Add UI support for fund manager blacklist and dev fund coupon mint delay

zheli:
Follow-up PR for #6793. It will fix the rest of the issue #6722, as well as #6904. It will close the milestone https://github.com/canton-network/splice/milestone/59

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
@@ -433,6 +433,101 @@ class DevelopmentFundCouponIntegrationTest
     }
   }
 
+  "Delaying the claiming of a development fund coupon until its mintAfter" in { implicit env =>
+    onboardWalletUser(aliceValidatorWalletClient, aliceValidatorBackend)
+    val sv1UserId = sv1WalletClient.config.ledgerApiUser
+    val bobParty = onboardWalletUser(bobWalletClient, bobValidatorBackend)
+    val beneficiary = bobParty
+    val initialUnclaimedDevelopmentFundCouponAmount = BigDecimal(SpliceUtil.damlDecimal(1000))
+    val developmentFundCouponAmount = BigDecimal(SpliceUtil.damlDecimal(40.0))
+    val expiresAt = CantonTimestamp.now().plus(Duration.ofDays(1))
+    val reason = "Bob has contributed to the Daml repo"
+    val mintingDelay = Duration.ofSeconds(30)
+
+    val bobUserName = bobWalletClient.config.ledgerApiUser
+    val bobMergeAmuletsTrigger =
+      bobValidatorBackend
+        .userWalletAutomation(bobUserName)
+        .futureValue
+        .trigger[CollectRewardsAndMergeAmuletsTrigger]
+
+    archiveExistingUnclaimedDevelopmentFundCoupons()
+    actAndCheck(
+      "Mint one unclaimed development fund coupon", {
+        createUnclaimedDevelopmentFundCoupon(
+          sv1ValidatorBackend.participantClientWithAdminToken,
+          sv1UserId,
+          initialUnclaimedDevelopmentFundCouponAmount,
+        )
+      },
+    )(
+      "The unclaimed development fund coupon is created",
+      _ => {
+        getUnclaimedDevelopmentFundCouponTotal(
+          sv1ValidatorBackend
+        ) shouldBe initialUnclaimedDevelopmentFundCouponAmount
+      },
+    )
+
+    val bobBalanceBefore = bobWalletClient.balance().unlockedQty
+    val (mintAfter, _) = setTriggersWithin(
+      triggersToPauseAtStart = Seq(bobMergeAmuletsTrigger)
+    ) {
+      actAndCheck(
+        "Allocate one development fund coupon that is not mintable yet", {
+          val mintAfter = CantonTimestamp.now().plus(mintingDelay)
+          aliceValidatorWalletClient.allocateDevelopmentFundCoupon(
+            beneficiary,
+            developmentFundCouponAmount,
+            expiresAt,
+            reason,
+            Some(mintAfter),
+          )
+          mintAfter
+        },
+      )(
+        "The coupon is created and carries the requested mintAfter",
+        allocatedMintAfter => {
+          val coupons = bobWalletClient.listActiveDevelopmentFundCoupons()
+          coupons should have size 1 withClue "bob coupons"
+          coupons.head.payload.mintAfter shouldBe java.util.Optional.of(
+            allocatedMintAfter.toInstant
+          )
+        },
+      )
+    }
+
+    clue("The coupon is left alone while its mintAfter is in the future") {
+      always(durationOfSuccess = 10.seconds) {
+        bobWalletClient
+          .listActiveDevelopmentFundCoupons() should have size 1 withClue "bob coupons before mintAfter"
+        bobWalletClient.balance().unlockedQty shouldBe bobBalanceBefore
+      }
+      CantonTimestamp
+        .now()
+        .isBefore(mintAfter) shouldBe true withClue "still before mintAfter"
+    }
+
+    clue("The coupon is collected once its mintAfter has passed") {
+      eventually(60.seconds) {
+        bobWalletClient
+          .listActiveDevelopmentFundCoupons() shouldBe empty withClue "bob coupons after mintAfter"
+        bobWalletClient.balance().unlockedQty shouldBe
+          (bobBalanceBefore + developmentFundCouponAmount)
+      }
+    }
+
+    clue("The collected coupon is listed in listDevelopmentFundCouponHistory as claimed") {
+      eventually() {
+        assertListDevelopmentFundCouponHistoryStatuses(
+          bobWalletClient,
+          beneficiary,
+          Seq(httpDef.ArchivedDevelopmentFundCoupon.Status.Claimed -> None),
+        )
+      }
+    }
+  }
+
   "Expiring a development fund coupon" in { implicit env =>
     val sv1UserId = sv1WalletClient.config.ledgerApiUser
     onboardWalletUser(aliceValidatorWalletClient, aliceValidatorBackend)
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
@@ -497,16 +497,6 @@ class DevelopmentFundCouponIntegrationTest
       )
     }
 
-    clue("The coupon is left alone while its mintAfter is in the future") {
-      always(durationOfSuccess = 10.seconds) {
-        bobWalletClient
-          .listActiveDevelopmentFundCoupons() should have size 1 withClue "bob coupons before mintAfter"
-        bobWalletClient.balance().unlockedQty shouldBe bobBalanceBefore
-      }
-      CantonTimestamp
-        .now()
-        .isBefore(mintAfter) shouldBe true withClue "still before mintAfter"
-    }
 
     clue("The coupon is collected once its mintAfter has passed") {
       eventually(60.seconds) {
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
@@ -433,7 +433,7 @@ class DevelopmentFundCouponIntegrationTest
     }
   }
 
-  "Delaying the claiming of a development fund coupon until its mintAfter" in { implicit env =>
+  "Minting a development fund coupon once its mintAfter has passed" in { implicit env =>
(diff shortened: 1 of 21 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
```diff
@@ -433,6 +433,101 @@ class DevelopmentFundCouponIntegrationTest
...
+  "Delaying the claiming of a development fund coupon until its mintAfter" in { implicit env =>
+    onboardWalletUser(aliceValidatorWalletClient, aliceValidatorBackend)
+    val sv1UserId = sv1WalletClient.config.ledgerApiUser
+    val bobParty = onboardWalletUser(bobWalletClient, bobValidatorBackend)
+    val beneficiary = bobParty
+    val initialUnclaimedDevelopmentFundCouponAmount = BigDecimal(SpliceUtil.damlDecimal(1000))
+    val developmentFundCouponAmount = BigDecimal(SpliceUtil.damlDecimal(40.0))
+    val expiresAt = CantonTimestamp.now().plus(Duration.ofDays(1))
+    val reason = "Bob has contributed to the Daml repo"
+    val mintingDelay = Duration.ofSeconds(30)
```
meiersi-da: When testing aspects related to timing we usually use a TimeBased integration test, which allowS changing the clock using `advanceTime` without busy waiting.

However, all the logic wrt minting delays is already tested at the Daml level. So what we need to test here is that a coupon with a minting delay gets eventually minted without raising unexpected errors. 

I'd suggest we set a delay of `10s`, which is long enough so it would result in a retry if the minting would be attempted, but short enough so its CI cost is not egregious. I'd then add 10s extra wait time also to the `eventually` that checks whether the coupon gets minted.

For bonus points, I'd change the minting code once to not respect the mint after, and check what kind of errors are raised in the logs (I expect retries); and then check again with the proper filtering that these are gone.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
```diff
@@ -433,6 +433,101 @@ class DevelopmentFundCouponIntegrationTest
...
+
+    val bobUserName = bobWalletClient.config.ledgerApiUser
+    val bobMergeAmuletsTrigger =
+      bobValidatorBackend
+        .userWalletAutomation(bobUserName)
+        .futureValue
+        .trigger[CollectRewardsAndMergeAmuletsTrigger]
+
+    archiveExistingUnclaimedDevelopmentFundCoupons()
```
meiersi-da: this looks fishy. We usually do test isolation by relying on fresh parties being allocated for new test instances. So if you scope your queries wrt coupons being created to the test-instance-specific party, then you have proper isolation.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
```diff
@@ -433,6 +433,101 @@ class DevelopmentFundCouponIntegrationTest
...
+    clue("The coupon is left alone while its mintAfter is in the future") {
+      always(durationOfSuccess = 10.seconds) {
+        bobWalletClient
+          .listActiveDevelopmentFundCoupons() should have size 1 withClue "bob coupons before mintAfter"
+        bobWalletClient.balance().unlockedQty shouldBe bobBalanceBefore
+      }
+      CantonTimestamp
+        .now()
+        .isBefore(mintAfter) shouldBe true withClue "still before mintAfter"
+    }
```
meiersi-da: I don't think we need to test this, as that's already tested at the Daml level.
```suggestion
```

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
```diff
@@ -433,6 +433,101 @@ class DevelopmentFundCouponIntegrationTest
...
+
+    clue("The coupon is collected once its mintAfter has passed") {
+      eventually(60.seconds) {
```
meiersi-da: This is very expensive! Let's avoid it.

meiersi-da: With the above suggestion, this would be 30s, as the `eventually` has a 20s default and we add the 10s extra from the delay to get some extra protection from flakes.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
```diff
@@ -659,6 +659,119 @@ class WalletMintingDelegationTimeBasedIntegrationTest
       }
     }
 
+    "not collect a development fund coupon before its mintAfter" in { implicit env =>
```
meiersi-da: For the same reason as explained above, I don't think we need an additional test, which increases both CI and maintenance cost. We just want to check that the coupons are ignored when their time is not due, and collected afterwards.

I'd suggest to restructure the existing test to do so as follows:
1. create two dev fund coupons: one with `mintAfter = None` and another one with `mintAfter = Some 24h`.
2. adjust https://github.com/bitdynamics-ab/splice/blob/860446c52b0247b6c3720bc7b7949300aad85fad/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala#L591-L612 to expect to not collect the dev fund coupon that  has a `mintAfter`
3. add an `advanceTime(25h)` afterwards and then check that the coupon with `mintAfter` is also collected.

apps/sv/frontend/src/__tests__/utils/buildAmuletRulesConfigFromChanges.test.ts
```diff
@@ -346,6 +346,50 @@ describe('buildAmuletRulesConfigFromChanges', () => {
...
+        fieldName: 'minDevelopmentFundMintingDelay',
+        label: 'Min Development Fund Minting Delay',
+        currentValue: '604800000000',
+        newValue: '',
+      },
+    ];
+
+    const result = buildAmuletRulesConfigFromChanges(changes);
+
+    expect(result.developmentFundManagerBlacklist).toEqual([]);
```
meiersi-da: wdyt about setting it to `null` if it is emptied?

apps/sv/frontend/src/utils/buildAmuletRulesConfigFromChanges.ts
```diff
@@ -104,8 +110,13 @@ export function buildAmuletRulesConfigFromChanges(
         ? null
         : { microseconds: externalPartyConfigStateTickDuration },
     transferPreapprovalBaseDuration: null,
-    developmentFundManagerBlacklist: null,
-    minDevelopmentFundMintingDelay: null,
+    // The frontend will never send null for developmentFundManagerBlacklist, an empty list is sent as `Some []`
+    // so that we can distinguish DSO that have not upgraded yet (they will have None for the field)
```
meiersi-da: we do we need to distinguish that? My expectation was that forcing the upgrade happens by setting an actual minting delay.

apps/wallet/frontend/src/__tests__/useDevelopmentFundAllocationForm.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+    vi.useFakeTimers();
+    vi.setSystemTime(openedAt);
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  // Happy
+  test('treats an absent minting delay as a zero delay and still prefills mintAfter', () => {
```
meiersi-da: Note that this forces the recipient to have upgraded. We generally try to avoid this. How difficult would it be to not set `mintAfter` if no `mintingDelay` is set?

apps/wallet/frontend/src/components/DevelopmentFundAllocation.tsx
```diff
@@ -124,6 +128,26 @@ const DevelopmentFundAllocation: React.FC = () => {
                   }}
                 />
               </Stack>
+
+              <Stack spacing={1} sx={{ flex: 1 }}>
```
meiersi-da, line 133: Please add screenshots of all relevant states to the PR description to simplify review, as done here: https://github.com/canton-network/splice/pull/6387

meiersi-da (commented):
Thanks a lot. Direction looks good.

apps/sv/frontend/src/utils/buildAmuletRulesConfigFromChanges.ts
zheli: > we do we need to distinguish that? My expectation was that forcing the upgrade happens by setting an actual minting delay.

Yeah, you are right. It is better without the distinction. Fixing it now

apps/sv/frontend/src/__tests__/utils/buildAmuletRulesConfigFromChanges.test.ts
zheli: Fixed

apps/wallet/frontend/src/__tests__/useDevelopmentFundAllocationForm.test.tsx
zheli: Fixed. mintAfter is optional now if minDelay is not set.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
zheli: Good point! Fixing it now.

zheli: @meiersi-da should we fix the other tests that should not use `archiveExistingUnclaimedDevelopmentFundCoupons()` as well?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
zheli: Shortened the delay. I will leave the bonus point to the last.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
zheli: Merged. Thanks!

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
zheli: Fixed.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/WalletMintingDelegationTimeBasedIntegrationTest.scala
zheli: Merged mintAfter checks to `collect rewards for all coupons owned by the beneficiary` test

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
zheli: I also ran the integration tests without the mintAfter filter and the test failed as expected

meiersi-da (commented):
Thx a lot. This looks good. I'll approve once the release notes and the screenshots are also here.

docs/src/release_notes_upcoming.rst
```diff
@@ -67,9 +67,27 @@ release-notes:: Upcoming
           that omit ``mintAfter``. Coupons allocated before that vote have no ``mintAfter``.
           A beneficiary can mint those coupons with no delay.
 
+    - Wallet UI
+
+        - The development fund allocation form now has a ``Mint After`` field, which sets the
+          earliest time at which the beneficiary can mint the coupon. When the field is 
+          optional, the coupons are allocated without a mint-after time as before.
```
meiersi-da:
```suggestion
          earliest time at which the beneficiary can mint the coupon. While
          the ``minDevelopmentFundMintingDelay`` stays unset in the ``AmuletConfig``, the field is 
          optional and the coupons are allocated without a mint-after 
          constraint.
```

meiersi-da (approved):
Thank you for this great work. The screenshots were very helpful to quickly get a feel for the feature. Much appreciated!

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/DevelopmentFundCouponIntegrationTest.scala
meiersi-da: if it was already there, then leave it, unless you see a reason of why it would cause a flake. Even then, I'd suggest to do that in a separate PR.

PR #6953, opened 2026-08-25
Public Sequencer: Add automation to revoke ParticipantSynchronizerPermission based on ValidatorUnpermision. Extend integration test to to check both temporary revocation and permanent revocation

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/6938

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
@@ -11,7 +11,6 @@ import org.lfdecentralizedtrust.splice.scan.admin.api.client.commands.HttpScanAp
 import org.lfdecentralizedtrust.splice.util.*
 
 import java.time.Instant
-import scala.concurrent.duration.*
 import java.util.Optional
 import com.digitalasset.canton.data.CantonTimestamp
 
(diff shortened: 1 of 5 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
```diff
@@ -120,5 +128,78 @@ class PermissionedSynchronizerIntegrationTest
...
+      }
+    }
+    def manuallyUnpermissionValidator(
+        participantId: String,
+        loginAfter: Option[Instant],
+        revoked: Boolean,
+    ): Unit = {
+      val dsoParty = sv1Backend.getDsoInfo().dsoParty
+
+      Seq(sv1Backend, sv2Backend, sv3Backend).foreach { sv =>
```
moritzkiefer-da: I don't understand why you need this, why not just create a vote? You're already making 3 submissions here so there isn't really a meaningufl performance difference

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
```diff
@@ -120,5 +128,78 @@ class PermissionedSynchronizerIntegrationTest
...
+        eventuallySucceeds(timeUntilSuccess = 40.seconds, maxPollInterval = 1.second) {
+          val svParty = sv.getDsoInfo().svParty
+          val dsoRules = sv.appState.dsoStore.getDsoRules().futureValue
+
+          clue(s"${sv.participantClient.name} votes for UnpermissionValidator(revoked=$revoked)") {
```
moritzkiefer-da: how do I repermission them after this?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExecuteConfirmedActionTrigger.scala
```diff
@@ -272,6 +273,32 @@ class ExecuteConfirmedActionTrigger(
...
               )
               .map(_.isEmpty)
+          case unpermissionAction: SRARC_UnpermissionValidator if config.permissionedSynchronizer =>
+            // MergeUnpermissionValidatorContractsTrigger merges the new contract with any previous contract,
+            // hence we have to consider the merged contract state when calculating isStaleAction
+            val payload = unpermissionAction.dsoRules_UnpermissionValidatorValue
+            val proposedIsRevoked = payload.revoked
+            val proposedLoginAfter = payload.loginAfter.toScala
+
+            store.listValidatorUnpermissions(payload.participantId).map { contracts =>
```
moritzkiefer-da: I'm confused, why do we need a staleness check for ExecuteConfirmedAction? I was expecting this gets created from a vote? What user flow do you have in mind where we'll have confirmation contracts for this?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/automation/delegatebased/ExecuteConfirmedActionTrigger.scala
pasindutennage-da: Fixed

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
pasindutennage-da: @moritzkiefer-da 

Repermission is automatic if it is `loginAfter` based `ValidatorUnpermission`. Validator will get `PERMISSIONED_DENIED` until `loginAfter`, but then will work normally

 Repermission with `revoke` based `ValidatorUnpermission` is more subtle, and needs more work, IMO
 
 1. Validator X is revoked permeently, and SVs don't want it ever again. However, if someone buys traffic for X, then `GrantValidatorPermissionTrigger` will grant topology again, and hence bypass the unpermissioning. For this, we need to update `GrantValidatorPermissionTrigger` to not grant topology, if there is a matching `ValidatorUnpermission`

2. Allow rejoining after permenent revocation - to implement this we will need another seperate vote action where SVs vote to archive `ValidatorUnpermission` contract. Only then someone can buy member traffic to submit topology from SV.

And anything else you have in mind for repermission?

pasindutennage-da: @moritzkiefer-da 
Also, to simulate Repermission after `loginAfter` based `ValidatorUnpermission`, we need to add a TimeBased(simulated time)Integration Test. should we?

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
```diff
@@ -120,5 +128,89 @@ class PermissionedSynchronizerIntegrationTest
...
+        },
+      )(
+        "vote request has been created",
+        _ => sv1Backend.listVoteRequests().filter(_.payload.action == action).head,
+      )
+
+      Seq(sv2Backend, sv3Backend, sv4Backend).foreach { sv =>
+        clue(s"${sv.participantClient.name} accepts the vote request") {
+          eventuallySucceeds() {
+            sv.castVote(
```
moritzkiefer-da, line 206: isn't that racy? the vote may already be archived before sv4 can cast, you only need 3/4 votes

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
moritzkiefer-da: > Validator X is revoked permeently, and SVs don't want it ever again.

I don't think that is gonna fly. In some cases, the SVs just want to say "unpermission this node" without thinking about a time but maybe later change their mind. You can argue they shoudl set loginAfter to very far in the future but that seems like a mess to me. I guess maybe you can hide it in the UI but I'd say if we do have the option to fully revoke we also need to have the option to revert that.

moritzkiefer-da (approved):
thanks, I think we'll need a bit more work for repermissioning but let's do that in another pr.

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerIntegrationTest.scala
pasindutennage-da: Fixed

PR #6958, opened 2026-08-25
TestTokenV2SettlementIntegrationTest prevent validators from using shutting-down canton

OriolMunoz-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/9758

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TestTokenV2SettlementIntegrationTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TestTokenV2SettlementIntegrationTest.scala
@@ -181,7 +181,7 @@ class TestTokenV2SettlementIntegrationTest
         bobValidatorBackend,
         splitwellValidatorBackend,
         ttAdminValidator,
-      ).reverse.foreach(_.stop())
+      ).foreach(_.stop())
     }
   }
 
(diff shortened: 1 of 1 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TestTokenV2SettlementIntegrationTest.scala
```diff
@@ -157,6 +157,34 @@ class TestTokenV2SettlementIntegrationTest
...
+    try {
+      f()
+    } finally {
+      Seq(
+        sv1ValidatorBackend,
+        aliceValidatorBackend,
+        bobValidatorBackend,
+        splitwellValidatorBackend,
+        ttAdminValidator,
+      ).reverse.foreach(_.stop())
```
moritzkiefer-da: why would you reverse a hardcoded list? At that point just write it the other way around? I also don't understand why you even care about the order though.

moritzkiefer-da (approved):
thanks

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/TestTokenV2SettlementIntegrationTest.scala
OriolMunoz-da: same order as they were initialized but in reverse, which clearly I failed in making obvious
either way as you said, the order shouldn't matter after thinking about it

PR #6978, opened 2026-08-26
Port from Canton: give native protoc plugins a stable launcher path

OriolMunoz-da:
Fixes https://github.com/DACH-NY/cn-test-failures/issues/4700 (hopefully)

The file is copied verbatim from Canton's repo

[scratch deployment](https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/79877/details)

### DIFF
--- a/project/ProtocNixPlugin.scala
+++ b/project/ProtocNixPlugin.scala
@@ -1,0 +1,71 @@
+// Copyright (c) 2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import sbt.*
+import sbt.Keys.*
+
+/** Points `protocGenerate` at a stable launcher for every native protoc plugin.
+ *
+ * When `NIX_CC` is set (it always is, since sbt runs inside the nix shell) sbt-protoc wraps each
+ * native plugin binary in a fresh `/tmp/nix<random>` script per task run and passes that path as
+ * `--plugin=`. The path is part of the task's cache key, so codegen and the doc rendering rerun on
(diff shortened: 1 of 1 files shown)


project/ProtocNixPlugin.scala
```diff
@@ -0,0 +1,71 @@
...
+/** Points `protocGenerate` at a stable launcher for every native protoc plugin.
+  *
+  * When `NIX_CC` is set (it always is, since sbt runs inside the nix shell) sbt-protoc wraps each
+  * native plugin binary in a fresh `/tmp/nix<random>` script per task run and passes that path as
+  * `--plugin=`. The path is part of the task's cache key, so codegen and the doc rendering rerun on
+  * every build even when no proto changed, and the scripts pile up in `/tmp` for the life of the
+  * sbt server. Writing the launcher ourselves under a digest-named path keeps the key stable, and
+  * the `.sh` suffix stops sbt-protoc from wrapping it again.
+  */
+object ProtocNixPlugin extends AutoPlugin {
```
moritzkiefer-da, line 16: is this magically picked up? I thought we need to configure it somewhere to actually be used in the sbt config?

moritzkiefer-da (approved):
thx

project/ProtocNixPlugin.scala
OriolMunoz-da: I had the same doubt :smile: I added a println right below this line and in nativePluginArtifact and they got executed
afaiu it's because of the AutoPlugin extends:
> An AutoPlugin defines a group of settings and the conditions where the settings are automatically added to a build (called "activation"). The requires and trigger methods together define the conditions, and a method like projectSettings defines the settings to add.

PR #6969, opened 2026-08-26
Add `globalSynchronizerHealth` alerts

martinflorian-da:
See https://github.com/DACH-NY/canton-network-internal/issues/6371

Regular alerts for now, plan is to promote at least some of them to sev-2 and/or sev-1 after some testing.

Preview with more or less the queries of the alerts:

[Discarded requests / replays](https://grafana.global.canton.network.digitalasset.com/explore?schemaVersion=1&panes=%7B%22siq%22:%7B%22datasource%22:%22prometheus%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22sum%28rate%28daml_sequencer_block_events_total%7Bnamespace%3D%5C%22sv-1%5C%22,type%3D%5C%22send-confirmation-request%5C%22%7D%5B$__rate_interval%5D%29%29%20-%20sum%28rate%28daml_mediator_requests_total%7Bnamespace%3D%5C%22sv-1%5C%22%7D%5B$__rate_interval%5D%29%29%5Cn%5Cn%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22prometheus%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22from%22:%22now-30d%22,%22to%22:%22now%22%7D,%22compact%22:false%7D%7D&orgId=1): :point_left: also adds this as panels on the Global Synchronizer Utilization dashboard

image

[...]

PR #6992, opened 2026-08-27
Exclude duplicate requests for confirmation request failure alert

martinflorian-da:
And also the dashboard panel, and add a new panel that shows duplicate confirmation requests.

On TestNet it would look like this (except for the vertical alerting bar and :broken_heart:):

image

Follow-up to #6969


moritzkiefer-da (approved):
thanks!

PR #6995, opened 2026-08-27
Add new ACS snapshot endpoints using opaque pagination tokens

### LINKED ISSUE #6884: Add new ACS snapshot endpoints using `after: (Opaque)String` pagination tokens
The [pagination token is defined as a Long](https://github.com/canton-network/splice/blob/306014bc81fa2045c60c7fee50b599f8218bf0f4/apps/scan/src/main/openapi/scan.yaml#L2810). This worked fine with the first implementation, but we're finding that:
* The query plan is less than ideal
* Maintaining backwards compatibility with that in the new implementation also makes the query plan equally bad

We would like to:
* Deprecate the existing ACS snapshot endpoints that define `after` like that
* Replace it with `after: String` which would be entirely opaque to the user
Which might (because pending to test) allow us to have a better query plan here.

Our plan would be:
* 0.7.x: Introduce a v2 API that uses after: String 
* 0.7.x: Deprecate the v1 API that uses after: Long
* 0.9.0: Remove the v1 API

OriolMunoz-da:
Fixes https://github.com/canton-network/splice/issues/6884

### DIFF
--- a/apps/scan/src/main/openapi/scan.yaml
+++ b/apps/scan/src/main/openapi/scan.yaml
@@ -667,6 +667,37 @@ paths:
         "500":
           $ref: "../../../../common/src/main/openapi/common-external.yaml#/components/responses/500"
 
+  /v2/state/acs:
+    post:
+      tags: [ external, scan ]
+      x-jvm-package: scan
+      operationId: "getAcsSnapshotAtV2"
+      description: |
+        Returns the ACS in creation date ascending order, paged, for a given migration id and record time.
+        Unlike /v0/state/acs, every contract is identified by an (optional) update_id
+        (as opposed to the event ID in /v0/state/acs, which was not BFT-safe).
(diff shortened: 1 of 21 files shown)


rautenrieth-da (approved):
Thanks!

apps/scan/src/main/openapi/scan.yaml
```diff
@@ -719,10 +754,14 @@ paths:
 
   /v1/holdings/state:
     post:
+      deprecated: true
       tags: [external, scan]
       x-jvm-package: scan
       operationId: "getHoldingsStateAtV1"
       description: |
+        Deprecated. Please use /v2/state/acs instead.
```
```suggestion
        Deprecated. Please use /v2/holdings/state instead.
```

OriolMunoz-da: yikes, thanks

ray-roestenburg-da (approved):
Thanks!

PR #7011, opened 2026-08-28
Revert "Backport PR #6902 to release-line-0.7.0 (#6903)"

martinflorian-da:
[static]

Currently this causes the watchdog to be deployed on the legacy migration on DevNet, which just causes a lot of noise.

Revert this seems like the easiest way to denoise until we're ready to take down those nodes.

See https://daholdings.slack.com/archives/C09LHGB8GUB/p1787881032667409

This reverts commit c023a98d94b06c2cfd203ce9476b272f8816f784.


pasindutennage-da (approved):
@martinflorian-da 

For my understanding: 0.7.1 is on main

Why do we bother about 0.7.0? any place where this still runs?

PR #7017, opened 2026-08-28
Helm charts hardening: don't mount k8s API tokens by default

martinflorian-da:
Part of https://github.com/DACH-NY/canton-network-internal/issues/6745

Nothing we deploy uses the Kubernetes API from inside a pod.

[static]


isegall-da (approved):
Nice, thank you

nicu-da (approved):
One would've imagined this is the default 🤷

PR #7018, opened 2026-08-28
Fix non-exhaustive match

rautenrieth-da:
Found during an unrelated review. Rewrites the partitioning to be exhaustive.

### DIFF
--- a/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
+++ b/apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
@@ -382,16 +382,29 @@ class BftScanConnection(
         .map(result => connection -> result)
     )
   } yield {
-    val withData = results.collect {
-      case (connection, BftScanConnection.SuccessfulResponse(Some(info))) => connection -> info
-    }.toMap
-    val withoutData = results.collect {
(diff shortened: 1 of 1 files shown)


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -382,16 +382,29 @@ class BftScanConnection(
...
-      case (connection, BftScanConnection.HttpFailureResponse(_, _)) => connection
-      case (connection, BftScanConnection.ExceptionFailureResponse(_)) => connection
-    }.toSet
+    val (withData, withoutData, unknownStatus) =
+      results.foldLeft(
+        (
+          Map.empty[SingleScanConnection, SourceMigrationInfo],
+          Set.empty[SingleScanConnection],
+          Set.empty[SingleScanConnection],
+        )
```
stephencompall-DA: I suggest using `partitionMap` twice instead. Changes here: 6b64eeec1215de5cc86fd2ef430a0e3831c47d82

stephencompall-DA (approved):

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
rautenrieth-da: Cherry-picked your change, thanks!

PR #7022, opened 2026-08-28
Public Sequencer: Add SV UI changes to trigger unpermission votes + tests  

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/7012

image

image

image

image

image

image


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/integration/tests/PermissionedSynchronizerSvFrontendIntegrationTest.scala
```diff
@@ -0,0 +1,344 @@
+package org.lfdecentralizedtrust.splice.integration.tests
+
```
pasindutennage-da: @moritzkiefer-da 

The test fails in the post-test logs, due to the following.

Permanently closing sequencer subscription due to error::  GrpcPermissionDeniedError(Request failed for sequencer.\n  GrpcRequestRefusedByServer: PERMISSION_DENIED/Authentication

Token refresh encountered error: Status***code=PERMISSION_DENIED, description=Member PAR::aliceValidator::1220a00154


Request failed for server-SEQ::sv2::1220795a9d215bd39a2d34ae8a1f49fd529a0586be2dec53405fd1613086b7b3ff77-0.\n  GrpcRequestRefusedByServer: PERMISSION_DENIED/Authentication

:"Token refresh encountered error: Status***code=PERMISSION_DENIED, description=Member PAR::aliceValidator::1220a00154c4... access is disa

These are arising once we unpermission alice/bob. To reduce noise, I stop alice and bob, however, their participants keep running and cause these logs.

I faced simmilar stuff in the backend test, and i first tried to supress logs, which didn't work. Then I added them to canton-ignore file (like 5-6 lines). Is that the correct approach (if so I will add these to canton-ignore as well)

moritzkiefer-da: Usually the cleanest option is to use `withCanton` for this type of stuff so you can actually stop the participant. The problem with ignores is that you may end up the next test after that if things fail which gets quite confusing.

apps/sv/frontend/src/utils/config.tsx
```diff
@@ -33,6 +34,7 @@ const configScheme = z.object({
   services: z.object({
     sv: serviceSchema,
   }),
+  permissioned: z.boolean().optional(),
```
moritzkiefer-da, line 37:
```suggestion
  permissionedSynchronizer: z.boolean().optional(),
```

Wondering if we should also avoid the permissioned naming here but I also am struggling to come up with another good config name. publicSynchronizer? Maybe our head of name bikeshedding @martinflorian-da has an idea.

apps/sv/frontend/src/components/forms/UnpermissionValidatorForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+    expiryDate: initialExpiration.format(dateTimeFormatISO),
+    effectiveDate: {
+      type: 'custom',
+      effectiveDate: initialEffectiveDate.format(dateTimeFormatISO),
+    },
+    url: '',
+    summary: '',
+    participantId: '',
+    revoked: false,
+    loginAfter: dayjs().add(1, 'day').format(dateTimeFormatISO),
```
moritzkiefer-da: why is a day a sensible default?

apps/sv/frontend/src/components/forms/UnpermissionValidatorForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+                      <Checkbox
+                        checked={field.state.value}
+                        onChange={e => field.handleChange(e.target.checked)}
+                        id="unpermission-validator-revoked"
+                      />
+                    }
+                    label={<Typography fontWeight="bold">Permanent Revocation</Typography>}
+                  />
+                  <Typography variant="body2" color="text.secondary" sx={{ ml: 4, mt: -0.5 }}>
+                    If checked, the validator is permanently unpermissioned.
```
moritzkiefer-da: it's not permanent but it needs an sv vote to repermission them

moritzkiefer-da (changes requested):
Thanks, the UI changes look fine. However following up on @martinflorian-da's testing comments from yesterday I do think it's worth tweaking that a bit.

Generally the rule of thumb is that we want each API endpoint to be integration tested once and the rest should be vitests. There is no new API endpoint here so imho there is no reason to integration test anything.

And even for cases where we do want to integration test UI stuff I think we need to carefully look at what we actually test.

E.g., for this particular test your current test is a 4 SV test that actually unpermissions a node. But the UI doesn't really care that you have 4 SVs nor does it care that a node actually gets unpermissioned. So an integration test would just create the vote and then check that the vote is created on-ledger.

apps/sv/frontend/src/utils/config.tsx
martinflorian-da: I feel attacked...

IMHO if we want to reduce the usage of that word also in the code we should do a bigger sweep and not just stop at this config. Using different words here and there adds confusion that also has a cost.

That said, the technical thing we're actually using is not even called permissioned, but `RestrictedOpen`. Not sure if `restrictedSynchronizer` is a lot better than `permissionedSynchronizer` (I do think it is slightly better), but that would be my best idea this morning.

So IMO:

- let's keep whatever naming variant with permissioned you prefer here
- make an issue to grep for permissioned in the code base and bulk-rename to something better we like more; ideally before merging the feature branch (or just decide that we won't do it)

pasindutennage-da: Thanks, added to tracking issue

apps/sv/frontend/src/components/forms/UnpermissionValidatorForm.tsx
pasindutennage-da: Set to 10 days, still arbitary though

.github/workflows/build.scala_test.yml
```diff
@@ -26,6 +26,10 @@ on:
         type: boolean
         required: false
         default: false
```
pasindutennage-da, line 28: because I removed the front end test, this change is redundent; however, i would like to keep it, because in future we will need to have a mechanism to pass args to frontend test.

apps/sv/frontend/src/__tests__/governance/forms/unpermission-validator-form.test.tsx
```diff
@@ -0,0 +1,339 @@
...
+    await user.type(summaryInput, 'Summary of the proposal');
+
+    const urlInput = screen.getByTestId('unpermission-validator-url');
+    await user.type(urlInput, 'https://example.com');
+
+    const revokedCheckbox = screen.getByRole('checkbox');
+    await user.click(revokedCheckbox);
+
+    const submitButton = screen.getByTestId('submit-button');
+    await user.click(actionInput); // using this to trigger the onBlur event which triggers the validation
```
moritzkiefer-da, line 370: I would expect that we assert that the mock for the submission gets hit?

.github/workflows/build.scala_test.yml
moritzkiefer-da: ime that rarely works out well, if you have untested code it just is guaranteed to break and it ends up being more confusing than helpful when you need it next. So I'd lean towards removing it.

build.sbt
```diff
@@ -2452,7 +2452,7 @@ updateTestConfigForParallelRuns := {
   def isPreflightIntegrationTest(name: String): Boolean = name.contains("PreflightIntegrationTest")
   def isEnterpriseIntegrationTest(name: String): Boolean = name.contains("Enterprise")
   def isPermissionedSynchronizerTest(name: String): Boolean =
-    name.contains("PermissionedSynchronizer")
+    name.contains("PermissionedSynchronizer") && !isFrontEndTest(name)
```
moritzkiefer-da: do you still need this?

moritzkiefer-da (approved):
nice thanks!

.github/workflows/build.scala_test.yml
pasindutennage-da: Fixed

build.sbt
pasindutennage-da: Fixed

apps/sv/frontend/src/__tests__/governance/forms/unpermission-validator-form.test.tsx
pasindutennage-da: Fixed

PR #7028, opened 2026-08-31
Add ability to override rate limiter per IP CIDR

nicu-da:
[ci]

fixes https://github.com/DACH-NY/canton-network-internal/issues/6672

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/config/SpliceConfigTest.scala
+++ b/apps/app/src/test/scala/org/lfdecentralizedtrust/splice/config/SpliceConfigTest.scala
@@ -100,6 +100,11 @@ class SpliceConfigTest extends AsyncWordSpec with BaseTest {
           |  ip-overrides {
           |    "10.0.0.0/8" = { rate-per-second = 100 }
           |    "192.0.2.0/24" = { enabled = false, rate-per-second = 0 }
+          |    "10.1.2.3" = { rate-per-second = 200 }
+          |    "2001:db8::/32" = { rate-per-second = 300 }
+          |    "2001:db8:1::/48" = { enabled = false, rate-per-second = 0 }
+          |    "2001:db8:2:3:4:5:6:7" = { rate-per-second = 400 }
+          |    "::/0" = { rate-per-second = 500 }
           |  }
           |}
           """.stripMargin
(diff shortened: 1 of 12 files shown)


apps/app/src/test/scala/org/lfdecentralizedtrust/splice/config/SpliceConfigTest.scala
```diff
@@ -86,6 +86,72 @@ class SpliceConfigTest extends AsyncWordSpec with BaseTest {
     }
   }
 
+  "rate limiting config" should {
+
+    def perClientIpOf(cfg: SpliceConfig) =
+      cfg.scanApps.values.headOption.value.parameters.rateLimiting.global.perClientIp
+
+    "parse the per client IP CIDR overrides" in {
```
martinflorian-da, line 94: Shall we also add some IPv6 test cases to demonstrate this works already here?

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/IpCidrRateLimits.scala
```diff
@@ -0,0 +1,101 @@
...
+import java.net.InetAddress
+import scala.util.Try
+
+/** An IP network in CIDR notation, e.g. `10.0.0.0/8` or `2001:db8::/32`. A value without a prefix
+  * length denotes a single host, i.e. a `/32` (IPv4) or `/128` (IPv6) network.
+  *
+  * @param network
+  *   the network address with all host bits zeroed out, 4 bytes for IPv4 and 16 bytes for IPv6.
+  */
+final case class IpCidr(network: Seq[Byte], prefixLength: Int) {
```
martinflorian-da: A `Seq[Byte]` doesn't sound like a very efficient way to store an IP address and do operations on it... Have you considered reusing `InetAddress` (since you already import that?) and implementing mask as an appropriate operation on `.getAddress`?

More generally I wonder how there isn't some Scala library for IP addresses / ranges that we could use. Implementing IP address parsing and CIR matching does not sound like something we should be doing in splice.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/http/HttpRateLimiterTest.scala
```diff
@@ -434,8 +434,88 @@ class HttpRateLimiterTest extends AnyWordSpec with BaseTest with ScalatestRouteT
...
 
-  private def perClientIp(ratePerSecond: Double): PerAttributeRateLimitConfig =
-    PerAttributeRateLimitConfig(limit = SpliceRateLimitConfig(ratePerSecond = ratePerSecond))
+  "the per client IP CIDR overrides" should {
+
+    "apply the custom limit to clients of a matching network" in {
+      withRoutes(
+        globalPerClientIp = perClientIp(
+          1,
+          cidrOverrides = Map("10.0.0.0/8" -> SpliceRateLimitConfig(ratePerSecond = 5)),
```
martinflorian-da: I think we should add *some* test that involves IPv6.

The reality is that we're unlikely to use the IPv6 codepaths soon anyway. But we do have them, so might as well have some basic testing to confirm they are not broken?

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/util/SpliceRateLimiterTest.scala
```diff
@@ -97,35 +97,33 @@ class SpliceRateLimiterTest
       }
     }
 
-  }
-
-  "the per attribute rate limiter" should {
-
-    "gate the per attribute limit on the overall limiter being enabled" in {
```
martinflorian-da, line 104: Is this deliberate? We don't want to be able to disable via a single flag anymore?

martinflorian-da (approved):
Thanks!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/util/IpCidrRateLimits.scala
nicu-da: Switched to more efficient binary ops.
unfortunately could only find the SubnetUtils in apache commons but it doesn't work out for our usecase, and I would not involve any third party library just for this.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/util/SpliceRateLimiterTest.scala
nicu-da: Yes. 
Did this because the behavior gets quite confusing if you explicitly enable the ip rate limiter but disable the global one, especially considering that the per ip one is disabled in most cases. 
Not feeling strongly about going back to the old behavior, but felt like that coupling doesn't really make sense

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/config/SpliceConfigTest.scala
nicu-da: Expanded all the testing with ipv6 :robot:

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/util/SpliceRateLimiterTest.scala
martinflorian-da: I don't think I have a strong opinion either, if it's deliberate and we're aware of it SGTM to keep as is :+1:

PR #7042, opened 2026-09-01
Improve istio rate limits

nicu-da:
Previosly we had the global limit and the per ip limit, but a request would consume only from one bucket which made us vulnerable

We now have the following limits, mimicking the apps (in order of being applied, short circuiting):

- ability to configure global limit per endpoint per client ip (with overrides)
- global limit per client ip (with overrides)
- ability to configure global limit per endpoint for all the clients
- global limit for all clients, always consumed

these 4 rate limiters all apply at the same time and all must hold

Extend metrics and logs:
- logs now inlude the external address and the masked address
- metrics are distinguished for each filter and we have a relabelng rule to add a label with the filter name so we know which ones are enforced

[static]

fixes https://github.com/DACH-NY/canton-network-internal/issues/2115

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


cluster/configs/shared/scratch-default-synchronizer-migration.yaml
```diff
@@ -4,6 +4,8 @@ synchronizerMigration:
   active:
     id: 0
     version: "local"
+    sequencer:
```
nicu-da, line 7: enable by default for scratches to not deal with cometbft

martinflorian-da: how was that not the default already 😕 thanks!

martinflorian-da (approved):
Thank you!

I couldn't find anything concerning and I think fully understanding will take me hours... so let's merge and let the experience on the clusters show us what we got wrong (if anything).

PR #7043, opened 2026-09-01
Don't return featured app right in token standard contexts from scan

moritzkiefer-da:
[ci]

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/util/SvUtil.scala
```diff
@@ -241,25 +248,37 @@ object SvUtil {
...
+      defaultSynchronizerNodeConfigLimits,
+      1024, // maxTextLength
+      defaultDsoDecentralizedSynchronizerConfig(
+        synchronizerId,
+        acsCommitmentReconciliationInterval,
+      ), // decentralizedSynchronizerConfig
+      Optional.empty(), // nextScheduledSynchronizerUpgrade
+      voteCooldownTime.map(t => new RelTime(t.duration.toMicros)).toJava,
+      Optional.empty(), // nextScheduledLogicalSynchronizerUpgrade
+      // We silently drop a value that is set when it's not supported as making it an error doesn't work well with setting it as the default.
```
meiersi-da:
```suggestion
      // We silently drop all switchover values when switchover is not supported as making it an error doesn't work well with setting it as the default.
```

meiersi-da: Is this what we are doing?

docs/src/release_notes_upcoming.rst
```diff
@@ -31,6 +31,15 @@ release-notes:: Upcoming
           ``/v0/dso`` endpoint. The scan is configured via the new ``.joinWithKeyOnboarding.sponsorScanUrl`` Helm value.
           SVs who set the ``.joinWithKeyOnboarding`` key config must set it before upgrading.
 
+    - Scan App
+
+        - *Breaking* Scan will no longer return featured app rights as part of the choice context for CC transfers and allocations
+          once ``no-featured-app-choice-context`` has been set in ``svOperationsSwitchOverTimes`` and the switchover time has been reached.
+          This change was made to avoid apps accidentally not complying with the marker guidance set by the Canton foundation. If you
+          really do intend to feature a CC transfer, you can query for the contract through ``/api/scan/v0/featured-apps/{provider_party_id}`` on Scan
+          and add it to the choice context under the ``featured-app-right`` choice context key.
```
meiersi-da, line 40: checking: because we set the switch-over time on `DsoRules`, validators do not have to upgrade to the new amulet version? and that's why the typical language of they must upgrade is not here?

I'm though not sure that's fully the case: won't we hit the problem of `DsoRules` not being downgradable when acting on `splice-amulet` templates once the `svOperationsSwitchOver` field is set?

meiersi-da: also should we add a note explaining the motivation? That seems like an obvious question that people have. Something along the lines of "Tokenomics guidelines for featured apps require them precisely control what transfers they feature and how much weight they apply to them. Scan therefore no longer implicitly adds ... "

meiersi-da (approved):
Very nice work. Thanks a lot!

docs/src/release_notes_upcoming.rst
moritzkiefer-da: > checking: because we set the switch-over time on DsoRules, validators do not have to upgrade to the new amulet version? and that's why the typical language of they must upgrade is not here?

fair let me add it. It won't break validator submitted transactions but it will break featured app conversions which seems important enough to mention.


> also should we add a note explaining the motivation? 

I have this line already but let me try to make it a bit more explicit: This change was made to avoid apps accidentally not complying with the marker guidance set by the Canton foundation.

meiersi-da: sgtm

PR #7068, opened 2026-09-02
feat: add checks for FROM directives

krzysztofczyz-da:
Part of https://github.com/DACH-NY/canton-network-internal/issues/5015

To show that it fails when needed: [link](https://github.com/canton-network/splice/actions/runs/33748189292/job/100625439883?pr=7068) 🔴 
```
docker inspect europe-docker.pkg.dev/da-images/public/docker/cometbft-canton-network:3.5.0-snapshot.20260203.17930.0.v8a849517-stable-20260204@sha256:4fee6d941517fa25634bef416d439ec2fd61c81765478d210cca12f9df91d803 | jq '.[].Architecture'
"amd64"
```
To show that it passes when ignores are applied: [link](https://github.com/canton-network/splice/actions/runs/33749290837/job/100628892509?pr=7068) 🟢 

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


.github/workflows/pr_multiarch_image_check.yml
```diff
@@ -0,0 +1,40 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.head.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check_multiarch_images:
+    name: Check that pinned Docker images are multi-arch
+    # Only run for PRs from the same repo (not forks), because the check needs
+    # self-hosted runners with docker proxy access.
```
martinflorian-da: Why do we need docker proxy access? Aren't the images public?

scripts/check-multiarch-images.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def _file_is_ignored(file: str) -> bool:
+    for pattern in IGNORED_FILE_PATTERNS:
+        if pattern.search(file):
+            return True
+    return False
+
+
+def main() -> int:
```
martinflorian-da: nit: Isn't a big feature of `if __name__ == "__main__":` that you can lead with your `main()` function and only after that unroll all the more private worker functions? Just better readability...

scripts/check-multiarch-images.py
```diff
@@ -0,0 +1,202 @@
...
+    failures = 0
+
+    # Dockerfiles may use variable digests (e.g. ARG $cometbft_sha), so parse all of them
+    for file in _tracked_dockerfiles():
+        path = Path(file)
+        if not path.is_file():
+            continue
+        if _file_is_ignored(file):
+            continue
+        refs = _refs_from_dockerfile(path)
```
martinflorian-da: How about just making that a set or otherwise deduplicating it explicitly, instead of that error-prone extra `checked` set?

Makefile
martinflorian-da, line 1: Do really need this additional indirection? We call scripts from CI directly most of the time it seems?

scripts/check-multiarch-images.py
```diff
@@ -0,0 +1,202 @@
...
+        print(f"ERROR: invalid JSON from docker buildx for {ref}: {e}", file=sys.stderr)
+        return False
+
+    return info.get("image", {}).get("architecture") is None
+
+
+def _extract_refs_from_text(text: str) -> set[str]:
+    """Extract full image:tag@sha256:<digest> references from arbitrary text."""
+    return {
+        f"{image}:{tag}@sha256:{digest}"
```
martinflorian-da: Hm hm so this assumes that we always pin the image. I guess we always do, but I do wonder if we can make this a bit more robust.

martinflorian-da: Can `DockerfileParser` already do the resolving for us perhaps?

nix/shell.nix
```diff
@@ -121,6 +121,7 @@ let
     ammonite
     curl
     dpm-sdk
+    docker
```
martinflorian-da: `skopeo` is already on that list - would that also work?

martinflorian-da (commented):
Thanks thanks! I have comments 🤓

Did you test this, for example by editing / "breaking" some digests locally to confirm they get caught?

Also about

> Fixes

I suggest keeping the issue open until we have made sure that upstream will adopt something equivalent: https://github.com/DACH-NY/canton-network-internal/issues/5015#issuecomment-4398816814

.github/workflows/pr_multiarch_image_check.yml
krzysztofczyz-da: They are but we will get rate-limited - thus this is needed.

scripts/check-multiarch-images.py
krzysztofczyz-da: changed

scripts/check-multiarch-images.py
krzysztofczyz-da: changed

Makefile
krzysztofczyz-da: removed

scripts/check-multiarch-images.py
krzysztofczyz-da: We do pin, and if we don't we have a check that will fail so should not be an issue.

nix/shell.nix
krzysztofczyz-da: Changed, thanks 👌

.github/workflows/pr_multiarch_image_check.yml
martinflorian-da: OK... But then does this need to be a dedicated workflow with its own startup cost for setting up nix? Can this be part of, say, `deployment_test` instead?

martinflorian-da (approved):
Thanks!

PR #7072, opened 2026-09-02
Mark `splice-domain` Helm chart as deprecated

martinflorian-da:
Will follow up with a PR to 0.9.0 staging that removes it; here it is: https://github.com/canton-network/splice/pull/7073

[static]

### DIFF
--- a/docs/src/release_notes_upcoming.rst
+++ b/docs/src/release_notes_upcoming.rst
@@ -7,16 +7,6 @@
 
 release-notes:: Upcoming
 
-    - Deployment
-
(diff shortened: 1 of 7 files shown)


docs/src/release_notes_upcoming.rst
```diff
@@ -7,16 +7,6 @@
 
 release-notes:: Upcoming
 
-    - Deployment
-
```
martinflorian-da, line 11: just moved this block as a drive-by fix

moritzkiefer-da (approved):
thanks

PR #7092, opened 2026-09-03
Add script for SV operators to collect resolved node configs

martinflorian-da:
Extracts the startup config dump from pod logs in a namespace; optionally restarts pods whose logs no longer contain the dump.

Tested on CILR; it works beautifully and also for validators.

Main part of https://github.com/DACH-NY/canton-network-internal/issues/6847 but before I close it I want to:

- test it with external SVs (after merging this)
- add docs for it - that also make it clear that this is super useful also for validator support cases

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines

### DIFF
--- a/scripts/collect-node-configs.sh
+++ b/scripts/collect-node-configs.sh
@@ -17,9 +17,7 @@
 # out of the Kubernetes log buffer. In that case the script offers to restart
 # the node's Deployment (once) so that the config gets logged again. This
 # causes a short downtime of that node. The restart is done via
-# `kubectl rollout restart`; since all Splice Helm charts use the `Recreate`
-# strategy, the old pod is fully stopped before the new one starts, so there
-# are never two instances of a node running at the same time.
+# `kubectl rollout restart`.
 #
 # Requirements: bash, kubectl (configured for the target cluster), jq, grep,
 # sed, and `zip` if --zip is used.
@@ -75,9 +73,8 @@ fi
 # Helpers
 # ---------------------------------------------------------------------------
 
-# The container images (without registry and tag) of the nodes we collect
-# configs from. Everything else in the namespace (postgres, cometbft, web UIs,
-# ...) is skipped.
+# The container images of the nodes we collect configs from. Everything else in
+# the namespace (postgres, cometbft, web UIs, ...) is skipped.
 node_images=(
   canton-participant
   canton-sequencer
--- a/scripts/collect-node-configs.sh
+++ b/scripts/collect-node-configs.sh
@@ -13,10 +13,13 @@
 # with `kubectl logs`, extracts the config from them and writes one `.conf`
 # file per node.
 #
-# If a pod has been running for a long time, the config line may have rotated
-# out of the Kubernetes log buffer. In that case the script offers to delete
-# the pod (once) so that its Deployment/StatefulSet recreates it and the config
-# gets logged again. Deleting a pod causes a short downtime of that node.
+# If a node has been running for a long time, the config line may have rotated
+# out of the Kubernetes log buffer. In that case the script offers to restart
+# the node's Deployment (once) so that the config gets logged again. This
+# causes a short downtime of that node. The restart is done via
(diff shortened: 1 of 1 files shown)


scripts/collect-node-configs.sh
```diff
@@ -0,0 +1,229 @@
...
+mkdir -p "$out_dir"
+echo "Collecting node configs from namespace '$namespace' into '$out_dir'"
+
+# List all pods with their main container. We only care about containers
+# running a Canton image (canton-participant, canton-sequencer, ...) or a
+# Splice app image (sv-app, validator-app, scan-app, ...); everything else
+# (postgres, cometbft, web UIs, ...) is skipped. For each such container we
+# print: pod name, container name, the pod's `app` label, and whether the pod
+# is managed by a controller (only then is it safe to delete it).
+pods="$(kubectl get pods -n "$namespace" -o json | jq -r '
```
moritzkiefer-da: wondering if we should query deployments and properly restart those instead of killing given how many discussions we had around concurrently running instances being bad (it _shouldn't_ be bad but 🤷)

scripts/collect-node-configs.sh
```diff
@@ -0,0 +1,229 @@
...
+  .items[]
+  | . as $pod
+  | .spec.containers[]
+  | (.image | split("/") | last | split("@")[0] | split(":")[0]) as $image
+  | select($image | test("^canton-") or test("-app$"))
```
moritzkiefer-da: -app seems quite broad

scripts/collect-node-configs.sh
```diff
@@ -0,0 +1,229 @@
...
+# Deletes the pod, waits for its replacement (found via the shared `app`
+# label) to log its config, and prints that config. Returns non-zero on
+# timeout.
+restart_and_extract() {
+  local pod="$1" container="$2" app="$3"
+  local timeout_seconds=300 interval_seconds=5 waited=0
+  local new_pod config
+
+  echo "  Deleting pod $pod ..." >&2
+  kubectl delete pod -n "$namespace" "$pod" --wait=false >/dev/null
```
moritzkiefer-da: what do you gain from wait=false?

scripts/collect-node-configs.sh
```diff
@@ -0,0 +1,229 @@
...
+  if config="$(extract_config "$pod" "$container")"; then
+    printf '%s\n' "$config" > "$target"
+    echo "  Saved to $target"
+    collected+=("$app")
+    continue
+  fi
+
+  echo "  Logs of $pod no longer contain the startup config."
+  if [[ "$has_owner" != "true" ]]; then
+    echo "  Pod is not managed by a controller and would not be recreated; skipping." >&2
```
moritzkiefer-da: you can also skip that if you query deployments as I suggested

moritzkiefer-da (approved):
thx

scripts/collect-node-configs.sh
martinflorian-da: The best alternative I can think of is to unroll the full list. Works too I guess, maybe better to not be too smart. Then I'd also unroll the canton part.

scripts/collect-node-configs.sh
martinflorian-da: Well in principle we don't care when the delete finished, we care when the pod comes back up. I don't think it matters much in practice either way.

martinflorian-da: Fixed by moving to deployments

scripts/collect-node-configs.sh
martinflorian-da: That option didn't even occur to me, old habits die hard 😅

But after some playing around I do think it's better. The code also got slightly less convoluted. Thanks!

scripts/collect-node-configs.sh
martinflorian-da: Did that now. In the future we might want to make the list extendable but I'll wait for someone to complain about this first.

PR #7100, opened 2026-09-03
Pin postgres version in compose, localnet, splice-postgres & pulumi

OriolMunoz-da:
Fixes https://github.com/DACH-NY/canton-network-internal/issues/6633

### DIFF
--- a/cluster/compose/localnet/compose.env
+++ b/cluster/compose/localnet/compose.env
@@ -4,7 +4,7 @@
 IMAGE_REPO=${IMAGE_REPO:-ghcr.io/digital-asset/decentralized-canton-sync/docker/} # Publicly released images
 
 
-POSTGRES_VERSION=${POSTGRES_VERSION:-14}
+POSTGRES_VERSION=${POSTGRES_VERSION:-14.24-trixie}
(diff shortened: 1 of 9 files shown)


cluster/compose/localnet/compose.env
```diff
@@ -4,7 +4,7 @@
 IMAGE_REPO=${IMAGE_REPO:-ghcr.io/digital-asset/decentralized-canton-sync/docker/} # Publicly released images
 
 
-POSTGRES_VERSION=${POSTGRES_VERSION:-14}
+POSTGRES_VERSION=${POSTGRES_VERSION:-14.24-trixie}
```
moritzkiefer-da, line 7: hm usually we pin up to the hash, kinda feels like we should also do that here. Although doing that without automation to bump the hash is probably also annoying. So maybe let's keep this PR as is and create an issue for that.

moritzkiefer-da (approved):
thx

cluster/compose/localnet/compose.env
OriolMunoz-da: https://github.com/canton-network/splice/issues/7108

PR #7111, opened 2026-09-03
Implement String interning store

### LINKED ISSUE #7075: Implement String interning store
Follow design from: https://docs.google.com/document/d/13buMz1orZi5NrCuyrTfBo4OcwdOh-SmyvP8qm1DQqe0/edit?tab=t.0

Should be able to:
- Insert into the store
- Read from it

No need to wire it up (yet), as that can happen in the other tickets that define new tables, and thus we avoid costly/annoying migrations

OriolMunoz-da:
Fixes https://github.com/canton-network/splice/issues/7075


apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/InternedStringStore.scala
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  class DbInternedStringStore(storage: DbStorage, protected val loggerFactory: NamedLoggerFactory)(
+      implicit
+      ec: ExecutionContext,
+      close: CloseContext,
+  ) extends InternedStringStore
+      with NamedLogging {
+
+    override def getOrIntern(value: String)(implicit tc: TraceContext): Future[InternedId] = {
```
OriolMunoz-da, line 109: the query proposed in the design was:
```
with new_row as ( 
  insert into interned_strings (value) values ($v) 
  on conflict (value) do nothing returning id 
) 
select id from new_row union all select id from interned_strings where value = $v limit 1;
```
this does not work with contention:
```
[info] - handle concurrent interning of the same value without cache *** FAILED ***
[info]   java.util.NoSuchElementException: empty result setSome( when invoking statement
[info]             with new_row as (
[info]               insert into interned_strings (value) values (?)
[info]               on conflict (value) do nothing returning id
[info]             )
[info]             select id from new_row union all select id from interned_strings where value = ? limit 1;
[info]           )
```

the reason for that is that if there's two queries, you can end up in this situation (as far as I understand):
- q1 runs the insert successfully
- q2 runs the insert which conflicts and returns nothing
- q2 runs the select which returns nothing
- q2 returns nothing
- q1 commits

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -519,7 +520,9 @@ class UpdateHistory(
     val safeSignatories = event.createdEvent.getSignatories.asScala.toSeq.map(lengthLimited)
     val safeObservers = event.createdEvent.getObservers.asScala.toSeq.map(lengthLimited)
     metrics.UpdateHistory.assignments.mark()
-    sqlu"""
+    for {
+      _ <- DBIO.from(internTemplateId(templateId))
```
OriolMunoz-da: this is wrapping the future: i.e., it's not transactional with the rest of the statements. That should be fine since 99.99% of the time we're using the cache anyway

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
```diff
@@ -654,6 +660,17 @@ class UpdateHistory(
         $recordTime, $safeUpdateId, $safeDomainId, $migrationId
       )
     """
+    } yield result
+  }
+
+  // This does not need to be transactional with the rest of the transactions in UpdateHistory
+  private def internTemplateId(identifier: Identifier)(implicit tc: TraceContext): Future[Unit] = {
+    // TODO (#6257): use the returned ids in the partitioned table
```
OriolMunoz-da: we can start to fill up the store now and use it later in the partitioned tables
from the scan pruning doc:
> The dualWrite/readFromId rollout in the String Interning design covers only the legacy tables, which are dropped wholesale, so it is no longer needed.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/InternedStringStore.scala
ray-roestenburg-da: And q1 returns the right set?

ray-roestenburg-da: yeah but q2 returns nothing

OriolMunoz-da: yes, q1 behaves properly, but q2 blows up because nothing is returned

ray-roestenburg-da: yeah your version is safer, is fine.

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/InternedStringStore.scala
```diff
@@ -0,0 +1,148 @@
...
+    private[InternedStringStore] def fetchAll(maxSize: Long)(implicit
+        tc: TraceContext
+    ): Future[Vector[(String, InternedId)]] = {
+      storage
+        .query(
+          sql"select value, id from interned_strings limit $maxSize".as[(String, InternedId)],
+          "fetchAllInternedStrings",
+        )
+        .map { all =>
+          if (all.size >= maxSize) {
```
ray-roestenburg-da, line 137: why don't you just do a count(1) on interned_strings (or similar on metadata of postgres), and then compare so maxSize? Also, we do not necessarily need all strings in cache, this depends on access pattern, will this not just lead to unnecessary warnings?

OriolMunoz-da: from the design doc, this warning should never be hit and everything should be fine to be in the cache:
image

if it is hit, we have a problem

ray-roestenburg-da: ah right, thanks lol

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/InternedStringStore.scala
ray-roestenburg-da: Wait, it is a union right, so why does it get nothing? 
`select id from new_row union all select id from interned_strings where value = $v limit 1;`

ray-roestenburg-da: 'select from new_row union select from the interned strings'

OriolMunoz-da: (this is my assumption from the failure i posted above)
because q1 hasn't committed yet, so q2 cannot read it from the select. The concurrent inserts between q1 and q2 do conflict, and q2 loses

ray-roestenburg-da: I don't get why the insert flatmap select works and the insert then select does not.

ray-roestenburg-da: but in the new way you can still have concurrent insert/select no? the flatmap is not a tx boundary or is it?

ray-roestenburg-da: (if it is, then it explains it, otherwise you might still have a problem?)

ray-roestenburg-da: ah ok by default that is autocommit per statement

OriolMunoz-da: > the flatmap is not a tx boundary

it is; it's separate txs (storage.query twice, as opposed to DBIO composition)

ray-roestenburg-da: ok cool thanks I get it

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/InternedStringStore.scala
OriolMunoz-da: ~actually from this i realize i'm missing the package name and choice name in the updatehistory intern call~ done

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/UpdateHistoryIngestionPerformanceTest.scala
```diff
@@ -48,6 +50,13 @@ class UpdateHistoryIngestionPerformanceTest(
       participantId = mkParticipantId(this.getClass.getSimpleName),
       updateStreamParty = dsoParty,
       backfillingRequired = UpdateHistory.BackfillingRequirement.BackfillingNotRequired,
+      internedStringStore = InternedStringStore.createWithoutWarmup(
```
ray-roestenburg-da, line 53: Are you doing withoutWarmup in tests just to speed up the tests? Any idea how long this takes in ScanApp with the warmup?

OriolMunoz-da: > Are you doing withoutWarmup in tests just to speed up the tests? 

doesn't really matter, the DB will be empty there

> Any idea how long this takes in ScanApp with the warmup?

it's a query fetching 0-1000 rows, that shouldn't be too long

ray-roestenburg-da: so we don't need the withoutWarmup version?

OriolMunoz-da: I guess not, I can remove it

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/UpdateHistory.scala
ray-roestenburg-da: https://github.com/canton-network/splice/issues/6312

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/UpdateHistoryIngestionPerformanceTest.scala
ray-roestenburg-da: not a huge deal, was just wondering

ray-roestenburg-da (approved):
Looks good, thanks!

apps/app/src/test/scala/org/lfdecentralizedtrust/splice/performance/tests/UpdateHistoryIngestionPerformanceTest.scala
OriolMunoz-da: it does have the annoyance that warmup returns a Future, whereas the other one can just return the cache. Will leave as-is unless you disagree

ray-roestenburg-da: is fine, leave as is

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/store/db/InternedStringStore.scala
```diff
@@ -0,0 +1,148 @@
...
+      close: CloseContext,
+  ) extends InternedStringStore
+      with NamedLogging {
+
+    override def getOrIntern(value: String)(implicit tc: TraceContext): Future[InternedId] = {
+      storage
+        .query(
+          sql"""
+            insert into interned_strings (value) values ($value)
+            on conflict (value) do nothing returning id;
```
rautenrieth-da, line 114: This might consume an entry from the auto-increment sequence on each read. We should still be fine for the next billion years before the bigint overflows.

apps/common/src/test/scala/org/lfdecentralizedtrust/splice/store/db/InternedStringStoreTest.scala
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  override protected def cleanDb(
+      storage: DbStorage
+  )(implicit traceContext: TraceContext): FutureUnlessShutdown[?] = {
+    for {
+      _ <- resetAllAppTables(storage)
+      _ <- storage.queryAndUpdate(
+        sqlu"truncate interned_strings",
```
rautenrieth-da: Aren't we already truncating the table in `resetAllAppTables`?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/ScanApp.scala
```diff
@@ -217,13 +216,21 @@ class ScanApp(
         config.acsStoreDescriptorUserVersion,
         config.txLogStoreDescriptorUserVersion,
       )
+      internedStringStore <- InternedStringStore.createAndWarmupCache(
```
rautenrieth-da, line 219: Loading all interned strings from the database will delay application startup slightly, do you know by how much?

apps/common/src/main/resources/db/migration/canton-network/postgres/stable/V074__string_interning.sql
```diff
@@ -0,0 +1,5 @@
+create table interned_strings
```
rautenrieth-da, line 4: Most (but not all) migration files have a copyright header.

rautenrieth-da (approved):
Thanks!

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/ScanApp.scala
OriolMunoz-da: same answer as in https://github.com/canton-network/splice/pull/7111#discussion_r3933267261, there will be <10k strings that should fit in the cache and be trivially fetchable from a single query

PR #7169, opened 2026-09-08
fix(ci): anchor release-line branch check to prevent TODO/FIXME check…

sashaphmn:
Fixed a bug where the TODO/FIXME checker used a substring check (.contains("release-line")) to detect release-line branches, allowing any branch whose name merely contained that text (e.g. feature/release-line-bypass) to fully skip the checker, including hard-blocked FIXMEs. Replaced it with an anchored regex matching the actual release-line-<major>.<minor>.<patch> naming convention (confirmed via LATEST_RELEASE and common_setup.sh), so only real release-line branches bypass the check.

### DIFF
--- a/scripts/todo/src/checkTodos.sc
+++ b/scripts/todo/src/checkTodos.sc
@@ -31,8 +31,18 @@ val baseBranch = {
   }
 }
 val runningInCI = sys.env.contains("CI")
-val releaseLineStr = "release-line"
-val onReleaseLine = baseBranch.exists(_.contains(releaseLineStr)) || branch.contains(releaseLineStr)
+// Release-line branches follow the naming convention `release-line-<major>.<minor>.<patch>`
+// matching the exact contents of the LATEST_RELEASE file (e.g. release-line-0.8.0).
+// See .github/actions/scripts/common_setup.sh, which fetches
+// `refs/heads/release-line-${latest_release}` with no fallback and hard-fails
+// if that exact branch does not exist. Matching must be exact/anchored:
+// a substring check (previously `.contains("release-line")`) would let a PR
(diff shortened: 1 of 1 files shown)


scripts/todo/src/checkTodos.sc
```diff
@@ -31,8 +31,18 @@ val baseBranch = {
   }
 }
 val runningInCI = sys.env.contains("CI")
-val releaseLineStr = "release-line"
-val onReleaseLine = baseBranch.exists(_.contains(releaseLineStr)) || branch.contains(releaseLineStr)
+// Release-line branches follow the naming convention `release-line-<major>.<minor>.<patch>`
```
martinflorian-da: This comment is unnecessarily verbose (I don't think we need any comment here at all) and likely 100% AI written.

Consider shortening or removing it and please see our AI policy: https://github.com/canton-network/splice/blob/main/AI_POLICY.md

Specifically, note that we don't generally welcome fully autonomous / AI-only contributions and might close, ignore and block contributions and contributors that appear to violate our policy, without further comment.

martinflorian-da (commented):
Thanks! The change look correct to me.

- Please see my other comment about your 🤖 comment; happy to approve and merge once you have addressed that.
- It seems like you are solving a problem we don't really have... please check our contributor guidelines before picking up a bigger issue in the future: https://github.com/canton-network/splice/blob/main/CONTRIBUTING.md#picking-up-issues

scripts/todo/src/checkTodos.sc
sashaphmn: Yes, I was familiar with these terms before making the commit, I simply discovered a few issues I wanted to help you resolve. Thank you for understanding. Comment deleted.

martinflorian-da (approved):

PR #7181, opened 2026-09-08
Docker mirror: move cache fix to the right place...

martinflorian-da:
Follow-up to #7148 ... so part of DACH-NY/canton-network-internal#4580

Yes I should be fired for not noticing this during my testing...

[static]


isegall-da (approved):
Tnx

PR #7228, opened 2026-09-10
Expose SV app `/version` to validators

martinflorian-da:
Also adds support for (other) endpoints that are defined in the OpenAPI yaml via `$ref:`.

Tested on scratch; it correctly fixes the authorization policy there.

Part of DACH-NY/canton-network-internal#6966 ; I also want to do a docs fix for this (+ will also just add the release notes there right away as I'm also prepping the next release).

[static]


nicu-da (approved):
Thanks!

Lets make sure we have a ticket to actually remove this when it's no longer needed. 
imo the validators should always do version checks only through scan not the sv app

PR #7244, opened 2026-09-10
Set linear canton bft blaklisting on scratches

OriolMunoz-da:
...before setting it on cimain/ciperiodic
Part of https://github.com/DACH-NY/cn-test-failures/issues/10057

[preflight](https://app.circleci.com/pipelines/github/DACH-NY/canton-network-internal/81831/details?job=8886f6a4-eae5-4baf-ac50-aedf54a102e8&workflowId=71d6bb3d-8bef-4939-8a06-b8aa290378bf&buildNumber=492535&jobType=build)

### DIFF
--- a/cluster/configs/shared/scratchnet-sv.yaml
+++ b/cluster/configs/shared/scratchnet-sv.yaml
@@ -15,23 +15,9 @@ svApp:
     # complaining that all the other keys are missing
     - name: ADDITIONAL_CONFIG_CANTONBFT_EXPONENTIAL_BLACKLISTING
       value: |
-        canton.sv-apps.sv.canton-bft-sequencing-parameters {
-          blacklist-leader-selection-policy-config {
-              how-long-to-blacklist {
-                  # Reduced from the default of 250 by 4x to match the increase in 4x in segment length.
-                  maximum-epoch-blacklisted=62
-                  type=linear
-              }
-              how-many-can-we-blacklist=num-faults-tolerated
-          }
-          # Increased by 4x as there is a synchronization point at the end of each epoch which can be expensive.
-          segment-length=40
-          max-requests-in-batch=32
-          max-batches-per-block-proposal=16
-          pbft-view-change-timeout="5 seconds"
-          pbft-view-change-timeout-step="2 seconds"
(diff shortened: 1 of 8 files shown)


cluster/configs/shared/scratchnet-sv.yaml
```diff
@@ -10,6 +10,30 @@ svApp:
     requests:
       cpu: "1"
       memory: 2Gi
+  additionalEnvVars:
+    # A naive `canton.sv-apps.sv.canton-bft-sequencing-parameters.blacklist-leader-selection-policy-config.how-long-to-blacklist.type=linear` doesn't work,
+    # complaining that all the other keys are missing
```
OriolMunoz-da: Image

I don't understand how the current cilr configuration is working given this

nicu-da: could it be that there's some weird behavior in the config reader that ignores the code defaults if you provide any overrides?

cluster/configs/shared/scratchnet-sv.yaml
nicu-da, line 1: should we create some CI cluster config that gets used by the scratches and all the CI clusters? 
I often find myself wanting to change something just for those clusters

nicu-da (approved):
Thx!

cluster/configs/shared/scratchnet-sv.yaml
OriolMunoz-da: 90% sure that's the case, maybe it's because it's `Option[CaseClass] = Some(CaseClass.default)`? not sure

nicu-da: would be nice if we can figure out a way to fix it, this is going to be not so great UX if we wanna update anything

OriolMunoz-da: https://pureconfig.github.io/docs/overriding-behavior-for-case-classes.html#default-field-values likely explains this: `BftSequencingParameters` has no default parameters, so the moment you try to set one of the values pureConfig tries to get the default for other parameters and fails. I will attempt to fix it.

nicu-da (approved):
Nice, thanks!

PR #7265, opened 2026-09-11
Fix pg health check test signal

jagathweerasinghe-da:
Fix: https://github.com/DACH-NY/cn-test-failures/issues/10074

[static]


moritzkiefer-da (approved):
thx change looks sensible but I'm a bit confused, don't we retry for quite some time? how long is the difference betweeen the socket and the port being available?

PR #7294, opened 2026-09-14
token-standard: add splice-api-token-conditional-lock-v1 (CIP draft)

tankcdr:
Adds the interface package for the Conditional Holding Lock CIP: an owner attaches a release policy (guards paired with outcomes, expiry with fallback) to a holding. Interface package only; TestTokenV2 and Canton Coin implementations follow in separate PRs. Depends on splice-api-token-metadata-v1 and splice-api-token-holding-v2.

CIP draft and reference proofs: https://github.com/tankcdr/conditional-holding-lock

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
```diff
@@ -0,0 +1,286 @@
...
+
+-- | Hash function for a hashlock. Digests and preimages are hex-encoded byte strings.
+data HashAlgorithm
+  = Sha256
+    -- ^ SHA-256 over the raw preimage bytes. Registries MUST support this algorithm.
+  | Keccak256
+    -- ^ Keccak-256 over the raw preimage bytes. Registries MUST support this algorithm.
+  deriving (Eq, Ord, Show, Serializable)
+
+-- | A condition the ledger can check when a rule is enacted.
```
meiersi-da: It's the signatories of the lock contract that will check this.

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
```diff
@@ -0,0 +1,286 @@
...
+        -- ^ Unique list of parties.
+      threshold : Int
+        -- ^ Satisfied when at least `threshold` of `parties` are among the enacting actors.
+        -- MUST satisfy 1 <= threshold <= length parties.
+  | Guard_All with
+      guards : [Guard]
+        -- ^ Satisfied when every guard is satisfied. MUST be non-empty.
+  | Guard_Any with
+      guards : [Guard]
+        -- ^ Satisfied when at least one guard is satisfied. MUST be non-empty.
```
meiersi-da: allows arbitrarily deep nesting. Problematic for wallets.

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
```diff
@@ -0,0 +1,286 @@
...
+    preimages : [Text]
+      -- ^ Hex-encoded 32-byte preimages. Each `Guard_Preimage` node is satisfied if any
+      -- listed preimage hashes to its digest.
+  deriving (Eq, Ord, Show, Serializable)
+
+-- | A destination for locked funds.
+data Leg = Leg with
+    receiver : Account
+    amount : Decimal
+      -- ^ MUST be positive.
```
meiersi-da, line 79: metadata?

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
```diff
@@ -0,0 +1,286 @@
...
+      -- ^ Unique within the terms. Reported in events.
+    enactors : [Party]
+      -- ^ Parties entitled to enact this rule. At least one MUST be among the actors.
+    guard : Guard
+    outcome : Outcome
+  deriving (Eq, Ord, Show, Serializable)
+
+-- | The release policy, provided by the owner's wallet.
+data LockTerms = LockTerms with
+    owner : Account
```
meiersi-da: consider `authorizer` as this seems a similar case as for allocations.

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
```diff
@@ -0,0 +1,286 @@
...
+    instrumentId : InstrumentId
+    amount : Decimal
+      -- ^ Locked amount. On a continuation, the remaining amount.
+    rules : [Rule]
+      -- ^ Enactable strictly before `expiresAt`. Each fires at most once.
+    expiresAt : Time
+      -- ^ Absolute, inclusive ledger time at which the rules stop being enactable and
+      -- the fallback becomes enactable.
+    fallback : Outcome
+      -- ^ Enactable by the owner at or after `expiresAt`. Typically `Outcome_Unlock`.
```
meiersi-da: when would you use a different outcome?

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
```diff
@@ -0,0 +1,286 @@
...
+
+interface ConditionalLockInstruction where
+  viewtype ConditionalLockInstructionView
+
+  conditionalLockInstruction_acceptImpl : ContractId ConditionalLockInstruction -> ConditionalLockInstruction_Accept -> Update ConditionalLockResult
+  conditionalLockInstruction_rejectImpl : ContractId ConditionalLockInstruction -> ConditionalLockInstruction_Reject -> Update ConditionalLockResult
+  conditionalLockInstruction_withdrawImpl : ContractId ConditionalLockInstruction -> ConditionalLockInstruction_Withdraw -> Update ConditionalLockResult
+
+  nonconsuming choice ConditionalLockInstruction_Accept : ConditionalLockResult
+    -- ^ A receiver accepts. Result is `Locked` once every receiver has accepted,
```
meiersi-da: We found that it's not always the receiver, e.g., it might also be a custodian that needs to approve a transfer. Generalize the comment.

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
```diff
@@ -0,0 +1,286 @@
...
+
+  nonconsuming choice ConditionalLock_Enact : ConditionalLockResult
+    -- ^ Fire one rule. MUST fail at or after `terms.expiresAt`, unless at least one of
+    -- the rule's enactors is among `actors`, unless the guard is satisfied, or
+    -- unless `legs` are valid for the outcome.
+    with
+      ruleId : Text
+      actors : [Party]
+      witness : Witness
+      legs : Optional [Leg]
```
meiersi-da: Are there cases where a fixed leg would make sense too?

meiersi-da (commented):
Thanks a lot for sharing this draft. Directionally this looks right.

Given that interfaces are immutable once defined their polishing and validation needs to be done up front. I think this one will require quite a few rounds until we're sure that it is worth the effort for all asset registries to implement. I'm writing all, as it is the universally supported APIs that create the most value.

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
tankcdr: Correct. The `Guard` doc comment now reads "A condition checked by the lock contract's signatories when a rule is enacted", and the module header says the same rather than attributing the check to the ledger. CIP section 3.3 names those signatories: the registry admin, the parties the registry requires to move funds out of `terms.authorizer`, and for every approved receiver account the parties that gave that approval, and it states that the enactor supplies the `witness` without being trusted to evaluate it. Section 3.5 gives the evaluation rule for each guard kind. The phrase "ledger-checkable" is gone from the module, the CIP and the Splice docs.

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
tankcdr: Agreed, and the recursion is gone. `Guard` is now a flat leaf type with four constructors and no self-reference; a `Rule` carries `anyOf : [Alternative]` and each `Alternative` carries `allOf : [Guard]`, so a wallet renders a rule as a fixed two-level list with no recursive component and no depth budget. CIP section 3.5 states the evaluation rule and says outright that guards are a closed set with no nesting beyond `anyOf`/`allOf`; section 3.8 replaces `max-guard-depth` with `max-alternatives-per-rule` and `max-guards-per-alternative`, which registries MUST support at 8 or more. Nothing in the CIP lost expressiveness: every disjunction in the worked terms of section 4 was already a single `Guard_Parties ... 1` leaf or a separate rule.

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
tankcdr: Added, and `Leg` was missing an identifier for the same reason. It now carries `legId : Text` and `meta : Metadata` alongside `receiver` and `amount`, so it lines up with `AllocationV2.TransferLeg`, which carries `transferLegId` and `meta`. CIP section 3.7 defines the mapping: each enacted leg to a receiver other than the authorizer produces a `TransferLegSide` on each side with the identifier `<lockId>/<ruleId>/<legId>`, and `TransferLegSide.meta` MUST contain every key of `Leg.meta`. Because `legId` is free text where the old encoding used a positional index, section 3.7 also requires `lockId`, `Rule.id` and `Leg.legId` to be non-empty and to contain no `/`; two of those three fields are outside the line you commented on, so I am calling it out rather than slipping it into the diff. `Leg.meta` MUST NOT set `splice.lfdecentralizedtrust.org/tx-kind`, `splice.lfdecentralizedtrust.org/conditional-lock/rule-id`, or any other reserved key under `splice.lfdecentralizedtrust.org/conditional-lock/`, and registries reject such terms at creation and amendment.

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
tankcdr: Renamed. `LockTerms.authorizer : Account` is now documented as "Account that funds the lock and receives unlocked funds", which is the same dual role `AllocationSpecification.authorizer` plays in `AllocationV2`. The two result fields followed the rename: `ownerHoldingCids` is `authorizerHoldingCids` on `ConditionalLockResult_Enacted`, and `ownerChangeCids` is `authorizerChangeCids` and moves up from `ConditionalLockResult_Locked` to `ConditionalLockResult` itself, so change holdings are reported whatever the output, as CIP section 3.1 now requires. `receiverHoldingCids` keeps its name to match `TransferInstructionResult_Completed`, and all three stay `[ContractId Holding]` rather than the allocations' `TextMap` keyed by instrument, since a lock is single-instrument. The CIP uses `authorizer` throughout sections 2, 3.1 and 3.3, and the only `owner` left in the module is the prose reference to an account's owner in `pendingApprovals`, which is `HoldingV2` vocabulary and stays.

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
tankcdr: You would not, so it is gone. `LockTerms.fallback` and `ConditionalLock_Expire.legs` are removed, and `ConditionalLock_Expire` now returns the full remaining amount to `terms.authorizer`, unlocked, and terminates the lock with no continuation. Every case for a directed fallback is better written as a rule with `Guard_After T` whose `enactors` name the beneficiary, with `expiresAt` behind it as the safety valve, which is the shape the worked terms in CIP section 4 already use. CIP section 3.6 carries the new rule and adds that registries MUST NOT fail an `Expire` for reasons attributable to any party other than the authorizer, a guarantee that is now unconditional because there is no fallback left to qualify it. This also matches CIP-0112, where directed movement is `Allocation_Settle` and expiry only ever returns funds to the authorizer.

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
tankcdr: Generalized. The `ConditionalLockInstruction_Accept` doc comment now says that an approver accepts, that the approver is the account whose approval is pending, and that the parties who must act for it are registry-defined and reported in `availableActions`; the view field `pendingReceivers` is now `pendingApprovals : [Account]`, whose doc comment says the registry decides whether an account's owner, its provider, or both must act, and the choice argument `receiver` is now `approver : Account`. The choice keeps an `Account` argument rather than dropping it as `TransferInstruction_Accept` does, because a lock instruction can have several accounts pending at once. I also added `availableActions : Map.Map ConditionalLockInstructionAction [[Party]]` to `ConditionalLockInstructionView` and `availableActions : Map.Map ConditionalLockAction [[Party]]` to `ConditionalLockView`, with the field shape and doc wording of `TransferInstructionV2`, so wallets learn who may act from a field they already parse. CIP section 3.2 covers the instruction and section 3.3 the lock, where the `CLA_Enact` entries mean a rule exists and the listed groups may enact it, not that any alternative is currently satisfied.

token-standard/splice-api-token-conditional-lock-v1/daml/Splice/Api/Token/ConditionalLockV1.daml
tankcdr: Yes, and you could not write one in a single rule. `Outcome_Transfer` and `Outcome_Distribute` are merged into `Outcome_Release with fixedLegs : [Leg]; receivers : [Account]`, where `receivers = []` is the old `Outcome_Transfer`, `fixedLegs = []` is the old `Outcome_Distribute`, and both non-empty is the case that was missing: an arbiter fee fixed by the terms with the remainder apportioned at the arbiter's discretion, atomically. `fixedLegs` and `receivers` MUST NOT both be empty, and conservation is checked over the union: the `Outcome_Release` doc comment, reproduced in CIP section 2, requires the fixed legs plus the enactor-supplied legs to be positive and at most the remaining amount, and CIP section 3.6 states the continuation, with any remainder continuing the lock without that rule. `ConditionalLock_Enact.legs` is a plain `[Leg]` rather than `Optional [Leg]`, since `receivers` now says directly when enactor-supplied legs are admissible. This is the shape allocations already have, with `AllocationSpecification.transferLegSides` fixed at authorization and `Allocation_Settle.extraTransferLegSides` supplied at settlement and validated against one budget.

---

tankcdr:
## Changes since round one

Round two is a second commit on this branch, so the review comments map onto hunks you can read by diff. Each of your seven comments and where it landed:

| Your comment | Round two |
| --- | --- |
| L23, guards are checked by the lock's signatories | `Guard` doc comment and the module header; CIP sections 3.3 and 3.5 |
| L47, arbitrary guard nesting | `Guard` is a flat leaf type; new `Alternative.allOf : [Guard]` and `Rule.anyOf : [Alternative]`; CIP sections 3.5 and 3.8 |
| L61, `Leg` metadata | `Leg.legId : Text` and `Leg.meta : Metadata`; CIP section 3.7 |
| L91, `owner` versus `authorizer` | `LockTerms.authorizer`, `ConditionalLockResult.authorizerChangeCids`, `ConditionalLockResult_Enacted.authorizerHoldingCids`; CIP sections 2, 3.1, 3.3 |
| L102, a different expiry outcome | `LockTerms.fallback` and `ConditionalLock_Expire.legs` removed; `ConditionalLock_Expire` always unlocks to the authorizer; CIP sections 3.3 and 3.6 |
| L194, acceptance is not always by the receiver | `ConditionalLockInstructionView.pendingApprovals`, `ConditionalLockInstruction_Accept.approver`, `availableActions` on both views; CIP sections 3.2 and 3.3 |
| L250, fixed legs alongside enactor legs | `Outcome_Release with fixedLegs; receivers` replaces `Outcome_Transfer` and `Outcome_Distribute`; `ConditionalLock_Enact.legs : [Leg]`; CIP section 3.6 |

Changes with no inline comment behind them, so you do not meet them cold in the diff:

- `ConditionalLockInstructionView.lockId`, so a wallet can correlate an instruction across approvals with the lock it becomes. `ConditionalLockView.lockId` already existed; it now MUST NOT contain `/`, because it is the first component of the transfer leg identifier.
- `ConditionalLockView.enactedRuleIds`, so wallets can validate an amendment before submitting and registries can enforce fire-once across continuations.
- `holdingCid : ContractId Holding` becomes `holdingCids : [ContractId Holding]` on `ConditionalLockResult_Locked` and on `ConditionalLockView`, and MAY be empty for registries that do not represent holdings on-ledger.
- `ConditionalLock_Amend` never returns `Pending`: every approval `newTerms` introduces MUST be given in the amending transaction, and `Amend` is forbidden at or after the current `expiresAt`. Rule ids in `enactedRuleIds` MUST NOT be reused.
- Seven new `*ExtraObservers` functions joining the one round one already had, so every choice except `ConditionalLockFactory_PublicFetch` has one, with a one-view-per-choice budget in CIP section 3.9. `conditionalLock_enactExtraObservers` MUST NOT widen beyond the parties of the accounts the outcome pays, because the arguments of `ConditionalLock_Enact` carry `witness.preimages`.
- The advertised registry limits in CIP section 3.8 change: `max-guard-depth` goes, and `max-alternatives-per-rule`, `max-guards-per-alternative`, `max-preimages` and `max-named-parties` join, leaving exactly `max-rules`, `max-legs`, `max-alternatives-per-rule`, `max-guards-per-alternative`, `max-preimages`, `max-named-parties`, `min-duration`, `max-duration`.
- `max-legs` is redefined from per outcome to per enactment, `length fixedLegs + length ConditionalLock_Enact.legs`, and the same bound applies to `Outcome_Release.receivers`.
- `Leg.meta` MUST NOT set a key the registry is required to set, per CIP section 3.7.
- `ConditionalLockResult_Failed` gains `authorizerHoldingCids`, so a rejected or withdrawn instruction reports the returned holdings the same way `ConditionalLockResult_Enacted` does.
- `daml/dars.lock` moves from `9e362fc605d36864ee08d69ef95db6937ab8f920419874f37951cc3a40105c30` to `cc541d14181e265667ea06c6e738e2415881ec49f849474da63319fcfb10d5ac`.
- `docs/src/app_dev/token_standard/index.rst` rewrites the Conditional Lock paragraph for the above.

Everything else is unchanged. The package layout is the same: one module, `Splice.Api.Token.ConditionalLockV1`, in `splice-api-token-conditional-lock-v1` at 1.0.0, depending only on `splice-api-token-metadata-v1` and `splice-api-token-holding-v2`, with the same `daml.yaml`, `build.sbt` and `project/BuildCommon.scala` wiring as round one.

`conditional-lock-v1.yaml` ships in this PR and is unchanged apart from two clarifications that round two forced: the path parameter of the four lock endpoints is renamed from `lockId` to `lockContractId`, because it is the contract ID of the lock and not the `lockId` of its view, and the use-case list names DvP between registries. There is no `Guard`, `Outcome`, `Leg` or `LockTerms` schema in it, because lock terms cross the wire as Daml values, so none of the type changes above touch it.

The reference implementation, the test suite and a mirror of the OpenAPI file live in https://github.com/tankcdr/conditional-holding-lock, along with the CIP draft. This PR remains the interface package only.

tankcdr:
## Round three

Two commits on this branch, nothing that reverses round two:

- **Recorded approvals.** `ConditionalLock_Approve` records approvals on the lock for an exact rule and list of legs, and `ConditionalLock_Enact` counts them toward `Guard_Parties` and the enactor requirement. Parties of a threshold approve one at a time from their own wallets, with no registry delegation contract, in the same way the lock controllers in canton-foundation/cips#250 authorize one after another. New: `Approval`, `ConditionalLockView.approvals`, `CLA_Approve`, a ninth `*ExtraObservers` function, and an `approve` choice-context route in the OpenAPI file. Approvals are irrevocable and `Amend` clears them; the CIP's Security Considerations state that a quorum recorded for a rule lets any party trigger exactly that enactment, and section 4 shows the venue-settled DvP form.
- **Version.** The package is `0.2.0` rather than `1.0.0`, matching the reference release, since Canton will not vet two packages with the same name and version and 1.0.0 should mark a stable interface. `daml/dars.lock` carries the new package ID.

The CIP also now states that time comparisons MUST be bounds on ledger time (`isLedgerTimeLT`/`GE`), per CIP-0062, so the CIP-0107 submission delay holds, and the reference evaluator was changed accordingly. It explains why an allocation with namespaced metadata, the approach of cips#250, does not carry fact-checked guards, several outcomes over one pool, or a receiver bound on discretion, and that governance locks stay out of scope. The Canton Coin paragraph was corrected against CIP-0078 and the `LockedAmulet` choices.

Reference implementation and tests: https://github.com/tankcdr/conditional-holding-lock (94 Daml Scripts on Canton 3.5.16 and 3.5.17; the DvP against real Amulet on the Mainnet-configuration localnet).

PR #7295, opened 2026-09-14
Public Sequencer: Move minMemberTrafficToOnboardValidator, devNetPublicSetupTrafficAmount, which is currently in SvAppBackendConfig to DAML.

pasindutennage-da:
Fix https://github.com/canton-network/splice/issues/7293


daml/splice-dso-governance/daml/Splice/DsoRules.daml
```diff
@@ -520,6 +520,8 @@ data DsoRulesConfig = DsoRulesConfig with
     nextScheduledSynchronizerUpgrade: Optional SynchronizerUpgradeSchedule
     voteCooldownTime : Optional RelTime -- ^ The minimum time between two votes by the same SV.
     nextScheduledLogicalSynchronizerUpgrade: Optional LogicalSynchronizerUpgradeSchedule
+    minMemberTrafficToOnboardValidator: Optional Int -- ^ Minimum amount of traffic a validator must purchase to be granted a ParticipantSynchronizerPermission.
```
moritzkiefer-da, line 523: intuitively I would have expected those to be in `SynchronizerFeesConfig` or at least in `SynchronizerConfig`. Probably the latter is better to avoid blowing up amulet rules even more. Any reason it's not?

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/util/SvUtil.scala
```diff
@@ -258,6 +258,8 @@ object SvUtil {
     Optional.empty(), // nextScheduledSynchronizerUpgrade
     voteCooldownTime.map(t => new RelTime(t.duration.toMicros)).toJava,
     Optional.empty(), // nextScheduledLogicalSynchronizerUpgrade
+    Optional.of(100000L: java.lang.Long), // minMemberTrafficToOnboardValidator
```
moritzkiefer-da, line 261: Why do you need to set this explicitly? You already have a default if it's not set so setting it to None avoids duplicating the default and allows for downgrades. If you set it explicitly you would also need to make it conditional on the version.

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
```diff
@@ -912,7 +912,8 @@ class HttpSvPublicHandler(
         participantId.toProtoPrimitive,
         dsoRules.payload.config.decentralizedSynchronizer.activeSynchronizerId,
         dsoStore.domainMigrationId.toInt,
-        config.devNetPublicSetupTrafficAmount,
+        dsoRules.payload.config.devNetPublicSetupTrafficAmount.toScala
+          .getOrElse(10000000L: java.lang.Long),
```
moritzkiefer-da, line 916: inlining defaults in random places makes the code quite hard to follow. I would suggest to have one object probably in SvUtil where you define all the defaults with sensible names and then reference those.

moritzkiefer-da (approved):
thanks

daml/splice-dso-governance/daml/Splice/DsoRules.daml
pasindutennage-da: You mean DecentrazlizedSynchronizer.SynchronizerConfig right?

Sure, I can change.

pasindutennage-da: @moritzkiefer-da 

Well, that was mainly for convenience in using the existing DsoRulesConfig voting, with minimal changes.

If DecentrazlizedSynchronizer.SynchronizerConfig also has a well defined vote path, then sure.

daml/splice-dso-governance-test/daml/Splice/Scripts/DsoTestUtils.daml
```diff
@@ -88,6 +88,8 @@ initDecentralizedSynchronizerWithAmuletPrice isDevNet initialRound amuletPrice o
         nextScheduledSynchronizerUpgrade = None
         voteCooldownTime = None -- use default value of 1 minute
         nextScheduledLogicalSynchronizerUpgrade = None
+        minMemberTrafficToOnboardValidator = Some 100000
```
julientinguely-da: you seem to hardcode these two values at many places

daml/splice-dso-governance/daml/Splice/DsoRules.daml

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/util/SvUtil.scala
pasindutennage-da: Fixed

apps/sv/src/main/scala/org/lfdecentralizedtrust/splice/sv/admin/http/HttpSvPublicHandler.scala
pasindutennage-da: Fixed

moritzkiefer-da (approved):
thanks

PR #7298, opened 2026-09-15
support eventual-consistency in scan BFT reads

isegall-da:
Used only in `getBulkObjectChecksums` for now, not entirely sure if it fits the others eventual-consistency cases.
My confidence in this PR is fairly low, please feel free to tell me all the things I got wrong here.

I tried covering:
- logging
- metrics
- retries
- not breaking anything existing

#### Cluster Testing

#### PR Guidelines

#### Merge Guidelines


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/commands/HttpScanAppClient.scala
```diff
@@ -3414,6 +3414,9 @@ object HttpScanAppClient {
         http.GetBulkObjectChecksumsResponse,
         definitions.GetBulkObjectChecksumsResponse,
       ] {
+    // Keep 404 as an HTTP-level error for this call so BFT can classify "not yet" by status code.
```
rautenrieth-da: Some endpoints return HTTP 200 for "not yet" responses.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/commands/HttpScanAppClient.scala
```diff
@@ -3433,6 +3436,7 @@ object HttpScanAppClient {
       definitions.GetBulkObjectChecksumsResponse,
     ]] = {
       case http.GetBulkObjectChecksumsResponse.OK(response) => Right(response)
+      case http.GetBulkObjectChecksumsResponse.NotFound(err) => Left(err.error)
```
rautenrieth-da: How about making the result of this method optional, and returning `Right(None)` here?

Failure responses are a mess. AFAICT:

- A response that doesn't conform to the OpenAPI spec gets coverted to `Failure(unexpected: BaseAppConnection.UnexpectedHttpJsonResponse)` and then to `BftScanConnection.HttpFailureResponse`. This at least keeps the status code.
- A response that conforms to the OpenAPI spec, but returns a `Left` in `HttpCommand.handleOk` gets converted to `Future.failed(Status.UNAVAILABLE.withDescription(err).asRuntimeException())` and then to `BftScanConnection.ExceptionFailureResponse`. The HTTP status code is lost. Which makes me wonder how your `isNotYet` even works.
- `BftScanConnection` can reach consensus on `BftScanConnection.HttpFailureResponse` if a majority of scans return the same HTTP error, and the resulting `HttpFailureResponse` preserves the HTTP error code. However, if you were to return this response in a http handler, the `HttpErrorHandler` will not recognize this exception and will return a HTTP 500 internal server error.

I would prefer the following model:
- All expected responses are listed in the OpenAPI spec (at least things like "404 item not found" or "404 data not available yet")
- The result of the corresponding `HttpCommand` is a Scala type that includes all expected responses (i.e., use Option, Either, or a sealed trait if more than one expected response). No expected response should result in a failed Future.

FWIW, TBAR code uses a HTTP 200 response with a body that can be one of the following responses: "ok, here's the data", "wait a bit, data not available yet", and "I'll never be able to answer that, stop asking me". I don't have a strong opinion on whether that's better than using 404 for "not available yet", as long as we don't use the failed futures in scala.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1184,48 +1245,90 @@ object BftScanConnection {
 
     val responses =
       new ConcurrentHashMap[BftScanConnection.ScanResponse[T], List[Uri]]()
+    val notYetResponses = new ConcurrentHashMap[Uri, Unit]()
     val nResponsesDone = new AtomicInteger(0)
+    val nNotYetResponses = new AtomicInteger(0)
```
ray-roestenburg-da, line 1250: nrNotYetResponses, or notYetResponsesCount or something

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1904,7 +2040,7 @@ object BftScanConnection {
           trustedScanDetails = trustedScans
             .map(s => s"  - Name: ${s.svName}, URL: ${s.publicUrl}")
             .mkString("\n")
-          _ = logger.info(s"all available trusted scans on booststrap:\n$trustedScanDetails")
+          _ = logger.info(s"all available trusted scans on boosttrap:\n$trustedScanDetails")
```
ray-roestenburg-da, line 2043:
```suggestion
          _ = logger.info(s"all available trusted scans on bootstrap:\n$trustedScanDetails")
```

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1184,48 +1245,90 @@ object BftScanConnection {
 
     val responses =
       new ConcurrentHashMap[BftScanConnection.ScanResponse[T], List[Uri]]()
+    val notYetResponses = new ConcurrentHashMap[Uri, Unit]()
```
ray-roestenburg-da, line 1248: Rather use a concurrent Set if you don't use values

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
ray-roestenburg-da: Do you really need a separate counter for the map/set?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -2178,6 +2314,49 @@ object BftScanConnection {
       extends RuntimeException(s"Scan $url has no answer to contribute to consensus")
       with NoStackTrace
 
+  private case class NotYetResponse(url: Uri)
```
rautenrieth-da, line 2317: Isn't this the same as the existing `IgnoreResponse`?

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
ray-roestenburg-da: I think I would just use .size()

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1160,13 +1177,56 @@ class BftScanConnection(
       _.getBulkObjectChecksums(requiredCatchupTimestamp, objectKeys),
```
rautenrieth-da: Counter-proposal:

### Ignoring 404 responses

In `getBulkObjectChecksums`, map individual responses that returned 404 originally to `IgnoreResponse`/`NotYetResponse`. This counts as an error response and will never lead to consensus. 

```scala
_.getBulkObjectChecksums(requiredCatchupTimestamp, objectKeys).transform {
  case Success(None) => Failure(IgnoreResponse)
  case Success(Some(r)) => Success(Some(r)) 
  case Failure(e) => Failure(e)
}
```

### Logging disagreeing hashes

Change `logDisagreements` to not warn if the disagreeing responses are those coming from ignored responses.

Set `disagreementLogLevel` such that actual hash mismatches are logged at warn level.

### Remove double retry loop

Inner retry loop already retries a few times if consensus is not reached because too many scans are not ready yet.

Outer loop should IMO be at the call site. You already have one in `BulkStorageCommitFromStaging`.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
ray-roestenburg-da: You could use `ConcurrentHashMap.newKeySet[Uri]()`

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1160,13 +1177,56 @@ class BftScanConnection(
...
+      operationDescription: String,
+      task: => Future[T],
+      logger: TracedLogger,
+      disagreementRetryFor: RetryFor = RetryFor.ClientCalls,
+      waitingForReadyScansRetryFor: RetryFor = RetryFor.Automation,
+  )(implicit
+      ec: ExecutionContext,
+      tc: TraceContext,
+  ): Future[T] = {
+    // External retry loop with long timeouts, retrying on NotEnoughAvailableResponsesToReachConsensus
```
rautenrieth-da, line 1209: `BulkStorageCommitFromStaging` already has an external retry loop, why do we need this one? IMO, such external, long retry loops are the responsibility of the caller.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1184,48 +1245,90 @@ object BftScanConnection {
...
+          val notYetCouldEnableConsensus =
+            nNotYetResponses.get() > 0 && (maxAgreement + nNotYetResponses.get()) >= nTargetSuccess
+          val exception =
+            if (notYetCouldEnableConsensus) {
+              logger.debug(
+                s"Not enough Scan responses to reach consensus, but some scans are not yet ready"
+              )
+              logNotYetReadyResponses(logger, Level.DEBUG, notYetResponses.keys().asScala.toList)
+
+              BftScanConnection.NotEnoughAvailableResponsesToReachConsensus(
```
rautenrieth-da, line 1267: This seems to be the main benefit of this PR - the result of `executeCall` makes it possible to decide if waiting a bit longer could solve consensus failures.

What do we do with this information? `BulkStorageCommitFromStaging` already has infinite retries for any kind of consensus failure. Presumably we also have metrics and alerts that would notify us if the bulk storage subsystem was stuck.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
```diff
@@ -1263,12 +1366,27 @@ object BftScanConnection {
...
+         |${notYetResponses.map(url => s"  $url").mkString("\n")}""".stripMargin,
+    )
+
+  }
+
   private def logDisagreements[T](
       logger: TracedLogger,
       consensusResponse: Try[T],
       responses: ConcurrentHashMap[BftScanConnection.ScanResponse[T], List[Uri]],
       disagreementLogLevel: Level,
```
rautenrieth-da, line 1387: Maybe this method should have two modes?

1. Immutable data ("return hash for round 13"): log at warn level if any two successful responses differ, otherwise log at info
2. Mutable/volatile data ("return the latest round"): always log at info

rautenrieth-da (commented):
The PR is rather big and I don't see yet what exactly it solves. I think we can get to a reasonable behavior with a much smaller change:

### Ignoring 404 responses

In `getBulkObjectChecksums`, map individual "data not available yet" responses to `IgnoreResponse`/`NotYetResponse`. This will make sure that "data not available yet" is never the consensus. 

```scala
_.getBulkObjectChecksums(requiredCatchupTimestamp, objectKeys).transform {
  case Success(None) => Failure(IgnoreResponse(...))
  case Success(Some(r)) => Success(Some(r)) 
  case Failure(e) => Failure(e)
}
```

### Logging disagreeing hashes

Two scans returning different hashes is a serious issue. Some scans being unavailable or slightly behind is normal.
 
Change `logDisagreements` to not warn if the disagreements are coming from failed responses.

Set `disagreementLogLevel` such that actual hash mismatches are logged at warn level.

### Remove double retry loop

Inner retry loop already retries a few times if consensus is not reached.

Outer retry loop should IMO be the responsibility of the caller. You already have infinite retries in `BulkStorageCommitFromStaging`. TBAR code also has infinite retries in the form of a polling trigger. The trigger was [recently changed](https://github.com/canton-network/splice/commit/1da36f32a5372cded2cfc9356c374132e12c589e#diff-67b7af8f037e70bad9d943d65878bda8ca078a57a722d7df0b0f17e6cf0abb32R13-R138) to never warn if the task failed because of unavailable data.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/BftScanConnection.scala
isegall-da: At some point I convinced myself that I need them separately, but I can't figure out why any more, so I think I just confused myself...

isegall-da:
Re

> In getBulkObjectChecksums, map individual responses that returned 404 originally to IgnoreResponse/NotYetResponse

I intentionally wanted to make it part of the API: give me a function that determines whether a response is a "not-yet"; rather than the implementation assuming something about what responses it should treat as such. Makes it more explicit for the caller that this is a supported feature.

isegall-da:
Converting to draft for now while I'm revisiting this.

apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/admin/api/client/commands/HttpScanAppClient.scala
isegall-da: Hmm... One thing that would make your suggestion awkward is the "bulk storage is disabled" response. For now, I want to make that a "not-yet", but once we finish roll-out, I want to turn it into a failure. If I follow your proposal, then that future change is an actual API change, which feels wrong.

isegall-da: > A response that conforms to the OpenAPI spec, but returns a Left in HttpCommand.handleOk 

That's where I believe you're missing something.
* A response that conforms to the OpenAPI spec and is listed in `nonErrorStatusCodes` goes to handleOk, where what you say is correct.
* However, a failuire response that confirms to the OpenAPI spec and is not listed in `nonErrorStatusCodes` never goes to handleOk, it remains a `HttpCommandException`. That's why I had to remove 404 from `nonErrorStatusCodes` for this endpoint, since it's in there by default.

(I do realize that I'm handling both 404 and 501 in `handleOk`, but now that I read it closer, I don't think that's ever actually executed)

PR #7308, opened 2026-09-15
Participant: set `queue-uncommitted-warn-threshold = 14000`

martinflorian-da:
[static]

Fixes DACH-NY/canton-network-internal#7066

It's the default starting from Canton 3.6; see https://github.com/DACH-NY/canton/pull/33829

(Overriding in splice seemed more economical than backporting above PR.)

I tested this config string on a local Canton.


moritzkiefer-da (approved):
thx

PR #7311, opened 2026-09-15
Don't wait forever on a non-active psid in `ensureSynchronizerRegisteredAndConnected`

martinflorian-da:
[ci]

Fixes DACH-NY/cn-test-failures#10088 , which is a rare race condition between validator app init and an actual LSU.

Not super happy with this fix; my best alternative suggestion would be to crash the validator app if that happens (which will make more sense on clusters) and tweak the `LsuIntegrationTest` so this race is more unlikely there.

The race seems too rare to motivate a bigger intervention where we bubble up the error to retry the affected init step with a different `psid`...


martinflorian-da:
The LSU test failure is unrelated: https://github.com/DACH-NY/cn-test-failures/issues/9445#issuecomment-5684175331

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminSynchronizerConnection.scala
```diff
@@ -196,6 +196,13 @@ trait ParticipantAdminSynchronizerConnection {
         s"participant registered ${config.synchronizerAlias} with config $config",
         lookupRegisteredSynchronizer(config.synchronizerAlias, config.psid).map {
           case Some(_) if !overwriteExistingConnection => Right(())
+          // Canton rejects modifications of non-active connections and the connection will never become active again.
+          // Not hard-crashing as the connect might still work and reconciliation automation will eventually fix the config.
+          case Some(existing) if existing.status != data.RegisteredSynchronizer.Status.Active =>
+            logger.error(
```
nicu-da: any reason to have this as an error and not a warn at most? it's a legitimate race condition.

nicu-da (approved):
thanks!

apps/common/src/main/scala/org/lfdecentralizedtrust/splice/environment/ParticipantAdminSynchronizerConnection.scala
martinflorian-da: Hm fair I think we could say warn.

IMHO it should be more than INFO because in theory we *could* have some follow up problem that is due to us not actually setting the config.

PR #7350, opened 2026-09-16
Update ACSSnapshotStore.insertNewSnapshot to use production code

OriolMunoz-da:
The big SQL statement was from before incremental snapshots, which means it wasn't testing anything useful


apps/scan/src/main/scala/org/lfdecentralizedtrust/splice/scan/store/AcsSnapshotStore.scala
```diff
@@ -853,6 +777,10 @@ class AcsSnapshotStore(
       s"acs_snapshot_stakeholders_v1_${historyId}_${snapshot.targetRecordTime.toEpochMilli}"
 
     for {
+      _ <-
+        sqlu"drop table if exists #$createsTableName"
```
rautenrieth-da: Why do we need to drop the table before creating it? Isn't this whole DBIO action run in a transaction? So if the table exists, it will already contain all the data, so dropping the table seems wrong?

rautenrieth-da (approved):
Thanks!

PR #7383, opened 2026-09-17
Copy `logback.xml` from Canton upstream

martinflorian-da:
Confirmed fix for what is failing in https://github.com/canton-network/splice/actions/runs/35247351678/job/105322057041?pr=7381

Confirmed locally by running `sbt bundle` twice:

- once without fix: repro of above issue
- once with fix: it works!

Created https://github.com/canton-network/splice/issues/7385 so that we improve our testing here so we catch such things earlier next time...

S.a. https://github.com/DACH-NY/canton/pull/35775


nicu-da (approved):
Trusting you for the logback changes as properly reviewing them would take a long time and considering this is not the logback we deliver as part of the docker images it's low risk.

I think we need to sort this out honestly, we have a huge amount of logback files scattered all through the project and at this point it's confusing which one is used for what.

martinflorian-da:
@nicu This is just a copy from the Canton repo, into our Canton fork! Nothing to review in detail... should have made this a bit more clear in the description.

> I think we need to sort this out honestly

Here I do agree; but in this particular case the "sorting out" should IMO be "kill the Canton fork". Why we use that particular logback file for anything at all is still not fully clear to me (some weird sbt dependencies?).

(cc @stephencompall-DA in case this interests you)

PR #7413, opened 2026-09-18
Fix performance test benchmarks

jagathweerasinghe-da:
fix: https://github.com/DACH-NY/cn-test-failures/issues/9829
[static]


ray-roestenburg-da (approved):
Thx

